PR #21, opened 2019-07-02
Fix Travis git checkout

anicholls:


anicholls:
These changes have already been put in master! Closing 👍

PR #24, opened 2019-07-03
refactor(forms): Split CSS forms module into individual modules

stephanerangaya:
## Summary

As part of our breaking migration, we are factoring @workday/canvas-kit-css-form out into individual input modules to match the react equivalent. This PR rebases https://github.com/Workday/canvas-kit/pull/10 on new `master`.

## Checklist

- [x] `yarn test` passes
      applicable

## Additional References

### Breaking Changes

**Code:**

- `@workday/canvas-kit-css-forms` has been split to match the react module breakdown. It has been decomposed into the following modules:
  - `@workday/canvas-kit-css-checkbox`
  - `@workday/canvas-kit-css-common`
  - `@workday/canvas-kit-css-form-field`
  - `@workday/canvas-kit-css-radio`
  - `@workday/canvas-kit-css-select`
  - `@workday/canvas-kit-css-textarea`
  - `@workday/canvas-kit-css-textinput`


modules/form-field/css/package.json
```diff
@@ -17,7 +17,13 @@
   },
   "dependencies": {
     "@workday/canvas-kit-css-core": "^3.0.0-alpha.2",
-    "@workday/canvas-system-icons-web": "^1.0.1"
+    "@workday/canvas-kit-css-checkbox": "^3.0.0-alpha.0",
```
anicholls: These are only dev dependencies

modules/radio/css/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "node ../../../utils/css-build.js index.scss"
+  },
+  "keywords": [
+    "canvas",
+    "canvas-kit",
+    "css",
+    "scss",
+    "sass",
+    "workday"
```
anicholls: Can you add keywords for each component?

modules/form-field/css/package.json
stephanerangaya: I'm not sure I understand, don't we need to reference them as dependencies to be able to use them?

stephanerangaya:
These changes have been moved to https://github.com/Workday/canvas-kit/pull/39

anicholls:
Note to self: We should be able to change `wdc-form-field-wrapper` to `wdc-form-field` and eliminate the need for the nested div

modules/form-field/css/package.json
anicholls: So right now we're re-exporting all of these in `modules/form-field/css/index.scss`. That's the old style though (where all form modules are available in this package). 

I'm about to push a commit that removes those re-exports. Since we don't use any of these for any of the exported code in `modules/form-field/css/lib`, they're not dependencies. However, since we are using them for demos in the storybook, we need to include them as devDependencies (also included in my commit).

PR #30, opened 2019-07-03
chore: move icon list under icon folder

mannycarrera4:
## Summary

Move icon list to icon css folder. References https://github.com/Workday/canvas-kit/pull/14


mannycarrera4:
There wasn't a readme for the icon list, still need to add this!

modules/icon/css/README.md
```diff
@@ -160,3 +160,24 @@ selector when injecting.
 ```js
 initializeIcons(null, '.wdc-custom-icon');
 ```
+
+# Canvas Kit CSS Icon List
+
```
lychyi: Can we add a description here? And usage below? I don't know if a user will be able to know what this section is for without those two things.

mannycarrera4: yep! just found an example to use

modules/core/css/index.scss
```diff
@@ -7,5 +7,5 @@
 @import './lib/accessibility.scss';
 
 // Components
-@import './lib/components/icon-list.scss';
+@import '~@workday/canvas-kit-css-icon/lib/icon-list';
```
lychyi: Question: if we're moving it from `core` to `icon` isn't this line supposed to be deleted rather than modified? Or was there a specific reason we need to keep this in core?

lychyi: Oh wait, I see. I guess `core` relies on this.

mannycarrera4: yep yep!

anicholls: @lychyi had it right here. Everything related to `icon-list` in core should be moving into the icon module. Everything in `core/css/lib/components/accessibility.scss` now pertains to icon-list and should just be moved over. We're trying to completely eliminate the `components` folder from `core/`

modules/page-header/css/index.scss
```diff
@@ -1,4 +1,4 @@
 @import '~@workday/canvas-kit-css-core/lib/mixins';
 @import '~@workday/canvas-kit-css-core/lib/variables';
-@import '~@workday/canvas-kit-css-core/lib/components/icon-list';
+@import '~@workday/canvas-kit-css-icon/lib/icon-list';
```
anicholls: Also need to add this import in `table` since it's not in core anymore.

modules/core/css/index.scss
mannycarrera4: oh i see! I'll make the change!

Issue #38, opened 2019-07-09
Fonts module should be opt-in


anicholls:
## 🐛 Bug Report

The fonts module should be opt-in so consumers can choose whether or not they consume things from our CDN (it's the only module that pulls directly from that URL). Currently it's being exported by default from `@workday/canvas-kit-react` and it should not be.

PR #40, opened 2019-07-10
Fix fonts module export

anicholls:
## Summary

The fonts module should be opt-in so consumers can choose whether or not they consume things from our CDN (it's the only module that pulls directly from that URL). Currently it's being exported by default from @workday/canvas-kit-react and it should not be.

Closes https://github.com/Workday/canvas-kit/issues/38

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/_canvas-kit-react/README.md
+++ b/modules/_canvas-kit-react/README.md
@@ -1,8 +1,8 @@
 # Canvas Kit React
 
-The bundle package containing all modules of the Canvas React Kit.
+The bundle package containing all modules of the Canvas Kit React.
 
 > Note: By default, no fonts are included with Canvas Kit modules. To use official Canvas Kit fonts,
 > install and import the `@workday/canvas-kit-react-fonts` module in addition to
-> `@workday/canvas-kit-react`. The fonts files are served from the Workday CDN so this allows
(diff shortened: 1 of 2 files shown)


modules/_canvas-kit-react/README.md
```diff
@@ -1,3 +1,8 @@
-# Canvas Kit
+# Canvas Kit React
 
 The bundle package containing all modules of the Canvas React Kit.
```
lychyi: Haha I saw this while on npm... this should be Canvas Kit React.

modules/_canvas-kit-react/README.md
```diff
@@ -1,3 +1,8 @@
...
+
+> Note: By default, no fonts are included with Canvas Kit modules. To use official Canvas Kit fonts,
+> install and import the `@workday/canvas-kit-react-fonts` module in addition to
+> `@workday/canvas-kit-react`. The fonts files are served from the Workday CDN so this allows
```
lychyi: I feel like the language here is odd since you probably are a consumer looking at this. Maybe something like:

> By installing `@workday/canvas-kit-react-fonts` you are opting in to using the Workday CDN.

Issue #42, opened 2019-07-17
Header Bugs


vibdev:
## 🐛 Bug Report

1) Header spreads html attributes props to multiple elements. This causes issues with style, id, auto-ids, etc
2) Search magnifying glass is not clickable to search
3) Search clear button (x) is not accessible and can not be tabbed to.

PR #43, opened 2019-07-17
fix(header): Header Search Accessibility & Prop Spread Fixes

vibdev:
## Summary

Closes https://github.com/Workday/canvas-kit/issues/42

Extra html attributes sent to header no longer spread to multiple elements.
Magnifying glass icon is now to clickable to submit search
Search clear button (x) is now accessible via keyboard controls

## Checklist

- [x] `yarn test` passes
      applicable

### DIFF
--- a/modules/header/react/lib/parts/Search.tsx
+++ b/modules/header/react/lib/parts/Search.tsx
@@ -371,7 +371,7 @@ export class Search extends React.Component<SearchProps, SearchState> {
           value={this.state.value}
           aria-label="Reset Search Input"
           icon={xSmallIcon}
-          buttonType={IconButtonTypes.Circle}
+          buttonType={iconButtonType(!!this.state.value && !this.state.focused && themeColor !== HeaderTheme.White)}
           onClick={this.resetSearchInput}
           type="reset"
           toggled={false}
--- a/modules/header/react/lib/parts/Search.tsx
+++ b/modules/header/react/lib/parts/Search.tsx
@@ -294,6 +294,7 @@ export class Search extends React.Component<SearchProps, SearchState> {
           unmountOnExit={true}
         >
           <SearchContainer
+            role="search"
             rightAlign={rightAlign}
             collapse={collapse}
             onSubmit={this.onSearchSubmit}
@@ -303,7 +304,6 @@ export class Search extends React.Component<SearchProps, SearchState> {
               placeholder={placeholder}
               themeColor={themeColor}
               collapse={collapse}
-              role="search"
               innerRef={this.inputRef}
               onFocus={this.setFocused.bind(this, true)}
               onBlur={this.setFocused.bind(this, false)}
@@ -341,6 +341,7 @@ export class Search extends React.Component<SearchProps, SearchState> {
 
     return (
       <SearchContainer
+        role="search"
         rightAlign={rightAlign}
         collapse={collapse}
         onSubmit={this.onSearchSubmit}
@@ -358,7 +359,6 @@ export class Search extends React.Component<SearchProps, SearchState> {
           placeholder={placeholder}
           themeColor={themeColor}
           collapse={collapse}
-          role="search"
           value={this.state.value}
           innerRef={this.inputRef}
           onMouseEnter={() => this.handleHover(true)}
--- a/modules/header/react/lib/parts/Search.tsx
+++ b/modules/header/react/lib/parts/Search.tsx
@@ -5,7 +5,7 @@ import {CSSTransition} from 'react-transition-group';
 import {HeaderHeight, HeaderTheme} from '../shared/types';
 import {colors, spacing, spacingNumbers, type} from '@workday/canvas-kit-react-core';
 import {focusRing} from '@workday/canvas-kit-react-common';
-import {Button, IconButton, IconButtonTypes} from '@workday/canvas-kit-react-button';
+import {IconButton, IconButtonTypes} from '@workday/canvas-kit-react-button';
 import {searchIcon, xIcon, xSmallIcon} from '@workday/canvas-system-icons-web';
 
 export interface SearchProps extends React.HtmlHTMLAttributes<HTMLFormElement> {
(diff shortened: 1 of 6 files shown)


modules/header/react/lib/parts/Search.tsx
```diff
@@ -298,40 +319,41 @@ export class Search extends React.Component<SearchProps, SearchState> {
...
+        {...elementProps}
+      >
+        <IconButton
+          aria-label="Search"
           icon={searchIcon}
-          style={{...iconStyle, pointerEvents: 'none'}}
-          color={iconColor}
-          colorHover={iconColor}
+          style={{...searchIconStyle}}
+          buttonSize={Button.Sizes.Small}
```
anicholls: What was the reason for using `Small` for these? This decreases the size of the icon so it no longer visually matches what we had before.

modules/header/react/lib/parts/Search.tsx
```diff
@@ -163,22 +163,24 @@ const iconStyle: React.CSSProperties = {
   left: spacing.xxs,
 };
 
-const SearchReset = styled('span')<Pick<SearchState, 'value'>>(
+const searchIconStyle: React.CSSProperties = {
```
anicholls: Do you mind using a styled `IconButton` for this so the treatment between this and `SearchReset` is the same?

modules/header/react/spec/__snapshots__/GlobalHeader.snapshot.tsx.snap
```diff
@@ -576,7 +576,7 @@ exports[`App GlobalHeader Snapshots renders a header with SystemIcon and IconBut
 `;
 
 exports[`App GlobalHeader Snapshots renders a header with a search bar 1`] = `
-.emotion-18 {
+.emotion-21 {
```
anicholls: Often these classname updates are because you didn't run `yarn build` before updating the tests. Can you try that first and see if it resolves this?

modules/header/react/lib/parts/Search.tsx
```diff
@@ -255,13 +258,23 @@ export class Search extends React.Component<SearchProps, SearchState> {
...
 
-    const {onSearchSubmit, ...props} = this.props;
+    const {
+      themeColor,
+      placeholder,
+      value,
+      rightAlign,
+      collapse,
+      onSearchSubmit,
+      ...elementProps
```
anicholls: Nit: not sure it's even worth calling out, but we've been using `elemProps` everywhere else.

modules/header/react/lib/parts/Search.tsx
vibdev: the focus ring fit super weird on the non small version

modules/header/react/spec/__snapshots__/GlobalHeader.snapshot.tsx.snap
vibdev: running `yarn build` then `npm test -u` incremented the number again

modules/header/react/lib/parts/Search.tsx
anicholls: I just pushed a commit with some fixes after reviewing with Jesse. I basically overrode the container size so we could use 32px with a 24px icon. The only point of contention is that we decided to bump the text input height up 4px (but this is specific to the header). Let me know what you think.

vibdev: I think this is fine, don't think it will block moving search to a standard form component down the line. I also pushed a update to the reset button to match the plain styes and fix the case where there is a value but no focus on the blue header.

modules/header/react/lib/parts/Search.tsx
```diff
@@ -5,11 +5,10 @@ import {CSSTransition} from 'react-transition-group';
 import {HeaderHeight, HeaderTheme} from '../shared/types';
 import {colors, spacing, spacingNumbers, type} from '@workday/canvas-kit-react-core';
 import {focusRing} from '@workday/canvas-kit-react-common';
-import {SystemIcon} from '@workday/canvas-kit-react-icon';
 import {IconButton} from '@workday/canvas-kit-react-button';
 import {searchIcon, xIcon, xSmallIcon} from '@workday/canvas-system-icons-web';
 
-export type SearchProps = {
+export interface SearchProps extends React.HtmlHTMLAttributes<HTMLFormElement> {
```
NicholasBoll: I didn't see anything that required this change. Is this change to be consistent with `HeaderProps`? Do we have documentation around prop passing as an API convention? What about child element attributes?

vibdev: https://github.com/Workday/canvas-kit/blob/master/API_PATTERN_GUIDELINES.md#prop-spread-behavior

modules/header/react/lib/parts/Search.tsx
```diff
@@ -274,19 +293,27 @@ export class Search extends React.Component<SearchProps, SearchState> {
...
+            onSubmit={this.onSearchSubmit}
+            {...elemProps}
+          >
             <SearchInput
-              {...props}
-              type="search"
+              placeholder={placeholder}
+              themeColor={themeColor}
+              collapse={collapse}
+              role="search"
```
NicholasBoll: Do you mean to put `role="search"` on the `<SearchContainer>` instead? I couldn't find any documentation for `role="search"` on inputs. https://www.w3.org/TR/wai-aria-practices/examples/landmarks/search.html

vibdev: whoops you are correct, will fix

modules/header/react/lib/parts/Search.tsx
```diff
@@ -255,13 +264,23 @@ export class Search extends React.Component<SearchProps, SearchState> {
...
+      rightAlign,
+      collapse,
+      onSearchSubmit,
+      ...elemProps
+    } = this.props;
 
     return (
       <React.Fragment>
         <IconButton
+          aria-label="Search"
```
NicholasBoll: This makes me wonder if a label should always be required for `IconButton`s

vibdev: This seems really reasonable to me, but is probably outside the scope of this PR. @anicholls should this be filed as an issue?

modules/header/react/lib/parts/Search.tsx
vibdev: updated! 😄 nice catch

modules/header/react/lib/parts/Search.tsx
anicholls: Yeah, that sounds like a great idea! If you could file that separately, that would be great 👍

modules/header/react/spec/__snapshots__/GlobalHeader.snapshot.tsx.snap
anicholls: Yeah, I'm seeing the same thing. Trying to figure out what's going on. I'd prefer to figure it out before merging. I tried updating the snapshots with `--no-cache` as well, but no luck. Still investigating... 🔍

PR #44, opened 2019-07-22
fix(Popper): Spread remainder props on wrapper div

anicholls:
## Summary

A request was made to be able to support adding a z-index to the wrapping div. This was solved generically using the styled components pattern to spread remainder props on the top-level element.

## Checklist

- [x] `yarn test` passes
- [x] code follows the
      [style guide](https://ghe.megaleo.com/UIC/wd-components/blob/master/docs/STYLEGUIDE.md)

### DIFF
--- a/modules/common/react/README.md
+++ b/modules/common/react/README.md
@@ -54,6 +54,10 @@ jest.mock('@workday/canvas-kit-react-common', () => ({
 
 ## Component Props
 
+This component extends the HTML `div` element. All additional props that are passed to this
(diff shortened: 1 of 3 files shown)


anicholls:
@NicholasBoll Yeah, it's definitely a bit open ended this way. However, we thought that was easier than trying to manage all the attributes we want to support (aria, data, style, id, className, etc.). The list starts to add up quick. We really want to limit the number of DOM nodes we create so I'm against component wrappers where possible.

We also need to start implementing our box component for simple styling similar to https://material-ui.com/system/spacing/. We have this available, but it hasn't been used for most components.

PR #39, opened 2019-07-10
DNM: CSS Audit: align styles to React components

stephanerangaya:
This is a WIP https://docs.google.com/spreadsheets/d/1OaEB6-ySup-kkMhQkWEMk9d93seeXeeqmqfCZ-d2fnM/edit#gid=0

PR #45, opened 2019-07-24
fix: specificity issue when you click on a primary button in the CSS …

roblevintennis:
@lychyi hopefully submitting PR's is ok! I wanted to have a look at Canvas so sort of just diving in. Feel free to ignore and/or delete this PR if it doesn't follow proper guidelines!

Unfortunately, when I ran `yarn test` I saw many failures locally. I've ran `yarn install` so I'm not certain why but it seems like it passes CI checks here.

## Issue

If you run `yarn start` off `master` and then visit the storybook locally and view the Button --> All of the CSS section, the primary button doesn't properly reflect the active/focus states (it works in the React version however) in that you see absolutely no box-shadow.

The reason is, even though those rules come after line #179 `&:hover:not([disabled]):not(.wdc-btn-disabled)`, they do not have the _:not_'s and so have lower specificity.

As it seems there's no reason to not have these in the focus/active as well, I added thus bumping up the specificity to be equal, and so now the top down specificity that I believe was originally intended kicks in:

AFTER
image

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/button/css/lib/button.scss
+++ b/modules/button/css/lib/button.scss
@@ -187,11 +187,11 @@ $_wdc-btn-caret-down-icon-hover: svg-load(
     }
 
     &.wdc-btn-active,
-    &:active {
+    &:active:not([disabled]):not(.wdc-btn-disabled) {
       background-color: $wdc-btn-bg-primary;
     }
 
-    &:focus,
+    &:focus:not([disabled]):not(.wdc-btn-disabled),
     &.wdc-btn-focus {
       background-color: $wdc-btn-bg-primary;
       box-shadow: 0 0 0 2px $wdc-btn-border-color-focus;


modules/button/css/lib/button.scss
```diff
@@ -187,11 +187,11 @@ $_wdc-btn-caret-down-icon-hover: svg-load(
     }
 
     &.wdc-btn-active,
-    &:active {
+    &:active:not([disabled]):not(.wdc-btn-disabled) {
```
roblevintennis: These have to override:
https://github.com/Workday/canvas-kit/blob/4c97e3789716469aae19292d61c5b96bc16c02c4/modules/button/css/lib/button.scss#L178

lychyi:
@roblevintennis Thanks for the contribution! The test failures could be a few things but did you also run `yarn build` before running your tests?

NicholasBoll:
@roblevintennis Thanks for bringing this up and taking initiative with a pull request. @stephanerangaya is currently working on the CSS kit to match the React kit. It looks like his Draft PR addresses the same issue.

It is a bit hard to understand the diffs on #39, but the fix is https://github.com/Workday/canvas-kit/pull/39/files#diff-fc945502a8e9527ec3570ce67e61df93R167 (I hope that link holds up) which effects original lines 190 and 194 making the specificity match.

@stephanerangaya could chime in to verify he is addressing it.

roblevintennis:
Hi @NicholasBoll and @lychyi and thanks for such swift feedback!

Regarding #39 I was able to see the diff and that fix is much better as it just removes the :not's (amongst other things) which keeps the code simpler—especially if they are not needed! I'll close this PR, but thanks again for getting back Nicholas!

Regarding `yarn build` I bet I didn't! I'll have to try > work later, thanks Lynn!

roblevintennis:
For posterity, I did get the tests passing after running `yarn build` 👍

PR #46, opened 2019-07-25
Reduce number of manual steps required after running create-module.sh

roblevintennis:
## Context
 I read the [Exporting a Module](https://github.com/Workday/canvas-kit/blob/master/CONTRIBUTING.md#exporting-a-module) section, and then had a look at the `create-module.sh` script and tried it out from the command line creating an imaginary `alert` module

 I noticed that while it sets up a lot of boiler-plate, there were the additional manual steps required in this section's list of steps. Wondering why, I noticed that shell script doesn't write to those files because they are updates to existing files rather then just scaffolding that can be written out with the general HEREDOC and then redirect to file strategy it's currently using.

I thought it would be nice if it _did_ set those up for you. Put differently, here it is in use case form:

> "As a user of the canvas-kit module generator, I'd like to be able to have boiler-plate created with as little manual steps as possible".

## Approach

Since we need to read in json and imports/exports, I thought this would be much more suitable for JavaScript, and so I have the shell script calling out to the corresponding node script.

## Issue

I _am_ blocked at the moment on the very last step of writing the import statement as the guide currently advises:

>Import your module in modules/_canvas-kit-css/index.scss.

When I do this via my script in `addImportToSassIndex`, everything else work, but, I get the following error:

image

Per the error and also googling for `npm @workday/canvas-kit-css` I see that these are in fact published to the NPM registry, so I'm wondering if that's a missing step in the guide, and/or, if I should remove that automated step and leave it back in the CONTRIBUTING.md guide (as I have it currently worded).

## Summary

Of course my above all assumes that my idea is of merit to the canvas-kit team :) If so, I'd love to get your thoughts on direction overall, if I'm breaking any conventions and/or following proper procedures and protocols, how I can rectify that if needed, and what we'd like to do for the sass imports question if you'd like me to proceed. Thanks!

Fwiw, I seem to be able to run `yarn lint` and `yarn test` successfully.

## Checklist

- [ ] `yarn test` passes
      applicable


create-module.sh
```diff
@@ -263,6 +263,14 @@ cat > $readme << EOF
 # Canvas Kit CSS $upperName
 EOF
 
+# Adds module dependency and export for React and CSS
```
roblevintennis: This part of the conditional is the `then` if they've replied they want to add CSS module (not visible in the diff)

utils/create-module.js
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Add the module as a dependency in modules/_canvas-kit-<TYPE>/package.json.
+ */
+const addModuleAsDependencyByType = (cwd, moduleName, type) => {
+  const packagePath = path.join(cwd, `modules/_canvas-kit-${type}/package.json`);
+  const packageContents = fs.readFileSync(packagePath);
+  const json = JSON.parse(packageContents);
+
+  // Add module, sort dependencies, then write back out
+  json.dependencies[`@workday/canvas-kit-${type}-${moduleName}`] = '^3.0.0-alpha.2';
```
roblevintennis: Maybe there's a better way to pull the `'^3.0.0-alpha.2';` part from a versions constant somewhere?

utils/create-module.js
```diff
@@ -0,0 +1,131 @@
...
+    if (arg.toLowerCase() === 'css') {
+      wantsCss = true;
+    }
+  }
+  addModuleAsDependency(cwd, moduleName, wantsCss);
+  addExportToIndex(cwd, moduleName);
+  if (wantsCss) {
+    // If we add the import by calling this, it throws error:
+    // error Couldn't find package "@workday/canvas-kit-css-alert@^3.0.0-alpha.2" required by "@workday/canvas-kit-css@3.0.0-alpha.2" on the "npm" registry.
+    addImportToSassIndex(cwd, moduleName);
```
roblevintennis: This is the area in question

NicholasBoll:
Thanks for looking into this. I'll take a look at it today.

utils/create-module.js
NicholasBoll: Agreed. The idea is to version lock packages, so you can pull from the root `package.json`. Is that correct @lychyi?

utils/create-module.js
NicholasBoll: I think you want `addExportToIndex` for 'react' module type and `addExportToSassIndex` for 'css' module type.

utils/create-module.js
```diff
@@ -0,0 +1,131 @@
...
+  sortedImports.unshift(
+    "$wdc-system-icons-path: '../../node_modules/@workday/canvas-system-icons-web/dist/svg' !default;"
+  );
+  fs.writeFileSync(indexPath, sortedImports.join('\n'), 'utf8');
+};
+
+// Only allow this script to be ran off command line
+if (require.main === module) {
+  const cwd = process.cwd();
+  const moduleName = process.argv[2];
```
NicholasBoll: The following 8 lines are a lot of logic to parse. I find the following much easier to understand:

```ts
const [_, __, moduleName, moduleType] = process.argv;
const wantsCss = moduleType.toLowerCase() === 'css'
```

And a friendlier error-handling version:
```ts
const [_, __, moduleName, moduleType = 'react'] = process.argv

if (!['css', 'react'].includes(moduleType)) {
  throw Error(`Module type must be either 'react' or 'css'`)
}
```

Going a little further, a boolean argument usually indicates you have 2 separate functions. Refactoring a bit:
```ts
if (require.main === module) {
  const cwd = process.cwd();
  const [_, __, moduleName, moduleType = 'react'] = process.argv

  if (!['css', 'react'].includes(moduleType)) {
    throw Error(`Module type (second argument) must be either 'react' or 'css'`)
  }

  addModuleAsDependencyByType(cwd, moduleName, moduleType);
  addExportToIndexByType(cwd, moduleName, moduleType);
}
```

utils/create-module.js
roblevintennis: Hey thanks for having a look @NicholasBoll :)

When you say root `package.json` are you referring to the one at the very top level? I don't see a version specified there or in the yarn.lock.

However, it occurs, the generator could just take a look at what's already in:
`modules/_canvas-kit-${type}/package.json`

so for React ones would be:
`modules/_canvas-kit-react/package.json`

and CSS would be:
`modules/_canvas-kit-css/package.json`

And just use whatever's there since I don't see how you'd be needing create a new module AND bump versions. Seems reasonable to me if you're good with it. If there was a `version: VERSION` constant somewhere it would be cleaner to use that, but, if not, I can do proposed solution.

Thoughts?

utils/create-module.js
roblevintennis: Hrm, I was replicating what's currently advised at:
https://github.com/Workday/canvas-kit/blob/master/CONTRIBUTING.md#css

I think for the CSS one it IS and `import` no?

Here's essentially what I'm writing in that method:

```js
  const importStatement = `@import '~@workday/canvas-kit-css-${moduleName}/index.scss';`;
```

The issue I was seeing was I believe the shell script does some sort of `yarn install` after I've added the import and then it can't find it cause it hasn't been published.

What's kind of funny as I type this is _rubber ducky_ realization that if I were to follow the current instructions as it's written today, the shell script would complete _first_, and _then_ I'd go manually add the import (and it wouldn't fail the `yarn install`) 🤣 

So, hrm, answering my own question I'm thinking I might have to do one of:

1. See if I can somehow pull out the CSS part in to it's own script that strategically doesn't get called until > the shell script pulls in the `yarn install` stuff.
1. If that doesn't work, I could just add that back to the read me and we're still "better" in terms of less steps

Lmk if that sounds correct, but otherwise I'll try to find some spare cycles to make these changes. Thanks again @NicholasBoll 🙏

utils/create-module.js
roblevintennis: I really like the destructuring approach you've outlined, I'll incorporate it for sure!

I'm thinking we not need the error handling as the 'css' is optional, but the 'react' step will always happen. Also, I think it would never fail...

If I'm not misunderstanding:
```js
  const [_, __, moduleName, moduleType = 'react'] = process.argv
```
Is an ES6 destructuring with also a default parameter being set for `moduleType` such that if `process.argv` is falsy at `[3]`, then `moduleType === 'react'` will always be true, correct? If so,

```js
 !['css', 'react'].includes(moduleType)
```
will never fail no?

If agreed, I'll just refactor to use pretty much everything in your example less the error handling. Sound good? LMK, thanks Nicholas!

utils/create-module.js
NicholasBoll: I meant the module type package. For instance, `json.version` will be `3.0.0-alpha.2` here. The new module gets the same version as the exported module.

NicholasBoll: When we add a new module, the export module will have to be bumped to include it (minor bump once out of alpha). When we bump a new version, all packages should be versioned the same. I think `json.version` is safe to use. There is no other source of truth.

utils/create-module.js
NicholasBoll: My thought was changing what you have in `create-module.sh` to call `create-module.js` multiple times - once for 'css' and once for 'react'. In the case you only want react, call it with 'react'. If you want both, call it for 'react' and another to call it with 'css'

> I'm thinking we not need the error handling as the 'css' is optional, but the 'react' step will always happen. Also, I think it would never fail...

The following will fail:
```
node ./utils/create-module.js myModule angular
```

The only commands that will pass this check are as follows:
```
node ./utils/create-module.js myModule
node ./utils/create-module.js myModule react
node ./utils/create-module.js myModule css
```

Personally I'd get rid of the default and require `react` and `css`

utils/create-module.js
NicholasBoll: I suppose the `index.scss` only has imports where the `index.ts` has both imports and exports. The name `addImportToSassIndex` is fine.

utils/create-module.js
roblevintennis: Gotcha, thanks for clarification 👍 
> require react and css
Makes sense

utils/create-module.js
NicholasBoll: I rearranged some things and it works. I run the `yarn` command before invoking your JS script. This seems to follow all the necessary steps outlined in https://github.com/Workday/canvas-kit/blob/master/CONTRIBUTING.md#creating-a-module

The chicken/egg problem is the CI will fail unless that package is published, but that would be an issue without automating anything. This is a problem.

NicholasBoll: @roblevintennis So, you'll have to hard-code `0.0.0` as the package version when writing the dependency to `_canvas-kit-css/package.json` and `_canvas-kit-react/package.json`. That version is special and `yarn` will pass. When we go to create a release, lerna will set the version number correctly for us.

NicholasBoll:
Here's a PR against this branch with some suggested changes. I tested it and it works: https://github.com/roblevintennis/canvas-kit/pull/1

It is currently conflicting with some of your recent changes here though. Take what you will from it.

utils/create-module.js
roblevintennis: Thanks for guidance @NicholasBoll ... I believe this is now working for the nominal cases. 

Running the node script directly without 'react' or 'css' works as intended:

image

utils/create-module.js
NicholasBoll: It seems the version should be a hard-coded `0.0.0`. That is treated special - it isn't actually pulled from npmjs.org and allows us to deploy through lerna when we're ready to publish

roblevintennis:
Really cool Nicholas, and I think we're on same page! I think we should go with what's here since I had to address:
1. EOL on one of the package.json's was getting removed
1. Kicking the node script has to come > `yarn install` step so it doesn't try to find our unregistered NPM modules (for both React and CSS it turns out).

I think they're the same in spirit. Hope you agree! If not, we could also add ^ to yours :) This was fun, thanks sir!

I better jump back on some other stuff, but I'll have a look for notifications!

utils/create-module.js
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Add the module as a dependency in modules/_canvas-kit-<TYPE>/package.json.
+ */
+const addModuleAsDependencyByType = (cwd, moduleName, type) => {
+  const packagePath = path.join(cwd, `modules/_canvas-kit-${type}/package.json`);
+  const packageContents = fs.readFileSync(packagePath);
+  const json = JSON.parse(packageContents);
+
+  // Add module, sort dependencies, then write back out
+  json.dependencies[`@workday/canvas-kit-${type}-${moduleName}`] = `^${json.version}`;
```
roblevintennis: Lmk if we want to switch this to that `0.0.0` one or if this will suffice

NicholasBoll: Yes, please use `0.0.0`. We use Lerna to publish. The `0.0.0` will be bumped to the real version during this step. `0.0.0` solves the `yarn install` problem. `0.0.0` is treated special. A symlink will be created to `../../modules/${moduleName}/react` and everything will pass as expected.

utils/create-module.js
```diff
@@ -14,7 +14,7 @@ const addModuleAsDependencyByType = (cwd, moduleName, type) => {
   const json = JSON.parse(packageContents);
 
   // Add module, sort dependencies, then write back out
-  json.dependencies[`@workday/canvas-kit-${type}-${moduleName}`] = `^${json.version}`;
+  json.dependencies[`@workday/canvas-kit-${type}-${moduleName}`] = `0.0.0`;
```
roblevintennis: Thanks @NicholasBoll I did mean to come back and do this but thanks for pulling it through and best of luck :)

PR #48, opened 2019-07-26
feat(menu): Menu Item can skip onClose

jayscheidt:
## Summary
The menu always fires the onClose event at the end of the onClick event (if onClose is provided). 
This change gives each menu item the ability to to skip the onClose event from firing, giving the menu more flexibility.

In our use case, a menu item has the ability to replace the contents of the menu with different menu items when it is clicked. This is difficult to do, because `onClose` will always fire. We were previously handling this by maintaining state in our application to determine if the `onClose` callback was actually supposed to close the menu.

## Checklist
- [x] `yarn test` passes

### DIFF
--- a/modules/menu/react/README.md
+++ b/modules/menu/react/README.md
@@ -134,7 +134,8 @@ Default: `true`
 #### `onClose: () => void`
 
 > If specified, this callback is executed when the menu should close. Called after an item is
-> selected or the escape shortcut key is used.
+> selected or the escape shortcut key is used. This will not fire if the menu item overrides 
+> the action with canSkipClose.
 
 ---
 
@@ -229,3 +230,8 @@ Default: `false`
 > Allow you to disable a menu item so it is not clickable.
 
 Default: `false`
+
+#### `canSkipClose: boolean`
+
+> Allows you to skip the onClose Menu callback that is fired at the end of
+> the onClick event.
--- a/modules/menu/react/README.md
+++ b/modules/menu/react/README.md
@@ -134,8 +134,8 @@ Default: `true`
 #### `onClose: () => void`
 
 > If specified, this callback is executed when the menu should close. Called after an item is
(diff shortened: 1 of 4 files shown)


NicholasBoll:
Should we extend the menu to support hierarchical menus instead? I didn't find much information on Menus in our Design system other than a single-level menus in Figma.

@anicholls @lychyi I see you had comments on the original internal Menu PR for CKR. Do you have opinions on this? My primary concern is hierarchy is a feature many teams will need and pushing that functionality to the application layer will make accessibility an exercise of the reader rather than simply built-in. It also means testing and implementation details are exposed to the consumer.

This fix is pretty simple, and maybe it is okay to add here, but I think hierarchy is something we should consider bringing into the Design System.

jayscheidt:
I was under the impression that there was a new menu in the works that will include nested hierarchical menus and that this menu component was the mid term plan. It is a weird situation, because we are striving for feature parity with the existing Workday App Header that doesn't match 100% to the newest designs

NicholasBoll:
Ah, yes. We talked about that a bit on Monday. Lots to catch up on. I'll talk with Robert, but we'll also schedule some time with him to go over the menu component he was working on.

Thanks!

modules/menu/react/README.md
```diff
@@ -229,3 +230,8 @@ Default: `false`
 > Allow you to disable a menu item so it is not clickable.
 
 Default: `false`
+
+#### `canSkipClose: boolean`
```
NicholasBoll: I didn't see a blurb about naming conventions on boolean variables. I see precedence for `hasX` and `isX`. Should this be named `shouldSkipClose` or maybe `shouldClose` defaulted to `true` to avoid the negative?

anicholls: Yep, I agree with @NicholasBoll, let's go with `shouldClose` that defaults to true 👍

jayscheidt: Done. Thanks for the feedback!

PR #49, opened 2019-07-26
refactor(Tooltip): Update CSS styles to match React implementation

stephanerangaya:
This PR updates the Tooltip CSS styles so they match the React component. 

Note : The story doesn't perfectly match the React one because it relies on a Button style that doesn't exists yet in CSS. It will be added to the `css-update` later and we'll update the story of `Tooltip CSS` after that.

### Breaking Changes

**Visual:**

- Updated styling to match React implementation. No longer includes an arrow since the origin of the tooltip should be obvious when your mouse is on it (there's also an animation that helps).

**Code:**

- Removed `wdc-tooltip-menu` implementation. This is moving to the Popup module
- Removed tooltip max width
- Removed a large majority of the tooltip variables as they are no longer needed

### DIFF
--- a/modules/tooltip/css/lib/tooltip.scss
+++ b/modules/tooltip/css/lib/tooltip.scss
@@ -1,12 +1,10 @@
 .wdc-tooltip {
   @include wdc-type();
   @include wdc-type-body();
-  @include wdc-depth-3();
   display: inline-block;
   background-color: $_wdc-tooltip-bg-color;
-  padding: $wdc-spacing-s;
-  border-radius: 3px;
-  border: 1px solid $_wdc-tooltip-border-color;
+  padding: 0;
+  border-radius: $wdc-spacing-xxxs;
   position: relative;
   width: max-content;
   max-width: $_wdc-tooltip-max-width;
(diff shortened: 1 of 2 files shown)


modules/tooltip/css/lib/tooltip.scss
```diff
@@ -1,12 +1,10 @@
 .wdc-tooltip {
   @include wdc-type();
   @include wdc-type-body();
-  @include wdc-depth-3();
   display: inline-block;
   background-color: $_wdc-tooltip-bg-color;
-  padding: $wdc-spacing-s;
-  border-radius: 3px;
-  border: 1px solid $_wdc-tooltip-border-color;
+  padding: 0;
```
anicholls: Duplicate padding

modules/tooltip/css/lib/tooltip.scss
```diff
@@ -24,142 +22,31 @@
...
-    &:after {
-      margin-left: -$_wdc-tooltip-arrow-size;
-      border-bottom-color: $_wdc-tooltip-bg-color;
-    }
+    margin-top: $wdc-spacing-xxxs;
   }
 
-  padding: $wdc-spacing-xs $wdc-spacing-s;
-  font-size: 14px;
+  padding: $wdc-spacing-xxs;
```
anicholls: Can you move these styles up with the other base `.wdc-tooltip` styles?

modules/tooltip/css/lib/tooltip.scss
```diff
@@ -24,142 +22,31 @@
...
 
   &-top {
-    margin-bottom: $wdc-spacing-s;
-
-    &:before,
-    &:after {
-      top: 100%;
-      left: $_wdc-tooltip-arrow-offset;
-    }
+    margin-bottom: $wdc-spacing-xxxs;
```
anicholls: Can you horizontally center the `top` & `bottom` variations? Can you also add `topleft`, `topright`, `bottomleft`, and `bottomright`?

modules/tooltip/css/lib/tooltip.scss
```diff
@@ -1,12 +1,10 @@
...
   display: inline-block;
   background-color: $_wdc-tooltip-bg-color;
-  padding: $wdc-spacing-s;
-  border-radius: 3px;
-  border: 1px solid $_wdc-tooltip-border-color;
+  padding: 0;
+  border-radius: $wdc-spacing-xxxs;
   position: relative;
   width: max-content;
   max-width: $_wdc-tooltip-max-width;
```
anicholls: There's no min/max width in the react version. Let's just drop all of these.

modules/tooltip/css/lib/variables.scss
```diff
@@ -1,8 +1,5 @@
-$_wdc-tooltip-bg-color: $wdc-color-french-vanilla-100;
+$_wdc-tooltip-bg-color: rgba(0, 0, 0, .85);
 $_wdc-tooltip-border-color: $wdc-color-soap-500;
```
anicholls: Can you remove this and the min/max width variables?

modules/tooltip/css/lib/tooltip.scss
stephanerangaya: What do you mean?

anicholls: [Line 47](https://github.com/Workday/canvas-kit/pull/49/files#diff-e517748a8012b36b2b0623e4b4bb0444R47) overrides this. When you move those styles up, you'll see it haha.

stephanerangaya: Oh I see

PR #52, opened 2019-07-29
ci(travis): tagged publish; re-enable master storybook

d-bye:
Changes to travis config allowing:
- npm publish on tagged commits (on master)
- master branch storybook builds to be pushed to gh-pages (if this was disabled for a reason let me know!)


.travis.yml
```diff
@@ -5,18 +5,14 @@ cache:
...
-  except:
-  - gh-pages
 stages:
 - name: pr-previews
   if: "(branch = master AND type = pull_request) OR branch != master OR NOT branch =~ /^release-.+$/"
 - name: npm publish
-  if: branch =~ /^release-.+$/ AND type != pull_request
-- name: storybook
-  if: branch =~ /^release-.+$/ AND type != pull_request
+  if: branch = master AND type != pull_request AND tag IS present
```
anicholls: @lychyi Do we want to constrain the content of tags that get released at all?

lychyi: Tagging in general is fine for now. We can reduce the scope by a regex matcher later when we find reasons why we need to tag master other than for releases.

.travis.yml
```diff
@@ -29,9 +25,14 @@ jobs:
...
-    script:
-    - "./utils/travis/getGit.sh"
-  - stage: storybook
+    deploy:
+      provider: npm
+      email: "$GITHUB_EMAIL"
+      api_key: "$NPM_TOKEN"
+      on:
+        branch: master
+        tags: true
```
lychyi: Looking over the docs, does this mean that `branch` is moot if we set `tags`? https://docs.travis-ci.com/user/deployment/#conditional-releases-with-on

Also do we need `if` in the stages, if these are set?

lychyi: I guess we'll always need `if: branch = master AND type != pull_request` at least if `on.tags` invalidates `on.branch`

PR #53, opened 2019-07-29
Update Button CSS styles to match React

stephanerangaya:
This PR updates the Button CSS styles so they match the React component. 

Notes:
`.wdc-btn-icon` has been **removed**, `.wdc-btn-icon-square` and `.wdc-btn-icon-circle` replace it. A size must be set for icon buttons, using either `.wdc-btn-medium` or `.wdc-btn-small`.


modules/icon/css/lib/icon-list.scss
```diff
@@ -1,7 +1,10 @@
 .wdc-icon-list {
   display: flex;
-  margin-left: -$wdc-spacing-xxxs;
-  margin-right: -$wdc-spacing-xxxs;
+  margin-left: -$wdc-spacing-l;
```
anicholls: What did this get changed for?

modules/button/css/stories.tsx
```diff
@@ -2,248 +2,254 @@ import React from 'react';
...
+  .add('Primary', () => (
     <div className="story">
-      <section>
-        <button className="wdc-btn wdc-btn-primary">Primary</button>
-        <button className="wdc-btn">Secondary</button>
-        <button className="wdc-btn wdc-btn-delete">Delete</button>
-      </section>
+      <h3>Large Primary</h3>
+      <button className="wdc-btn wdc-btn-primary">Primary Button</button>
+      <button className="wdc-btn wdc-btn-primary wdc-btn-disabled">Primary Button</button>
```
anicholls: Can you add a `disabled` attributed to all of the ones we're using the disabled class on? Right now they're all tab focusable.

anicholls: Actually all of our styles shouldn't require the class if the attribute is there, so you should be able to drop `wdc-btn-disabled` as well.

modules/button/css/lib/icon-button.scss
```diff
@@ -0,0 +1,244 @@
...
+    &-square-filled {
+      border-radius: $wdc-btn-border-radius-icon;
+    }
+
+    border-radius: 999;
+
+    &:not([disabled]):not(.wdc-btn-disabled) {
+      &.wdc-btn-active,
+      &:active {
+        box-shadow: 0 0 0 2px white, 0 0 0 4px $wdc-btn-border-active;
```
anicholls: Just calling this out. I noticed this is the only one in CKR that uses `0 0 0 2px $wdc-btn-border-active`. Everything else has the 2px white ring. This is inconsistent right now, so we may want to update the React version. **Follow up needed**

anicholls: Another note: We should probably support the toggled states as well, but I think we can add that later down the road.

modules/button/css/lib/button.scss
```diff
@@ -22,27 +22,30 @@ $_wdc-btn-caret-down-icon-hover: svg-load(
...
   height: $wdc-btn-height-l;
-  line-height: $wdc-btn-height-l;
   padding: $wdc-btn-padding-l;
   background-color: $wdc-btn-bg;
-  border: none;
+  border: 1px solid $wdc-btn-border;
   color: $wdc-btn-color;
   border-radius: $wdc-btn-border-radius;
-  transition: 0.2s background, 0.2s border-color, 0.2s box-shadow;
+  transition: all 120ms linear;
```
anicholls: Because of the `focusRing` helper, this doesn't actually match. Can you do `all 120ms, box-shadow 100ms` instead? (linear easing is the default)

modules/button/css/README.md
```diff
@@ -51,6 +51,7 @@ buttons are used in a sequence of screens (e.g. Task Orchestration) and on mobil
   - Small
 - [Colors](#colors)
   - Default
+  - Filled (Icons)
```
anicholls: Can you explain what the `Filled` color is in the colors section?

modules/button/css/README.md
```diff
@@ -60,7 +61,7 @@ buttons are used in a sequence of screens (e.g. Task Orchestration) and on mobil
   - Disabled
 - [Types](#types)
```
anicholls: Link is broken. Should update title, or change this to `[Variants](#variants)`

modules/button/css/README.md
```diff
@@ -159,15 +160,23 @@ applied to form controls using the classes below.
 > Use when space is limited and the action is clearly communicated through an icon. For instance, an
 > edit button.
 
-Use by adding `.wdc-btn-icon`. A Workday icon should be used as the `<button>`'s content.
+Use by adding `.wdc-btn-icon-square` or `.wdc-btn-icon-circle`. A Workday icon should be used as the
```
anicholls: Can you add a note about accessibility? (aria-label, etc.)

modules/icon/css/lib/icon-list.scss
mannycarrera4: I was trying to match this style: https://github.com/Workday/canvas-kit/blob/master/modules/page-header/react/lib/PageHeader.tsx#L74

modules/button/css/lib/variables.scss
```diff
@@ -22,47 +25,57 @@ $wdc-btn-padding-l: 0 $wdc-spacing-l;
...
+$wdc-btn-border-radius-icon: $wdc-spacing-xxxs;
 
 $wdc-btn-bg: $wdc-color-soap-200;
 $wdc-btn-bg-hover: $wdc-color-soap-500;
 $wdc-btn-bg-active: $wdc-color-soap-200;
 $wdc-btn-bg-focus: $wdc-color-soap-200;
 $wdc-btn-bg-disabled: $wdc-color-french-vanilla-100;
-$wdc-btn-bg-icon-active: $wdc-color-blueberry-500;
-$wdc-btn-bg-icon-disabled: $wdc-color-french-vanilla-100;
+$wdc-btn-bg-icon: $wdc-color-french-vanilla-100;
```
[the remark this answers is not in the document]
mannycarrera4: hey @anicholls i did some refactoring, would love a second pair of eyes on this to see if this is what you meant!

modules/icon/css/lib/icon-list.scss
anicholls: @mannycarrera4 This also gets used in `Table` can you make sure it didn't break that use case? If this change is specific to the page header the margin should probably get added in that module.

modules/button/css/lib/variables.scss
```diff
@@ -1,71 +1,88 @@
...
+$wdc-btn-border-active: $wdc-color-blueberry-400;
+$wdc-btn-border-focus: $wdc-color-blueberry-400;
+$wdc-btn-border-disabled: $wdc-color-soap-500;
+$wdc-btn-border-primary: $wdc-color-cantaloupe-400;
+$wdc-btn-border-primary-hover: $wdc-color-cantaloupe-400;
+$wdc-btn-border-primary-active: $wdc-color-blueberry-400;
+$wdc-btn-border-primary-disabled: $wdc-color-cantaloupe-400;
+$wdc-btn-border-delete: $wdc-color-cinnamon-500;
+$wdc-btn-border-delete-hover: $wdc-color-cinnamon-500;
+$wdc-btn-border-delete-active: $wdc-color-blueberry-400;
```
anicholls: Should be `$wdc-color-cinnamon-500`

modules/button/css/lib/variables.scss
```diff
@@ -1,71 +1,88 @@
...
-$wdc-btn-border-color-delete-disabled: $wdc-color-cinnamon-200;
-
 $wdc-btn-split-border-radius: 3px;
 $wdc-btn-split-icon-height: $wdc-spacing-s;
 $wdc-btn-split-icon-width: $wdc-spacing-xxxs + $wdc-spacing-m + $wdc-spacing-xxs;
 $wdc-btn-split-icon-padding-v: ($wdc-btn-height-l - $wdc-btn-split-icon-height)/2;
 $wdc-btn-split-icon-padding: $wdc-btn-split-icon-padding-v $wdc-spacing-xxs
   $wdc-btn-split-icon-padding-v $wdc-spacing-xxxs;
+
+$wdc-btn-icon-inverse: $wdc-color-french-vanilla-100;
```
anicholls: Can you add `-bg` to all of the inverse variables as well?

modules/button/css/lib/variables.scss
anicholls: This is actually a bug in the React button. Do you mind updating it there as well?

modules/button/css/lib/variables.scss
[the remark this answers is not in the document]
anicholls: @mannycarrera4 I left a couple comments, but this looks way better and is much clearer to understand! Thanks 🙂

modules/button/css/lib/icon-button.scss
anicholls: Focus outline update captured here: https://github.com/Workday/canvas-kit/issues/97

modules/button/css/lib/variables.scss
mannycarrera4: will do!

mannycarrera4:
@anicholls `plain` and `inverse filled` should be added now!

modules/button/css/lib/button.scss
```diff
@@ -210,87 +224,59 @@ $_wdc-btn-caret-down-icon-hover: svg-load(
...
     &.wdc-btn-disabled,
     &:disabled {
-      @include wdc-btn-box-shadow($wdc-btn-border-color-delete-disabled);
+      box-shadow: none;
       color: $wdc-btn-color-delete-disabled;
       background-color: $wdc-btn-bg-delete-disabled;
-    }
-  }
-
-  &.wdc-btn-icon {
```
anicholls: Last change - I promise haha. Can we add this back in and make it match the default in react? (i.e. `wdc-btn-icon` inherits all the styles from `wdc-btn-icon-circle`). 

Can you also include the `.wdc-btn-medium` size by default?

Can you put these changes in the icon button file instead of this one?

This will require much less markup for the most common use cases.

modules/button/css/lib/variables.scss
```diff
@@ -1,71 +1,91 @@
 $_wdc-btn-icons-path: $wdc-system-icons-path !default;
 
-$wdc-btn-icon-size: $wdc-spacing-s;
-$wdc-btn-icon-size-l: $wdc-spacing-m;
+$wdc-btn-font-weight: 500;
```
anicholls: Can you document which variables have been removed/renamed in the PR description? We will have to add this to the migration guide.

e.g.
"Several SCSS variables have been added, removed or renamed:
- ➕ `$wdc-btn-font-weight`
- ✖️ `$wdc-btn-icon-size-l`
...
- ✏️ `$wdc-btn-bg-icon-active` > `$wdc-btn-icon-active-bg`

etc."

mannycarrera4: will do!

modules/button/css/lib/icon-button.scss
```diff
@@ -7,6 +7,7 @@
   &-plain,
   &-inverse,
   &-inverse-filled {
+    @include wdc-btn();
```
mannycarrera4: this is so we don't have to use `wdc-btn wdc-btn-icon-square`

modules/button/css/lib/icon-button.scss
```diff
@@ -60,7 +61,7 @@
 
       .wd-icon-accent,
       .wd-icon-fill {
-        fill: $wdc-btn-icon-disabled-color;
+        fill: $wdc-btn-icon-disabled-color !important;
```
mannycarrera4: Because of JS injector for icons, we need to add !important to override those styles that are coloring the icons.

modules/button/css/lib/icon-button.scss
```diff
@@ -149,10 +150,10 @@
       background-color: transparent;
       .wd-icon-accent,
       .wd-icon-fill {
-        fill: $wdc-btn-icon-inverse-bg;
+        fill: $wdc-btn-icon-inverse-bg !important;
       }
 
-      &.wdc-btn-hover:not([disabled]),
+      &.wdc-btn-hover:not([disabled]):not(.wdc-btn-disabled),
```
mannycarrera4: i can probably remove `wdc-btn-disabled`

modules/button/css/lib/icon-button.scss
```diff
@@ -325,14 +326,11 @@
       }
     }
 
-    &.wdc-btn-medium {
-      width: 40px;
-      height: 40px;
-    }
-
```
mannycarrera4: made this the default size for icon buttons

modules/button/css/lib/icon-button.scss
```diff
@@ -325,14 +326,11 @@
...
     &.wdc-btn-small {
       width: 32px;
       height: 32px;
+      min-width: 32px;
+      max-width: 32px;
```
mannycarrera4: because we're extending `wdc-btn` it's .wdc-btn-small has a min width and max width that is different for icon buttons, so I added this here to override those styles.

PR #66, opened 2019-08-01
fix(TextArea): align border color to match other inputs

neilpelow:
Simple style fix for `TextArea` comp. Behaviour now matches `TextInput`


lychyi:
Thanks for the contribution Neil! 🎉 

You may need to update some snapshots for this change.
1. Make sure `yarn` is run and deps are up to date
2. `yarn build`
3. `yarn updateSnaps` (or `yarn test -u`)

It should generate some snapshots representing the change you made.

PR #72, opened 2019-08-01
chore: Release 3.0.0-alpha.4

anicholls:
## Summary
Now that we've set up the tag-based release infrastructure, I'm opening the PR for our first one! This should serve as a template for the future releases.

- Added a `yarn bump` command. Note: until we graduate v3, you need to use `yarn bump prerelease`
- Bumped the versions across the board to `3.0.0-alpha.4`
- Added a changelog that describes the branches merged since the last release

**Next Steps**
- Create a [new release](https://github.com/Workday/canvas-kit/releases/new)
- Use `v3.0.0-alpha.4` for the tag version and release name
- Copy the changelog contents for `v3.0.0-alpha.4` into the description of the new release
- Check "This is a pre-release"
- Publish ✅


package.json
```diff
@@ -102,7 +102,8 @@
     "format:tslint": "tslint --project tsconfig.json --fix",
     "format": "yarn format:prettier && yarn format:tslint",
     "test": "TZ=UTC jest",
-    "updateSnaps": "TZ=UTC jest -u"
+    "updateSnaps": "TZ=UTC jest -u",
+    "version": "yarn lerna version --no-git-tag-version --no-push --no-changelog"
```
lychyi: Not necessary for this release but we could probably relocate this to `lerna.json` and delete `command.publish`

lychyi: Just found out that `version` script here will run twice since `lerna` runs `yarn version` as a lifecycle. I'll make the changes in `lerna.json` for the above comment and rename this script to something else.

Issue #73, opened 2019-08-02
SidePanel never removes resize event


lychyi:
## 🐛 Bug Report

We need to remove the event listener when `SidePanel unmounts. Right now it erroneously adds another event listener.

https://github.com/Workday/canvas-kit/blob/master/modules/side-panel/react/lib/SidePanel.tsx#L141

PR #74, opened 2019-08-02
fix: remove event handler for resize when component unmounts

mannycarrera4:
## Summary

Addresses: https://github.com/Workday/canvas-kit/issues/73

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/side-panel/react/lib/SidePanel.tsx
+++ b/modules/side-panel/react/lib/SidePanel.tsx
@@ -138,7 +138,7 @@ export default class SidePanel extends React.Component<SidePanelProps, SidePanel
     window.addEventListener('resize', throttle(this.handleResize, 150));
   }
   public componentWillUnmount() {
-    window.addEventListener('resize', throttle(this.handleResize, 150));
+    window.removeEventListener('resize', throttle(this.handleResize, 150));
   }
 
   public render() {


lychyi:
@mannycarrera4

Adding relevant component scope ("SidePanel" in this case) to your conventional commit messages will help us quickly scan PRs and gives a good indication to what the final commit should look like when we squash and merge. I know @NicholasBoll added the scope when we merged, but having it in the title/commit initially makes all the difference!

Another thing to consider is to leverage keywords to automatically close the referenced issue: https://help.github.com/en/articles/closing-issues-using-keywords. In this case, instead of "Addresses", I suggest using "Fixes" or "Resolves". This little automation win will help us keep the issues neat and tidy. ✨

lychyi:
@mannycarrera4 @NicholasBoll 

I don't think this works either. We need to have a reference to the function `throttle` generates in order for this to get cleaned up correctly.

```js
constructor() {
  this.throttledHandleResize = throttle(this.handleResize, 150);
}
...

public componentDidMount() {
  window.addEventListener('resize', this.throttledHandleResize);
}
public componentWillUnmount() {
  window.removeEventListener('resize', this.throttledHandleResize);
}
```

PR #75, opened 2019-08-02
fix(menu): Allow aria role of menu item to be overridden

jayscheidt:
## Summary
There were some menu fixes from the old repository that did not make its way into this one. This 
PR includes four commits that have an assortment of bug fixes that resolved the issue with these comments: 
- When you first load the first story, there is no hover state for the first item. Once you've tabbed into the menu, this goes away.
- Disabled menu item icons should always use iconColors.disabled. Right now they flash between that and the standard color depending on where you've hovered.
- Every once in a while I see white text on a gray background on hover. It only happens on the first item when it's focussed.

## Checklist
- [x] `yarn test` passes

## Additional References
N/A


modules/menu/react/stories/stories.tsx
```diff
@@ -11,27 +11,29 @@ import {CanvasSystemIcon} from '@workday/design-assets-types';
...
 import Menu from '../lib/Menu';
 import MenuItem, {MenuItemProps} from '../lib/MenuItem';
 
 import README from '../README.md';
 
 const FocusableButton = React.forwardRef(
-  (props: ButtonProps<BetaButtonTypes>, ref: React.RefObject<HTMLButtonElement>) => (
-    <DropdownButton buttonRef={ref} {...props}>
+  (props: ButtonProps, ref: React.RefObject<HTMLButtonElement>) => (
+    <Button buttonRef={ref} {...props}>
```
NicholasBoll: @vibdev Did we mean to remove `DropdownButton`? Here's what it looks like before and after:

Before:
https://workday.github.io/canvas-kit/?path=/story/menu--with-control-button
Screen Shot 2019-08-06 at 12 08 05 PM

After:
Screen Shot 2019-08-06 at 12 05 39 PM

@anicholls Can you confirm if this change is intentional?

vibdev: @jayscheidt was this something that got eaten by the cherry pick?

jayscheidt: Not sure if I can link the storybook from ghe, but the `After` screenshot matches what is in master of the old repo

vibdev: I wonder if @anicholls made the change when the lib moved?

anicholls: So the reason this was done was due to the lack of availability of the blue buttons in our product. They are coming, but not widely available yet. I forget who made the call, but either @jpante or Jesse Spencer requested these be changed back to the original orange style to more closely align with what's in product today for the short term. 

We don't have a dropdown button version in orange, so we just used the regular `Button`. With the blue buttons becoming available soon, I'd say we could probably keep this as `DropdownButton`.

NicholasBoll: The clarification I got was to use the `<Button>` component (orange buttons) and we'll change as part of the the migration to the new blue buttons later. I'm approving and merging.

PR #79, opened 2019-08-05
build(travis): skip cleanup on npm publish

lychyi:
## Summary
We need to skip the cleanup phase for Travis since we rely on the artifacts generated from the `install` phase to release.

This PR also tags every publish as prerelease. This models how Lerna does it for us today. We currently need to manually push `latest` as a dist-tag since we have no official `latest` version. 

We'll eventually need to change this up and write a script that pushes the appropriate dist tags during a publish.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ git:
   depth: 3
 stages:
 - name: pr-previews
-  if: "(branch = master AND type = pull_request) OR branch != master OR NOT branch =~ /^release-.+$/"
+  if: "(branch = master AND type = pull_request) OR branch != master"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,7 +7,7 @@ git:
   depth: 3
 stages:
 - name: pr-previews
-  if: "(branch = master AND type = pull_request) OR branch != master OR NOT branch =~ /^release-.+$/"
+  if: "(branch = master AND type = pull_request) OR branch != master"
```
lychyi: no longer need this branch check.

PR #82, opened 2019-08-07
ci: Break out testing to stages

d-bye:
- Changing lint/test scripts to run on PRs and non-tagged master


.travis.yml
```diff
@@ -6,11 +6,11 @@ cache:
 git:
   depth: 3
 stages:
-- name: pr-previews
-  if: "(branch = master AND type = pull_request) OR branch != master"
+- name: testing
+  if: branch = master AND type = pull_request
```
lychyi: I just realized that this won't build PRs that don't target master, so this won't work. What was the reason we couldn't just do `type = pull_request` again? I forgot how we came to that conclusion.

Issue #96, opened 2019-08-09
Export missing enums as static variables


anicholls:
## 🚀 Feature Proposal

A few components are not exporting all their enums as static variables. `Banner`, `Select`, and `SidePanel` are the culprits.

Pulled from https://github.com/Workday/canvas-kit/issues/51, but is not breaking.

### Example
See https://github.com/Workday/canvas-kit/blob/master/API_PATTERN_GUIDELINES.md#static-class-variables

PR #110, opened 2019-08-12
fix: Add missing static variables and fix Popup padding static variable

anicholls:
## Summary

We generally export all enums that you would use with a given component as static variables to reduce the number of imports consumers have to use. A few components are not exporting all their enums as static variables. `Banner`, `Select`, and `SidePanel` are the culprits. This resolves those issues. Bonus: Popup had some documentation issues with it's static vars and those have been remedied.

Closes https://github.com/Workday/canvas-kit/issues/96

### Breaking Change
- Renamed `Popup.padding` to `Popup.Padding`

## Checklist

- [ ] `yarn test` passes
      applicable


anicholls:
Not sure why the `push` build is triggering. This seems like a travis issue, so the PR is blocked on that.

I don't see this check on any other forked PRs and it seemed to only happen on this one for me (the first PR of my personal fork).

modules/popup/react/README.md
```diff
@@ -40,7 +40,11 @@ import {Popup} from '@workday/canvas-kit-react-popup';
 
 ## Static Properties
 
-> None
+#### `padding: PopupPadding`
```
lychyi: Since we're doing breaking changes anyway, can we change this from `padding` to `Padding`?

PR #113, opened 2019-08-13
refactor(Card): Update CSS styles to match React implementation

mannycarrera4:
## Summary

Most of the styles for the card was matching the react card. The story and readme needed to be updated to match the styles that are the same in the react card component.

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/card/css/README.md
+++ b/modules/card/css/README.md
@@ -33,25 +33,25 @@ Card sizes follow the 12 column grid system. To adjust sizes append the size mod
 ```html
 <div class="wdc-card-container">
   <div class="wdc-card-4">
-    <div class="wdc-card-header">
+    <h3 class="wdc-card-header">
       Card Header
-    </div>
+    </h3>
     <div class="wdc-card-body">
       .wdc-card-4
     </div>
   </div>
   <div class="wdc-card-4">
-    <div class="wdc-card-header">
+    <h3 class="wdc-card-header">
       Card Header
-    </div>
+    </h3>
     <div class="wdc-card-body">
       .wdc-card-4
     </div>
   </div>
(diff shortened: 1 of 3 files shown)


modules/card/css/lib/card.scss
```diff
@@ -30,6 +31,11 @@ $_wdc-container-positions: 'start' flex-start, 'end' flex-end, 'center' center,
   }
 }
 
+.wdc-card {
+  @include card-common();
+  width: 100%;
```
anicholls: Let's drop this since it's inherent to a block element.

modules/card/css/README.md
```diff
@@ -95,12 +95,33 @@ Card positioning can be changed by adding a modifier to the container class,
...
     </div>
   </div>
 </div>
 ```
+
+#### Card Depth
+
+Five levels of depth available in CSS classes
+
+> The CSS classes are implemented using the equivalent SASS mixin.
```
anicholls: Let's drop this.

modules/card/css/README.md
```diff
@@ -95,12 +95,33 @@ Card positioning can be changed by adding a modifier to the container class,
...
     <div class="wdc-card-body">
       .wdc-card-container .wdc-card-container-space-around
     </div>
   </div>
 </div>
 ```
+
+#### Card Depth
+
+Five levels of depth available in CSS classes
```
anicholls: In case we ever change the number of depths available we shouldn't use the exact number. Can we change this to:

"Depth 2 is the default. You can use the [depth classes from @workday/canvas-kit-css-core](../../core/css/README.md#depth) to set a custom depth for your card."

PR #115, opened 2019-08-13
fix(TextButton): Update TextButton focus outline for accessibility

jstin:
## Summary

This PR changes the Focus state of Text Buttons, to stay consistent with the Workday Accessibility Guidelines.

## Checklist

- [x] `yarn test` passes

### New focus state shown below:
---
Screen Shot 2019-08-13 at 4 35 20 PM
Screen Shot 2019-08-13 at 4 35 26 PM

### Old focus state shown below:
---

Screen Shot 2019-08-13 at 4 43 09 PM
Screen Shot 2019-08-13 at 4 43 16 PM


modules/button/react/lib/ButtonStyles.ts
```diff
@@ -453,10 +453,11 @@ function getButtonFocusRing(buttonType: AllButtonTypes): CSSObject {
     case BetaButtonTypes.OutlineInverse:
       return focusRing(2, 2, true, false, buttonColors.focusRingInner, buttonColors.focusRingOuter);
     case TextButtonTypes.Default:
-    case TextButtonTypes.Inverse:
     case TextButtonTypes.AllCaps:
+      return focusRing(2, 0, true, false, buttonColors.focusRingInner, buttonColors.focusRingOuter);
```
anicholls: This can just use the default params so you can combine these cases with the ones above for `focusRing(2, 0)`

modules/button/react/lib/ButtonStyles.ts
```diff
@@ -453,10 +453,11 @@ function getButtonFocusRing(buttonType: AllButtonTypes): CSSObject {
...
+    case TextButtonTypes.Inverse:
     case TextButtonTypes.InverseAllCaps:
-      return {};
+      return focusRing(2, 0, true, false, buttonColors.focusRingInner, buttonColors.focusRingOuter);
```
anicholls: All of the cases returning this function should be combined into one (`TextButtonTypes.Inverse`, `TextButtonTypes.InverseAllCaps`, `IconButtonTypes.Inverse` and `IconButtonTypes.InverseFilled`)

modules/button/react/lib/ButtonColors.ts
```diff
@@ -200,8 +200,10 @@ export const ButtonColors: ButtonColorCollection = {
     disabledBackground: 'transparent',
     disabledBorder: 'transparent',
     disabledText: 'rgba(8, 117, 225, 0.5)',
-    focusBackground: canvas.colors.blueberry100,
+    focusBackground: 'transparent',
     focusText: canvas.colors.blueberry400,
+    focusRingInner: 'currentColor',
```
anicholls: Since this will use the defaults, let's drop `focusRingInner` and `focusRingOuter` here.

modules/button/react/lib/ButtonStyles.ts
jstin: Changing this to `focusRing(2, 2)`, otherwise the inverse will not get the white border. Is that ok?

modules/button/react/lib/ButtonStyles.ts
```diff
@@ -456,7 +456,7 @@ function getButtonFocusRing(buttonType: AllButtonTypes): CSSObject {
     case TextButtonTypes.Inverse:
     case TextButtonTypes.AllCaps:
     case TextButtonTypes.InverseAllCaps:
-      return {};
+      return focusRing(2, 2);
```
jstin: Changing this to `focusRing(2, 2)`, otherwise the inverse will not get the white border. Is that ok?

anicholls: This makes the `Default` and `AllCaps` versions' focus ring 2px too large. Those two types should be moved up to line 452. We needed to delete the `focusRingInner` and `focusRingOuter` in `Default`, but not in `Inverse`. Looks like it's disappeared in both. 

While experimenting with it locally I ended up just making the change so I'll just push my commit 🙂

PR #117, opened 2019-08-14
refactor(Menu): Update CSS styles to match React implementation

jamesfan:
This PR updates the Menu CSS styles so they match the React component. 

### Visual
- Directional modifiers are no longer supported (see below).
- Padding increased
- Font size increased
- Hover color is now gray instead of blue. Focus color is blue (matches our a11y behavior from other components)

### Code
- Removed all references to directional modifiers since they doesn't exist in the Default React Menu. Thus, the `.wdc-menu-top`, `.wdc-menu-right`, `.wdc-menu-bottom`, and `.wdc-menu-left` classes no longer exist.
- Removed Display on hover `.wdc-menu-container` class since it was dependent on the now-removed Directional Modifier classes.

refactor(Menu): Update CSS styles to match React implementation (#117)


modules/menu/css/lib/menu.scss
```diff
@@ -1,15 +1,15 @@
...
+  @include wdc-depth-2();
   display: inline-block;
   background-color: $_wdc-menu-bg-color;
-  padding: $wdc-spacing-s;
   border-radius: 3px;
   border: 1px solid $_wdc-menu-border-color;
   position: relative;
   width: max-content;
-  max-width: $_wdc-menu-max-width;
+  -webkit-font-smoothing: auto;
```
jamesfan: What's the correct way to override font smoothing here? I had to do this in order to get the CSS styles to match the React styles. Currently, I'm overriding styles from the `wdc-type` mixin directly in this Sass file, but I'm not sure that's the right way to go about this.

modules/menu/css/stories.tsx
```diff
@@ -23,41 +17,17 @@ storiesOf('CSS/Menu', module)
...
-            <li>
-              <a href="#">Item 1</a>
-            </li>
-            <li>Item 2</li>
-          </ul>
-        </div>
-        <div className="wdc-menu wdc-menu-bottom">
-          <ul>
+            <li>Second Item with really really really long label</li>
+            <li className="is-disabled">Third Item (disabled)</li>
```
jamesfan: Is the `is-disabled` class properly named?

modules/menu/css/stories.tsx
```diff
@@ -23,41 +17,17 @@ storiesOf('CSS/Menu', module)
...
             <li>
-              <a href="#">Item 1</a>
+              <em>
+                Fourth Item (<b>with markup</b>)
+              </em>
             </li>
-            <li>Item 2</li>
+            <hr />
```
jamesfan: I tried to implement the divider purely using CSS, but I'm not sure it's possible. Or I haven't figured it out yet. In the meantime, I added an `<hr />` tag for the divider (which is how the React Menu implements the divider).

modules/menu/css/README.md
```diff
@@ -39,70 +39,3 @@ Use `.wdc-menu` to create a menu. This can be used in conjunction with any direc
   </ul>
 </div>
 ```
```
jamesfan: I'm guessing I'll need to update the CSS README to mention the `is-disabled` class and how to add a divider?

modules/menu/css/lib/menu.scss
anicholls: I think this may be incorrect in the React implementation then. We generally try to use `*-font-smoothing: subpixel-antialiased;` for everything.

modules/menu/css/stories.tsx
anicholls: We generally prefix our classes. Can we call this `wdc-menu-item-disabled`?

modules/menu/css/stories.tsx
anicholls: I think an HR is fine 👍

anicholls:
Can you also add a `wdc-menu-item-focused` class that uses the blue background just so that's available if they're using JS to control the menu?

jamesfan:
@anicholls Awesome, sounds good! I'll add an example with icons, and I'll address the rest of your feedback as well.

modules/menu/css/stories.tsx
jamesfan: Done.

modules/menu/css/README.md
jamesfan: Done.

Issue #84, opened 2019-08-07
TypeError with conditional Layout Column children


nikolasjchaconas:
## 🐛 Bug Report

TypeError when child of layout returns null

## To Reproduce

```javascript
<Layout gutter={0}>
  {
    conditional &&
    <Layout.Column>
      <div>
        <Component />
      </div>
    </Layout.Column>
  }
  <Layout.Column>
    <Component>
      <h2>
        {conditional ?  ${variable} : ""}
      </h2>
      {
        condtional && <Component />
      }
    </Component>
  </Layout.Column>
  {
    conditional && (
      <Layout.Column>
        {moreCode}
      </Layout.Column>
    )
  }
</Layout>
```

This code is okay:

```javascript
<Layout gutter={0}>
  <>
    {
      conditional &&
      <Layout.Column>
        <div>
          <Component />
        </div>
      </Layout.Column>
    }
    <Layout.Column>
      <Component>
        <h2>
          {conditional ?  ${variable} : ""}
        </h2>
        {
          conditional && <Component />
        }
      </Component>
    </Layout.Column>
    {
      conditional && (
        <Layout.Column>
          {moreCode}
        </Layout.Column>
      )
    }
  </>
</Layout>
```

## Expected Behavior

`<Layout.Column>` should handle null children

## Actual Results

`TypeError` is thrown

## Browser (if applicable)

Chrome

### Error Output

```javascript
TypeError: Cannot read property 'type' of null
Layout._this.renderChild
node_modules/@workday/canvas-kit-react-layout/dist/es6/lib/Layout.js:58
  55 | function Layout() {
  56 |     var _this = _super !== null && _super.apply(this, arguments) || this;
  57 |     _this.renderChild = function (child) {
> 58 |         if (typeof child.type === typeof Column) {
     | ^  59 |             var childProps = child.props;
  60 |             if (childProps.spacing || childProps.spacing === 0) {
  61 |                 return child;
```

anicholls:
Hi @nikolasjchaconas! Thanks for reporting. I have opened a PR to address this issue: https://github.com/Workday/canvas-kit/pull/119

PR #119, opened 2019-08-14
fix(Layout): Add check for null children

anicholls:
## Summary

There was a type error happening when a child of `Layout` returned null. It should be possible to have conditional columns, so this change makes that possible.

Closes https://github.com/Workday/canvas-kit/issues/84

## Checklist

- [x] `yarn test` passes


lychyi:
@anicholls can we generalize this title/commit msg to: `fix(Layout): add null check`?

This fix doesn't really allow or disallow conditional children, it just fixes a bug in our code.

We probably have these elsewhere in our code too where we associate compile-time checking with run-time checking.

lychyi:
Unearthed #125 from this one too. Though, I think these should be two separate fixes. This one should go in ASAP.

Issue #105, opened 2019-08-09
SidePanel API Updates


anicholls:
## 🚀 Feature Proposal

SidePanel needs a breaking change to bring it in line with our [API & Pattern Guidelines](https://github.com/Workday/canvas-kit/blob/master/API_PATTERN_GUIDELINES.md):

- Change `sidePanelBackgroundColor` to `backgroundColor`
- **Not Breaking**: Add `backgroundColor` as a static variable (https://github.com/Workday/canvas-kit/pull/110)
- **Not Breaking:** Change `open` param in `onBreakpointChange` to `aboveBreakpoint`. We shouldn't suggest the behavior - consumers can do what they want with the return value.
- **Bonus**: There are lots of undefined styles being applied based on conditionals in the styled components. Let's fix this 🙂 

Ensure you document all breaking changes that are made in your PR.

Pulled from https://github.com/Workday/canvas-kit/issues/51 and will be a breaking change.

## Motivation

We want our APIs to be as consistent as possible. Let's bring `SidePanel` back in line by making these changes so our APIs feel more natural from one component to the next.

NicholasBoll:
It looks like @anicholls is working on this in #123

PR #123, opened 2019-08-14
refactor(SidePanel): API update and fixes

anicholls:
## Summary

SidePanel needs a changes to bring it in line with our [API & Pattern Guidelines](https://github.com/Workday/canvas-kit/blob/master/API_PATTERN_GUIDELINES.md):

- Change `sidePanelBackgroundColor` to `backgroundColor`
- **Not Breaking:** Change `open` param in `onBreakpointChange` to `aboveBreakpoint`. We shouldn't suggest the behavior - consumers can do what they want with the return value.
- **Bonus**: There were lots of undefined styles being applied based on conditionals in the styled components. Note: use `undefined` instead of `''` in emotion ternaries.

Closes https://github.com/Workday/canvas-kit/issues/105.

| Breaking Changes | No   | Yes   |

### Code
- `sidePanelBackgroundColor` prop has been renamed to `backgroundColor`.

## Checklist

- [x] `yarn test` passes
      applicable


modules/side-panel/react/lib/SidePanel.tsx
```diff
@@ -52,13 +52,13 @@ const SidePanelContainer = styled('div')<SidePanelProps>(
     position: 'absolute',
   },
   ({open}) => ({
-    alignItems: open ? '' : 'center',
-    boxShadow: !open ? '0 8px 16px -8px rgba(0, 0, 0, 0.16)' : '',
+    alignItems: !open ? 'center' : undefined,
```
stephanerangaya: Don't have a strong opinion on this, but I wonder if it is not better to use `alignItems: open ? undefined : 'center',` instead of `!open`.

modules/side-panel/react/lib/SidePanel.tsx
```diff
@@ -166,10 +166,9 @@ export default class SidePanel extends React.Component<SidePanelProps, SidePanel
         role="region"
         aria-orientation="vertical"
         padding={padding}
-        onBreakpointChange={this.handleResize}
```
stephanerangaya: Why is this appearing as a delete (it is present in the file)

anicholls: `handleResize` is getting enabled in `componentDidMount`. It wasn't used at all in `SidePanelContainer` so this line wasn't doing anything.

Issue #97, opened 2019-08-09
Button API Changes


anicholls:
## 🚀 Feature Proposal

Button needs several breaking changes to bring it in line with our [API & Pattern Guidelines](https://github.com/Workday/canvas-kit/blob/master/API_PATTERN_GUIDELINES.md):

- Rename `buttonType` to `type`
- Rename `buttonSize` to `size`
- Rename all types to singular
- Rename all static variables to singular
- Re-evaluate prop spread behavior (let's discuss further)
- Make aria-label required for IconButton
- Consider scaling down `Small` `SquareFilled` Icon Button icon to match the other `Small` icon
- Fix the focus outline on square buttons to match others (it's the only one that doesn't have the white ring inside the blue) (see https://github.com/Workday/canvas-kit/pull/53#discussion_r312263943).

Ensure you document all breaking changes that are made in your PR.

Pulled from https://github.com/Workday/canvas-kit/issues/51 and will be a breaking change.

## Motivation

We want our APIs to be as consistent as possible and `Button` is currently one of the most divergent components. Let's bring it back in line by making these changes so our APIs feel more natural from one component to the next.

lychyi:
> Rename buttonType to type

Unfortunately, `type` is a reserved keyword in TS. Can we use `variant` instead?

lychyi:
> Re-evaluate prop spread behavior (let's discuss further)

@anicholls what were the discussion points around this?

anicholls:
@lychyi It depends on how it's used (i.e. `const type = 'a'` works just fine). I don't believe the key for an object vs. the name of a variable would be any different, so I don't think it's reserved in this context. Will have to look into it though.

> Re-evaluate prop spread behavior (let's discuss further)
> @anicholls what were the discussion points around this?

Remember when we ran into an issue a few months back because `elemProps` was being spread on the container, the label icon and the label data? I don't remember exactly what it was, but I know it's caused us problems and it's definitely non-standard. Just something we may want to reconsider.

anicholls:
@lychyi So after experimenting with it, there's no conflict with the TS `type` connotation. However, we do have  a conflict with the `React.ButtonHTMLAttributes<HTMLButtonElement>` type member (i.e. `<button type="reset">`. We should still support this, so I think we will have to go with `variant` for now.

Issue #51, opened 2019-07-29
Canvas Kit React API Change Proposal


anicholls:
As part of the 3.0 Alpha, we are making some breaking changes to solidify our APIs, particularly in accordance with our [API & Pattern Guidelines](https://github.com/Workday/canvas-kit/blob/master/API_PATTERN_GUIDELINES.md). 

The following are proposed API changes. This should probably move to our RFC repo once a public one is created.

# API Changes

## Universal Changes
- (https://github.com/Workday/canvas-kit/issues/95 / https://github.com/Workday/canvas-kit/pull/150) ~~**Not Breaking**: Rename `props` (ActionBar), `inputProps` (TextArea, TextInput) and `otherProps` to `elemProps`~~

- ~~**Maybe Breaking** (https://github.com/Workday/canvas-kit/issues/162) Change `HTMLAttributes` usage to `HTMLProps`~~ @lychyi 

- ~~Add a beta/sandbox/unstable module~~ @anicholls

### Avatar

- (https://github.com/Workday/canvas-kit/pull/194) ~~Change `themeColor` to `variant`~~ @stephanerangaya 

### Banner 

- (https://github.com/Workday/canvas-kit/issues/96 / https://github.com/Workday/canvas-kit/pull/110) ~~**Not Breaking**: Export enums as static variables~~
- (https://github.com/Workday/canvas-kit/pull/195) ~~Change `theme` to `variant`.~~ @stephanerangaya 

### ~~Button~~

- (https://github.com/Workday/canvas-kit/issues/97 / https://github.com/Workday/canvas-kit/pull/129) ~~Rename `buttonType` to `type`~~
- (https://github.com/Workday/canvas-kit/issues/97 / https://github.com/Workday/canvas-kit/pull/129) ~~Rename `buttonSize` to `size`~~
- (https://github.com/Workday/canvas-kit/issues/97 / https://github.com/Workday/canvas-kit/pull/129) ~~Rename all types to singular~~
- (https://github.com/Workday/canvas-kit/issues/97 / https://github.com/Workday/canvas-kit/pull/129) ~~Rename all static variables to singular~~
- (https://github.com/Workday/canvas-kit/issues/97 / https://github.com/Workday/canvas-kit/pull/129) ~~Re-evaluate prop spread behavior~~
- (https://github.com/Workday/canvas-kit/issues/97 / https://github.com/Workday/canvas-kit/pull/129) ~~Make aria-label required for IconButton~~
[...]

PR #129, opened 2019-08-16
refactor(Button): API Updates

anicholls:
- [x] Rename `buttonType` to `variant`
- [x] Rename `buttonSize` to `size`
- [x] Rename all types to singular
- [x] Rename all static variables to singular
- [x] Properly destructure and spread props
- [x] Remove nested prop spreading
- [x] Make aria-label required for IconButton
- [x] Scale down `Small` `SquareFilled` Icon Button icon to match the other `Small` icon
- [x] Fix the focus outline on square buttons to match others (it's the only one that doesn't have the white ring inside the blue) (see https://github.com/Workday/canvas-kit/pull/53#discussion_r312263943).
- [ ] Change defaultProps to required in interface. **Note:** _This can't be done until we upgrade to Emotion 10 (see https://github.com/emotion-js/emotion/issues/654). We have this issue in several places, so we will do a sweeping change when we upgrade that dep.

Closes https://github.com/Workday/canvas-kit/issues/97

### Code changes
- Rename `buttonType` to `variant`
- Rename `buttonSize` to `size`
- Rename `ButtonSizes` enum to `ButtonSize`
- Rename `ButtonTypes` enum to `ButtonVariant`
- Rename `BetaButtonTypes` enum to `BetaButtonVariant`
- Rename `IconButtonTypes` enum to `IconButtonVariant`
- Rename `IconPositions` enum to `IconPosition
- Rename `TextButtonTypes` enum to `TextButtonVariant`
- Rename `AllButtonTypes` type to `AllButtonVariants`
- Rename `Button.Types` static variable to `Button.Variant` for all buttons (`Button`, `beta_Button`, `DropdownButton`, `IconButton`, `TextButton`)
- Rename `Button.Sizes` static variable to `Button.Size` (`Button`, `beta_Button`, `DropdownButton`, `IconButton`, `TextButton`)
- Rename `TextButton.IconPositions` to `TextButton.IconPosition`
- `IconButton.Size` static variable and `IconButton` `size` prop now use `IconButtonSize` enum instead of `ButtonSize` (Because `ButtonSize.Large` isn't supported) 
- Extra props are no longer spread on child elements. They are only spread on the outer `<button>` element.
- `aria-label` is now a required prop for `IconButton` for accessibility reasons.
- The icon for `<IconButton variant={IconButton.Variant.SquareFilled} size={IconButton.Size.Small} />` has been scaled down to match the icon in other small icon buttons

## Checklist

- [x] `yarn test` passes
      applicable

### DIFF
--- a/modules/action-bar/react/spec/__snapshots__/ActionBar.snapshot.tsx.snap
+++ b/modules/action-bar/react/spec/__snapshots__/ActionBar.snapshot.tsx.snap
@@ -164,9 +164,11 @@ exports[`Action Bar Snapshots renders a action bar with one button 1`] = `
   >
     <button
       className="emotion-2 emotion-3"
+      size="large"
     >
       <span
         className="emotion-0 emotion-1"
+        size="large"
       >
         Button
       </span>
@@ -424,18 +426,22 @@ exports[`Action Bar Snapshots renders a action bar with two buttons 1`] = `
   >
     <button
       className="emotion-2 emotion-3"
+      size="large"
     >
       <span
         className="emotion-0 emotion-1"
+        size="large"
       >
         Button
       </span>
     </button>
     <button
       className="emotion-6 emotion-3"
+      size="large"
     >
       <span
         className="emotion-0 emotion-1"
+        size="large"
       >
         Button
       </span>
@@ -612,9 +618,11 @@ exports[`Action Bar Snapshots renders a fixed action bar with one button 1`] = `
   >
     <button
       className="emotion-2 emotion-3"
+      size="large"
     >
       <span
         className="emotion-0 emotion-1"
+        size="large"
       >
         Button
       </span>
@@ -876,18 +884,22 @@ exports[`Action Bar Snapshots renders a fixed action bar with two buttons 1`] = 
   >
     <button
       className="emotion-2 emotion-3"
+      size="large"
     >
       <span
         className="emotion-0 emotion-1"
+        size="large"
       >
         Button
       </span>
     </button>
     <button
       className="emotion-6 emotion-3"
+      size="large"
     >
       <span
         className="emotion-0 emotion-1"
+        size="large"
       >
         Button
       </span>
--- a/modules/action-bar/react/spec/__snapshots__/ActionBar.snapshot.tsx.snap
+++ b/modules/action-bar/react/spec/__snapshots__/ActionBar.snapshot.tsx.snap
@@ -164,7 +164,6 @@ exports[`Action Bar Snapshots renders a action bar with one button 1`] = `
   >
     <button
       className="emotion-2 emotion-3"
-      size="large"
     >
       <span
         className="emotion-0 emotion-1"
@@ -426,7 +425,6 @@ exports[`Action Bar Snapshots renders a action bar with two buttons 1`] = `
   >
     <button
       className="emotion-2 emotion-3"
-      size="large"
     >
       <span
         className="emotion-0 emotion-1"
@@ -437,7 +435,6 @@ exports[`Action Bar Snapshots renders a action bar with two buttons 1`] = `
     </button>
     <button
       className="emotion-6 emotion-3"
-      size="large"
     >
       <span
         className="emotion-0 emotion-1"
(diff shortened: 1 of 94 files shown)


modules/button/react/lib/types.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * The different button sizes.
  */
-export enum ButtonSizes {
+export enum ButtonSize {
```
NicholasBoll: Should we convert these to object exports instead of enums and update the the API guidelines to ban the use of enums? (mentioned in https://github.com/Workday/canvas-kit/issues/51)

The reasoning is enums are Typescript-only. The enum is completely compiled away during the Typescript process (requires multi-file compilation) and are not supported in Javascript or the `@babel/plugin-transform-typescript` plugin.

This PR would be a good place to do that.

modules/action-bar/react/spec/__snapshots__/ActionBar.snapshot.tsx.snap
```diff
@@ -164,9 +164,11 @@ exports[`Action Bar Snapshots renders a action bar with one button 1`] = `
   >
     <button
       className="emotion-2 emotion-3"
+      size="large"
```
NicholasBoll: I wonder if this was an intended change. `size` isn't a valid attribute for `button`

anicholls: We have this issue with some of our other components too (e.g. `Avatar`). I'm not aware of a way to omit props we want for styling from the resulting dom attributes that emotion generates. Do you have any ideas?

anicholls: nvm, found it 🙂 https://emotion.sh/docs/styled#customizing-prop-forwarding

modules/button/react/lib/Button.tsx
```diff
@@ -41,27 +41,27 @@ export interface ButtonProps<T = ButtonTypes | BetaButtonTypes>
...
   public render() {
-    const {buttonRef, ...elemProps} = this.props;
+    const {variant, size, buttonRef, dataLabel, icon, children, ...elemProps} = this.props;
 
     return (
-      <ButtonBaseCon {...elemProps} innerRef={buttonRef}>
-        {elemProps.icon && <ButtonLabelIcon {...elemProps} />}
-        <ButtonBaseLabel buttonSize={elemProps.buttonSize} buttonType={elemProps.buttonType}>
-          {elemProps.children}
+      <ButtonBaseCon variant={variant} size={size} innerRef={buttonRef} {...elemProps}>
```
NicholasBoll: By default, a `<button>` has a `type` of "submit". Most developers I've encountered assume the default type is "button" which has no special connotation inside of a `<form>` element. Normally this might not matter, but in IE11 if there is no `<form>` element, the document will be treated like a form. At my last place, we set the default `type="button"` to make it so that submit and reset buttons were explicit changes to how a button functions inside a form.

What is our stance on how this should work. The difference is subtle, but profound if type is omitted and "button" is assumed default:

```tsx
<form onSubmit={onSubmit}>
  {/* oops. This is the first "submit" type. If a user hits the enter key in an input, this will be "clicked" and both `onCancel` AND `onSubmit` will be executed. */}
  <Button onClick={onCancel}>Cancel</Button>
  <Button type='submit' variant={Button.Variant.Primary}>Submit</Button>
</form>
```

It is probably fine to fallback to the default in how HTML works, but this is a weird gotcha from HTML's colorful history.

anicholls: Interesting! I've always assumed `type="submit"` by default. I'm not against adding `type="button"`, but it does feel a little weird to override the HTML default

lychyi: Technically React does this with `onChange`. I'm on the fence about this too. There are drawbacks either way. The most notable one I can think of is that if we want to reverse this behavior it'll be a fairly large (and impactful) breaking change. Something to chew on before we decide.

lychyi: Though... if we do end up with a `<SubmitButton>` then yes, we should be defaulting to `type=button`... again just has to be well-documented. Maybe IN ALL CAPS or something ;)

anicholls: So is the consensus to think on this a bit more? I don't see it as a huge blocker for this PR.

Issue #104, opened 2019-08-09
Change marketing prop in PageHeader


anicholls:
## 🚀 Feature Proposal

Rename `marketing` prop in `PageHeader` to `capWidth` (matches `Layout`)

This is from https://github.com/Workday/canvas-kit/issues/51 and is a breaking change.

## Motivation

We are dropping all notions of the "Marketing" type of components. Workday Canvas is for all contexts are we are trying to remove the differentiation between product and non-product components. Previously we had used the term `marketing` to denote non product patterns since they were usually used in marketing sites. We would like to move away from this and be more descriptive about what's actually being affected (in this case, "capping" the width with a max-width).

PR #143, opened 2019-08-21
refactor(PageHeader): Update marketing and breakpoint props

sahlhoff:
# Summary

Renames `marketing` prop in `PageHeader` to `capWidth` (matches `Layout`)

This renames the `marketing` prop in `PageHeader` and simplifies it's `breakpoint` prop.

### Visual:
- Middle padding breakpoints have been removed

### Code:
- `marketing` prop changed to `capWidth`
- `breakpoint` prop changed from an object with three points to a single number.

Closes https://github.com/Workday/canvas-kit/issues/104. 

This is from https://github.com/Workday/canvas-kit/issues/51 and is a breaking change.

## Motivation

We are dropping all notions of the "Marketing" type of components. Workday Canvas is for all contexts are we are trying to remove the differentiation between product and non-product components. Previously we had used the term `marketing` to denote non product patterns since they were usually used in marketing sites. We would like to move away from this and be more descriptive about what's actually being affected (in this case, "capping" the width with a max-width).

## Checklist

- [x] `yarn test` passes
      applicable


modules/page-header/react/lib/PageHeader.tsx
```diff
@@ -31,10 +31,10 @@ const Container = styled('div')<PageHeaderProps>(
     overflow: 'hidden',
     padding: `0 ${spacing.s}`,
   },
-  ({marketing, breakpoints}) => {
+  ({capWidth, breakpoints}) => {
```
anicholls: I forgot that the marketing version had different padding at larger sizes than the normal version. This doesn't make sense when we change it to `capWidth` and I don't believe it's necessary anymore.

Can we change the `breakpoints` prop to `breakpoint?: number` instead and just have one media query (regardless of `capWidth`)?

sahlhoff: @anicholls should be good for another look– we're using a single breakpoint now that defaults to `768`.

PR #147, opened 2019-08-21
Css update menu icons

jamesfan:
This PR adds icon and grow support to the Menu CSS styles. 

| Breaking Changes | No    | No   |

fix(Menu): Add Menu example with icons and grow support (#147)

### DIFF
--- a/modules/menu/css/README.md
+++ b/modules/menu/css/README.md
@@ -40,6 +40,20 @@ Use `.wdc-menu` to create a menu.
 </div>
 ```
 
+### Variants
+
+#### Grow
+
+Use `.wdc-menu-grow` to expand the menu to fit its container.
+
+```html
+<div className="wdc-menu wdc-menu-grow">
+  <ul>
+    <li>This menu will expand to fit its container</li>
+  </ul>
+</div>
+```
+
 ### Menu Item Modifiers
 
 Use `.wdc-menu-item-disabled` to create a disabled menu item.
@@ -78,7 +92,38 @@ Menu item modifiers may be combined.
 </div>
 ```
 
-## Divider
+### Icons
+
+Menu items can include an icon as well as a secondary icon (if necessary).
+
+Wrap the text for the menu item using a `<span>` with `.wdc-menu-item-label`, and wrap the icon
+using a `<span>` with `.wdc-menu-item-icon`.
+
+The `<span>` for the secondary icon will need `.wdc-menu-item-icon-secondary` as well.
+
+```html
+<div class="wdc-menu">
+  <ul>
+    <li>
+      <span className="wdc-menu-item-icon">
+        <svg ...>
+      </span>
+      <span className="wdc-menu-item-label">Item with icon</span>
+    </li>
+    <li>
+      <span className="wdc-menu-item-icon">
+        <svg ...>
+      </span>
+      <span className="wdc-menu-item-label">Item with icon and secondary icon</span>
+      <span className="wdc-menu-item-icon wdc-menu-item-icon-secondary">
+        <svg ...>
+      </span>
+    </li>
+  </ul>
+</div>
+```
+
+### Divider
 
 Use `<hr />` to create a divider between menu items.
 
--- a/modules/menu/css/lib/menu.scss
+++ b/modules/menu/css/lib/menu.scss
@@ -8,8 +8,6 @@
   border: 1px solid $_wdc-menu-border-color;
   position: relative;
   width: max-content;
-  -webkit-font-smoothing: auto;
-  -moz-osx-font-smoothing: auto;
 
   // Fallback for IE11
   @media screen and (-ms-high-contrast: active),
--- a/modules/menu/css/lib/menu.scss
+++ b/modules/menu/css/lib/menu.scss
@@ -6,6 +6,7 @@
   background-color: $_wdc-menu-bg-color;
   border-radius: 3px;
   border: 1px solid $_wdc-menu-border-color;
+  box-sizing: border-box;
   position: relative;
   width: max-content;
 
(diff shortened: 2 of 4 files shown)


modules/menu/css/README.md
```diff
@@ -40,6 +40,20 @@ Use `.wdc-menu` to create a menu.
 </div>
 ```
 
+### Variants
+
+#### Grow
```
jamesfan: I snuck in support for the grow prop in the React Menu since it was missing from the previous refactor(Menu) PR. Let me know if this should be handled as a separate PR.

modules/menu/css/README.md
```diff
@@ -78,7 +92,38 @@ Menu item modifiers may be combined.
 </div>
 ```
 
-## Divider
+### Icons
```
jamesfan: I had to add some extra markup and classes to replicate the way icons are displayed with the React Menu. I copied the markup structure rendered by the React Menu with separate spans used for the icon, text, and secondary icon (each span is sized using `flex`).

I've been trying to find a way to do this with less markup, but I haven't been successful so far. What do you think? Is this ok or should I keep digging?

modules/menu/css/stories.tsx
```diff
@@ -1,22 +1,20 @@
 import React from 'react';
 import {storiesOf} from '@storybook/react';
 import withReadme from 'storybook-readme/with-readme';
+import {SystemIcon} from '@workday/canvas-kit-react-icon';
+import {extLinkIcon, setupIcon, uploadCloudIcon, userIcon} from '@workday/canvas-system-icons-web';
 import README from './README.md';
 import {css} from 'emotion';
```
anicholls: Nit: This is unused

modules/menu/css/README.md
anicholls: So unfortunately with the text overflow stuff, the label needs a container here unless we wanted to remove `display: flex` from the container `li` (which is even worse IMO).

Rather than using a wrapper for the icons, I would just add a class to them (i.e. `<SystemIcon icon={uploadCloudIcon} className="wdc-menu-item-icon" />`)

modules/menu/css/README.md
anicholls: Good catch! Fine to include it here.

modules/menu/css/README.md
jamesfan: Using CSS Icon instead of SystemIcon

modules/menu/css/stories.tsx
```diff
@@ -1,22 +1,71 @@
-import React from 'react';
+import React, {Component} from 'react';
 import {storiesOf} from '@storybook/react';
 import withReadme from 'storybook-readme/with-readme';
 import README from './README.md';
-import {css} from 'emotion';
 import './index.scss';
+// @ts-ignore
+import initializeIcons from '../../icon/css/lib/canvas-kit-css-icon';
```
jamesfan: I had to add a `@ts-ignore` directive here to prevent a `TS7016: Could not find a declaration file for module...` error (copied this from the CSS Icon stories). I hope this is ok?

modules/menu/css/stories.tsx
```diff
@@ -1,22 +1,71 @@
...
+  }
+
+  render() {
+    return (
+      <div className="wdc-type">
+        <div className="wdc-menu" style={{width: '250px'}}>
+          <ul>
+            <li className="wdc-menu-item-focused">
+              <i
+                className="wdc-icon wdc-menu-item-icon"
```
jamesfan: Following up on our slack convo earlier: technically, I could remove the `wdc-menu-item-icon` class and just target `wdc-icon`. Or do you want me to leave it there in case we can't safely assume the icon will have a `wdc-icon`class?

modules/menu/css/lib/menu.scss
```diff
@@ -40,32 +43,72 @@
...
+      }
+
+      a {
+        color: inherit;
+        text-decoration: none;
+      }
+
+      .wd-icon-fill {
+        fill: $wdc-color-icon-standard;
+      }
```
jamesfan: Is there a code style guideline I should follow when it comes to nesting? Right now, I have:

```
.wdc-menu {
  ul {
    li {
      .wd-icon-fill {
        ...
      }

      .wdc-icon,
      .wdc-menu-item-icon {
         ...
      }
    }
  }
}
```

I could reduce the amount of nesting and do this instead:

```
.wdc-menu {
  ul {
    li {
      ...
    }
  }

  .wd-icon-fill {
    ...
  }

  .wdc-icon,
  .wdc-menu-item-icon {
     ...
  }
}
```

In general, I prefer less nesting (even though that's not what I've currently done, whoops) since I find it easier to read/debug, but I'm happy to follow whatever guidelines we've set for the project.

modules/menu/css/stories.tsx
jamesfan: Sounds good, I've updated the README accordingly.

modules/menu/css/lib/menu.scss
jamesfan: Got it, I'll leave this as is then.

modules/menu/css/README.md
```diff
@@ -78,7 +92,32 @@ Menu item modifiers may be combined.
...
+```html
+<div class="wdc-menu">
+  <ul>
+    <li>
+      <i className="wdc-icon wdc-menu-item-icon" data-icon="uploadCloud" data-category="system" />
+      <span className="wdc-menu-item-label">Item with uploadCloud icon</span>
+    </li>
+    <li>
+      <i className="wdc-icon wdc-menu-item-icon" data-icon="uploadCloud" data-category="system" />
+      <span className="wdc-menu-item-label"
```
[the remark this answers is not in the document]
jamesfan: Fixed.

jamesfan:
@anicholls I think I've addressed all of your feedback. Let me know if find any other issues -- otherwise, I think this is good to go.

PR #155, opened 2019-08-28
refactor(popup): Refactor CSS to match React implementation

mannycarrera4:
This PR refactors the popup css to match the react implementation

### Code Changes
- Popup now extends styles from `card.scss` for basic styles similar to how react implementation does.
- **Removed** arrow modifiers and styles for arrows (`.wdc-popup-left/right/top/bottom`)
- **Removed** Ability to show on hover (`.wdc-popup-container`). Hover rendering should only be used for tooltips and Popups should always be triggered by a click or keyboard event.
- **Removed** `lib/variables.scss`


mannycarrera4:
Looking for some feedback regarding the animation and how we should handle position especially for the story?

modules/popup/css/lib/popup.scss
```diff
@@ -1,203 +1,120 @@
...
-
-  // Fallback for IE11
-  @media screen and (-ms-high-contrast: active),
-  screen and (-ms-high-contrast: none) {
-    width: auto;
-    min-width: $_wdc-popup-menu-min-width;
+@keyframes popupAnimationTopCenter {
+  from {
+    opacity: 0;
+    transform: translate(0, -50%);
```
anicholls: We should only be translating `spacing.xxs` (https://github.com/Workday/canvas-kit/blob/master/modules/popup/react/lib/Popup.tsx#L31)

modules/popup/css/stories.tsx
```diff
@@ -2,47 +2,76 @@ import React from 'react';
...
+                Cancel
+              </Button>
+            </div>
+          ) : null}
+        </div>
+      </div>
+    );
+  }
+
+  private handleClose = () => {
```
anicholls: Not used

modules/popup/css/stories.tsx
```diff
@@ -2,47 +2,76 @@ import React from 'react';
...
+    this.setState({
+      open: false,
+    });
+  };
+
+  private handleSubmit = () => {
```
anicholls: Can you rename these?

modules/popup/css/stories.tsx
```diff
@@ -2,47 +2,76 @@ import React from 'react';
...
+          display: 'flex',
+          justifyContent: 'center',
+        }}
+      >
+        <div className="wdc-popup-container">
+          <Button buttonType={Button.Types.Delete} onClick={this.handleClick}>
+            Delete Item
+          </Button>
+          {open ? (
+            <div className="wdc-popup wdc-popup-animation-origin-top-center wdc-depth-2">
```
anicholls: Missing a close button

modules/popup/css/lib/popup.scss
```diff
@@ -1,203 +1,120 @@
...
+  }
+}
 
-    &:before {
-      margin-left: -($_wdc-popup-arrow-size + 1px);
-      border-top-color: $_wdc-popup-border-color;
-    }
+.wdc-popup {
+  position: relative;
+  background-color: $_wdc-popup-bg-color;
```
anicholls: Can you update this to depend on card like the react version does? It would be nice if we didn't have the card-like styling here duplicated. (i.e. `@include card-common()`)

modules/popup/css/lib/popup.scss
```diff
@@ -1,203 +1,120 @@
...
-      left: $_wdc-popup-arrow-offset;
-    }
+  to {
+    opacity: 1;
+    transform: translate(0);
+  }
+}
 
-    &:before {
-      margin-left: -($_wdc-popup-arrow-size + 1px);
```
anicholls: Can you remove the variables like these that we are no longer using?

modules/popup/css/lib/popup.scss
```diff
@@ -1,203 +1,120 @@
...
-  position: relative;
-  width: max-content;
-  max-width: $_wdc-popup-max-width;
-
-  // Fallback for IE11
-  @media screen and (-ms-high-contrast: active),
-  screen and (-ms-high-contrast: none) {
-    width: auto;
-    min-width: $_wdc-popup-menu-min-width;
+@keyframes popupAnimationTopCenter {
```
anicholls: This is the right approach, but we will also need motion for TopLeft, TopRight, BottomLeft, BottomRight. You could look at an approach like cards and grids to programmatically generate these

modules/popup/css/lib/popup.scss
```diff
@@ -1,203 +1,120 @@
...
+  &-animation-origin-left-center {
+    animation-name: popupAnimationLeftCenter;
+    transform-origin: left center;
   }
 
-  p {
-    margin-top: 0;
-    margin-bottom: 0;
+  // Fallback for IE11
+  @media screen and (-ms-high-contrast: active), screen and (-ms-high-contrast: none) {
```
anicholls: We don't use `width: max-content` in card, which means we'll be able to drop these.

modules/button/css/lib/icon-button.scss
```diff
@@ -117,6 +117,7 @@
 
     &-plain {
       background-color: $wdc-btn-icon-bg;
+      margin: -8px;
```
mannycarrera4: to match the react version

modules/popup/css/lib/popup.scss
anicholls: Do you mind moving the animation stuff to another file as well?

mannycarrera4: sure thing!

PR #157, opened 2019-08-29
WIP Combobox in progress

vibdev:
## Checklist

- [ ] `yarn test` passes
      applicable


modules/text-input/react/lib/Combobox.tsx
```diff
@@ -0,0 +1,265 @@
...
+    const listBoxIdPart = `listbox`;
+    const labelIdPart = `label`;
+    const optionIdPart = `option`;
+
+    return (
+      <Container
+        grow={grow}
+        role='combobox'
+        aria-haspopup='listbox'
+        aria-owns={`${accessibleId}-${listBoxIdPart}`}
```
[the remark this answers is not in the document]
NicholasBoll: Yeah. I assume only one combobox could be opened at a time. Just wanted to make sure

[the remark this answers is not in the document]
vibdev: Since they close on blur only one should ever show at a time

modules/text-input/react/lib/TextInput.tsx
```diff
@@ -72,10 +72,36 @@ export default class TextInput extends React.Component<TextInputProps> {
...
+    // We also can not just update .value= here because React library overrides input value setter
+    // but we can call the function directly on the input as context.
+    // This will onChange events to fire no matter how value is updated.
+    if (this.props.inputRef && this.props.inputRef.current) {
+      const nativeInputValue = Object.getOwnPropertyDescriptor(Object.getPrototypeOf(this.props.inputRef.current), 'value')
+      if (nativeInputValue && nativeInputValue.set) {
+        nativeInputValue.set.call(this.props.inputRef.current, this.props.value);
+      }
+      const event = new Event('input', { bubbles: true });
+      this.props.inputRef.current.dispatchEvent(event);
```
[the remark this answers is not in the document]
NicholasBoll: I don't mean passing a `value` prop without an `onChange`, I mean forcing an `onChange` to be called when the parent component changes the `value` prop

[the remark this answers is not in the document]
vibdev: Gotcha. So far in my testing everything has been working correctly. 
The main use case for this is being able to clear the input field via a button or keyboard shortcut and let any top level listeners know something changed. Without this any listeners you are using to set the autocomplete menu items won't know the field was changed.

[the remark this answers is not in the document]
vibdev: One more use case is to change the value of the input to the text of the autocomplete item on click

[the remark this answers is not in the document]
NicholasBoll: I've solve those problems in the past by invoking the `onChange` from those user actions explicitly. This guarantees event handlers are only called due to events and not data changes. Otherwise some use-cases require special state tracking to tell the difference between handlers called from user events vs data changes.

Sometimes we violate this for easier data fetching like fetching data on mount of a component. A component mount isn't necessarily tied to a user event.

Always forcing events on any data change is probably easier since you don't have the explicitly call `onChange` for the times when you change the value, but it prevents accidental firings on `onChange` when it was not intended. The bugs can be subtle. It might manifest itself more when side-effects happen inside an `onChange` like fetching data from a server. You may find `onChange` being fired with data that is never actually rendered.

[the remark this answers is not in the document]
vibdev: Ok I will move this code into the combobox and explicitly call it when programmatically changing the data

modules/text-input/react/lib/Combobox.tsx
```diff
@@ -0,0 +1,265 @@
...
+            onChange={this.handleSearchInputChange}
+            onKeyDown={this.handleKeyboardShortcuts}
+            onFocus={this.setFocused}
+            onBlur={this.setFocused}
+          />
+        </FormField>
+        {
+          this.state.showingAutocomplete && (
+            <MenuContainer padding={spacing.zero} depth={depth[1]}>
+              <Autocomplete
```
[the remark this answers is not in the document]
NicholasBoll: That's more of an overall design decision and not something addressed in this PR. It was just something I noticed that spurred the question with a specific example.

I don't think 2 lines vs 1 line change in a diff if we want to switch and element from a span to a div matters much.

Code is read more often than it is written and by people that don't share the same context as the developer who wrote it (when they wrote it). The abstraction is offering very little value and just requires cognitive overhead of another (and often many) lookups. Also if I'm debugging components, that abstraction is harmful. This is what it looks like in my React dev tools:

Screen Shot 2019-08-29 at 4 18 00 PM

I now have to manually map "div `Styled`" to a randomly chosen component name.

I'm not a huge fan of the Emotion 10 API requiring a JSX pragma change to add `css=`, but it is very readable. Using the `styled` API, I have a hard time telling if `Autocomplete` is a Component or an element with styling applied. I have to jump sometimes hundreds of lines around a file to understand a render function. Doable? Sure. It just requires me to understand multiple abstraction levels simultaneously which is an indicator of a bad abstraction. Maybe I just haven't drank enough of the styled-components kool-aid yet.

I wouldn't necessarily change anything in this PR except maybe the name of the variable holding the styled element (`AutocompleteList`?). We'll bring it up with the team. I've found sometimes decisions are made arbitrarily and others with a specific set of requirements - it just takes digging.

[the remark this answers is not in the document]
vibdev: Totally understandable. Will change to AutocompleteList. 
Curious to hear where y'all will land on this.
I wonder if it could be something as simple as prepending styled components with the word "Styled".
e.g.
`StyledAutocompleteList`, that would communicate that it is an emotion component and wouldn't risk a mismatch like `StyledUl` ending up pointing to a `dl`

modules/text-input/react/lib/TextInput.tsx
```diff
@@ -9,18 +9,11 @@ import InputIconContainer from './InputIconContainer';
...
-  disabled?: boolean;
   error?: ErrorType;
-  hasIcon?: boolean;
-  inputRef?: React.Ref<HTMLInputElement>;
-  onChange?: React.ChangeEventHandler<HTMLInputElement>;
-  placeholder?: string;
-  readOnly?: boolean;
-  type?: string;
-  value?: any;
+  inputRef?: React.RefObject<HTMLInputElement>;
```
vibdev: since these were all normal props that are just being spread on the input we don't need to call them out. Also looks like hasIcon was only included for use in the type below. It's value is computed using the error prop.

modules/form-field/react/lib/FormField.tsx
```diff
@@ -79,22 +80,19 @@ export default class FormField extends React.Component<FormFieldProps> {
...
-  private renderChildren = (child: React.ReactChild): React.ReactNode => {
+  private renderChildren = (child: React.ReactElement): React.ReactNode => {
     if (React.isValidElement<any>(child)) {
       const props: GrowthBehavior &
         FormFieldErrorBehavior &
         React.HTMLAttributes<HTMLInputElement> = {
         ...child.props,
       };
 
-      if (this.props.grow && React.isValidElement<GrowthBehavior>(child)) {
```
vibdev: I don't think this check was doing anything since we are already wrapped in a React.isValidElement<any>(child) check

modules/form-field/react/lib/FormField.tsx
```diff
@@ -106,7 +104,7 @@ export default class FormField extends React.Component<FormFieldProps> {
         }
       }
 
-      if (!child.props.id && this.props.inputId) {
+      if (this.props.inputId) {
```
vibdev: The way this was setup before you could have a label and its input have mismatch ids
```
<FormElement inputId='foo'><TextInput id='bar' /></FormElement>
```
=>
```
<label for='foo' />
<input id='bar />
```

modules/menu/react/lib/Menu.tsx
```diff
@@ -239,7 +239,7 @@ export default class Menu extends React.Component<MenuProps, MenuState> {
   private setFirstCharacters = (): void => {
     const getFirstCharacter = (child: React.ReactNode): string => {
       let character = '';
-      if (!child || typeof child === 'boolean' || child === {}) {
+      if (child == null || typeof child === 'boolean' || child === {}) {
```
vibdev: this was failing on 0

modules/text-input/react/lib/Combobox.tsx
```diff
@@ -0,0 +1,265 @@
+
```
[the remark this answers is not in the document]
vibdev: I moved to form-field and made it so the combobox expects a formField with a TextInput to be passed in

modules/form-field/react/stories/stories_TextInput.tsx
```diff
@@ -1,16 +1,52 @@
...
+        autocompleteItems={
+          Array.apply(null, Array(this.state.currentText.length))
+            .map((x: any, i: string) => autocompleteResult(i))
+            .splice(0, 5)
+        }
+        onChange={this.autocompleteCallback}
+      >
+        <FormField label="Autocomplete example">
+          <TextInput placeholder="Autocomplete" />
+        </FormField>
```
vibdev: What does controlComponent do?

modules/text-input/react/lib/TextInput.tsx
anicholls: It's been conventional for us to explicitly call out these normal props that are used for functionality within the component (we do the same on checkbox, radio, etc. as well). If we want to change this, I don't think it should happen in this PR

Good catch on `hasIcon` 👍

modules/form-field/react/lib/Combobox.tsx
```diff
@@ -0,0 +1,317 @@
...
+    return React.cloneElement(formElement, {...formElement.props, ...newFormElementProps});
+  }
+
+  render() {
+    const {
+      accessibleId,
+      autocompleteItems,
+      children,
+      grow,
+      onChange,
```
anicholls: Missing `initialId` here. We should be pulling out all props the component accepts even if they're not used so `elemProps` is explicitly extra props passed that we're not expecting.

modules/form-field/react/stories/stories_TextInput.tsx
```diff
@@ -1,16 +1,52 @@
...
+  };
+
+  render() {
+    const autocompleteResult = (textModifier: string) => (
+      <MenuItem onClick={action(`Went to link Result ${textModifier}`)}>
+        Result <span>num<span>ber</span></span> {textModifier}
+      </MenuItem>
+    );
+    return (
+      <Combobox
```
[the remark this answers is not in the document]
vibdev: That make sense, in that case should it live with text input. or be all on its own in `modules/combobox/react/...`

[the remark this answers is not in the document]
anicholls: I think it should be its own component 👍 Honestly it could render the `TextInput` itself, but I guess this does provide a bit more flexability in terms of styling for the input.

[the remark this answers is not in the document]
vibdev: Yep I agree 👍 will do

modules/text-input/react/lib/TextInput.tsx
vibdev: are they used for functionality? Looks like they are just getting spread into inputProps

modules/form-field/react/lib/Combobox.tsx
```diff
@@ -0,0 +1,317 @@
+
```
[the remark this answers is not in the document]
vibdev: moving to its own

modules/header/react/lib/parts/SearchBar.tsx
```diff
@@ -0,0 +1,245 @@
...
+    event.preventDefault();
+    this.setState({ searchQuery: event.target.value });
+    if (this.props.onInputChange) {
+      this.props.onInputChange(event)
+    }
+  };
+
+  clearForm = (): void => {
+    if (this.formRef.current) {
+      this.formRef.current.reset()
```
christine-m: will this fire react change events for inputs? (does it need to?)

modules/header/react/lib/parts/SearchBar.tsx
```diff
@@ -0,0 +1,245 @@
...
+    }
+  }
+
+  clearButtonType = iconButtonType(
```
christine-m: might be more readable to have this method take in whether its a clear/submit button and evaluate the type based on that

modules/header/react/spec/__snapshots__/Header.snapshot.tsx.snap
```diff
@@ -232,41 +193,24 @@ exports[`Dub Header Snapshots renders a default header 1`] = `
...
-                "__html": "<img src=\\"//design.workday.com/images/ck-dub-logo-blue.svg\\">",
-              }
-            }
-            title=""
-          />
-          
-        </div>
-      </div>
+      MockDubLogoTitle
+	themeColor: 0
```
christine-m: @vibdev @anicholls 
mocked some components from other modules so that we can see the important details (name and props) but not have the snapshot blown up with all the internals of that component (css and structure). i can revert if you guys prefer but personally find that it makes the snapshot for this component more readable

modules/combobox/react/lib/Combobox.tsx
```diff
@@ -0,0 +1,348 @@
...
+});
+
+const MenuContainer = styled(Card)<Pick<ComboboxProps, 'grow'>>(
+  {
+    position: 'absolute',
+    zIndex: 1,
+    left: 0,
+    top: '100%',
+    borderRadius: 4,
+    minWidth: 280,
```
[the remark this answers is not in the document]
vibdev: It is hardcoded on all the input types :(

modules/combobox/react/lib/Combobox.tsx
```diff
@@ -0,0 +1,348 @@
...
+        nativeInputValue.set.call(this.inputRef.current, newValue);
+      }
+      const event = new Event('input', { bubbles: true });
+      this.inputRef.current.dispatchEvent(event);
+    }
+  }
+
+  getTextFromElement = (children?: React.ReactNode) => {
+    let text = '';
+    React.Children.map(children, (child) => {
```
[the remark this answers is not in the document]
vibdev: maybe? this PR is so big anyway and I think it would only save one line here since we need to account to all kinds of markup

modules/combobox/react/lib/Combobox.tsx
```diff
@@ -0,0 +1,348 @@
...
+    return (
+      <Container
+        role='combobox'
+        aria-haspopup='listbox'
+        aria-owns={`${id}-${listBoxIdPart}`}
+        aria-expanded={autocompleteItems && autocompleteItems.length > 0}
+        {...elemProps}
+      >
+        <Status role="status" aria-live="polite">
+          {autocompleteItems ? this.state.showingAutocomplete && this.buildStatusString(autocompleteItems.length) : ''}
```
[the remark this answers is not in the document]
vibdev: This should only get read when showingAutocomplete toggles to true and autocompleteItems has changed so I think it will work correctly. Is there a specific case you are thinking of

modules/header/react/lib/parts/SearchBar.tsx
vibdev: This got moved to combobox

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -14,7 +14,6 @@ import TextInput, {TextInputProps} from '@workday/canvas-kit-react-text-input';
 
 export interface ColorInputProps extends TextInputProps, GrowthBehavior {
   value: string;
-  inputRef?: React.Ref<HTMLInputElement>;
```
vibdev: This is already in TextInputProps

vibdev:
@anicholls @NicholasBoll @christine-m This is finally updated! Sorry for the delay, but I think it is complete now 💃

vibdev:
Not sure the best way to squash commits. Should I just drop them down to one?
`feat(Combobox): Add new combobox component and add to header`?

lychyi:
@vibdev We can squash on merge, no need to take care of it on your end.

NicholasBoll:
@vibdev There are some visual changes that will have to be approved. From the looks of it, the baseline is correct. You can view the changes by clicking on the "Details" link in the check list (named `ci/chromatic`

modules/_labs/combobox/react/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "peerDependencies": {
+    "react": ">= 16.8 < 17"
+  },
+  "dependencies": {
+    "@workday/canvas-kit-labs-react-menu": "^3.0.0-alpha.6",
+    "@workday/canvas-kit-react-button": "^3.0.0-alpha.6",
+    "@workday/canvas-kit-react-card": "^3.0.0-alpha.6",
+    "@workday/canvas-kit-react-common": "^3.0.0-alpha.6",
+    "@workday/canvas-kit-react-core": "^3.0.0-alpha.6",
+    "@workday/canvas-kit-react-text-input": "^3.0.0-alpha.6",
```
NicholasBoll: These should be at version `^3.0.0-beta.1`

modules/_labs/combobox/react/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@workday/canvas-kit-labs-react-combobox",
+  "version": "3.0.0-beta.1",
```
NicholasBoll: I thought the version for new modules needed to start at `0.0.0`, but the CI seemed to pass...

modules/_labs/combobox/react/spec/Combobox.spec.tsx
```diff
@@ -0,0 +1,325 @@
+import * as React from 'react';
+import Combobox from '../lib/Combobox';
+import {MenuItem} from '../../../menu/react/index';
+import {TextInput} from '../../../../text-input/react';
+import {render, fireEvent} from '@testing-library/react';
```
NicholasBoll: 👍

modules/_labs/combobox/react/spec/Combobox.spec.tsx
```diff
@@ -0,0 +1,325 @@
...
+    const input = await findByPlaceholderText(placeholderText);
+    const down = {keyCode: 40, key: 'ArrowDown'};
+
+    fireEvent.keyDown(input, down);
+    expect(input.getAttribute('aria-activedescendant')).toEqual(`${id}-option-0`);
+    fireEvent.keyDown(input, down);
+    expect(input.getAttribute('aria-activedescendant')).toEqual(`${id}-option-1`);
+    fireEvent.keyDown(input, down);
+    expect(input.getAttribute('aria-activedescendant')).toEqual(`${id}-option-0`);
+  });
```
NicholasBoll: We'll probably migrate a lot of these tests to Cypress. It's fine as a unit test for now

modules/_labs/header/react/README.md
```diff
@@ -384,3 +360,143 @@ Default: `HeaderTheme.White`
...
+import {MenuItem} from '@workday/canvas-kit-labs-react-menu';
+
+const handleInputChange = event => console.log('Adjust menu items here')
+const handleSearchSubmit = event => {
+  const query = (event.target as HTMLFormElement).getElementsByTagName('input')[0].value;
+  console.log("Submitted query: ", query)
+}
+
+<SearchBar
+  autocompleteItems={[<MenuItem>Item 1</MenuItem>]}
```
NicholasBoll: I'm guessing this will only ever be a small amount of items (no paging involved). To refine autocomplete items, type more characters. Would it ever happen where a search phase doesn't show up and scrolling is required?

modules/_labs/header/react/lib/Header.tsx
```diff
@@ -191,45 +175,30 @@ const navStyle = ({themeColor}: Pick<HeaderProps, 'themeColor'>) => {
...
-        },
-      },
-      [mq.lg]: {
-        flexGrow: centeredNav ? 1 : 'unset',
-      },
-    };
+const ChildrenSlot = styled('div')<Pick<HeaderProps, 'centeredNav' | 'themeColor' | 'isCollapsed'>>(
+  {
+    marginRight: spacing.m,
+    // TODO: remove this when we get real icon buttons
```
NicholasBoll: Do we have real icon buttons yet?

modules/_labs/header/react/README.md
vibdev: I guess it would be up to up-takers, but for example, workday's autocomplete api limits results to 7. Not sure what you mean by scrolling is required. If search phase isn't found I would guess you would either `set autocompleteItems={[]}` or to some empty state.

modules/_labs/header/react/lib/Header.tsx
vibdev: TBH I have no idea, but it seemed out of scope for this PR

modules/_labs/header/react/README.md
NicholasBoll: If the API limits to only 7 results, I don't see a reason to support paging. Most autocomplete only show a small amount of results. I have seen UIs with unbounded results. If our UX says there should be a finite amount of results, this is perfect.

modules/_labs/combobox/react/README.md
```diff
@@ -0,0 +1,107 @@
...
+    autocompleteItems={[<MenuItem>Item 1</MenuItem>]}
+    isCollapsed={false}
+    onInputChange={this.autocompleteCallback}
+    placeholder='Search with Autocomplete'
+    grow={true}
+    searchTheme={SearchBar.Theme.Dark}
+    onSubmit={handleSearchSubmit}
+    labelId='id-123'
+  />
+    <TextInput placeholder="Autocomplete" autoFocus={true} />
```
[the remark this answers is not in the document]
vibdev: I did not. I think you would still need to override props if ones passed in didn't match what is needed for accessibility? Worth considering.

modules/form-field/react/lib/FormField.tsx
```diff
@@ -78,24 +82,22 @@ export default class FormField extends React.Component<FormFieldProps> {
   static defaultProps = {
     labelPosition: FormField.LabelPosition.Top,
     useFieldset: false,
+    inputId: `form-field-${uuid()}`,
```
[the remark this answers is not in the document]
vibdev: updated 👍

vibdev:
Visual diff is down to change in placeholder 💃

modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -0,0 +1,422 @@
...
+  state: Readonly<ComboboxState> = {
+    isFocused: false,
+    value: '',
+    showingAutocomplete: false,
+    selectedAutocompleteIndex: null,
+  };
+
+  componentDidMount() {
+    if (this.props.initialValue) {
+      this.setInputValue(this.props.initialValue);
```
NicholasBoll: This seems familiar to me. I think I've asked this before - why do we need to fire an event when the component is initialized with a value? Would this cause problems if an event handler is expecting to change only based on user interactions? They'd have to do some crazy logic to ignore changes the first time the component is rendered.

modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -0,0 +1,422 @@
...
+      return;
+    }
+    const currentIndex = this.state.selectedAutocompleteIndex;
+    const autoCompleteItemCount = this.props.autocompleteItems.length;
+    const firstItem = 0;
+    const lastItem = autoCompleteItemCount - 1;
+    let nextIndex = null;
+
+    switch (event.key) {
+      case 'ArrowUp':
```
NicholasBoll: These strings are not supported in IE11. See here for IE11 mappings: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Example

modules/_labs/combobox/react/lib/Combobox.tsx
vibdev: Without it an initialValue won't be able to fetch autocomplete items

modules/_labs/combobox/react/lib/Combobox.tsx
vibdev: Dang Ⓜ️💲. Fixed!

PR #161, opened 2019-08-30
feat: Add watch support for faster development

NicholasBoll:
## Summary

This feature adds watch support for development mode. Some changes require a rebuild of Typescript files. On a dev machine, `yarn build` takes about 120s. `yarn watch` now will reduce that time down to about 5s. The command takes a lot of memory (multiple typescript compiler instances) and so each module has its own `yarn watch` for more targetted watching at a fraction of the memory. Each Typescript watch instance takes about 120-250MB and that is per module. Perhaps in the future we will change the way modules work so that a single Typescript instance could run for builds rather than an instance per module.

## Checklist

- [x] `yarn test` passes

## Additional References

yarn-watch

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -147,6 +147,11 @@ fix: Add missing static class variable to IconButton and Avatar
 2.  Run `yarn build`
 3.  Run `yarn start` to start [Storybook](https://storybook.js.org/)
 4.  Visit [http://localhost:9001/](http://localhost:9001/)
+5.  To quickly rebuild on file changes, run `yarn watch` in another terminal. This will rebuild
+    modules for each file change. **Note** this command will take a lot of memory (1-2GB). If you
+    are working on a specific module, you can navigate to the module your are working on and run
+    `yarn watch` from there. This will only watch a specific module at a fraction of the memory
+    requirements
 
 ### Creating a module
 
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
   },
   "scripts": {
     "start": "node --max-old-space-size=2048 node_modules/.bin/start-storybook -p 9001 -c .storybook",
+    "watch": "TSC_NONPOLLING_WATCHER=1 lerna run watch --parallel --stream",
     "clean": "lerna run clean",
     "build": "lerna run build",
     "build:rebuild": "lerna run build:rebuild",
--- a/create-module.sh
+++ b/create-module.sh
@@ -58,6 +58,7 @@ cat > $packageJson << EOF
     "index.ts"
   ],
   "scripts": {
+    "watch": "yarn build:es6 -w",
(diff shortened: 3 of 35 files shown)


package.json
```diff
@@ -90,6 +90,7 @@
   },
   "scripts": {
     "start": "node --max-old-space-size=2048 node_modules/.bin/start-storybook -p 9001 -c .storybook",
+    "watch": "TSC_NONPOLLING_WATCHER=1 lerna run watch --parallel --stream",
```
NicholasBoll: `TSC_NONPOLLING_WATCHER` uses OS-level directory change events. Without this, polling is used and it takes up a lot of CPU (thousands of files are polled for changes). This might now work for every OS.

The CPU is made worse by how this project is set up - the same file is watched multiple times. For instance, the `Button` module is used by many other modules and so it and its dependencies are watched multiple times. If the project was re-architected to have 1 build command in the base `package.json`, this issue would be mitigated and possible build times could be improved (120s build times currently).

CONTRIBUTING.md
```diff
@@ -147,6 +147,11 @@ fix: Add missing static class variable to IconButton and Avatar
 2.  Run `yarn build`
 3.  Run `yarn start` to start [Storybook](https://storybook.js.org/)
 4.  Visit [http://localhost:9001/](http://localhost:9001/)
+5.  To quickly rebuild on file changes, run `yarn watch` in another terminal. This will rebuild
```
mannycarrera4: 👍

lychyi: This is only specific to tsc compilation for dist folders right? Step 3 already runs a storybook which can also be interpreted as "quickly rebuilds on file changes".

Should this be a separate statement? I don't know if I would run both `yarn start` and `yarn watch` at the same time really.

NicholasBoll: @lychyi It is a subtle issue that doesn't always show up when developing.

Webpack (used by storybook) does not recompile everything. I'll lay out a scenario:

You're tasked with adding accessibility to the `Modal` component and you need add a `aria-labelledby` to the `Popup` component and an `id` to the `Header` component in the `Card` component and you need `Card` to accept a new prop for that to work.

You open up the `Popup.tsx` file and add `aria-labelledby` and save. You notice Storybook hot-reloads, but the Modal still doesn't show the attribute. Wha? You load up the Popup story and see that the `aria-labelledby` attribute shows up on the Popup example. Why?

You are working on the `Modal` storybook story which imports `@workday/canvas-kit-react-popup`. This module is actually coming from the `modules/popup/react/dist/es6/lib` folder and not `modules/popup/react/lib` folder. You only see changes in Storybook that don't go through an `import '@workday/canvas-kit-react-*'` dependency. In order to see that change show up, you'll need to run `yarn build` again which takes 120s. So this common type of development will need a `yarn watch` to get those types of changes faster (about 5s on my machine for the slowest module to update)

PR #168, opened 2019-09-05
refactor(layout): Update CSS naming to match React implementation

jamesfan:
This PR updates the Layout (formerly known as Grid) CSS styles so they match the React component. 

### Code Changes
- `lib/grid-system.scss` moved to `lib/layout.scss`
- `$wdc-grid-system-breakpoints` > `$wdc-layout-breakpoints`
- `$wdc-grid-system-columns` > `$wdc-layout-columns`

refactor(layout): Update CSS styles to match React implementation (#168)

### DIFF
--- a/modules/layout/css/lib/layout.scss
+++ b/modules/layout/css/lib/layout.scss
@@ -9,9 +9,9 @@
 @each $breakpoint in $wdc-grid-system-breakpoints {
   $_name: nth($breakpoint, 1);
   $_size: nth($breakpoint, 2);
-  $_gutter: nth($breakpoint, 3);
-  $_half-gutter: $_gutter * 0.5;
-  $_gutter-compensation: $_half-gutter * -1;
+  $_spacing: nth($breakpoint, 3);
+  $_half-spacing: $_spacing * 0.5;
+  $_spacing-compensation: $_half-spacing * -1;
 
   @media only screen and (min-width: $_size) {
     .wdc-row {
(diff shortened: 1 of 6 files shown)


jamesfan:
Here's what I changed:
- [x] Rename `grid-system.scss` to `layout.scss`
- [x] Rename `Grid` to `Layout` in stories
- [x] Rename `Grid` to `Layout` in README
- [x] Rename `gutter` to `spacing` in Sass variables
- [x] Rename `gutter` to `spacing` in README

I don't believe this is considered a breaking change since the module was already renamed before (though I did have to update the installation instructions in the `README`), but please correct me if I'm mistaken.

modules/layout/css/stories.tsx
```diff
@@ -5,9 +5,9 @@ import README from './README.md';
 import './index.scss';
 import './stories.scss';
 
-storiesOf('CSS/Grid System', module)
+storiesOf('CSS/Layout', module)
   .addDecorator(withReadme(README))
-  .add('Responsive Grids', () => (
+  .add('Responsive Layout', () => (
     <div className="story story-grid">
```
jamesfan: @sahlhoff btw, is `story-grid` needed? I don't see it referenced anywhere else, but I wanted to double-check with you before removing it.

sahlhoff: @jamesfan good to remove 👍

modules/layout/css/lib/layout.scss
```diff
@@ -9,9 +9,9 @@
 @each $breakpoint in $wdc-grid-system-breakpoints {
```
sahlhoff: @jamesfan We should probably update all references of `grid-system` to `layout`

jamesfan:
Pushed out changes addressing Chad's feedback above:
- [x] Remove `story-grid` class
- [x] Rename references of `grid(-system)` to `layout`

Let me know if you see any other issues. Otherwise, everything should be good to go.

jamesfan:
@anicholls Got it, thanks for updating the description. I never added `BREAKING CHANGE` to my commit messages, though. Should I do something about that?

anicholls:
Generally, yes - but since all of these commits went into `css-update` which contains a bunch of `BREAKING CHANGE` commits already, that will get caught when `css-update` is merged into `master`

PR #174, opened 2019-09-09
feat: Add Cypress tests to canvas-kit

NicholasBoll:
* Added documentation on how to write and run Cypress tests
* Added a simple example test

## Summary

This PR adds the ability to have specifications exist in living code, enforced by a CI system. Cypress runs in a browser environment and is easy to debug and offers a superior developer experience.

## Checklist

- [x] `yarn test` passes

## Additional References

Nothing interesting yet


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
NicholasBoll: Hmm. I should probably remove this fixture. It was added by the Cypress command

jest.config.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   collectCoverageFrom: [
     'modules/**/*.{tsx,ts}',
     '!modules/**/*.d.ts',
+    '!cypress/**/*',
```
NicholasBoll: I don't think it makes sense to have code coverage of Cypress test files

jest.config.js
```diff
@@ -26,6 +27,7 @@ module.exports = {
   moduleFileExtensions: ['tsx', 'ts', 'js', 'jsx', 'json'],
   snapshotSerializers: ['jest-emotion/serializer'],
   testMatch: ['**/?(*.)+(spec|test|snapshot).ts?(x)'],
+  testPathIgnorePatterns: ['/cypress/'],
```
NicholasBoll: We don't want Jest to run on Cypress test files

package.json
```diff
@@ -82,7 +84,7 @@
     "sass-loader": "^7.0.1",
     "storybook-readme": "^5.0.8",
     "style-loader": "^0.20.3",
-    "ts-loader": "^3.5.0",
+    "ts-loader": "^4.0.0",
```
NicholasBoll: I ran into a weird error with Cypress without this change. @lychyi @anicholls Should we upgrade to the latest (6.0.4)?

tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "strictNullChecks": true,
     "noUnusedLocals": true,
     "allowSyntheticDefaultImports": true,
+    "types": ["jest"],
```
NicholasBoll: This prevents all namespaces (globals) from being automatically imported. Cypress includes mocha and chai which are also namespaces and conflict with each other.

tsconfig.json
```diff
@@ -18,7 +18,8 @@
...
     "incremental": true
   },
-  "exclude": ["node_modules", "modules/**/dist/*", "modules/**/ts-tmp/*", "docs"]
+  "exclude": ["cypress", "node_modules", "modules/**/dist/*", "modules/**/ts-tmp/*", "docs"]
```
NicholasBoll: Storybook + Webpack will look at the Cypress folder for Typescript errors, but that should be handled inside the `cypress/tsconfig.json` instead.

package.json
```diff
@@ -14,6 +14,7 @@
     "@babel/preset-typescript": "^7.1.0",
     "@commitlint/cli": "^8.1.0",
     "@commitlint/config-conventional": "^8.1.0",
+    "@cypress/webpack-preprocessor": "^4.1.0",
```
NicholasBoll: This dependency allows for Cypress spec files to support imports in cypress files

cypress/plugins/index.js
```diff
@@ -0,0 +1,24 @@
...
+        extensions: ['.ts', '.tsx', '.js'],
+      },
+      module: {
+        rules: [
+          {
+            test: /\.tsx?$/,
+            use: [
+              {
+                loader: 'ts-loader',
+                options: {transpileOnly: true},
```
NicholasBoll: Cypress supports TS files by default, but doesn't support `imports` in spec files by default. This plugin adds support for imports by wrapping them with Webpack. Essentially every `spec` file is an entry point and a `script` tag that gets inserted into the test page. Only helper files should be imported so that helper code gets added as well. Another option could be that the helper folder gets added for every test automatically by adding it to `cypress/support/index.ts`, but that feels a bit too magical and may break "Jump to definition". We could try it...

CYPRESS_TESTS.md
```diff
@@ -0,0 +1,137 @@
+## Intro
```
NicholasBoll: Add a note about where Cypress files go

CYPRESS_TESTS.md
```diff
@@ -0,0 +1,137 @@
...
+
+Tests should be written in a BDD-style format as a list of specifications and should be readable by
+disciplines outside of Software Development Engineers (including Accessibility Specialists, Product
+Owners and UX Engineers). Cypress uses Mocha, Chai and jQuery under the hood allowing for an
+outline-style of test writing. It is very possible UX and Accessibility will give outlines for
+engineers to implement.
+
+## Running tests
+
+The following with run Cypress in headless mode and run all tests
```
anicholls: Typo

CYPRESS_TESTS.md
```diff
@@ -0,0 +1,143 @@
+## Intro
```
anicholls: Maybe add a `# Cypress Testing` header at the top?

CONTRIBUTING.md
```diff
@@ -66,9 +66,12 @@ within a single repository.
 
 #### Testing
 
-Canvas Kit uses [Jest](https://jestjs.io/) and [Enzyme](https://airbnb.io/enzyme/) to test our React
-components. Each and every component requires at least 97% unit test coverage and a thorough set of
-snapshot tests.
+Canvas Kit uses [Jest](https://jestjs.io/) and [Enzyme](https://airbnb.io/enzyme/) to unit test our
+React components. Each and every component requires at least 97% unit test coverage and a thorough
```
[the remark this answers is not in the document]
NicholasBoll: Sure

PR #175, opened 2019-09-10
feat: Add support for Canvas Kit Labs modules

anicholls:
## Summary
Add an unstable sandbox (Canvas Kit Labs) for work in progress components.

Since we have a rather rigorous process for getting components in at a production level, it can be
valuable to make them available earlier for teams who need them during development. Many of our
components are contributions from internal product teams. While these components can suit the
contributing product's use case, it takes time to iterate on the API/functionality as we evaluate
them in a more global context. The Labs module allows us to do that as needed.

## Checklist

- [x] `yarn test` passes


modules/_unstable/README.md
```diff
@@ -0,0 +1,44 @@
...
+them in a more global context. The Unstable module allows us to do that as needed.
+
+## Breaking Changes
+
+Due to the nature of this bundle of components, we will be continuously iterating, likely making
+breaking changes. The goal of this module is to work unconstrained by our design system's semantic
+versioning. This means that **breaking changes can be deployed to
+`@workday/canvas-kit-react-unstable` or `@workday/canvas-kit-css-unstable` at any time and it will
+not trigger a major version bump across our other modules**. Since we use a fixed version number for
+all components, all changes/additions to this module will only result in a patch version bump across
```
anicholls: Are we confirmed that our version bumps will always be patch versions for updates to `unstable`?

modules/_unstable/README.md
```diff
@@ -0,0 +1,44 @@
+# Canvas Kit Unstable
+
+This module is a sandbox for work in progress components.
+
+Since we have a rather rigorous process for getting components in at a production level, it can be
+valuable to make them available earlier for teams who need them during development. Many of our
+components are contributions from internal product teams. While these components can suit the
+contributing product's use case, it takes time to iterate on the API/functionality as we evaluate
+them in a more global context. The Unstable module allows us to do that as needed.
+
```
anicholls: Should we add any documentation around contributing to components in `unstable`? 

Should we be instructing in [CONTRIBUTING.md](https://github.com/Workday/canvas-kit/blob/master/CONTRIBUTING.md) that any new components being contributed go into `unstable` first? Or will we instruct contributors to move their component there during the PR process if we think it's not ready for primetime?

modules/_unstable/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Breaking Changes
+
+Due to the nature of this bundle of components, we will be continuously iterating, likely making
+breaking changes. The goal of this module is to work unconstrained by our design system's semantic
+versioning. This means that **breaking changes can be deployed to
+`@workday/canvas-kit-react-unstable` or `@workday/canvas-kit-css-unstable` at any time and it will
```
sahlhoff: Will this component act as a global namespace similar to how `@workday/canvas-kit-react` imports all CKR components? Or would all of the unstable modules share and live in the same namespace `@workday/canvas-kit-react-unstable` and `@workday/canvas-kit-css-unstable`?

anicholls: Good point of discussion. I haven't built that into this yet, so I guess right now they're independently published. However, IMO all of the components (of the same target) should be exported under the same namespace. I'm not sure how we'd restrict them being published under their own namespace though. I think adding `private: true` to the package.json will do it.

modules/_unstable/README.md
anicholls: I guess we don't really have a way to restrict this. If someone uses `feat: ...` for a commit message it will automatically be a minor bump

commitlint.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
 };
 
 // Extra scopes supported outside the `modules` folder
-const scopes = [];
+const scopes = ['unstable'];
```
NicholasBoll: :+1: for forethought

modules/_unstable/README.md
NicholasBoll: Hmm. I think we should encourage communication for new components earlier than a pull request. That's simple enough for internal developers, but we don't really have a public communication channel outside of github for external contributors...

We we encourage or allow new component development to external contributors?

modules/_unstable/README.md
NicholasBoll: Will `unstable` have a different cadence than the rest of the modules? Do we keep all modules on the same version? (right now they are all `3.0.0-alpha.6`)

anicholls: Now that we've switched to fixed versioning with Lerna, I don't think there is a way to release it with a different version/cadence unless we explicitly ignore it in lerna.json. Is this something we want to do?

NicholasBoll: 🤷‍♂ I could imagine if `unstable` was on a different and possibly more frequent cadence that `unstable` could be using code changes from "stable" that hasn't been released yet and that causing issues.

I think it still makes sense to tie `unstable` to the same fixed releases as the rest of the packages. If internal teams want a faster cadence than our standard releases, we can create internal releases or public releases with a special tag.

modules/_unstable/README.md
anicholls: I think it's unlikely/uncommon but possible. IMO not all new components need to start in `unstable` so maybe we skip this. I think we only move it over if there's a lot of debate/uncertainty over the PR.

anicholls: On second thought, maybe it's best to call out in CONTRIBUTING.md that some new components may be moved there while they're fleshed out, but not all need to be started there.

modules/_unstable/README.md
anicholls: After further discussion, we arrived at a different conclusion due to a few things:

- We can't publish a bundle module while other ones are private if we plan to re-export them (because they won't be available as deps to download)
- Embedding them in the bundle module would be a pain and could lead to a lot of potential manual errors when migrating into unstable, or promoting out of unstable

In the end, we decided to organize these modules into the `_unstable` folder, but keep them as their own independent modules. They will be released with a warning in the readme and a prerelease version (i.e. if the codebase is on `3.0.0`, the menu module would be released as `3.0.0-beta`)

modules/_unstable/README.md
```diff
@@ -0,0 +1,59 @@
...
+   > <a href="https://github.com/Workday/canvas-kit/tree/master/modules/_unstable/README.md">
+   >   <img src="https://img.shields.io/badge/UNSTABLE-alpha-orange" alt="UNSTABLE: Alpha" />
+   > </a>  This component is work in progress and currently in pre-release.
+3. Ensure a prerelease version is applied that matches the badge above
+
+## Deploying
+
+- Components in Unstable are given prerelease versions. Their general version will align with the
+  rest of the library (e.g. `3.0.0-alpha.0`)
+- Before deploying, ensure all unstable modules have a prerelease tag (`alpha` or `beta`).
```
anicholls: How do we enforce this?

We’ve just been running `yarn lerna version --no-git-tag-version --no-push --no-changelog prerelease` for releases I believe (@lychyi , correct me if this has changed - you’ve done the last couple). This works when everything is in prerelease, but when we do a final version (without `prerelease`), we’ll have to manually version each module in Unstable to add the prerelease tags.

Should I add a new yarn version command that bumps all modules and then versions everything in `modules/_unstable` as `alpha`? What if we want to support `beta`? This feels like it should be a CLI tool. However, for the sake of getting this PR in ASAP, I'm thinking we do this manually for now and add an issue to work on a script for this.

NicholasBoll: Prereleases can be tricky. If we release a `3.0.0` tomorrow, then `canvas-kit@^3.0.0` will no longer match pre-releases even if they occur later. We can use tags, however. By default, `npm publish` will do `npm publish --tag latest` including pre-releases. If we want to publish intermediate changes, we'd probably want more like a `-build-{hash}` and publish the tag as `next` or `master` or `unstable` or something.

I don't think we want to use `alpha` or `beta`, but more like `build`. No promises about `unstable` other than builds, right?

lychyi: @anicholls We don't run that command explicitly anymore. Here were the tweaks from last time:
* Removed `--no-changelog`. It not necessary without `--conventional-commits` ([link](https://github.com/lerna/lerna/tree/master/commands/version#--no-changelog))
* The semver bump positional arg: `prerelease` is now something that someone must select manually when using `yarn bump` (this way this command can be used for both prerelease bumps and regular semantic bumps)
* All options for `lerna version` are all moved into `lerna.json`

The `prerelease` tag is now specified in `travis.yml` as a [build step](https://github.com/Workday/canvas-kit/blob/master/.travis.yml#L30) as a `--pre-dist-tag` option for `lerna publish`. This allows publish to tag things as prerelease only if they have a prerelease version (`*-alpha.X` for example). Everything else should default to `latest` on publish.

modules/_unstable/README.md
```diff
@@ -0,0 +1,59 @@
...
+        /* Standard CSS component structure */
+      /react
+        /* Standard React component structure */
+```
+
+## Creating an Unstable Module
+
+1. Run `yarn create-module`
+2. When asked `Is this an unstable component? [Y/n]`, enter `y` or `Y`.
+3. Your new module will be generated in accordance with the file structure above.
```
sahlhoff: @anicholls Should we include a note about the storybook IA? 

eg, It may be good for visibility if all unstable components were in the storybook `unstable` category– or some other way to distinguish this component is unstable in storybook stories.

modules/_unstable/README.md
anicholls: This isn't about prereleases of any stable modules (or the bundle modules). Essentially we are talking about fixed versioning for everything _not_ in `_unstable`. Then manually versioning the items in `_unstable`. However, I'm not sure this is really possible without going back to independent versioning... (which we really don't want to do)

IMO, a warning in the readme, the folder the module is in within the source repo and an npm tag is super easy to miss from the consumer's perspective...

modules/_unstable/README.md
anicholls: Good call 👍 I will add that.

modules/_unstable/README.md
anicholls: So after a very long discussion about the best direction to go, we arrived at this:

## Short term
- We decided to call this area "Canvas Kit Labs". It's more concise and feels more positive than "unstable"
- All unstable modules will move into a `_labs` folder
- All unstable modules in this folder will have `-labs` put in their component name (i.e. `@workday/canvas-kit-labs-react-menu`. We are unconcerned about the pollution in the namespace because they're very long and names are cheap. It is extremely unlikely we will need this namespace in the future. This will be seen anywhere you see the version and will offer similar clarity to a prerelease version. It avoids the need for going back to independent versioning (for now).

## Long term
Consider one of two options:
- Switch back to "independent" versioning in Lerna, but write a Lerna integration script that keeps the majority of our components on a fixed version, while allowing us to manage select modules independently (e.g. unstable components)
- If we move to a slash-style import system (a single bundle, i.e. `@workday/canvas-kit/react/button`), it would make this very easy (e.g. `@workday/canvas-kit/labs/react/menu`)

create-module.sh
```diff
@@ -136,19 +153,39 @@ EOF
...
+  <img src="https://img.shields.io/badge/UNSTABLE-alpha-orange" alt="UNSTABLE: Alpha" />
+</a>  This component is work in progress and currently in pre-release.
+EOF
+fi
 
 # Create tsconfig.json
 tsconfig="$reactPath/tsconfig.json"
 echo -e "Creating ${CYAN}$tsconfig${NC}"
+
+if [ $stable ] ; then
```
lychyi: My eyes may be deceiving me but the if/else blocks are the same right? Is this just a preemptive measure in case we need separate logic?

create-module.sh
```diff
@@ -262,25 +308,41 @@ EOF
 # Create README.md
 readme="$cssPath/README.md"
 echo -e "Creating ${CYAN}$readme${NC}"
+
+if [ $stable ] ; then
 cat > $readme << EOF
 # Canvas Kit CSS $upperName
 EOF
+else
```
lychyi: Similar here, it's almost like the title can be extracted out. Either way, these are all nits, no big deal

modules/_labs/README.md
```diff
@@ -0,0 +1,64 @@
+# Canvas Kit Labs
+
+This is a group of unstable/work in progress components. Canvas Kit Labs is an incubator for new and
```
lychyi:
```suggestion
This is a group of unstable/work-in-progress components. Canvas Kit Labs is an incubator for new and
```

modules/_labs/README.md
```diff
@@ -0,0 +1,64 @@
...
+experimental components. Since we have a rather rigorous process for getting components in at a
+production level, it can be valuable to make them available earlier while we continuously iterate on
+the API/functionality. The Labs modules allow us to do that as needed.
+
+## Breaking Changes
+
+Due to the nature of this bundle of components, we will be continuously iterating, likely making
+breaking changes. The goal of these module is to work unconstrained by our design system's semantic
```
lychyi:
```suggestion
breaking changes. The goal of this module is to work unconstrained by our design system's semantic
```

modules/_labs/README.md
```diff
@@ -0,0 +1,64 @@
...
+versioning. This means that **breaking changes can be deployed to Labs modules at any time and it
+will not trigger a major version bump across our other modules**. Regardless of the fact that we use
+a fixed version number for all components, all changes/additions in this bundle will never result in
+a major version bump across our component library.
+
+**By consuming any of these modules, you are indicating that you are prepared for the consequences
```
lychyi: I think the wording sounds a bit harsh. What about a lighter tone?

```suggestion
**By consuming any of these modules, you acknowledge the potential complexity
```

modules/_labs/README.md
```diff
@@ -0,0 +1,64 @@
...
+5. Change the storybook path to add a `Labs` prefix (e.g. `Labs/Menu/Default`)
+
+## Promoting a Component out of Canvas Kit Labs
+
+1. Move the module folder from `modules/_labs` to `modules/`
+2. Rename the package in `package.json` to `@workday/canvas-kit-labs-<TARGET>-<COMPONENT>` (remove
+   `labs-`)
+3. Remove the warning from the README
+4. Update any necessary paths (links to storybook utils, tsconfig, etc.)
+
```
lychyi: We should add a reminder about deprecating the labs package.
```suggestion
5. Open an issue to have the `labs` component deprecated in npm.
```

create-module.sh
anicholls: Ah - good catch. It's left over from before I pulled out the `rootPath`

PR #181, opened 2019-09-11
fix: Extend component props with correct interfaces

### LINKED ISSUE #162: Fix incorrect extensions of HTMLAttributes
## 🐛 Bug Report

When we're extending props to contain attributes from HTML elements, we're currently using the type `React.HTMLAttributes<T>`. For some components we should be using a more specific type of `HTMLAttributes`. For example, when we're spreading element props to an `<input>` from the component (you'll see this as: `...elemProps`) we need to be more specific about the type of attributes. We are already doing this in some places correctly: `extends React.InputHTMLAttributes<HTMLInputElement>`.

Note:
There are a lot of resources on the web that specify that we use `HTMLProps` to extend our prop interfaces. This doesn't seem correct, it's too generic. `HTMLProps` contains `AllHTMLAttributes<T>` and `ClassAttributes<T>`. This makes me think that it's less for extending prop interfaces and more for describing props for HTML elements.

## To Reproduce

1. Add the `name` prop to `Checkbox`. It is an `<input>` that incorrectly extends `HTMLAttributes<HTMLInputElement>`
2. You'll get a TS error

## Actual Results

A clear and concise description of what actually happens

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

https://codesandbox.io/s/htmlattributes-vs-htmlprops-d34n9

### Error Output

Screen Shot 2019-08-30 at 11 28 01 AM

lychyi:
## Summary

Fixes #162

## Checklist

- [x] `yarn test` passes
      applicable

### DIFF
--- a/modules/banner/react/lib/Banner.tsx
+++ b/modules/banner/react/lib/Banner.tsx
@@ -6,7 +6,7 @@ import {SystemIcon} from '@workday/canvas-kit-react-icon';
 import {exclamationCircleIcon, exclamationTriangleIcon} from '@workday/canvas-system-icons-web';
 import {focusRing} from '@workday/canvas-kit-react-common';
 
-export interface BannerProps extends React.HTMLAttributes<HTMLButtonElement> {
+export interface BannerProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   /**
    * Action when interacting with the banner
    */
--- a/modules/checkbox/react/lib/Checkbox.tsx
+++ b/modules/checkbox/react/lib/Checkbox.tsx
@@ -11,7 +11,7 @@ import {SystemIcon} from '@workday/canvas-kit-react-icon';
 import {checkSmallIcon} from '@workday/canvas-system-icons-web';
 import uuid from 'uuid/v4';
 
-export interface CheckboxProps extends React.HTMLAttributes<HTMLInputElement> {
+export interface CheckboxProps extends React.InputHTMLAttributes<HTMLInputElement> {
   checked: boolean;
   disabled?: boolean;
   id?: string;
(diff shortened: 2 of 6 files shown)


modules/banner/react/lib/Banner.tsx
```diff
@@ -6,7 +6,7 @@ import {SystemIcon} from '@workday/canvas-kit-react-icon';
 import {exclamationCircleIcon, exclamationTriangleIcon} from '@workday/canvas-system-icons-web';
 import {focusRing} from '@workday/canvas-kit-react-common';
 
-export interface BannerProps extends React.HTMLAttributes<HTMLButtonElement> {
+export interface BannerProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
```
NicholasBoll: Ah. Now I understand what you mean. I guess I assumed `HTMLAttributes<HTMLButtonElement>` would automatically apply attributes from `HTMLButtonElement`, but I see that it doesn't.

NicholasBoll: I think we should use or make a lint rule to help with this. I see a pattern where the Element type should match the React element interface.

lychyi: @NicholasBoll Not all React Elements match the interface for that element. `HTMLAttributes<HTMLSpanElement>` is a good example... it's not `SpanHTMLAttributes<HTMLSpanElement>`.

It can be a lint rule or maybe we can make an indexable type so that it's maintained in one place? Something like:

```
// shared/types.ts

interface ElementProps {
  body: ...,
  br: ...,
  button: React.ButtonHTMLAttributes<HTMLButtonElement>,
  canvas: ...
  ...
}

// banner/react/Banner.tsx

type ButtonElementProps = ElementProps['button'];
export interface BannerProps extends ButtonElementProps {
  ...
}

PR #183, opened 2019-09-11
feat: Add faster testing of stories in Cypress

NicholasBoll:
## Summary

Added `window.setCurrentStory(story: string, variation: string)` to enable faster (re)loading of Storybook stories. The `story` parameter matches the string inside `storiesOf` and the `variation` matches the string inside the `.add` function

Use (in browser's console):
```ts
setCurrentStory('button', 'primary')
```

## Checklist

- [x] `yarn test` passes

## Additional References

<details>
    <summary>Animated GIF. Warning! Tests are fast and flash rapidly!</summary>
    Cypress tests
</details>

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -39,10 +39,10 @@ addParameters({
 configure(loadStories, module);
 injectGlobal(...fonts);
 
-function setCurrentStory(story, variation) {
+function setCurrentStory(categorization, story) {
   clearCurrentStory();
   addons.getChannel().emit(Events.SET_CURRENT_STORY, {
-    storyId: toId(story, variation),
+    storyId: toId(categorization, story),
   });
   forceReRender();
 }
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,8 +1,13 @@
-import {configure, addDecorator, addParameters} from '@storybook/react';
+import {configure, addDecorator, addParameters, forceReRender} from '@storybook/react';
 import {withKnobs} from '@storybook/addon-knobs/react';
 import {injectGlobal} from 'emotion';
 import fonts from '../modules/fonts/react';
 import {create} from '@storybook/theming';
+import addons from '@storybook/addons';
+import Events from '@storybook/core-events';
+import {toId} from '@storybook/router';
+import ReactDOM from 'react-dom';
+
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -118,11 +118,16 @@
   },
   "lint-staged": {
     "linters": {
-      "*.{ts,tsx}": [
+      "modules/**/*.{ts,tsx}": [
```
NicholasBoll: This prevents a TSLint warning about the file passed in is excluded by the `tsconfig.json` configuration settings. The project has 2 different Typescript projects - `modules` and `cypress`.

* `modules` - All modules source code goes here and includes the `jest` Typescript namespace and config to create a Storybook application and run Jest tests against glob patterns
* `cypress` - All source code here includes the `cypress` Typescript namespace because it conflicts with the `jest` namespace and requires a different config (mostly shared). The lint-staged needs to reference these different configs to be processed correctly

cypress/helpers/stories.ts
```diff
@@ -0,0 +1,33 @@
...
+    setCurrentStory(story: string, variation: string);
+  }
+}
+
+/**
+ * Load a story. This will invoke the storybook router,
+ * unmount a previous story, mount the current story and force a rerender
+ * This should be in a `beforeEach` block to force a fresh new story
+ * @param story Name of the story found in the `.add` function
+ * @param variation Variation of the story found in the `.add` function
```
anicholls: Nit: From the convo yesterday, can we call this a `variant`?

.storybook/config.js
```diff
@@ -33,3 +38,18 @@ addParameters({
 
 configure(loadStories, module);
 injectGlobal(...fonts);
+
+function setCurrentStory(story, variation) {
+  clearCurrentStory();
```
NicholasBoll: If this isn't here, loading the same story does not reset. The Storybook APIs will only clear if a story changes.

cypress/helpers/stories.ts
NicholasBoll: Oh, definitely. I'm sorry I missed that

NicholasBoll: I went with `categorization` and `story` with more clear descriptions

PR #184, opened 2019-09-11
test: Add Cypress modal specifications

### LINKED ISSUE #59: React: Accessibility improvements for Popup
## 🐛 Bug Report

`Popup` does not currently pass accessibility requirements. We cannot use it for many cases unless we add the ability to lock/cycle focus within them, and focus on the first focusable item when opened, etc.

We need to support everything in https://bitsofco.de/accessible-modal-dialog/ except for # 4:

> 1. Markup the Dialog and Dialog Overlay Appropriately
> 2. On Dialog Open, Set Focus
> 3. On Dialog Close, Return Focus to the Last Focused Element
> 5. While Open, Prevent Tabbing to Outside the Dialog
> 6. While Open?
> 7. Allow the ESC Key to Close the Dialog

NicholasBoll:
## Summary

Add the specifications for modals

This is the first browser-based specification of a canvas-kit component. It implements a specification style from #174 described in https://github.com/Workday/canvas-kit/blob/master/cypress/README.md

The specification format is meant to read like a set of acceptance criteria or specifications for how a component should behave. The format should be readable by many stakeholders including Software Development Engineers, Product Managers, Quality Assurance Engineers, UX and Accessibility Specialists. The specifications can eventually be part of a component's documentation in the raw or outline form.

## Checklist

- [x] `yarn test` passes

## Additional References

<details>
    <summary>Animated GIF. Warning! Tests are fast and flash rapidly!</summary>
    Cypress tests
</details>

Fixes #59

## Suggested Squash commit message:
```
test(modal): Add modal specification tests
* Added `@testing-library/react` for Jest. Modal doesn't work with Enzyme
* Added `headingId` to `Card` to support `aria-labelledby` accessibility
* Updated `Popup` story to use `Popper` component from `canvas-kit` instead of `material-ui`
BREAKING CHANGE:
* `heading` is no longer optional in a Modal component. This is for accessibility guarantees
* `transformOrigin` can no longer be passed to `Modal`. Animation is fixed
```


cypress/integration/Modal.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+describe('Modal', () => {
+  before(() => {
+    h.stories.visit();
+  });
+
+  ['Default', 'UseModal'].forEach(story => {
+    context(`given the ${story} story is rendered`, () => {
+      beforeEach(() => {
+        h.stories.load('Modal', story);
+        // cy.visit(`http://localhost:9001/iframe.html?id=modal--${story.toLowerCase()}`);
```
NicholasBoll: This should be removed

cypress/helpers/modal/modal.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @param testId Optional test id to target the desired modal
+ */
+export function get(testId?: string) {
+  const selector = testId ? `[data-testid='${testId}']` : `[role=dialog]`;
+
+  return cy.get(selector);
+}
+
+export function getTitle($modal: JQuery) {
+  return $modal.find('h3');
```
NicholasBoll: There was nothing special about the heading element. Is this okay, or should we add something more specific?

Perhaps `$modal.find('h3').first()`?

`$modal` is already a scoped element (only the modal), so we don't have to worry if there are other `h3` elements on the page. It just feels too generic to me.

cypress/helpers/stories.ts
```diff
@@ -24,7 +24,7 @@ declare global {
  */
 export function load(categorization: string, story: string) {
   cy.window().then(win => {
-    win.setCurrentStory(categorization, story);
+    win.setCurrentStory(categorization.toLowerCase(), story.toLowerCase());
```
NicholasBoll: `toLowerCase` allows us to use the same string as we find in `.add`. Easier for copy/paste

jest/setupTests.ts
```diff
@@ -1,6 +1,7 @@
 import Adapter from 'enzyme-adapter-react-16';
 import {configure} from 'enzyme';
 import {toHaveNoViolations} from 'jest-axe';
+import '@testing-library/jest-dom/extend-expect';
```
NicholasBoll: Oh, I SO recommend using these matchers. The errors are so much clearer when tests fail: 

https://github.com/testing-library/jest-dom

modules/common/react/lib/Popper.tsx
```diff
@@ -69,9 +69,15 @@ export class Popper extends React.PureComponent<PopperProps> {
...
-      ...this.props.popperOptions,
-    });
+    if (this.props.anchorElement) {
+      this.popper = new PopperJS(this.props.anchorElement, popperNode, {
+        placement: this.props.placement,
+        ...this.props.popperOptions,
+      });
+    } else {
+      console.warn(
+        `Popper: anchorElement was not defined. A valid anchorElement must be provided to render a Popper`
```
NicholasBoll: Hmm... I don't remember why I made this change...

cypress/helpers/modal/modal.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ *
+ * @param testId Optional test id to target the desired modal
+ */
+export function get(testId?: string) {
```
anicholls: Can we combine this file with `cypress/helpers/modal/index.ts`? It seems unnecessary to have two files for these.

modules/common/react/lib/Popper.tsx
anicholls: But why did we make it optional in the first place?

modules/modal/react/lib/Modal.tsx
```diff
@@ -10,12 +11,36 @@ export enum ModalWidth {
...
-  heading?: React.ReactNode;
+  /**
+   * Accessibility specifications state modals should be close when the escape key is pressed.
+   * However, we cannot guarantee that it is safe to simply bind an event listener and close in all
+   * cases. Some applications may use a Popup manager to make sure the correct popup is receiving
+   * the close command. If your application uses custom popup stacking, do not set this to true.
+   * Set this to true for simple applications and the modal will close when the escape key is pressed.
+   */
+  closeOnEscape: boolean;
+  heading: React.ReactNode;
```
anicholls: Should we support `headerId` as well?

modules/popup/react/lib/Popup.tsx
```diff
@@ -110,7 +117,13 @@ export default class Popup extends React.Component<PopupProps> {
             />
           </CloseIconContainer>
         )}
-        <Card depth={depth} heading={heading} width={width} padding={padding}>
+        <Card
+          depth={depth}
+          heading={heading}
+          headingId={heading ? this.id : undefined}
```
anicholls: Should we support this as a prop?

modules/modal/react/lib/Modal.tsx
```diff
@@ -10,12 +11,36 @@ export enum ModalWidth {
...
+   */
+  testId?: string;
   open: boolean;
-  padding: PopupPadding;
-  transformOrigin: TransformOrigin;
-  width: ModalWidth;
+  padding?: PopupPadding;
+  width?: ModalWidth;
+  /**
+   * Optional callback for the Modal handling closing. If this callback is provided
```
anicholls: Looks like the end of this sentence got cut off

modules/modal/react/lib/Modal.tsx
```diff
@@ -10,12 +11,36 @@ export enum ModalWidth {
...
+   */
   handleClose?: () => void;
-  heading?: React.ReactNode;
+  /**
+   * Accessibility specifications state modals should be close when the escape key is pressed.
```
anicholls: Nit: `close` > closed

modules/modal/react/lib/Modal.tsx
```diff
@@ -50,17 +116,32 @@ export default class Modal extends React.Component<ModalProps> {
     open: false,
     padding: Modal.Padding.l,
     width: Modal.Width.s,
+    closeOnEscape: false,
```
anicholls: Any reason in particular that you picked `false` for the default here?

modules/modal/react/lib/Modal.tsx
```diff
@@ -41,6 +66,47 @@ const Container = styled('div')({
...
+  if (firstFocusEl) {
+    return firstFocusEl;
+  } else {
+    const firstFocusable = modalEl && modalEl.querySelector<HTMLElement>('[aria-label=Close],h3');
+    if (firstFocusable) {
+      if (firstFocusable.tagName === 'H3') {
+        // If there is no close icon, we need to transfer focus to the header.
+        // Setting tabIndex allows the header to be focusable.
+        // We do the header instead of the next focusable element to prevent useful context from being skipped
+        firstFocusable.tabIndex = 0;
```
anicholls: Can you add `&:focus { outline: 'none' }` to the card title in `Card`? I believe we discussed hiding the focus outline if the heading receives focus with a11y.

modules/modal/react/lib/Modal.tsx
```diff
@@ -50,17 +116,32 @@ export default class Modal extends React.Component<ModalProps> {
...
     width: Modal.Width.s,
+    closeOnEscape: false,
     transformOrigin: {
       horizontal: 'center',
       vertical: 'top',
     },
   };
 
+  private handleKeydown = (event: KeyboardEvent) => {
+    if (this.props.closeOnEscape && this.props.handleClose && event.keyCode === 27) {
```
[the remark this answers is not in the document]
NicholasBoll: Sure. I think `key` has wide browser support. The `key` is `Esc` in IE11/Edge (pre-Blink) and `Escape` in other browsers

modules/modal/react/lib/Modal.tsx
anicholls: It may also be worth calling out the use case for a UI blocking modal. That's not immediately clear when reading this.

modules/common/react/lib/Popper.tsx
NicholasBoll: I'm going to try taking it out and see what happens.

modules/common/react/package.json
```diff
@@ -38,10 +38,12 @@
...
   },
   "dependencies": {
+    "@types/tabbable": "^3.1.0",
     "@workday/canvas-kit-react-core": "^3.0.0-alpha.7",
     "emotion": "^9.2.12",
     "lodash": "^4.17.14",
     "popper.js": "^1.15.0",
-    "react-emotion": "^9.2.12"
+    "react-emotion": "^9.2.12",
+    "tabbable": "^4.0.0"
```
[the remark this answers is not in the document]
NicholasBoll: Good catch. Originally I was using it directly, but `focus-trap` gives me `initialFocus` which allows me to set the first focus element without doing it explicitly. I'll remove this.

modules/core/react/README.md
```diff
@@ -375,8 +375,8 @@ handling.
 }
 ```
 
-We provide a [helper](../../common/react/lib/styles/hideMouseFocus.ts) to hide the focus
-outlines on mouse input. Simply spread it in your styles (i.e. `...hideMouseFocus`).
+We provide a [helper](../../common/react/lib/styles/hideMouseFocus.ts) to hide the focus outlines on
```
NicholasBoll: This change was my Prettier...

modules/common/react/lib/Popper.tsx
NicholasBoll: TS builds, Jest tests and Cypress tests all pass without this change. I'm reverting.

cypress/helpers/modal/modal.ts
NicholasBoll: We could move this file to `cypress/helpers/modal.ts`. I don't remember why I created a folder for this

NicholasBoll: I thought of a downside to the composable helper function vs Page Objects (or component class helpers). There is no strict type safety with this approach. At my last org, we used this style and didn't really run into issues, but I never experimented with if I should add a stricter type output.

For example, the return signature is `Cypress.Chainable<JQuery>` and the input of `getTitle` and `getCloseButton` are `$modal: JQuery`. Technically that could be any element. Should we be more strict about the type signature?

```ts
export interface Modal extends JQuery {
  __$el: 'Modal';
}

export function get(testId?: string): Cypress.Chainable<Modal> {
  const selector = testId ? `[data-testid='${testId}']` : `[role=dialog]`;

  return cy.get(selector) as Cypress.Chainable<Modal>;
}

export function getTitle($modal: JQuery): JQuery<HTMLHeadingElement> {
  return $modal.find('h3').first() as JQuery<HTMLHeadingElement>;
}

export function getCloseButton($modal: JQuery): JQuery<HTMLButtonElement> {
  return $modal.find('[aria-label=Close]') as JQuery<HTMLButtonElement>;
}
```

The typing is more strict, but the type error can be confusing and I don't think it adds much value.

modules/modal/react/lib/Modal.tsx
```diff
@@ -10,12 +11,36 @@ export enum ModalWidth {
 }
 
 export interface ModalProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * This will add a `data-testid` attribute to the Popup component, not the container
+   */
+  testId?: string;
```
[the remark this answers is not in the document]
NicholasBoll: Oh, yes. I was going to propose this as an official optional property with a helper function that includes for testing. I've done this in the past and it worked well. Some teams use `cy-test`, others use `data-testid`, others use `data-test-id` and I think I've seen `data-automation-id`.

Since our components mostly just mix extra attributes into the root component node, this isn't strictly needed and perhaps should not be introduced here.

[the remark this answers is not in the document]
NicholasBoll: Ah. So the `Container` mixes in `elemProps` and this adds the testid to the `[role=dialog]` element. The test ID could be added to the container and the modal helper instead tries to find the test id and then returns the modal element.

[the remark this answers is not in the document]
NicholasBoll: I wanted to explicitly call out the relationship between the `testId` in the code and the `testId` in the helper rather than it being an implicit relationship.

@lychyi @anicholls What do you think?

modules/modal/react/lib/Modal.tsx
```diff
@@ -50,17 +116,32 @@ export default class Modal extends React.Component<ModalProps> {
     open: false,
     padding: Modal.Padding.l,
     width: Modal.Width.s,
+    closeOnEscape: false,
     transformOrigin: {
```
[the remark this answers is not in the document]
NicholasBoll: True

modules/modal/react/lib/Modal.tsx
```diff
@@ -10,12 +11,36 @@ export enum ModalWidth {
...
 export interface ModalProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * This will add a `data-testid` attribute to the Popup component, not the container
+   */
+  testId?: string;
   open: boolean;
-  padding: PopupPadding;
-  transformOrigin: TransformOrigin;
-  width: ModalWidth;
+  padding?: PopupPadding;
```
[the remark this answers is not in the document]
NicholasBoll: I think these changes were a part of an experiment to refactor modal to a Function Component using Hooks. This change is needed for that style. I'll revert

modules/modal/react/lib/Modal.tsx
NicholasBoll: That is added to the `Card` readme

modules/modal/react/lib/Modal.tsx
NicholasBoll: That's not handled by the `InputProvider`?

modules/modal/react/lib/Modal.tsx
```diff
@@ -41,6 +66,47 @@ const Container = styled('div')({
...
+        firstFocusable.tabIndex = 0;
+
+        const changeTabIndex = () => {
+          // We no longer need to focus on the header after it looses focus
+          // We simply want to transfer focus inside
+          firstFocusable.removeEventListener('blur', changeTabIndex);
+          // We must wait one frame to ensure tabbable checks are satisfied
+          // by all focus libraries...
+          requestAnimationFrame(() => {
+            firstFocusable.removeAttribute('tabIndex');
```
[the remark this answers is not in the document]
NicholasBoll: Perhaps https://github.com/NicholasBoll/blog/blob/master/contents/articles/cypress-tab-plugin/index.md#header-with-tabindex0?

modules/modal/react/lib/Modal.tsx
NicholasBoll: Good question. Without a popup manager in Canvas Kit, I wanted to support it, but not force it on people. In very simple cases, this feature works, but it can do unexpected things when enabled. I figured defaulting to `false` was the safest bet.

Should we make this default to `true`?

modules/modal/react/lib/Modal.tsx
```diff
@@ -70,22 +151,80 @@ export default class Modal extends React.Component<ModalProps> {
...
+ *       <Button {...targetProps}>Delete Item</Button>
+ *       <Modal heading='Delete Item' {...modalProps}>
+ *         Are you sure?
+ *         <Button onClick={closeModal}>Cancel</Button>
+ *       </Modal>
+ *     </>
+ *   )
+ * }
+ */
+export function useModal() {
```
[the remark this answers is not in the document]
NicholasBoll: This used to have more logic in it. When I added `focus-trap`, it handled some things for me. Still a useful hook and an example of composing across components. Another useful place for this pattern might be tooltips.

modules/modal/react/lib/Modal.tsx
```diff
@@ -70,22 +147,80 @@ export default class Modal extends React.Component<ModalProps> {
...
+  const [open, setOpen] = React.useState(false);
+
+  return {
+    targetProps: {
+      onClick(event: React.SyntheticEvent<HTMLElement>) {
```
NicholasBoll: I guess I don't need to be so explicit on the signature here... I used to save the element to refocus on close. `handleClose` doesn't have any restrictions. Thoughts?

modules/modal/react/package.json
```diff
@@ -38,6 +38,7 @@
     "@workday/canvas-kit-react-common": "^3.0.0-alpha.7",
     "@workday/canvas-kit-react-popup": "^3.0.0-alpha.7",
     "emotion": "^9.2.12",
+    "focus-trap-react": "^6.0.0",
```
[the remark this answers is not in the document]
NicholasBoll: Sure. I believe `focus-trap-react`, `focus-trap` and `tabbable` all are by David and are all seeking co-maintainers:
https://github.com/davidtheclark/focus-trap-react/issues/48
https://github.com/davidtheclark/tabbable/issues/42

modules/popup/react/lib/Popup.tsx
NicholasBoll: Hmm. Is there a benefit to exposing this id? We could add it later if we need to.

modules/modal/react/lib/Modal.tsx
anicholls: I think the majority of consumers would want it `true` by default. I would prefer to cater to the simpler use cases by default and provide advanced configuration if necessary

modules/modal/react/lib/Modal.tsx
anicholls: No, `InputProvider` just wraps everything with a div that indicates the input type. It does not provide any overriding styling to avoid being too obtrusive. It's up to each individual component to manage their own styling based off of the data attrs it provides.

In this case I think we want to hide the focus outlines at all times, so it shouldn't really be dependent on the `InputProvider` state anyway.

cypress/helpers/modal/modal.ts
lychyi: > We could move this file to `cypress/helpers/modal.ts`. I don't remember why I created a folder for this

I'm guessing better import ergonomics when you have an index file? Or at least that's what I would do...

modules/modal/react/lib/Modal.tsx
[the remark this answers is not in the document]
lychyi: I'm indifferent, if this isn't a documented prop, it can really change at anytime.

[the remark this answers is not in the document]
anicholls: Personally, I'd prefer having a slightly more brittle test (using `elemProps`) than adding undocumented props that aren't meant to be used in prod, but I'm not too torn either way.

modules/popup/react/lib/Popup.tsx
anicholls: Yeah I guess in my head I always want human readable anchor links for headings but that doesn't make sense in a modal anyways. We can leave out 👍

modules/modal/react/lib/Modal.tsx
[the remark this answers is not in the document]
NicholasBoll: I'll remove and this and we can make an official proposal later. This would be something that all components would have as a documented optional prop that would be used in production with the option to enable/disable.

In the platform, automation IDs are present even in the production HTML code. I think it is preferable to opt-in during the testing phase, but I would be fine if they are always present like we currently have in the product.

modules/modal/react/lib/Modal.tsx
NicholasBoll: Is putting `&:focus { outline: 'none';  }` a leak if it goes in `Card`? The `Card` component doesn't necessarily know focus is possible on the `header`. I can add a `firstFocusable.style.outline = 'none';`. That is more DOM manipulation, but this parent knows about what is happening in the child rather than the child needing to know what parents might be doing.

NicholasBoll: I've added `firstFocusable.style.outline = 'none';` below this line.

modules/modal/react/README.md
```diff
@@ -78,21 +80,14 @@ Default: `PopupPadding.l`
 
 ---
```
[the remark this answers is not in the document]
NicholasBoll: The normal example in the Readme isn't very good. I'll update that and use a `useModal` example as well.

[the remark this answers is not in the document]
NicholasBoll: Added https://github.com/Workday/canvas-kit/pull/184/files#diff-d1af53e1e2dfa8c07fe483a3fc45bf04R19-R55

modules/modal/react/lib/Modal.tsx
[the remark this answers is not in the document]
NicholasBoll: I have some conversation planned internally to see if I can publicize such blog posts

modules/common/react/lib/Popper.tsx
NicholasBoll: This change is actually needed by this line: https://github.com/Workday/canvas-kit/pull/184/files#diff-a92e8ec0649e74da7d56e8238af87821R28

Changing `anchorElement` from `Element` to `Element | null` allows this component to work much better with refs without a bunch of defensive code in the render method of a consumer

modules/text-input/react/lib/TextInput.tsx
```diff
@@ -1,11 +1,11 @@
 import * as React from 'react';
```
lychyi: This looks like a prettier/format only change right?

NicholasBoll: Oh, yes. I'll remove this from the PR

NicholasBoll: A bit tricky as precommit wanted to fix, but I don't want conflicts with other changes

modules/core/react/lib/InputProvider.tsx
```diff
@@ -196,6 +196,7 @@ export default class InputProvider extends React.Component<InputProviderProps, I
   }
 
   componentWillUnmount() {
+    window.clearTimeout(this.eventTimer);
```
NicholasBoll: Without this I was getting React `setState` errors on unmounted components

modules/popup/react/stories/stories.tsx
```diff
@@ -25,7 +25,7 @@ class PopupWrapper extends React.Component<{}, PopupWrapperState> {
         <Button variant={Button.Variant.Delete} onClick={this.handleClick}>
           Delete Item
         </Button>
-        <Popper placement={'bottom'} open={open} anchorEl={anchorEl}>
+        <Popper placement={'bottom'} open={open} anchorElement={anchorEl}>
```
NicholasBoll: I'm assuming we like `anchorElement` over `anchorEl`? Our Popper API is a little different than the material-ui one

PR #185, opened 2019-09-12
feat(modal): Implement CSS Modal to match React implementation

jamesfan:
This PR implements the Modal CSS styles so they match the React component. 

| Breaking Changes | No    | No   |

feat(modal): Implement CSS styles to match React implementation (#185)

### DIFF
--- a/modules/modal/css/README.md
+++ b/modules/modal/css/README.md
@@ -66,24 +66,3 @@ or `wdc-modal-no-padding` to set the padding to `0`.
   <div class="wdc-modal-body">Modal content</div>
 </div>
 ```
-
-### Animation Origin Modifiers
-
-Eight directions are available. Each direction will change the the animation origin when the modal
-enters the DOM. Use origin classes in addition to `.wdc-modal` to specify the direction.
-
-- `.wdc-modal-animation-origin-top-center`
-- `.wdc-modal-animation-origin-right-center`
-- `.wdc-modal-animation-origin-bottom-center`
-- `.wdc-modal-animation-origin-left-center`
-- `.wdc-modal-animation-origin-top-left`
-- `.wdc-modal-animation-origin-top-right`
-- `.wdc-modal-animation-origin-bottom-right`
-- `.wdc-modal-animation-origin-bottom-left`
-
-```html
-<div class="wdc-modal wdc-modal-animation-origin-top-center">
-  <div class="wdc-modal-title">Modal Title</div>
-  <div class="wdc-modal-body">Modal content</div>
-</div>
-```
--- a/modules/modal/css/README.md
+++ b/modules/modal/css/README.md
@@ -1,0 +1,37 @@
+# Canvas Kit Modal
+
+A Modal component that allows you to render a Popup with a translucent overlay.
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-css
+```
+
+or
+
+```sh
+yarn add @workday/canvas-kit-css-modal
+```
+
+Add your `node_modules` directory to your SASS `includePaths`. You will then be able to import
+`index.scss`.
+
+```scss
+@import '~@workday/canvas-kit-css-modal/index.scss';
+```
+
+## Usage
+
+Use `.wdc-modal` to create a modal container with the translucent overlay. Use `.wdc-popup` within
+the `.wdc-modal` to create the popup. See [Popup](?path=/story/css-popup--all) for instructions on
+how to configure the popup.
+
+```html
+<div class="wdc-modal">
+  <div class="wdc-popup">
+    <div class="wdc-popup-title">Popup Title</div>
+    <div class="wdc-popup-body">Popup content</div>
+  </div>
+</div>
+```
--- a/modules/modal/css/README.md
+++ b/modules/modal/css/README.md
@@ -38,21 +38,18 @@ respectively.
 
 ### With Close Button
 
-```jsx
-import {xIcon} from '@workday/canvas-system-icons-web';
-import {SystemIcon} from '@workday/canvas-kit-react';
-
-<div className="wdc-modal-bg">
-  <div className="wdc-modal">
-    <div className="wdc-modal-close">
-      <button onClick={this.onCloseClick} className="wdc-btn-icon-plain">
-        <SystemIcon icon={xIcon} />
+```html
+<div class="wdc-modal-bg">
+  <div class="wdc-modal">
+    <div class="wdc-modal-close">
+      <button onClick="{this.onCloseClick}" class="wdc-btn-icon-plain">
+        <i class="wdc-icon" data-icon="x" data-category="system" />
       </button>
     </div>
-    <div className="wdc-modal-title">Modal Title</div>
-    <div className="wdc-modal-body">Modal content</div>
+    <div class="wdc-modal-title">Modal Title</div>
+    <div class="wdc-modal-body">Modal content</div>
   </div>
-</div>;
+</div>
 ```
 
 ### Padding
--- a/modules/modal/css/README.md
+++ b/modules/modal/css/README.md
@@ -23,15 +23,67 @@ Add your `node_modules` directory to your SASS `includePaths`. You will then be 
 
 ## Usage
 
-Use `.wdc-modal` to create a modal container with the translucent overlay. Use `.wdc-popup` within
-the `.wdc-modal` to create the popup. See [Popup](?path=/story/css-popup--all) for instructions on
-how to configure the popup.
+Use `.wdc-modal-bg` to create the background overlay and `.wdc-modal` to create the modal itself.
+The title and body content can be styled using `.wdc-modal-title` and `.wdc-modal-body`,
+respectively.
 
 ```html
-<div class="wdc-modal">
-  <div class="wdc-popup">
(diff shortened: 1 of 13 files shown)


modules/modal/css/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+## Usage
+
+Use `.wdc-modal` to create a modal container with the translucent overlay. Use `.wdc-popup` within
+the `.wdc-modal` to create the popup. See [Popup](?path=/story/css-popup--all) for instructions on
+how to configure the popup.
+
+```html
+<div class="wdc-modal">
```
jamesfan: I implemented the CSS Modal as a container element around the existing CSS Popup markup structure. Let me know if this is ok or if anyone else had another approach in mind.

modules/modal/css/README.md
```diff
@@ -0,0 +1,37 @@
...
+`index.scss`.
+
+```scss
+@import '~@workday/canvas-kit-css-modal/index.scss';
+```
+
+## Usage
+
+Use `.wdc-modal` to create a modal container with the translucent overlay. Use `.wdc-popup` within
+the `.wdc-modal` to create the popup. See [Popup](?path=/story/css-popup--all) for instructions on
```
jamesfan: What's the proper way to link to another component in the README? I noticed a few of the READMEs use the following syntax...

`[canvas-kit-core](../../core/css)`

... but the links don't actually work (at least not locally).

jamesfan:
@mannycarrera4 @anicholls I noticed a few issues with CSS Popup as I was implementing CSS Modal:

1. The font-families for the popup title and body aren't being set. Should we `@include wdc-type()` in `.wdc-popup`?

2. The README for CSS Popup doesn't apply `.wdc-popup-body` to the content `<div>`s. They're just plain classless `<div>`s.

3. The `Card` component used in React Popup sets `box-sizing: border-box`, but the `card-common()` mixin included in `.wdc-popup` doesn't set `box-sizing`, so setting a width of `400` in the React Popup vs the CSS Popup results in different size popups.

Let me know how/if you want to handle these.

modules/popup/css/lib/popup.scss
```diff
@@ -1,33 +1,37 @@
-.wdc-popup {
+@mixin wdc-popup-common() {
```
jamesfan: I named the mixin `wdc-popup-common` since it seemed like that was the convention, but I can remove `-common` if you'd prefer.

modules/popup/css/lib/popup.scss
```diff
@@ -1,33 +1,37 @@
...
+  }
+
+  &-close {
     position: absolute;
     right: $wdc-spacing-xs;
     top: $wdc-spacing-xs;
   }
 
-  .wdc-popup-title {
+  &-title {
```
jamesfan: Rather than define separate `wdc-popup-title` and `wdc-popup-body` mixins, I used the ampersand within the `wdc-popup-common` mixin.

jamesfan:
@anicholls see comments above. Let me know if you notice anything else.

modules/modal/css/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+```jsx
+import {xIcon} from '@workday/canvas-system-icons-web';
+import {SystemIcon} from '@workday/canvas-kit-react';
+
+<div className="wdc-modal-bg">
+  <div className="wdc-modal">
+    <div className="wdc-modal-close">
+      <button onClick={this.onCloseClick} className="wdc-btn-icon-plain">
+        <SystemIcon icon={xIcon} />
```
anicholls: Can we update this to use `<i className="wdc-icon" data-icon="xIcon" data-category="system" />` and update the example to be plain html?

modules/modal/css/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "index.scss"
+  ],
+  "style": "dist/canvas-kit-css-modal.min.css",
+  "main": "dist/canvas-kit-css-modal.min.css",
+  "repository": {
+    "type": "git",
+    "url": "http://github.com/Workday/canvas-kit/tree/master/modules/modal/css"
+  },
+  "dependencies": {
+    "@workday/canvas-kit-css-core": "^3.0.0-alpha.6"
```
anicholls: Missing deps here

modules/modal/css/lib/modal.scss
```diff
@@ -0,0 +1,26 @@
...
+  height: 100vh;
+  justify-content: center;
+  left: 0;
+  position: fixed;
+  top: 0;
+  width: 100vw;
+}
+
+.wdc-modal {
+  @include wdc-popup-common();
```
anicholls: Can you replicate the animation from `.wdc-modal-animation-origin-bottom-center`? This should match the React modal

modules/popup/css/README.md
```diff
@@ -39,21 +39,21 @@ title. `p` elements also have adjusted margins.
 import {xIcon} from '@workday/canvas-system-icons-web';
 import {SystemIcon} from '@workday/canvas-kit-react';
 
-<div class="wdc-popup">
+<div className="wdc-popup">
   <div className="wdc-popup-close">
     <button onClick={this.onCloseClick} className="wdc-btn-icon-plain">
       <SystemIcon icon={xIcon} />
```
anicholls: While you're at it do you mind updating this one too? We should always aim for ````html` code blocks in our CSS readmes.

modules/popup/css/README.md
```diff
@@ -39,21 +39,21 @@ title. `p` elements also have adjusted margins.
...
     </button>
   </div>
-  <div class="wdc-popup-title">Popup Title</div>
-  <div class="wdc-popup-body">Popup content</div>
+  <div className="wdc-popup-title">Popup Title</div>
```
anicholls: This will get reverted at the same time.

modules/modal/css/README.md
```diff
@@ -0,0 +1,89 @@
...
+or `wdc-modal-no-padding` to set the padding to `0`.
+
+```html
+<div class="wdc-modal wdc-modal-no-padding">
+  <div class="wdc-modal-title">Modal Title</div>
+  <div class="wdc-modal-body">Modal content</div>
+</div>
+```
+
+### Animation Origin Modifiers
```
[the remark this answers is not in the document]
jamesfan: Just confirming: the React Modal has a `transformOrigin` prop. Do you still want me to remove animation origins from the CSS Modal?

modules/modal/css/README.md
jamesfan: Sure, I'll update the example in the `README` -- should I need to update the code in `stories.tsx` as well?

modules/modal/css/package.json
jamesfan: I take it we need a `"@workday/canvas-kit-css-popup": "^3.0.0-alpha.6"` here? Does the CSS Popup `package.json` need `"@workday/canvas-kit-css-card": "^3.0.0-alpha.6"` as well then?

Is there an error or process that flags missing dependencies or is this something we need to be mindful of doing ourselves?

modules/modal/css/README.md
anicholls: I'm less concerned about the stories because they're already in jsx, but we might as well so they match

modules/modal/css/README.md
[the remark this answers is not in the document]
anicholls: Yes, I think this was an oversight in the development of `Modal`. We just added it because it was there on popup. I'm going to add a note to #51 to remove this before 3.0 final. Modals should always animate from the bottom up.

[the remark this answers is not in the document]
jamesfan: Got it 👍

modules/modal/css/README.md
jamesfan: Sounds good, the buttons will revert to their non-beta styles then since I don't think we have the beta buttons in CSS yet...?

modules/modal/css/package.json
anicholls: Yep exactly. You are correct - good catch! Do you mind adding it?

It's in our roadmap to add CI to check for missing deps, but we don't have it yet so we need to be mindful (for now 😅)

jamesfan: Sure thing. On a related note, does the CSS Popup `index.scss` need to import CSS Button?

https://github.com/Workday/canvas-kit/blob/37160e03a87948cea865f08abf5a13272a262669/modules/popup/css/index.scss#L3

The two are often used together, but unless I'm missing something I don't think Popup needs the Button styles.

modules/modal/css/README.md
jamesfan: Never mind, I realize you were probably just referring to updating the story code for the `xIcon` to be html and not the entire story itself. I left the beta buttons in.

modules/modal/css/package.json
jamesfan: I updated the dependencies in both `package.json`s. I also removed the CSS Button import from the CSS Popup `index.scss`.

jamesfan:
@anicholls I've implemented the latest round of fixes. See [comments](https://github.com/Workday/canvas-kit/pull/185#discussion_r324400325) regarding the importing of CSS Button into the CSS Popup `index.scss`.

Issue #158, opened 2019-08-29
Checkbox requires an id if label is enabled


jeffma123:
## 🐛 Bug Report

The checkbox should not require an id.

## To Reproduce

Create multiple checkboxes. Clicking on the label only affects the first checkbox.

## Expected Behavior

Each label should work independently.

lychyi:
@jeffma123 Is the bug that a checkbox shouldn't require an `id`? Or is it the fact that multiple checkboxes do not function correctly if you don't supply an `id`?

anicholls:
@lychyi from @NicholasBoll on slack:
> `defaultProps` sets the `id` to a `uuid`, but that function is only evaluated for the component and not for each instance.
> The fix is pretty simple - make `id` be an instance variable on the `Checkbox` class, get rid of `defaultProps.id` and update the render function to use `id || this.id`. That will guarantee a unique `id` for each instance that is overridable.

PR #192, opened 2019-09-16
fix(checkbox): Change default id to be unique per instance

mannycarrera4:
## Summary
Fixes: https://github.com/Workday/canvas-kit/issues/158

### DIFF
--- a/modules/checkbox/react/lib/Checkbox.tsx
+++ b/modules/checkbox/react/lib/Checkbox.tsx
@@ -265,7 +265,7 @@ export default class Checkbox extends React.Component<CheckboxProps> {
           </CheckboxBackground>
         </CheckboxInputWrapper>
         {label && (
-          <CheckboxLabel htmlFor={id} disabled={disabled}>
+          <CheckboxLabel htmlFor={id || this.id} disabled={disabled}>
             {label}
           </CheckboxLabel>
         )}
--- a/modules/checkbox/react/lib/Checkbox.tsx
+++ b/modules/checkbox/react/lib/Checkbox.tsx
@@ -229,7 +229,9 @@ export default class Checkbox extends React.Component<CheckboxProps> {
     checked: false,
     label: '',
   };
+
   private id = uuid();
+
   public render() {
     // TODO: Standardize on prop spread location (see #150)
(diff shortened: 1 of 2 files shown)


modules/checkbox/react/lib/Checkbox.tsx
```diff
@@ -251,7 +250,7 @@ export default class Checkbox extends React.Component<CheckboxProps> {
           <CheckboxInput
             checked={checked}
             disabled={disabled}
-            id={id}
+            id={id || this.id}
```
mannycarrera4: Should we add a story with multiple checboxes? @NicholasBoll

anicholls: Nope, but we should add a test for it @mannycarrera4

modules/checkbox/react/lib/Checkbox.tsx
```diff
@@ -228,9 +228,8 @@ export default class Checkbox extends React.Component<CheckboxProps> {
   public static defaultProps = {
     checked: false,
     label: '',
-    id: uuid(),
   };
-
+  private id = uuid();
```
anicholls: Nit: can you put new lines on either side of this? We generally put a space between all class members/functions

lychyi:
@mannycarrera4 this one's been kind of stale for a while, are we waiting on something for this?

stephanerangaya:
> @mannycarrera4 this one's been kind of stale for a while, are we waiting on something for this?

Looks like the only thing that was blocking this was writing a test, which @mannycarrera4 did. I think it's good to be merged.

modules/checkbox/react/lib/Checkbox.tsx
```diff
@@ -251,7 +252,7 @@ export default class Checkbox extends React.Component<CheckboxProps> {
           <CheckboxInput
             checked={checked}
             disabled={disabled}
-            id={id}
+            id={id || this.id}
```
lychyi: Nit: when we destructure `id` out of `props` (L240) we should just assign `this.id` as a default value. This way if this logic ever needs to change you're not changing it in two places (here and L270).

PR #193, opened 2019-09-16
DNM: Decompose text input

MackenzieBerliner-Glasser:
## Summary

The idea for this PR is to make an `Input` module that has the canvas styling and that's it. Then we will import the Input component into the `TextInput` component to prevent any breaking changes.


modules/input/react/lib/Input.tsx
```diff
@@ -0,0 +1,79 @@
...
+export interface InputProps extends GrowthBehavior, React.InputHTMLAttributes<HTMLInputElement> {
+  disabled?: boolean;
+  error?: ErrorType;
+  hasIcon?: boolean;
+  inputRef?: React.Ref<HTMLInputElement>;
+  onChange?: React.ChangeEventHandler<HTMLInputElement>;
+  placeholder?: string;
+  readOnly?: boolean;
+  type?: string;
+  value?: any;
```
NicholasBoll: I'm surprised we don't disallow `any`. It is a bit more work, but would a generic type be more appropriate? Otherwise `HTMLInputElement` types `value` as a `string` since the input element can only represent strings. 

Supporting other types as a higher level concept would require serializing and deserializing to satisfy the `<input>` element. And serializing/deserializing requires care to not change what the user is typing while typing. Ex: a DateInput wrapping this input serialize/deserialize to a `Date` object not changing the characters representing the date while the user is interacting. One could imagine the frustration. That might require a whole lifecycle subsystem of parsing, validation and serializing to get right.

I don't have much context here. How is this handled elsewhere?

modules/text-input/react/lib/TextInput.tsx
```diff
@@ -1,10 +1,10 @@
 import * as React from 'react';
-import styled from 'react-emotion';
-import {GrowthBehavior, ErrorType, errorRing} from '@workday/canvas-kit-react-common';
-import {inputColors, spacingNumbers, type} from '@workday/canvas-kit-react-core';
+import {GrowthBehavior, ErrorType} from '@workday/canvas-kit-react-common';
+import {inputColors} from '@workday/canvas-kit-react-core';
 import {SystemIcon} from '@workday/canvas-kit-react-icon';
 import {exclamationCircleIcon, exclamationTriangleIcon} from '@workday/canvas-system-icons-web';
 import InputIconContainer from './InputIconContainer';
+import Input from '../../../input/react/index';
```
NicholasBoll: This import should match the others in this file:

```suggestion
import Input from '@workday/canvas-kit-react-input';
```

modules/input/react/lib/Input.tsx
NicholasBoll: @anicholls @lychyi Do you have ideas or past experience with this before?

I've done low-level inputs that represented higher-level types like `number`, `Date` and other 2-way transformative types in the past. They are tricky, but doable. Best to solve at the lowest layer possible. AngularJS 1.x was originally created just to tackle forms (validation, parsing, etc)

modules/input/react/README.md
```diff
@@ -0,0 +1,119 @@
+# Canvas Kit Input
+
+Canvas-styled React components.
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-react
```
Parker-Ledoux: `yarn add @workday/canvas-kit-react-input`

modules/_canvas-kit-react/package.json
```diff
@@ -50,6 +50,7 @@
     "@workday/canvas-kit-react-form-field": "^3.0.0-alpha.6",
     "@workday/canvas-kit-react-header": "^3.0.0-alpha.6",
     "@workday/canvas-kit-react-icon": "^3.0.0-alpha.6",
+    "@workday/canvas-kit-react-input": "0.0.0",
```
Parker-Ledoux: This version should match the others.

modules/input/react/README.md
```diff
@@ -0,0 +1,119 @@
+# Canvas Kit Input
+
+Canvas-styled React components.
```
Parker-Ledoux: "A generic canvas-styled React input component."

modules/input/react/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@workday/canvas-kit-react-input",
+  "version": "0.0.0",
```
Parker-Ledoux: This version should also match the other modules.

modules/input/react/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "description": "generalized input component with canvas styling",
```
Parker-Ledoux: "Canvas styled input"

modules/input/react/README.md
NicholasBoll: `yarn add @workday/canvas-kit-react-input` is added on line 14

modules/_canvas-kit-react/package.json
NicholasBoll: @Parker-Ledoux @MackenzieBerliner-Glasser this has to be `0.0.0` before a release is on npm. This instructs yarn to only look at local files. Changing this will fail CI

modules/input/react/package.json
NicholasBoll: @Parker-Ledoux @MackenzieBerliner-Glasser this has to be `0.0.0` before a release is on npm. This instructs yarn to only look at local files. Changing this will fail CI

Parker-Ledoux:
We changed the approach so closing this PR.

Issue #89, opened 2019-08-07
Add required asterisk to input labels


anicholls:
## 🚀 Feature Proposal

Form labels should show an asterix for those fields that are required.

Color should use `cinammon500` (`canvas.inputColors.error`)

![Screen Shot 2019-08-07 at 4 20 18 PM](https://user-images.githubusercontent.com/908478/62664136-442e5780-b92f-11e9-9c2f-0ca0c268df22.png)

## Example

```
<FormField label="Label" inputId="my-checkbox-field" required={true}>
  <Checkbox id="1" label="Checkbox option" />
</FormField>
```

PR #196, opened 2019-09-17
feat(form-field): Add red asterisk support for required inputs

elliot-at-workday:
Adds a red astrisk to required formfields.
Closes https://github.com/Workday/canvas-kit/issues/89

I'm not sure of the process of going through design and a11y reviews when contributing this repo. Is there a section of the guide for that?

## Summary

As someone who like to indicate the status of fields in a UI to help people when filling out forms I would like the ability to mark some fields as required. This PR implements styling on those widgets.

## Checklist

- [x] `yarn test` passes

## Additional References
Required Text Input


lychyi:
@elliot-at-workday That's on me to change that on the checklist. It's really a reminder for the core team since we interface with the UX team at Workday in person. Don't worry about those ticks, we'll address them or provide guidance when required.

Thanks for the contribution!

modules/form-field/react/lib/Label.tsx
```diff
@@ -16,19 +17,32 @@ const labelStyles = [
...
+          }
+        : {
+            display: 'block',
+            marginBottom: spacing.xxxs,
+          }),
+      ...(props.required
+        ? {
+            ':after': {
+              content: '"*"',
+              color: inputColors.error.border,
```
anicholls: Nit: Do you mind using `cinnamon500` instead? We try to avoid using the semantic colors for things they're not intended for.

modules/form-field/react/lib/Label.tsx
```diff
@@ -16,19 +17,32 @@ const labelStyles = [
     padding: 0,
   },
   (props: LabelProps) => {
```
anicholls: Nit: I know it's not something you touched here, but do you mind destructuring the props like we do in other components? (i.e. `({labelPosition, required}) => ({`)

modules/form-field/react/lib/Label.tsx
```diff
@@ -16,19 +17,32 @@ const labelStyles = [
...
+            marginBottom: spacing.xxxs,
+          }),
+      ...(props.required
+        ? {
+            ':after': {
+              content: '"*"',
+              color: inputColors.error.border,
+              fontSize: '16px',
+              fontWeight: 400,
+              position: 'relative' as 'relative',
```
anicholls: What is the `as 'relative'` here for?

modules/form-field/react/stories/stories_ColorInput.tsx
```diff
@@ -23,6 +23,11 @@ storiesOf('Form Field/Color Picker/Color Input/Top Label', module)
       {controlComponent(<ColorInput disabled={true} />)}
     </FormField>
   ))
+  .add('Required', () => (
+    <FormField label="Label" inputId="input-required" required={true}>
+      {controlComponent(<ColorInput />)}{' '}
```
anicholls: What is the `{' '}` here for?

modules/form-field/react/stories/stories_ColorInput.tsx
```diff
@@ -23,6 +23,11 @@ storiesOf('Form Field/Color Picker/Color Input/Top Label', module)
       {controlComponent(<ColorInput disabled={true} />)}
     </FormField>
   ))
+  .add('Required', () => (
+    <FormField label="Label" inputId="input-required" required={true}>
```
anicholls: This isn't actually a variant of the various form elements, so do you mind reverting these extra stories and just adding another `FormField/Label` story? This will require you to rename the current one (i.e. `FormField/Label/Default` and `FormField/Label/Required`)

modules/form-field/react/stories/stories_ColorInput.tsx
elliot-at-workday: Likely unintended, I'll remove it

modules/form-field/react/stories/stories_ColorInput.tsx
elliot-at-workday: Makes sense! I noticed that there aren't a lot of knobs in the stories, why is that?

anicholls: Yeah - the main reason is that our storybook is integrated into the [Workday design site](http://design.workday.com). It's currently set up to just render different stories, with no access to the knobs. If we put everything behind knobs, we couldn't show the different variants in this context.

We're currently evaluating our entire developer experience (including documentation integrating with storybook), and this is one portion we're looking at. So it's on our radar, but we're not making much use of knobs yet for this reason. Hopefully that answers your question!

elliot-at-workday: It does, thanks!

modules/form-field/react/lib/Label.tsx
elliot-at-workday: Nothing, removing

elliot-at-workday:
Reacted to all PR comments, I'm most uncertain of the storybook changes.

modules/form-field/react/stories/stories.tsx
```diff
@@ -6,17 +6,22 @@ import withReadme from 'storybook-readme/with-readme';
...
-  .add('Hint', () => (
+  .add('Required', () => (
     <div className="story">
-      <Hint>Hint</Hint>
-      <Hint error={Hint.ErrorType.Error}>Hint</Hint>
-      <Hint error={Hint.ErrorType.Alert}>Hint</Hint>
+      <Label required={true}>Label</Label>
     </div>
   ));
+storiesOf('Form Field/Hint', module).add('Hint', () => (
```
anicholls: This is close, but now we have two levels of nesting for Hint when there's only one story at the bottom level. We're also missing the Readme here now. Can you change this to:
```jsx
storiesOf('Form Field', module)
  .add('Hint', () => (
```

anicholls:
I chatted with a11y about the best way to implement this and they said we should avoid using `:after` (see https://www.w3.org/WAI/WCAG21/Techniques/failures/F87.html)

The correct method would be to use a nested `<abbr>` element inside the label (see https://www.w3.org/WAI/WCAG21/Techniques/html/H90.html#example-2-using-an-asterisk-to-indicate-required-state). 

Do you mind updating to this method @elliot-at-workday ? Let me know if you have any questions.

PR #206, opened 2019-09-18
fix(avatar): Add AvatarButton component and convert Avatar into ordinary div

### LINKED ISSUE #180: Avatar component should render as a div by default and button only if onClick props is defined
## 🐛 Bug Report

The Avatar component is rendered as a button whether or not its onClick props is defined. 
However it can be used as a visual element only or it can already be included in a button in which case it would trigger the following warning error in the browser console:
`Warning: validateDOMNesting(...): <button> cannot appear as a descendant of <button>.`
The Avatar component would need to be rendered as a div by default and as a button only if its onClick props is defined.

## To Reproduce

Try to add the Avatar component with no onClick props as a child of a button element.

## Expected Behavior
If its onClick props is not defined, the Avatar component should be rendered as a div.
If its onClick props is defined, the Avatar component should be rendered as a button.

## Actual Results

Avatar component with no onClick props is rendered as a button.
The following warning error in the browser console:
`Warning: validateDOMNesting(...): <button> cannot appear as a descendant of <button>.`

### Error Output

Screenshot 2019-09-11 at 12 31 38

Screenshot 2019-09-11 at 12 31 05

6r3al:
## Summary

- Avatar component is now rendered as a `<div>`
- AvatarButton component has been introduced and is rendered as a `<button>`

### Breaking Code Change
- `Avatar` is now just a `div` container. To get the previous button functionality when using `onClick`, use `AvatarButton`

Closes #180 

## Checklist

- [x] `yarn test` passes
      applicable


modules/avatar/react/README.md
```diff
@@ -102,6 +101,98 @@ Default: `Avatar.Size.m`
...
+  onClick={() => { window.alert('AvatarButton Clicked') }}
+/>
+
+// Using AvatarVariant import directly
+<AvatarButton size={AvatarButton.Size.xs} variant={AvatarVariant.Dark} />
+```
+
+## Static Properties
+
+#### `variant: AvatarVariant`
```
anicholls: `variant` should be capitalized

modules/avatar/react/lib/Avatar.tsx
```diff
@@ -28,17 +27,9 @@ export interface AvatarProps extends React.ButtonHTMLAttributes<HTMLButtonElemen
...
-  /**
-   * Ref of button that the styled component renders.
-   */
-  buttonRef?: React.Ref<HTMLButtonElement>;
 }
 
-const Container = styled('button', {
+const Container = styled('div', {
   shouldForwardProp: prop => isPropValid(prop) && prop !== 'size',
 })<Omit<AvatarProps, 'altText'>>(
```
anicholls: Nit: Can you change this to `<Pick<AvatarProps, 'size'>>`?

modules/avatar/react/lib/AvatarButton.tsx
```diff
@@ -0,0 +1,91 @@
+import * as React from 'react';
+import styled from 'react-emotion';
+import isPropValid from '@emotion/is-prop-valid';
+import Avatar, {AvatarVariant} from './Avatar';
+import {colors} from '@workday/canvas-kit-react-core';
+import {focusRing, hideMouseFocus} from '@workday/canvas-kit-react-common';
+import {SystemIconCircleSize} from '@workday/canvas-kit-react-icon';
+
+export interface AvatarButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
```
anicholls: Can you update this to `React.HTMLAttributes`? This was an issue from before (see https://github.com/Workday/canvas-kit/issues/162)

anicholls: Also can you just extend `AvatarProps` instead of duplicating variant, size, altText and url?

modules/avatar/react/lib/AvatarButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+  onClick?: (e: React.SyntheticEvent) => void;
+  /**
+   * Ref of button that the styled component renders.
+   */
+  buttonRef?: React.Ref<HTMLButtonElement>;
+}
+
+const Container = styled('button', {
+  shouldForwardProp: prop => isPropValid(prop) && prop !== 'size',
+})<Omit<AvatarButtonProps, 'altText'>>(
```
anicholls: Nit: `<Pick<AvatarButtonProps, 'variant' | 'size' | 'onClick'>>`

modules/avatar/react/lib/AvatarButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+  {
+    display: 'flex',
```
anicholls: None of these base styles are needed except for `borderRadius`. They are all handled by the internal `Avatar`. All you need are the styles in the function below

modules/avatar/react/stories/stories_avatarButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+import * as React from 'react';
+import {storiesOf} from '@storybook/react';
+import withReadme from 'storybook-readme/with-readme';
+
+import {AvatarButton} from '../index';
+import README from '../README.md';
+
+const IMAGE_URL = 'https://s3-us-west-2.amazonaws.com/design-assets-internal/avatars/lmcneil.png';
+
+storiesOf('AvatarButton', module)
```
anicholls: Can you put all of these under Avatar (`storiesOf('Avatar/AvatarButton', module)`)? You'll have to update the other story file as well.
```
Avatar/
  Default/
    Light
    Dark
    Image
  AvatarButton/
    Light
    Dark
    Image
```

modules/avatar/react/stories/stories_avatarButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+  .addDecorator(withReadme(README))
+  .add('Light', () => (
+    <div className="story">
+      <h3>Extra-Extra Large</h3>
+      <AvatarButton size={AvatarButton.Size.xxl} />
```
anicholls: Do you mind adding a mock `onClick` to each of these so they're not disabled?

modules/avatar/react/README.md
6r3al: isn't `variant` a props there?

6r3al: sure

modules/avatar/react/lib/Avatar.tsx
6r3al: sure

modules/avatar/react/lib/AvatarButton.tsx
NicholasBoll: > Can you update this to React.HTMLAttributes? This was an issue from before (see #162)
@anicholls It is done correctly here. The style is:

```ts
React.xHTMLAttributes<HTMLxElement>
```

where `x` is the element type (should it exist). In this case, props like `form` will exist for `AvatarButton`. Using `HTMLElement`, those extra attributes for the `<button>` element won't exist and people will get type error using them even though it is valid to do so.

> Also can you just extend AvatarProps instead of duplicating variant, size, altText and url?
```suggestion
export interface AvatarButtonProps extends AvatarProps, React.ButtonHTMLAttributes<HTMLButtonElement> {
```

modules/avatar/react/lib/AvatarButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+   */
+  altText: string;
+  /**
+   * The url of the users avatar photo
+   */
+  url?: string;
+  /**
+   * An event handler function that gets called when the avatar is clicked
+   */
+  onClick?: (e: React.SyntheticEvent) => void;
```
NicholasBoll: This is typed incorrectly:
```ts
// `e.currentTarget` is an HTMLButtonElement
<button onClick={e => e.currentTarget.form}></button>

// Typescript error: "Property 'form' does not exist on type 'EventTarget & Element'. ts(2339)"
<AvatarButton onClick={e => e.currentTarget.form}></button>
```

Either we should omit properties already handled by the element we are extending (`React.ButtonHTMLElement<HTMLButtonElement>`) OR type it correctly (so we can control documentation).

```suggestion
  onClick?: React.MouseEventHandler<T>
```

This is how `<button>` is typed internally in React.

modules/avatar/react/lib/AvatarButton.tsx
NicholasBoll: @anicholls What do you think of using Emotion's `as` to add and change some styles? https://emotion.sh/docs/styled#as-prop

It seems like we are extending `Avatar` to add additional styles and change the tag used. As this is currently implemented, we will have this DOM structure: `button > div > img|svg`. We don't need that extra `div`

modules/avatar/react/lib/AvatarButton.tsx
anicholls: Whoops - sorry I had it backwards! You can follow @NicholasBoll recommendation above

modules/avatar/react/lib/AvatarButton.tsx
anicholls: The `as` prop is [new in Emotion 10](https://medium.com/emotion-js/announcing-emotion-10-f1a4b17b8ccd#f8b7), so unfortunately we can't do this yet 😞

6r3al: ok but seems like 
```
padding: 0,
border: 0,
```
are needed too, will leave those as well

modules/avatar/react/lib/AvatarButton.tsx
6r3al: cool

modules/avatar/react/stories/stories_avatarButton.tsx
6r3al: sure

modules/avatar/react/stories/stories_avatarButton.tsx
6r3al: cool

modules/avatar/react/lib/AvatarButton.tsx
anicholls: Ah, yep - sorry forgot about the default button styling 👍

NicholasBoll: Ah. I think the API is `withComponent` that would work in lower versions

modules/avatar/react/lib/AvatarButton.tsx
6r3al: Ok but by doing that, I receive couple of warnings like this one:
```
Interface 'AvatarButtonProps' cannot simultaneously extend types 'AvatarProps' and 'ButtonHTMLAttributes<HTMLButtonElement>'
Named property 'onAbort' of types 'AvatarProps' and 'ButtonHTMLAttributes<HTMLButtonElement>' are not identical.
```
Any idea?

modules/avatar/react/lib/AvatarButton.tsx
6r3al: Thanks, added:
`onClick?: React.MouseEventHandler<HTMLButtonElement>;`

modules/avatar/react/lib/AvatarButton.tsx
anicholls: Ah shoot - because `AvatarProps` extends the div attributes already. Not sure if `AvatarLocalProps` is the best naming for this, but you could do 

```tsx
// Avatar.tsx
export interface AvatarLocalProps {
  // Avatar specific props
}

export interface AvatarProps extends AvatarLocalProps, React.HTMLAttributes<HTMLDivElement> {}


// AvatarButton.tsx
export interface AvatarButtonProps extends AvatarLocalProps, React.ButtonHTMLAttributes<HTMLButtonElement> {
  // buttonRef and onClick
}
```

Another alternative is to use `Pick`. There's some duplication, but at least the comments and types are kept in `Avatar`:
```tsx
export interface AvatarButton extends <Pick<AvatarProps, 'variant' | 'size' | 'altText' | 'url'>>, React.ButtonHTMLAttributes<HTMLButtonElement> {
```
It's a bit easier to understand at first glance, but is more prone to errors in the future.

Thoughts @NicholasBoll @6r3al? The duplication in here currently isn't a blocker, I was just trying to avoid manual errors if/when this gets updated in the future.

modules/avatar/react/lib/AvatarButton.tsx
6r3al: You have a point @NicholasBoll, the current structure for the rendered AvatarButton is now:
```
<button aria-label="Avatar">
  <div aria-label="Avatar">
    <img src="imageUrl" alt="Avatar">
  </div>
</button>
```
Which is not the best. 
Any recommendation here @anicholls?

modules/avatar/react/lib/AvatarButton.tsx
6r3al: Thanks Alex, I went with the first approach.

modules/avatar/react/lib/AvatarButton.tsx
anicholls: Yes, let's use [`withComponent`](https://emotion.sh/docs/styled#change-the-rendered-tag-using-withcomponent)

modules/avatar/react/lib/AvatarButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+        size={size}
+        url={url}
+        altText={altText}
+        onClick={onClick}
+        disabled={onClick ? false : true}
+        innerRef={buttonRef}
+        aria-label={altText}
+        {...elemProps}
+      >
+        {url ? (
```
anicholls: So this works as is, but unfortunately we need to duplicate the children. @NicholasBoll and I came up with a better option using a factory pattern to generate the interface that would allow us to do something like:

```tsx
return (
  <Avatar
    as="button"
    variant={variant}
    size={size}
    url={url}
    altText={altText}
    onClick={onClick}
    disabled={onClick ? false : true}
    innerRef={buttonRef}
    aria-label={altText}
    {...elemProps}
  />
)
```
Note the `as="button"` and the lack of duplicated children.

However, this may take some experimenting and I'm in a ton of meetings this week as I'm visiting the Boulder office. I know this is blocking you right now, so I'm going to merge as is and come back to it.

PR #218, opened 2019-09-24
refactor(text-input): Remove error icon in React

MackenzieBerliner-Glasser:
As `text-input` are the only inputs to display an error/alert icon, we would like to remove it so all of our inputs are displaying errors and alerts in a similar way. This also affects the `color-picker` component because it relies on the `text-input`.

### Breaking Visual Changes
- Error and alert icons are not longer displayed inside `text-input` component
- `ColorInput` width no longer changes with an error

### Breaking Code Changes
- `TextInput` is no longer wrapped by a `div`. Extra props are passed to the `<input>` element which is now the container. If code previous relied on this organization, this will be a visual breaking change.

refactor(text-input): Remove error icon in React (#218)

BREAKING CHANGE:
- `TextInput` is no longer wrapped by a `div`. Extra props are passed to the `<input>` element which is now the container. If code previous relied on this organization, this will be a visual breaking change.
- Error and alert icons are not longer displayed inside `TextInput` component
- `ColorInput` width no longer changes with an error


modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -26,7 +26,6 @@ const swatchTileSpacing = spacing.xxs;
 const swatchTileSize = 20;
 const swatchCheckIconSpacing = 8;
 const colorInputWidth = 116;
-const colorInputErrorWidth = 144;
 
 const CustomHexInput = styled(TextInput)<Pick<ColorInputProps, 'disabled' | 'error' | 'grow'>>(
```
stephanerangaya: Because `error` is no longer needed, could you remove this mention of `'error'` here as well as the `ErrorType` import on line 7 and in the interface on line 17?

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -14,7 +9,6 @@ import TextInput, {TextInputProps} from '@workday/canvas-kit-react-text-input';
 
 export interface ColorInputProps extends TextInputProps, GrowthBehavior {
   value: string;
-  error?: ErrorType;
```
anicholls: Nit: Technically this works because we spread `elemProps` onto the input component, but I still think we should expose `error` as a prop here just to be clear. It's the same as any other input specific props already included in `TextInputProps` that we also call out here.

This also involves pulling it out of `this.props` in `render` (which it looks like we had missed before):
```tsx
        const {

          //...

          error,
          ...elemProps
        } = this.props;

        //...

        <CustomHexInput
          inputRef={inputRef}
          onChange={this.handleChange}
          type="text"
          placeholder={placeholder}
          value={formattedValue}
          error={error}
          spellCheck={false}
          disabled={disabled}
          maxLength={7} // 7 to allow pasting with a hash
          {...elemProps}
        />
```

stephanerangaya:
I'm going to make these changes to the CSS styles as well, please wait before merging.

stephanerangaya:
We're going to update our CSS styles to match these changes in this PR https://github.com/Workday/canvas-kit/pull/222

lychyi:
Blocked by #222

modules/color-picker/react/lib/ColorInput.tsx
stephanerangaya: I made this change

NicholasBoll:
> Blocked by #222

Is this blocked by or related to #222?

modules/color-picker/react/lib/ColorInput.tsx
NicholasBoll: It was moved down a few. It is still included now

Issue #226, opened 2019-09-27
Fix UUID behavior across all components that use it


lychyi:
## 🐛 Bug Report

Remove `uuid` calls in `defaultProps` (static) and move them to instance variables so that they'll actually be random per instance.

This was an error that wasn't caught for multiple components: `Radio`, `Menu`, and `Checkbox`. And `Checkbox` is being addressed in #158.

## To Reproduce

Stamp out multiple `Radio` components without passing in an `id` prop. Observe that each `Radio` id is the same.

## Expected Behavior

Every new component mentioned above should have a random id if one isn't assigned to it.

## Actual Results

Each instance will stamp out the same uuid if it is not assigned one.

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

## Link to repl or repo (highly encouraged)

https://codesandbox.io/s/uuid-bug-14ric

### Error Output

Paste the error output here

```bash

```

or

Include a screenshot of a visual bug

stephanerangaya:
Do we need to do that to `menu`?

PR #228, opened 2019-09-30
(fix): Move uuid from defaultProps to component instance

stephanerangaya:
## Summary

This PR fixes https://github.com/Workday/canvas-kit/issues/226.

Remove uuid calls in defaultProps (static) and move them to instance variables so that they'll actually be random per instance.

## Checklist

- [x] `yarn test` passes
      applicable
- [x]  code adheres to the [API & Pattern guidelines](../API_PATTERN_GUIDELINES.md)


mannycarrera4:
I added a simple test for checkbox, if you want to reference that if we want to add a test, otherwise this looks good!

lychyi:
@stephanerangaya can you update snaps for this change?

modules/text-area/react/spec/TextArea.spec.tsx
```diff
@@ -28,7 +36,7 @@ describe('Text Area Accessibility', () => {
     const html = ReactDOMServer.renderToString(
       <>
         <label id="123">Label</label>
-        <TextArea placeholder="Placeholder" value={'Hello'} aria-labelledBy="123" />;
+        <TextArea placeholder="Placeholder" value={'Hello'} aria-labelledBy="123" onChange={cb} />;
```
lychyi: Was this just to fix a warning in tests or does this address something in this particular PR?

PR #233, opened 2019-10-01
fix: Card, Popup and Modal cleanup

anicholls:
## Summary

I found a few issues with Cards, Popups and Modals. The heading margin issue was noted in our recent design review.
- All of their headings where the same, but named different things. Now we support `*-heading`, `*-header`, and `*-title` to be flexible.
- Card headings had the browser default margin top which was adding extra space to popups and modals as well
- `PopupProps` were missing their div attribute interface extension

### Breaking Visual Changes:
- Margin top on card headers removed which changes the height of the card. Could cause reflow issues

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/modal/react/lib/Modal.tsx
+++ b/modules/modal/react/lib/Modal.tsx
@@ -66,7 +66,7 @@ const Container = styled('div')({
 
 const transformOrigin = {
   horizontal: 'center',
-  vertical: 'top',
+  vertical: 'bottom',
 } as const;
 
 function onInitialFocus(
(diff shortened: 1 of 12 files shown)


modules/modal/react/lib/Modal.tsx
```diff
@@ -12,7 +12,6 @@ export enum ModalWidth {
 export interface ModalProps extends React.HTMLAttributes<HTMLDivElement> {
   open: boolean;
   padding: PopupPadding;
-  transformOrigin: TransformOrigin;
```
NicholasBoll: This will probably conflict with #184

anicholls: Good call. We can wait until that one is in.

anicholls: Rebased and removed my notes about `transformOrigin` 👍

Issue #145, opened 2019-08-21
Fix FormField spacing for checkboxes with varied label positions


anicholls:
## 🐛 Bug Report

In some cases, spacing between form fields can be inconsistent and occasionally too "spacey". This boils down to two things:

- `FormField`s with top labels have 16px bottom margin. `FormField`s with left labels have 24px bottom margin. This should be 24px in all cases.
- `FormField` currently has a minimum height of 40px (including the label). This works for every input type except for checkboxes with left labels, which have a total height of 24px. This coupled with the bottom margin make them appear very spaced out in comparison to other fields. We will need to reduce the `min-height` across the board to 24px, or add a special case for checkbox.

This update would be a **breaking visual change** and should go in before we go out of `3.0.0-alpha`

## To Reproduce

Render a mixture of `FormField` components with varied label positions including a few stacked checkboxes with left labels (see screenshot)

## Expected Behavior

Consistent padding

## Actual Results

Varied padding

### Error Output
![Screenshot 2019-08-21 10 13 16](https://user-images.githubusercontent.com/908478/63470307-24b02800-c421-11e9-9072-805efee9a92b.png)

PR #234, opened 2019-10-01
fix: Form field spacing

stephanerangaya:
Closes: https://github.com/Workday/canvas-kit/issues/145

### Breaking Visual Changes
- Previously, Form Fields with top labels had a 16px bottom margin. FormFields with left labels had a 24px bottom margin. This has been consolidated to 24px in all cases.
- Previously, FormField had a minimum height of 40px (including the label). This worked for every input type except for checkboxes, single radios and switches with left labels, which have a total height of 24px. This min-height has been removed


modules/form-field/react/spec/__snapshots__/Label.snapshot.tsx.snap
```diff
@@ -26,10 +26,9 @@ exports[`Label Snapshots renders left label 1`] = `
   font-family: "Roboto","Helvetica Neue","Helvetica",Arial,sans-serif;
   padding: 0;
   display: inline-block;
-  vertical-align: top;
-  margin-top: 10px;
```
stephanerangaya: This create an issue though. On `Checkbox` and `Switch`, the `Label`, when on the left, is vertically centered. But on other components, it’s not.

One way to fix this would be to restore `marginTop: 10` but it will therefore push the Label to the bottom for Checkbox and Switch as well.

We might need a way to position the Label vertically based on the component that is on the right, any thoughts?

You can see this on that temporary story that will be removed after we find a solution: http://localhost:9001/?path=/story/form-field-test--default

modules/switch/react/lib/Switch.tsx
```diff
@@ -21,14 +21,14 @@ const translateLength = spacing.s;
 
 const SwitchContainer = styled('div')({
   position: 'relative',
-  height: switchTapArea,
+  height: spacing.m,
```
stephanerangaya: This is aligning `Switch`'s height to `Checkbox` so it looks more harmonious when they're on the same page.

modules/common/css/lib/errors.scss
```diff
@@ -2,21 +2,23 @@
 @mixin wdc-form-validation($_type) {
   $_borderColor: $wdc-color-black-pepper-100 !default; // This should not be used, it is necessary for the @if / @else if to.
 
-  @if ($_type== 'error') {
+  @if ($_type=='error') {
```
lychyi: Is it possible to get rid of the prettier noise here? I want to make sure this diff is clean in case we need to go back and see it.

anicholls: @stephanerangaya I think you may have something in your IDE that's screwing with our prettier rules. I just had to re-save the file and it undid all of this using our default formatting rules.

@lychyi Just pushed a commit to revert these changes

modules/switch/react/lib/Switch.tsx
```diff
@@ -21,14 +21,14 @@ const translateLength = spacing.s;
...
   width: switchTapArea,
 });
 
 const SwitchInput = styled('input')<SwitchProps>(
   {
     position: 'absolute',
-    height: switchTapArea,
+    height: spacing.m,
```
anicholls: This decreases the tappable area by 8px. Why was this needed?

modules/common/css/lib/errors.scss
```diff
@@ -259,7 +259,7 @@
     margin: -4px -12px;
     padding: 4px 12px;
     text-align: left;
-    margin-top: -12px;
+    margin-top: -4px;
```
anicholls: Nit: We already set the top margin to `-4px` above. Can we just remove this line?

modules/common/css/lib/forms.scss
```diff
@@ -82,6 +82,7 @@
       color: $wdc-color-black-pepper-300;
       font-family: 'Roboto', 'Helvetica Neue', 'Helvetica', Arial, sans-serif;
       position: relative;
+      top: 2px;
```
anicholls: What was this for? This is the only difference between the react and css versions now (visually at least)

modules/form-field/css/lib/form-field.scss
```diff
@@ -1,13 +1,13 @@
...
-  margin-bottom: $wdc-spacing-s;
+  margin-bottom: $wdc-spacing-m;
   position: relative;
 
   .wdc-form-field {
     position: relative;
-    min-height: $wdc-spacing-l;
     display: flex;
     flex-direction: column;
+    justify-content: center;
```
anicholls: What was this needed for? Do we want all fields to be left aligned?

modules/form-field/css/lib/form-field.scss
```diff
@@ -53,9 +51,10 @@
   .wdc-form-group-fields {
     display: inline-block;
     flex-grow: 1;
-    margin-top: -$wdc-spacing-xxs;
+    margin-top: 0;
```
anicholls: This isn't overriding anything, let's just remove it.

modules/form-field/css/lib/form-field.scss
```diff
@@ -70,12 +69,17 @@
 }
 
 .wdc-form-group {
+  padding: 0;
```
anicholls: Can you add a comment that these styles override the default fieldset styles so they don't get accidentally removed again?

modules/form-field/css/lib/form-field.scss
```diff
@@ -70,12 +69,17 @@
...
+  border: 0;
+
+  margin-bottom: $wdc-spacing-m;
+
   @include wdc-form-group();
 }
 
 @mixin wdc-form-label-position-left() {
   .wdc-form-field-wrapper {
-    display: flex;
```
anicholls: Why did we drop `display: flex` and position the label absolutely? Can't this be laid out as flexbox the same as React?

modules/switch/react/lib/Switch.tsx
stephanerangaya: This is to make sure `Switch` has an height of 24px total and not 32px, so a form field row has the same height than a checkbox when the label is on the left. This is what it looks like with the original height:

image

lychyi:
@stephanerangaya Since this is a visual breaking change, can we indicate this in one of the commits as outlined [here](https://www.conventionalcommits.org/en/v1.0.0-beta.4/#commit-message-with-description-and-breaking-change-in-body)

modules/switch/react/lib/Switch.tsx
anicholls: Right, sorry - I was thinking checkbox was at 32 as well

modules/common/css/lib/forms.scss
stephanerangaya: So it seems the react version changed and they're not correct. The CSS is the correct version and I've made this change to align the CSS to the React. I'm still investigating.

modules/form-field/css/lib/form-field.scss
stephanerangaya: It's a vertical alignment (`flex-direction: column`)

modules/form-field/css/lib/form-field.scss
stephanerangaya: React:
```
<fieldset>
  <legend>
  <wrapper>
    <radiogroup>
    <hint>
  <wrapper>
</fieldset>
```

CSS:
```
<fieldset>
  <legend>
  <radiogroup>
  <hint>
</fieldset>
```

In React, we have a wrapping around `RadioGroup` and `Hint`, but not in CSS, so I had to refactor Label to allow us to have it be positioned properly for `RadioGroup`, `Checkbox` and `TextInput` (in CSS).

The alternative would have been to refactor the HTML to be closer to React, but it would mean having an extra div.

If I leave `display:flex;`, the error/alert border of the `RadioGroup` will grow.

Without `display: flex;`: 
image

With `display: flex;`:
image

anicholls: Ah, I see what you mean. I missed the fact that the HTML structure was slightly different.

Hmm 🤔. I get that this works, it just feels rather brittle to absolutely position it. It's also a shame that we're doing this for all labels/fields even though it's only an issue for `.wdc-form-group-fields` (for other input types, the hint message is inside `.wdc-form-field`).

I feel like in the future we're going to need to support variable width labels, and flexbox just feels like a more "flexible" option (ha). How do you feel about adding an extra wrapper around the group and hint in CSS so that it matches React so that the styling can be more consistent with the other input types when there's an alert/error?

stephanerangaya: The issue is that if I switch back to the way it was before, with `float: left;` on the Label, then it doesn't know the height of its parent component and I can't center it vertically.

stephanerangaya: This change also allows us to properly position `label` for form field that have a height of `24px` or `40px or higher`. If we change it back to what it was before for other fields than `.wdc-form-group-fields`, we would lose this ability.

We should revisit this in the future when we'll support variable width labels.

modules/radio/react/lib/Radio.tsx
```diff
@@ -44,6 +44,9 @@ const RadioContainer = styled('div')({
 const RadioInputWrapper = styled('div')<Pick<RadioProps, 'disabled'>>(
   {
     height: radioHeight,
+    width: radioWidth,
+    marginTop: '7px',
```
anicholls: How come this is 7px and checkbox is 3px?

stephanerangaya: It was because they must have a min height of 32px in RadioGroup, but it does make sense to have them have the same height than `Checkbox` and `Switch` when they're standalone. I've updated `Radio` and moved the spacing in the context of a radio group in `RadioGroup` itself.

Issue #125, opened 2019-08-15
Layout erroneously passes in spacing prop to non-Column children


lychyi:
## 🐛 Bug Report

Children in `Layout` will mistakenly get cloned with `spacing` as a prop set by `Layout`. Only `Column` should have `spacing` passed in by `Layout`.

The logic [here](https://github.com/Workday/canvas-kit/blob/master/modules/layout/react/lib/Layout.tsx#L64) doesn't check for the type of React child properly.

## To Reproduce

See Codesandbox

## Expected Behavior

`Layout` should ignore non `Column` type components and not clone it with the `spacing` prop.

## Actual Results

`Layout` incorrectly identifies children as `Column`s and clones that child with an extra `spacing` prop if `Layout` defines a `spacing` prop.

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

## Link to repl or repo (highly encouraged)

https://codesandbox.io/s/clever-haibt-vnkim

### Error Output

Paste the error output here

```bash

```

or

Include a screenshot of a visual bug

NicholasBoll:
Fixed by #238

PR #238, opened 2019-10-03
fix(layout): Compensate column width calculations

### LINKED ISSUE #205: Layout/Column Grid not sizing inner cells correctly
## 🐛 Bug Report

Building a `<Layout>` with `<Layout.Column columns=4>` with Images have a width of 100% in each cell, The center column is a smaller width than the outer columns because of inner padding.

## To Reproduce

Create a `<Layout>` with `<Layout.Column columns=4>` and put `` tags with widths of 100%.

## Expected Behavior

All cells should be the same width and be spaced correctly.

## Actual Results

Cells along the center column have smaller widths than cells in the outermost columns.

## Browser (if applicable)

Chrome

image

sahlhoff:
## Summary

Fixes #205  and #125 

See [CodeSandbox Example](https://codesandbox.io/s/canvas-kit-issue-reproduction-oeqer?fontsize=14). You will notice the miss-alignment of the last two rows.

## Checklist

- [x] `yarn test` passes
      applicable

## Additional References

Why do we need to subtract compensation from the widths?

The width of columns without siblings are calculated incorrectly. We need to proportionally apply a reduction to the widths of columns for the potentially missing padding of the sibling column. 

A potentially easier way to think of this would be to remove the spacing from the parent container and then doing the width percentage calculation. We are effectively doing this at the `Column` level with these lines:

```tsx
        const compensation = ((12 - columns) / 12) * (spacing! * 2);
        return {width: `calc(${(columns / 12) * 100}% - ${compensation}px)`}; 
```


modules/layout/react/lib/Layout.tsx
```diff
@@ -53,9 +53,14 @@ const LayoutContainer = styled('div')<LayoutProps>(
...
 );
 
+export function stripUnit(value: string): number {
+  return parseInt(`${value}`.replace('px', ''), 10);
+}
+
 export default class Layout extends React.Component<LayoutProps> {
   static defaultProps = {
     gutter: canvas.spacing.xs,
+    spacing: stripUnit(canvas.spacing.xs),
```
sahlhoff: Not 100% sure if this is the correct way to handle the default spacing for a child. What I am attempting to do is have the child spacing have a default value of `canvas.spacing.xs` But with the ability to override it at the `Layout` and `Column` level like this:

```tsx
       <Layout spacing={60}>
          <Layout.Column spacing={0}>
            <Card heading="Custom Spacing (0px)" />
          </Layout.Column>
          <Layout.Column>
            <Card heading="Custom Spacing (60px)" />
          </Layout.Column>
          <Layout.Column>
            <Card heading="Custom Spacing (60px)" />
          </Layout.Column>
        </Layout>
```

I found that if I set the default spacing in `Column` instead of the `Layout` [this check](https://github.com/Workday/canvas-kit/blob/eed62cf9591dc97fd98c305ac96caeb42acc0da9/modules/layout/react/lib/Layout.tsx#L76) would always be true.

modules/layout/react/lib/Column.tsx
```diff
@@ -35,19 +34,20 @@ const ColumnContainer = styled('div')<ColumnProps>(
...
     if (width) {
       return {width};
     }
 
     if (columns) {
-      return {width: `${(columns / 12) * 100}%`};
+      if (columns === 12) {
+        return {width: `100%`};
+      } else {
+        const compensation = ((12 - columns) / 12) * (spacing! * 2);
```
lychyi: I may be reading this wrong but doesn't this mean that `compensation` can be `NaN` if we don't define spacing?

modules/layout/react/lib/Layout.tsx
lychyi: I think you can use `spacingNumbers.xs` from `core` here instead of recalculating it.

modules/layout/react/lib/Column.tsx
NicholasBoll: Javascript is weird...

```ts
1 * null // 0
```

NicholasBoll: But I would suggest the following to not rely on this quirk:
```suggestion
        const compensation = ((12 - columns) / 12) * (spacing || 0 * 2);
```

NicholasBoll:
```ts
1 * undefined // NaN
```

sahlhoff: Updated! Ended up having to add in extra parentheses– 

```tsx
const compensation = ((12 - columns) / 12) * ((spacing || 0) * 2);
```

modules/layout/react/lib/Layout.tsx
```diff
@@ -81,7 +82,7 @@ export default class Layout extends React.Component<LayoutProps> {
   };
 
   public render() {
-    const {children, spacing, gutter, capWidth, ...elemProps} = this.props;
+    const {children, gutter, capWidth, ...elemProps} = this.props;
 
     return (
       <LayoutContainer gutter={gutter} capWidth={capWidth} {...elemProps}>
```
lychyi:
```suggestion
      <LayoutContainer gutter={gutter} capWidth={capWidth}>
```

I know we didn't do this before but it should be best practice that we destructure all props and then use them as needed.

modules/layout/react/lib/Column.tsx
NicholasBoll: Cool. I had parenthesis in originally and did some light testing and removed it.

modules/layout/react/lib/Layout.tsx
sahlhoff: Shouldn't we be spreading the props here thought? This will fail tests for `data-propspread` attribute test

lychyi: Oof you're right. I lost track of this one, I'll re-review and try to get this one in today. So sorry about that.

PR #240, opened 2019-10-03
ci: Create github action for CI and add Cypress to CI

NicholasBoll:
## Summary

We're evaluating Github Actions. The PR adds common CI tasks already covered by Travis, but also adds Cypress tests and uploads to the Cypress dashboard. Cypress runs can be found here: https://dashboard.cypress.io/#/projects/odida5/runs

## Checklist

## Additional References

It doesn't look like Github Actions runs if created from a fork. Here is an example of it working inside my fork: https://github.com/NicholasBoll/canvas-kit/pull/1. Screenshot below:

Screen Shot 2019-10-03 at 3 57 54 PM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,43 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  unit_test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-node@v1
+      with:
+        node-version: 10.x
+    - name: Install Packages
+      run: yarn install --production=false
+    - name: Linting
+      run: yarn lint
+    - name: Build
+      run: yarn build
+    - name: Unit tests
+      run: yarn test
+
+  integration_test:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - master
```
anicholls: We'll have to extend this, but fine for now.

cypress.json
```diff
@@ -1,4 +1,5 @@
 {
+  "projectId": "odida5",
```
anicholls: Can we customize this in cypress?

.github/workflows/ci.yml
NicholasBoll: I'm not sure this should be restricted to PRs to `master`. Visual regression might since that's more of a linear approval to a single branch.

cypress.json
NicholasBoll: I don't think so. This project ID is generated by the Cypress Dashboard when adding a project to an organization. The local Cypress runner can be connected to your Cypress account and this project ID is used to link a project to the Dashboard. From there a record key is generated (projectID is public, the record key is private)

anicholls: Okay no worries, was just wondering for URL purposes.

PR #241, opened 2019-10-04
ci: Fix fork failures

NicholasBoll:
Forks are failing because they do not have access to secrets. Cypress fixed this if `--key` is not used, but an environment variable is instead used: https://github.com/cypress-io/cypress/issues/1193

Hopefully this fixes the issue. If not, we'll have to either use the Cypress key in plain text or disable recordings altogether.


NicholasBoll:
I'll wait for #243

PR #242, opened 2019-10-07
build: Replace TSLint with ESLint

### LINKED ISSUE #154: Replace TSLint with ESLint
## 🚀 Feature Proposal

Replace TSLint with ESLint using [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint)

## Motivation

TSLint will be deprecated sometime in 2019. The TSLint team has decided to adopt ESLint longterm and the recommendation is to use ESLint and typescript-eslint. They've also provided a [CLI command](https://github.com/typescript-eslint/tslint-to-eslint-config) to assist with the migration.

See:
- https://eslint.org/blog/2019/01/future-typescript-eslint
- https://github.com/palantir/tslint/issues/4534
- https://medium.com/palantir/tslint-in-2019-1a144c2317a9

anicholls:
## Summary

TSLint is being deprecated in favor of ESLint (see https://github.com/palantir/tslint/issues/4534). We are also looking at migrating to Emotion 10, and the codemod for doing so uses ESLint. This update replaces TSLint with ESLint while keeping close to parity with our previous rules.

I left a lot of rules commented out that I think we should enable to prevent a large style update/refactor cluttering up this PR. We should spend some time talking over these and make the call on which recommended rules to uptake.

I also added Prettier to our ESLint config, so they are run at the same time. This simplified some of the details in our package.json. You still have the ability to run Prettier on it's own with `yarn format:prettier`

Closes #154 

## Checklist

- [ ] `yarn test` passes


eslintrc.js
```diff
@@ -0,0 +1,121 @@
+module.exports = {
+  env: {
+    browser: true,
+    es6: true,
+    node: true,
+    'jest/globals': true,
+  },
+  extends: [
+    // 'eslint:recommended',
```
anicholls: We should turn all of these on eventually

eslintrc.js
```diff
@@ -0,0 +1,121 @@
...
+        printWidth: 100,
+        proseWrap: 'always',
+        semi: true,
+        singleQuote: true,
+        tabWidth: 2,
+        trailingComma: 'es5',
+        useTabs: false,
+      },
+    ],
+    // NOTE: Commented out everything below that caused problems. A lot of this is likely included in the stuff above.
```
anicholls: All of the stuff below is generated by https://github.com/typescript-eslint/tslint-to-eslint-config. A lot of it is likely covered with things above, and our goal should be to remove this section completely.

Some of the items below caused problems so I just commented them out for now.

eslintrc.js
```diff
@@ -0,0 +1,118 @@
...
+    '@typescript-eslint',
+    '@typescript-eslint/tslint',
+    'jest',
+    'react',
+    'prettier',
+    'react-hooks',
+  ],
+  rules: {
+    '@typescript-eslint/class-name-casing': 'error',
+    '@typescript-eslint/indent': 'off',
```
NicholasBoll: I thought this one would be covered by one of the prettier configs

eslintrc.js
```diff
@@ -0,0 +1,118 @@
...
+    'prefer-const': 'error',
+    'quote-props': 'off',
+    'space-before-function-paren': 'off',
+    'react/jsx-no-bind': 'warn', // Update
+    curly: 'error',
+    radix: 'error',
+    'prettier/prettier': [
+      'error',
+      {
+        arrowParens: 'avoid',
```
NicholasBoll: These are all duplicated from `.prettierrc`. Are these not automatically inferred? If not, maybe we could just require that file and reference the config here?

Something like:
```ts
const fs = require('fs');
const prettierConfig = JSON.parse(fs.readFileSync('.prettierrc'))

/// ...

'prettier/prettier': ['error', prettierConfig]
```

package.json
```diff
@@ -110,10 +119,9 @@
     "commit": "git-cz",
     "commitmsg": "commitlint --edit $HUSKY_GIT_PARAMS",
     "build-storybook": "build-storybook -c .storybook -o docs",
-    "lint": "node utils/check-lockfile-links.js && tslint --project tsconfig.json",
+    "lint": "node utils/check-lockfile-links.js && eslint -c ./eslintrc.js --ext=jsx,ts,tsx .",
     "format:prettier": "prettier --config=./.prettierrc --ignore-path=./.prettierignore \"**/*.{js,jsx,json,ts,tsx,md}\" --write",
-    "format:tslint": "tslint --project tsconfig.json --fix",
-    "format": "yarn format:prettier && yarn format:tslint",
+    "format": "yarn lint --fix",
```
NicholasBoll: I'm a little confused. Are we removing Prettier in favor of `eslint --fix`?

eslintrc.js
NicholasBoll: Or maybe better, change `.prettierrc` to `.prettierrc.js` and just `require` it

eslintrc.js
anicholls: Yes, I think this was turned off by https://github.com/typescript-eslint/tslint-to-eslint-config to prevent conflicts with prettier

package.json
anicholls: Prettier is run as part of `eslint --fix` now.

NicholasBoll: Oh. I see. It is a eslint plugin now.

lychyi:
@anicholls I believe we still reference tslint in our storybook webpack config which will cause a storybook build error

Issue #244, opened 2019-10-07
Toast: Fix Link Background on Safari


stephanerangaya:
## 🐛 Bug Report

In Safari (macOS Catalina), when there's a link in the `Toast`, the background is gray

## To Reproduce

Create a `Toast` with a link, open in Safari

## Expected Behavior

The background of the link should be white

## Actual Results

The background of the link is gray

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

Safari, macOS Catalina 

## Link to repl or repo (highly encouraged)

Please provide a reference to a repo or branch to help us reproduce this bug.

### Error Output

<img width="1306" alt="image" src="https://user-images.githubusercontent.com/723109/66351534-b7eae180-e92b-11e9-8081-aed0e64a0c2f.png">

PR #245, opened 2019-10-07
fix(toast): Fix background color on Safari

stephanerangaya:
This PR fixes https://github.com/Workday/canvas-kit/issues/244


lychyi:
Going to remove the "BREAKING CHANGE" notice in the body when I squash. This is a bug fix it's a background change but it shouldn't have happened in the first place

PR #246, opened 2019-10-08
chore: Upgrade to Emotion 10

### LINKED ISSUE #137: Breaking: Upgrade to Emotion 10
## 🚀 Feature Proposal

Update to Emotion 10.

## Motivation

We depend on [emotion](https://emotion.sh) 9.x for our CSS-in-JS needs. They recently(-ish) launched a major new version and this causes errors if projects use emotion 10 and try to consume Canvas Kit React today. This would be a major breaking change for CKR if we upgrade because teams that consume CKR on emotion 9.x would have to stay on a previous version for compatibility reasons.

## Open Questions
- [ ] When is emotion 9 getting deprecated?
- [ ] How will this affect other teams using Canvas Kit?

anicholls:
## Summary

Upgrades Emotion from 9 to 10. 

Biggest changes:

- Now use `css` prop instead of `className`. There is a new linter that will automatically add the jsx pragma and import to the top of your file if you use this prop (see https://emotion.sh/docs/css-prop#jsx-pragma).
- Emotion is trying to prevent its users from thinking about styles being applied as classes and wants to avoid users thinking about the class name in general. Because of this, if you need to apply a custom class, the API is a bit gross (see https://github.com/Workday/canvas-kit/pull/246/files#diff-0fd78f5b5d3a2af13cc4cb142f45e174R32). More info here: https://emotion.sh/docs/class-names

### Breaking Code Changes:
- `injectGlobal` function for `@workday/canvas-kit-react-fonts` now uses their new [`Global` component](https://emotion.sh/docs/globals).
- Any exported types that used emotion's `CSSObject` or `Interpolation` types have changed slightly due to updates upstream
- May required consumers to be on Emotion 10. We should do some testing in Emotion 9 projects to see if this causes any problems. Maybe we don't need to wait for 4.x if we can get it in while we're still in beta.

Closes #137 

## Checklist

- [x] `yarn test` passes


modules/_labs/core/react/spec/type.snapshot.tsx
```diff
@@ -1,5 +1,6 @@
+/** @jsx jsx */
```
NicholasBoll: Are we going to add this to every file or change the pragma in our configs? As long as the pragma is not set to `preserve` in the Typescript config, this should work for downstream consumers.

modules/_labs/core/react/spec/type.snapshot.tsx
```diff
@@ -49,23 +50,23 @@ describe('Core Labs Type Snapshot', () => {
 
         <h3 style={type.h3}>Variants</h3>
 
-        <span className={css(type.body, type.variant.button)}>Button Text</span>
+        <span css={css(type.body, type.variant.button)}>Button Text</span>
```
NicholasBoll: Isn't this the same as:
```tsx
<span css={[type.body, type.variant.button]}>Button Text</span>
```
Otherwise, why use the jsx pragma + the `css` prop?

anicholls: You are correct, I will update these 👍

modules/_labs/core/react/spec/type.snapshot.tsx
anicholls: This was done automatically by their linting rules via https://github.com/emotion-js/emotion/blob/master/packages/eslint-plugin-emotion/docs/rules/jsx-import.md. I was planning to update the configs so I can disable this rule, but haven't got there yet.

modules/icon/react/lib/Svg.tsx
```diff
@@ -28,17 +28,24 @@ export default class Svg extends React.Component<SvgProps> {
...
-
     return (
-      <span
-        {...elemProps}
-        dangerouslySetInnerHTML={{__html: src.svg}}
-        // Need to combine iconStyle with the className prop, otherwise we'll clobber it
-        // (we'll need to do something like this for each HTML <span> prop we explicitly set in this component)
-        className={cx(assetStyle, elemProps.className)}
-      />
+      <ClassNames>
```
anicholls: This change makes it so that `shallow` in the unit tests no longer works for `Svg` and `Icon`.

`Svg` works if you change it to `mount`, but then you're checking the props rather than the resulting element's className. Needs some finessing.

Haven't been able to get the `toHaveStyleRule` check in `Icon.spec.tsx` to find the width and height with this `ClassNames` wrapper.

modules/_labs/core/react/spec/type.snapshot.tsx
anicholls: Just tried this and while I was able to move the pragma to the tsconfig successfully, it caused some other problems. Also, I'm not sure how to make Emotion's `jsx` import global without babel so you still need to `import {jsx} from '@emotion/core';` in every file that `css` is used.

anicholls:
Oh I also noticed I fixed the type conflicts on build, but now they're showing up in Storybook :( Maybe 5.2/3 will help? 🙏

modules/icon/react/spec/Icon.spec.tsx
```diff
@@ -8,18 +8,8 @@ import {matchers} from 'jest-emotion';
 expect.extend(matchers);
 
 describe('Icon component', () => {
-  test('Sets size correctly', () => {
```
lychyi: It makes more sense to me to test CSS style rules in a snapshot test (and it's way easier).

modules/_labs/core/react/spec/type.snapshot.tsx
anicholls: Just to track on the PR, I decided not to set the `jsxFactory` in `tsc` because we only need the `jsx` pragma for select files that use the `css` prop (mainly stories). This is added automatically by the ESLint rule linked above for files using the `css` prop, so it's not too big of a deal to require in the files as necessary. Also there doesn't seem to be any way to avoid the `jsx` import with `tsc` in these cases.

The biggest reason for not adding the pragma to all files is because it [downgrades fragment node experience](https://github.com/emotion-js/emotion/issues/1156).

PR #251, opened 2019-10-09
feat: Add support for translation

mannycarrera4:
## Summary

Add strings and defaults english strings to translations


modules/modal/react/lib/Modal.tsx
```diff
@@ -76,7 +76,11 @@ function onInitialFocus(
   if (firstFocusEl) {
     return firstFocusEl;
   } else {
-    const firstFocusable = modalEl && modalEl.querySelector<HTMLElement>('[aria-label=Close],h3');
+    const firstFocusable =
+      modalEl &&
+      modalEl.querySelector<HTMLElement>(
+        `[data-close=close],#${modalEl.getAttribute('aria-labelledby')}`
```
NicholasBoll: The test is failing with:
> Error: Failed to execute 'querySelector' on 'Element': '[data-close=close],#2f023d86-296a-457c-984f-8d7cbe3650b9' is not a valid selector.

A `uuid` can start with a number, but an id selector shorthand cannot. The following should fix the error:
```suggestion
        `[data-close=close],[id="${modalEl.getAttribute('aria-labelledby')}"]`
```

mannycarrera4: Ah, thank you so much!

modules/side-panel/react/README.md
```diff
@@ -124,6 +124,18 @@ class SidePanelExample extends React.Component<{}, SidePanelState> {
 
 ---
 
+#### `hideNavigationLabel: string`
```
anicholls: Since we use the term `open`, can we change this to `closeNavigationLabel` and `openNavigationLabel`?

modules/popup/react/lib/Popup.tsx
```diff
@@ -107,13 +110,14 @@ export default class Popup extends React.Component<PopupProps> {
         {handleClose && (
           <CloseIconContainer closeIconSize={closeIconSize}>
             <IconButton
+              data-close="close"
```
anicholls: Nit: Can you add a comment for why this is here?

modules/side-panel/react/README.md
mannycarrera4: that makes sense!!

PR #254, opened 2019-10-11
ci: Add ChromaticQA to CI

NicholasBoll:
## Summary

ChromaticQA will help us keep track of visual changes requiring approval for visual diffs. According to the documentation, multiple baselines should be supported. That should include `master` (the main baseline) and any visual diffs detected in pull request branches. I'm not sure how well this baselining will work. I was surprised it wasn't based on the name of a branch - we shall see how well it works.

There are 2 new jobs - one for pull requests and one for master. Master auto-accepts changes to the baseline because we only allow changes to be approved through pull requests.

## Additional References

https://www.chromaticqa.com/builds?appId=5d854c26ba934e0020f5e98a

### DIFF
--- a/.github/workflows/chromaticqa.yml
+++ b/.github/workflows/chromaticqa.yml
@@ -21,7 +21,11 @@ jobs:
     - name: Install Packages
       run: yarn install
     - name: ChromaticQA
-      run: yarn chromatic
+      # Chromatic has a separate Github check that handles changes. We need to exit
+      # zero, otherwise this Github workflow will have to be rerun _after_ a visual review.
+      # This extra step is unnecessary with the Chromatic check we have integrated.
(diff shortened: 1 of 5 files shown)


.github/workflows/chromaticqa.yml
```diff
@@ -0,0 +1,27 @@
...
+    - uses: actions/setup-node@v1
+      with:
+        node-version: 10.x
+    - name: Install Packages
+      run: yarn install
+    - name: ChromaticQA
+      run: yarn chromatic
+      env:
+        CI: true
+        CHROMATIC_APP_CODE: dlpro96xybh
```
NicholasBoll: We don't have a good way to hide these keys. I spoke with the ChromaticQA team and they recommended using the key in public until there is a better way to hide it. They can issue us a new app code on request.

PR #260, opened 2019-10-14
fix(button): Update missed css styles for buttons

mannycarrera4:
## Summary

Some styles where missed during the migration for css.


lychyi:
I think the small button has an incorrect anti-alias setting. For some reason font-smoothing is set to `subpixel-antialiased`. While everything else is just `antialiased`.

This is apparent when you closely compare small and medium buttons.

Screen Shot 2019-10-17 at 10 17 05 PM

I reverted the visual diff for the deprecated secondary button too. That should not be an accepted diff because that story shouldn't have been updated.

Once the deprecated button story and font-smoothing bug gets fixed we can merge.

PR #267, opened 2019-10-15
build: Upgrade Storybook to 5.2

sahlhoff:
Hi all, you'll notice two storybook changes in this PR: `Avatar` and `Card`. `Avatar` is using the legacy `storiesOf` api but with `addParameters({component: <component>})`. `Card` is using the new CSF format.

Some notes:

~I'm unable to get CSF to work with the `source-loader`. The `source-loader` is able to parse the legacy stories (and show the code) but is having issues showing the CSF code. There's similar issues on [github](https://github.com/storybookjs/storybook/issues/8104).~

~You'll notice for both the `Card` and `Avatar` story we are able to parse the props in the props section. However this appears to be just the default props from the component. I did have some success using the `docgenLoader` – you'll see commented out in `.storybook/webpack.config.js`. This however had a severe impact on build performance and also LOTS of props got passed because in most cases we are extending `HTMLAttributes`.~

The refactor to CSF is going to be a pretty large change to the codebase. CSF requires that there are only [one component per file](https://www.npmjs.com/package/@storybook/codemod#storiesof-to-csf)– this means potentially we could at least see double of the storybook files in components like `button` and `form-field`. I've got a lot of these changes stashed locally and I can follow up with them if we can figure out the webpack issues / `source-loader`.

Any thoughts or help, greatly appreciated! 🙏 

## Checklist

- [ ] `yarn test` passes
      applicable

## Additional References

Storybook's recommended docs configuration: https://github.com/storybookjs/storybook/tree/next/addons/docs#manual-configuration

## Performance

image

image


NicholasBoll:
It looks like the Chromatic CI check is failing on `yarn build-storybook`

NicholasBoll:
If we don't want to list out all the properties on the host Element, it looks like props can be filtered out: https://github.com/storybookjs/storybook/issues/8104#issuecomment-535383208

.storybook/webpack.config.js
```diff
@@ -21,12 +21,34 @@ const babelLoader = {
   },
 };
 
+// const docgenLoader = {
+//   loader: require.resolve('react-docgen-typescript-loader'),
+//   options: {
+//     tsconfigPath: path.join(__dirname, 'tsconfig.json'),
+//     skipPropsWithoutDoc: true,
+//   },
+// };
```
NicholasBoll: It looks like props coming from certain paths can be excluded:
```ts
const docgenLoader = {
  loader: require.resolve('react-docgen-typescript-loader'),
  options: {
    propFilter: prop => {
      // Filter out props from styled-components
      if (prop.name === 'as' || prop.name === 'ref' || prop.name === 'theme') {
        return false;
      }

      if (prop.parent == null) {
        return true;
      }

      // Filter out props which type definition is placed in react package
      return prop.parent.fileName.indexOf('node_modules/@types/react') < 0;
    }
  }
};
```

sahlhoff: Yeah I saw this. We can use this but we will not be able to use `HappyPack` for the loader since it will JSON sanitize the options. I can look at running it separate from `HappyPack`

sahlhoff:
@NicholasBoll this should be able `yarn build-storybook` after we upgrade to emotion 10

.storybook/docgen-plugin.js
```diff
@@ -0,0 +1,65 @@
+const docGen = require('react-docgen-typescript');
```
sahlhoff: Discovered from https://github.com/strothj/react-docgen-typescript-loader/issues/37

NicholasBoll:
@anicholls There was some dependency mismatches, so we simply updated the yarn.lock file will all the latest changes and that fixed the issue, but added a lot of diffs to the `yarn.lock` file. Is this okay or should we narrow it down by updating the `package.json` file as well?

lychyi:
@NicholasBoll Changes were manually introduced to yarn.lock?

anicholls:
@NicholasBoll I told you 🤣

.storybook/webpack.config.js
```diff
@@ -16,6 +16,17 @@ module.exports = ({config, mode}) => {
...
+  config.module.rules.forEach(rule => {
+    if (rule.test.toString().includes('scss|sass')) {
+      rule.use[2].options.plugins = () => [
+        require('postcss-discard-duplicates'),
+        require('autoprefixer'),
+        require('postcss-inline-svg'),
+      ];
+    }
+  });
+
```
sahlhoff: @anicholls so this is quite the hack but what's going on is when we use `react-app` as a preset we are unable to override their default config.

see: https://github.com/facebook/create-react-app/issues/5749

anicholls: Good find and fix! I think this is fine for now 👍

anicholls: Though I wonder if there's any way to reference the `.postcss.json` file here rather than just copying the plugins from the `development` key...

NicholasBoll:
@sahlhoff @anicholls What happened here? https://www.chromaticqa.com/snapshot?appId=5d854c26ba934e0020f5e98a&id=5db78ba417a1ee0020846163

It looks like now there is a background image? It wasn't showing before?

sahlhoff:
@NicholasBoll just looking at the resources panel in chrome– looks like previously it was incorrectly getting base64ed compared to current build. The story css has it explicitly set so going to assume we want it here.


image

.storybook/webpack.config.js
sahlhoff: Yup, I'll look into it 👍

sahlhoff: @anicholls Updated to just pass in config path.

PR #269, opened 2019-10-16
ci: Update ChromaticQA for better baselines

NicholasBoll:
According to my conversation with the ChromaticQA team, our baseline issue should be partially addressed by updating to the latest version of `storybook-chromatic`. The team is currently working on a Github action: https://github.com/chromaui/action (currently in beta).

Here's what the baselines now look like:
Screen Shot 2019-10-16 at 12 55 23 PM

Before:
```
7f607b on master
```

After:
```
 7f607b on chore/update-chromaticqa
```

Before all PRs used the `master` baseline which was a problem where each CI run against a PR potentially rebaselined another PR. Now we're able to send the baseline explicitly. I've tested this against this PR and another PR to make sure the baselines did not interfere with each other and it doesn't look like they do anymore.

## Proposed Squash Message:
```
ci: Update ChromaticQA for better baselines (#269)
```

### DIFF
--- a/.github/workflows/chromaticqa.yml
+++ b/.github/workflows/chromaticqa.yml
@@ -11,10 +11,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
-      with:
-        # Use 'head' instead of 'merge': https://docs.chromaticqa.com/setup_ci#pull-request-builds
-        # 'head' is the HEAD of the PR branch. 'merge' is the hidden merge commit created by github
-        ref: refs/pull/${{ github.event.pull_request.number}}/head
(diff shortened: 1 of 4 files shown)


.github/workflows/chromaticqa.yml
```diff
@@ -28,4 +24,6 @@ jobs:
       run: yarn chromatic --exit-zero-on-changes
       env:
         CI: true
+        CHROMATIC_BRANCH: ${{ github.event.pull_request.head.ref }}
```
NicholasBoll: The Chromatic team just added these yesterday. We're able to explicitly set the branch name as a baseline

Issue #237, opened 2019-10-02
Fix Icon Button Inverse Focus


jessespencer:
## 🐛 Bug Report
The focus state of `Icon Button Inverse` and `Icon Button Inverse Filled` should match the implementation of `Button Outline Inverse` and the original specs.

## To Reproduce
"Tab" on `Icon Button Inverse`.

## Expected Behavior
The background color and icon should invert like `Button Outline Inverse`'s focus state.

## Actual Results
The current focus state is just a white outline. 

## Link to repl or repo (highly encouraged)
https://workday.github.io/canvas-kit/?path=/story/css-button-icon-button--inverse-filled

![Screenshot 2019-10-02 11 10 06](https://user-images.githubusercontent.com/54119049/66070040-64306080-e505-11e9-84d1-4c594f86083b.png)

PR #271, opened 2019-10-17
fix(button): Address issues with various IconButton states

mannycarrera4:
Blocked on #377 

## Summary

When we moved some of the styles into utils, some of the icon button states broke. This PR addresses fixing the hover and focus states for the icon buttons as well as introducing a table to show all the current states for better regression testing. 

Also in this PR, we introduce a new style for inver and inverse filled icons where we change them to a white background when toggled on.
Closes https://github.com/Workday/canvas-kit/issues/237


modules/button/react/lib/ButtonStyles.ts
```diff
@@ -416,6 +412,20 @@ function getButtonStateStyle(variant: AllButtonVariants): CSSObject {
         'span .wd-icon-fill, span .wd-icon-accent': {fill: buttonColors.labelIconHover},
       }),
     },
+    '&:hover:focus': {
+      backgroundColor:
+        variant === IconButtonVariant.InverseFilled || IconButtonVariant.Inverse
```
anicholls: Why is this check needed? We use the `focusBackground` in both cases.

anicholls:
@jessespencer should the state change at all for `:focus:active`?  Usually we have a state change for this, but after these changes it does nothing.

mannycarrera4:
@anicholls for the icon buttons, i wanted to change the hover of the icon when it was focused and hovered, otherwise there would be no hover effect when the icon was focused, which seemed a little odd. For the toggleable icons this is where I wasn't sure what those states should be

jessespencer:
@mannycarrera4 was following the same implementation as `outline inverse` which doesn't have a state change on `focus:active` - do you think it should @anicholls? The icon is subtly changing grays but how would we approach the state change for white?

modules/button/react/lib/IconButton.tsx
```diff
@@ -129,45 +122,85 @@ export const IconButtonCon = styled('button', {
...
+          },
+          '&:hover': {
+            backgroundColor: colors.frenchVanilla100,
+            ...getFillSelector(colors.licorice500),
+            ...getAccentSelector(colors.licorice500),
           },
+          borderColor: colors.blueberry400,
+          ...getFillSelector(colors.frenchVanilla100),
+          ...getAccentSelector(colors.frenchVanilla100),
         };
```
mannycarrera4: I think this can be cleaned up for sure, but I finally got it to a reasonable state where they somewhat work, but wanted to get feedback on this before cleaning it up a bit.

anicholls:
Thanks for the continued work on this @mannycarrera4. I reviewed today with @jessespencer. It's getting close! Great job figuring out the weird transition issues with the icon colors. We have a few more notes:
- Button icons are no longer filled when toggled on
- Inverse icons are using a white background on hover & focus. They should be using the transparent black background (and even darker on click). When focussed they should just have the white ring. They should only have a white background when toggled on.
- Inverse icons should be filled and blue whenever there is a white background (when toggled on).

modules/button/react/lib/IconButton.tsx
```diff
@@ -65,11 +70,31 @@ function getAccentSelector(fillColor: string): CSSObject {
...
+      .replace(/:/g, '.')
+      .replace(/\.not/g, ':not')
+      .replace(/\.disabled/, ':disabled');
+    const value =
+      typeof obj[key] === 'object' ? changeStateSeparator(true, obj[key] as CSSObject) : obj[key];
+    const newObj = {...result, [newKey]: value};
+    return newObj;
+  }, {});
+};
+
```
mannycarrera4: @NicholasBoll and I wrote to this to be able to see the different states of IconButton. I think eventually we might want to move this into a utils?

modules/button/react/lib/IconButton.tsx
```diff
@@ -85,92 +110,161 @@ export const IconButtonCon = styled('button', {
...
+              '&:not([disabled]):focus': {
+                backgroundColor: colors.blueberry500,
+                ...(toggled ? focusRing(2, 2) : {}),
+              },
+              '&:hover': {
+                backgroundColor: colors.blueberry500,
+              },
+              ...mouseFocusBehavior({
+                '&:focus:active': {
+                  backgroundColor: `${colors.blueberry500} !important`,
```
mannycarrera4: Unfortunately, i have to override the mouseFocusBehavior for inverse and inverse filled since we no longer have a hover or active state for these buttons. I'm open to a different approach, thoughts?

modules/button/react/lib/utils.ts
```diff
@@ -115,6 +115,7 @@ export function getButtonFocusRing(variant: AllButtonVariants): CSSObject {
       return focusRing(2);
     case IconButtonVariant.Inverse:
     case IconButtonVariant.InverseFilled:
+      return focusRing(2, 2, true, false, buttonColors.focusRingInner, buttonColors.focusRingOuter);
```
mannycarrera4: New focus style for Inverse and Inverse Filled IconButton

modules/button/react/lib/IconButton.tsx
anicholls: @mannycarrera4 @NicholasBoll what do you think about adding it to the [wrapper `styled` function](https://github.com/Workday/canvas-kit/blob/master/modules/_labs/core/react/lib/theming/styled.ts) we added for theming/RTL? You could check for the `changeStateSeparator` prop there, rather than having to add the complexity of calling it for every interpolation function in the component files.

anicholls: This is awesome btw 👌

modules/_labs/core/react/index.ts
```diff
@@ -1,8 +1,8 @@
 import type from './lib/type';
 import space from './lib/space';
 import CanvasProvider from './lib/CanvasProvider';
-
 export default type;
 export {type, space, CanvasProvider};
 export * from './lib/type';
+export * from './lib/StaticStates';
```
mannycarrera4: we can either export this or just import it as `import {StaticStates} from '@workday/canvas-kit-labs-core/react/lib/StaticStates'` I'm fine with either

modules/button/react/lib/ButtonColors.ts
```diff
@@ -247,6 +251,7 @@ export const ButtonColors: ButtonColorCollection = {
     labelIconHover: canvas.colors.licorice500,
     labelIconActive: canvas.colors.licorice500,
     labelIconFocus: canvas.colors.licorice200,
+    labelIconFocusHover: canvas.colors.licorice500,
```
mannycarrera4: revisit

modules/button/react/lib/IconButton.tsx
```diff
@@ -39,6 +39,11 @@ export interface IconButtonProps extends React.ButtonHTMLAttributes<HTMLButtonEl
    * Callback that fires when a button changes toggled states
    */
   onToggleChange?: (toggled: boolean | undefined) => void;
+  /**
+   * Only used for visual snapshots and development
+   * @ignore
+   */
+  shouldChangeStateSeparator?: boolean;
```
mannycarrera4: remove

modules/button/react/lib/utils.ts
```diff
@@ -185,14 +192,122 @@ export function getButtonStateStyle(variant: AllButtonVariants): CSSObject {
...
+      ...(buttonColors.labelDataActive && {
+        ['.' + ButtonStyles.labelDataBaseStyles.classname]: {
+          color: buttonColors.labelDataActive,
+        },
+      }),
+      ...(buttonColors.labelIconActive &&
+        {
+          // 'span .wd-icon-fill, span .wd-icon-accent': {
+          //   fill: buttonColors.labelIconActive
+          // }
```
mannycarrera4: remove

Issue #229, opened 2019-09-30
Rewrite color utility to use ChromaJS


lychyi:
## 🚀 Feature Proposal

Upgrade [Chroma JS](https://github.com/gka/chroma.js/) to version 2.x (2.0.6 as of writing)

## Motivation

We use Chroma for some color manipulations and we also export a [function](https://github.com/Workday/canvas-kit/blob/5bdee566dbf9fb77b80957c08ddf31859e0c32cd/modules/common/react/lib/utils/colorUtils.ts) that calculates color usage based on accessibility which Chroma already does.

## Outline
* Update chroma-js to 2.x
* Update @types/chroma-js to latest
* Rewrite the color utility that picks foreground color to only use Chroma instead of calculating relative lum (no need to re-invent the wheel)

anicholls:
Updated chroma-js version to 2.1.0 as part of #272. Still need to rewrite the color utility though.

NicholasBoll:
Fixed by #700

PR #272, opened 2019-10-22
feat(labs): Theming (react)

anicholls:
## Summary

Spent last week working through how theming will work in Canvas Kit. This is a functional POC, but still needs some work. It also only adds theming to one existing component as a test.

Note 1: This does not have any breaking changes 🎉 
Note 2: Since I've put the theming work in labs (to allow for updates to the API/theme object), we now need to depend on `@workday/canvas-kit-labs-react-core` if we want to start adding theming support to our components.
Note 3: This does not address CSS theming. The implementation of that is still up in the air.

### Implementation

This adds a new [`CanvasProvider`](https://github.com/Workday/canvas-kit/pull/272/files#diff-902fb00cb9e4aa17deb52004e4e89e68) (which includes an `InputProvider`) that provides a theme using Emotion's `ThemeProvider`. You can pass a custom theme using the `createCanvasTheme` function:
```tsx
const theme = createCanvasTheme({
  palette: {
    primary: {
      main: 'orange'
    },
  }
});
<CanvasProvider theme={theme}>
  // Components
</CanvasProvider>
```

`createCanvasTheme` will expand a partial theme into a full one based on the default theme. If you only pass a main shade, all other shades will be generated automatically based on that.

[`CanvasTheme`](https://github.com/Workday/canvas-kit/pull/272/files#diff-ffbb205e8d0dc61da3dfc47e7ce6010fR7) also contains breakpoints (zero, s, m, l, xl) that can be accessed with the context. It also includes functions that return media queries based on the breakpoints (e.g. `theme.breakpoints.up('m')` will return `'@media (min-width:960px)'` by default).

Finally, since this is all based on React Contexts, we have an issue if anyone is not able to use contexts. This can be true for teams with multiple small React trees. To handle this, `CanvasProvider` can accept the `setThemeGlobal` prop. This sets the theme on `window.wdCanvas.theme`. Whenever we access the theme, we use the [`useTheme` hook](https://github.com/Workday/canvas-kit/pull/272/files#diff-ffbb205e8d0dc61da3dfc47e7ce6010fR75), which first looks for a context theme, then for a window theme, before falling back to the default. I [added an override/wrapper to Emotion's `styled` function](https://github.com/Workday/canvas-kit/pull/272/files#diff-6f9b7231fc4a056b835849871a96cb12R8) which calls `useTheme` behind the scenes.

## Checklist

- [x] `yarn test` passes
      applicable


modules/core/react/lib/theming/theme.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Typography
+  // Breakpoints
+  // Depth?
+};
+
+/**
+ * Currently a work around for a default theme if no ThemeProvider exists.
+ * Tracked on https://github.com/emotion-js/emotion/issues/1193.
+ */
+export function getTheme(theme: Object): CanvasTheme {
```
anicholls: I ran into some problems with using a default theme if users don't wrap everything in a higher order `ThemeProvider` (you can track [here](https://github.com/emotion-js/emotion/issues/1193)). This is a slightly awkward work around for now, but not too bad.

modules/core/react/lib/theming/theme.ts
```diff
@@ -0,0 +1,63 @@
+import colors from '@workday/canvas-colors-web';
+import deepmerge from 'deepmerge';
+import {CanvasTheme} from './types';
+
+/**
+ * Considerations:
```
anicholls: Lots of question around what we want to include in the theme object and how we want to calculate some of the colors. Needs more discussion.

modules/core/react/lib/theming/README.md
```diff
@@ -0,0 +1,23 @@
+### Custom Theme
```
anicholls: Need to build out this readme and link to it from the core readme.

modules/core/react/lib/CanvasProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+  static defaultProps = {
+    theme: defaultCanvasTheme,
+  };
+
+  render() {
+    const {children, theme} = this.props;
+
+    return (
+      <ThemeProvider theme={theme}>
+        <InputProvider>{children}</InputProvider>
```
anicholls: `CanvasProvider` would provide both the theme and the input type to avoid multiple HOCs

modules/core/react/lib/CanvasProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+  theme: CanvasTheme;
+}
+
+/**
+ * TODO:
+ * - Opt in for fonts
+ * - Make themes work without provider
+ */
+
+export default class CanvasProvider extends React.Component<CanvasProviderProps> {
```
anicholls: Not sure if we want this to be required or not. If we implicitly depend on `theme` in our components, consumers will get an error if they don't have a `CanvasProvider` HOC. Maybe we want to make this required anyway to enforce the `InputProvider`, but I added the `getTheme` function to work around [this](https://github.com/emotion-js/emotion/issues/1193) to keep it optional for now.

modules/banner/react/lib/Banner.tsx
```diff
@@ -47,23 +54,30 @@ const BannerWrapper = styled('button')<BannerProps>(
...
-  ({error, variant}) => ({
-    backgroundColor: error === ErrorType.Error ? colors.cinnamon500 : colors.cantaloupe400,
-    color: error === ErrorType.Error ? colors.frenchVanilla100 : colors.blackPepper400,
+  ({error, theme}) => {
+    theme = getTheme(theme); // eslint-disable-line no-param-reassign
+    return {
+      backgroundColor:
+        error === ErrorType.Error ? theme.palette.error.main : theme.palette.alert.main,
+      color: error === ErrorType.Error ? theme.palette.error.contrast : colors.blackPepper400,
+      '&:hover': {
```
anicholls: Note: I realized we don't have active states for our `Banner` component. The color/contrast gets pretty terrible based on our current system, so we should consider this.

modules/banner/react/lib/Banner.tsx
```diff
@@ -47,23 +54,30 @@ const BannerWrapper = styled('button')<BannerProps>(
...
-    },
     '&:hover': {
       cursor: 'pointer',
     },
   },
-  ({error, variant}) => ({
-    backgroundColor: error === ErrorType.Error ? colors.cinnamon500 : colors.cantaloupe400,
-    color: error === ErrorType.Error ? colors.frenchVanilla100 : colors.blackPepper400,
+  ({error, theme}) => {
+    theme = getTheme(theme); // eslint-disable-line no-param-reassign
```
anicholls: You can see `getTheme` usage here.

modules/core/react/lib/theming/README.md
anicholls: Add note about adding custom theming for a specific component (green `Switch` for example)

modules/core/react/lib/theming/theme.ts
anicholls: We should also consider the use of more transparency for flexibility in the future (dark modes, overlapping elements like selection control ripples, etc.)

modules/core/react/lib/theming/theme.ts
anicholls: After thinking about how this could scale to products that are not fully built in React and only use some React components, I think this function will be required regardless of whether this issue is fixed. 

React contexts do not span multiple trees. Depending on how an app is structured, this could involve a lot of wrapping `CanvasProviders`. I figured a better solution was a single provider with `setThemeGlobal={true}`. This would set it on the `window` object and then `getTheme` looks for the provider theme first, if that doesn't exist, it checks on the window. Finally, it returns the default theme if nothing else is defined. Thoughts on this approach?

modules/_labs/core/react/lib/theming/theme.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  if (window.wdCanvas && window.wdCanvas.theme) {
+    return window.wdCanvas.theme;
+  }
+
+  return defaultCanvasTheme;
+}
+
+// TODO: Should we use PartialCanvasTheme here?
```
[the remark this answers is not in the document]
anicholls: My thinking for leaving it out was that it may add value to allow consumers to add their own custom fields to the theme object without having to create their own provider.

modules/_labs/core/react/lib/theming/createCanvasTheme.ts
```diff
@@ -0,0 +1,97 @@
...
+    lightest,
+    light,
+    main: shades.main,
+    dark,
+    darkest,
+    contrast: shades.contrast || colors.frenchVanilla100,
+  };
+}
+
+export default function createCanvasTheme(partialTheme: PartialCanvasTheme): CanvasTheme {
```
anicholls: Should probably add some tests for this

modules/_labs/core/react/lib/theming/theme.ts
```diff
@@ -0,0 +1,94 @@
...
+ * NOTE: If you are using a class component, you MUST pass the theme.
+ * If not passed, the function will try to pull the theme from ThemeContext.
+ * If that does not work, it will try to retrieve it from the window object.
+ * As a last resort, it will return the default Canvas theme.
+ *
+ * Providing the default theme here is currently a work around for when no
+ * ThemeProvider or context exists.
+ * Tracked on https://github.com/emotion-js/emotion/issues/1193.
+ */
+export function useTheme(theme?: Object): CanvasTheme {
```
anicholls: And a test for this?

anicholls: I would like to break this out into it's own file as well.

modules/_labs/core/react/lib/theming/createCanvasTheme.ts
```diff
@@ -0,0 +1,97 @@
...
+  const {palette, breakpoints = {}} = partialTheme;
```
anicholls: Need to add `= {}` for palette.

modules/_labs/core/react/lib/theming/README.md
```diff
@@ -0,0 +1,167 @@
...
+    values: breakpoints,
+    up,
+    down,
+    between,
+    only,
+  },
+};
+```
+
+## Custom Theme
```
anicholls: Document breakpoints as well.

modules/_labs/core/react/README.md
```diff
@@ -83,3 +90,54 @@ const Box = styled('div')(space)
...
+  .add('All', () => <YourJSX />);
+```
+
+### Input Provider
+
+See the [@workday/canvas-kit-react-core docs](../../../core/react/README.md#input-provider)
+
+## Theming
+
+Theming documentation has it's own README. You can find it [here](./lib/theming/README.md)
```
stephanerangaya: Should be `its`

modules/_labs/core/react/lib/CanvasProvider.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+import {ThemeProvider} from 'emotion-theming';
+import {InputProvider} from '@workday/canvas-kit-react-core';
+import {CanvasTheme} from './theming/types';
+import {defaultCanvasTheme} from './theming/theme';
+
+export interface CanvasProviderProps {
+  theme: CanvasTheme;
+  setThemeGlobal: boolean;
```
lychyi: `setThemeGlobal` seems to encapsulate the concerns of two different components here. To me a 'provider' provides the subtree with some data that it needs to render properly (in this case the theme). But with `setThemeGlobal`, this component also can also modify the window which is a global side-effect.

Because of this, we can have some weird issues when `setThemeGlobal` is true on multiple providers. The global theme can flip-flop causing places that don't use a `CanvasProvider` for their theme to unexpectedly change.

Could we provide a separate component to explicitly take the role of setting the global instead? That way we're declaratively doing a side-effect with that one and the `CanvasProvider` can remain free from it (pure). Alternatively, we could just document that this component comes with a side-effect that users need to be aware of but my gut feeling is that this isn't a typical React pattern to have a component do two very distinct things b/c of the unpredictable behavior if used incorrectly.

Usage:
```jsx
// In the main App.jsx
// Sets global theme and other global things to your app maybe
<CanvasGlobal theme={...} />

// Somewhere else, I need a more specific theme for this subtree
<CanvasProvider theme={...}>
...
</CanvasProvider>
```

modules/_labs/core/react/lib/theming/breakpoints.ts
```diff
@@ -0,0 +1,64 @@
...
+  zero: number;
+  s: number;
+  m: number;
+  l: number;
+  xl: number;
+};
+
+export const breakpointKeys = ['zero', 's', 'm', 'l', 'xl'] as const;
+
+export const breakpoints: CanvasBreakpoints = {
```
lychyi: This may be a separate discussion but I think a breakpoint system like this might be a bit too stiff. It's assuming that all that matters is `width`. What if someone wants to have more complex [media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries) (height and width for example)?

One alternative is to just to keep this a simple `{[key: string]: string}` object and match breakpoint names to media queries directly:
```
// default breakpoints
{
  ...
  xl: '@media (min-width: 1920px)'
}
```

And then users can overwrite the media queries themselves or add new ones if they need to be more flexible.

modules/_labs/core/react/lib/CanvasProvider.tsx
anicholls: Totally agree - good call. I will revise.

modules/_labs/core/react/lib/theming/breakpoints.ts
anicholls: Hmm that's fair. I believe the idea was that this is for our own components rather than than for consumers to use (though it is available to them if they want it). I figured we'd only have media queries based on width and anything that's height based would be case-specific. So you could do something like ``@media (min-height: 500px) and (min-width: ${theme.breakpoints.l})`` if absolutely necessary.

Since this is going into labs, we have flexibility for now. IMO, this is good enough and we can iterate on it if we feel we need to after getting some feedback.

modules/_labs/core/react/lib/CanvasProvider.tsx
anicholls: @lychyi I just pushed a change for this. I ended up just [recommending that they add the theme to the window object themselves](https://github.com/Workday/canvas-kit/pull/272/commits/a4f0fa79ac3404a7d9a6facc39f6c754ebea24c3#diff-c895718dba9f9b5b971bb7cf5538a92fR167). It's easier for non-react apps to do this where needed. It seemed overkill to have a component just to do this. Let me know your thoughts.

lychyi: @anicholls I think that's a good idea. We can always think about it later if people are looking for a declarative way to handle theming on the window.

modules/core/react/README.md
```diff
@@ -312,17 +318,17 @@ type updates from the events above, the intent type will also be updated to the
...
-## Usage
+### Usage
 
 As an external consumer, you should reference the following example.
 
 If you are contributing a component, you must add the necessary styling (see below) and use the
-[`InputProviderDecorator`](#storybook-decorator) in your stories. _DO NOT_ wrap any canvas kit
-components in an `InputProvider`.
+[`InputProviderDecorator`](#storybook-decorator) in your stories. _DO NOT_ use an `InputProvider`
+directly within any canvas kit components.
```
lychyi: Nit: Canvas Kit

.storybook/config.js
```diff
@@ -1,18 +1,23 @@
...
+import { withKnobs } from "@storybook/addon-knobs/react";
+import { create } from "@storybook/theming";
+import addons from "@storybook/addons";
+import Events from "@storybook/core-events";
+import { toId } from "@storybook/router";
+import ReactDOM from "react-dom";
 
-import {commonColors, typeColors, fontFamily} from '../modules/core/react';
-import {InputProviderDecorator, FontsDecorator} from '../utils/storybook';
+import { commonColors, typeColors, fontFamily } from "../modules/core/react";
```
anicholls: Just noticed there's a bunch of weird formatting stuff going on in this file. Will fix.

anicholls:
So @stephanerangaya found a fun gotchya. Since both Canvas Kit and Storybook are now using emotion theming, we have conflicts due to both referencing the [same context](https://github.com/emotion-js/emotion/blob/master/packages/core/src/context.js#L17). We discovered this from an error in `@storybooks/addon-docs` when trying to render the docs tab.

After doing a bit more research into the Emotion team's plans for theming, it sounds like we should take this into our own hands. [This RFC](https://github.com/emotion-js/emotion/pull/973) discusses it and, while outdated, it has been continuously referenced in theming PRs and issues since it's creation. Since the Emotion API is so shallow, it seems like their plan is to encourage consumers to handle theming themselves with React's context and a `useTheme` hook.

In other words:
```tsx
import { createContext, useContext } from 'react'

let defaultTheme = {
  colors: {
    primary: 'hotpink',
    secondary: 'green'
  }
}

let ThemeContext = createContext(defaultTheme)

export let ThemeProvider = ThemeContext.Provider

export let useTheme = () => useContext(ThemeContext)
```

We've already done the `useTheme` bit. Rolling our own `ThemeProvider` should avoid this context conflict issue.

modules/_labs/core/react/lib/theming/createCanvasTheme.ts
```diff
@@ -0,0 +1,97 @@
...
+  const light = shades.light || shiftColor(shades.main, ColorDirection.Brighten);
+  const lightest = shades.lightest || shiftColor(light, ColorDirection.Brighten);
+
+  return {
+    lightest,
+    light,
+    main: shades.main,
+    dark,
+    darkest,
+    contrast: shades.contrast || colors.frenchVanilla100,
```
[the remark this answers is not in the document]
anicholls: The contrast is meant to go over the darkest colors. I would generally not ever expect to have a dark text color work on the darkest colors. So if they've entered a shade that's going to generate a super light palette, I think that's on them.

anicholls:
Update on the conflict with `@storybook/addon-docs`. I created our own theme context/provider and that still didn't resolve the issue. After a lot of _fun_ git bisecting, I ended up needing to enforce `emotion-theming@10.0.10` to keep the error away. I tried updating to the latest version of `@storybook/addon-docs` and that  did not fix it.

modules/_labs/core/react/lib/theming/useTheme.ts
```diff
@@ -0,0 +1,45 @@
...
+    const context = React.useContext(ThemeContext);
+    if (context) {
+      return context as CanvasTheme;
+    }
+  } catch (e) {
+    // Context not supported or invalid (probably called from within a class component)
+  }
+
+  if (window.wdCanvas && window.wdCanvas.theme) {
+    return window.wdCanvas.theme;
```
anicholls: This should change to `window.workday.canvas.theme`

modules/_labs/core/react/package.json
```diff
@@ -39,6 +39,8 @@
   "dependencies": {
     "@emotion/core": "^10.0.21",
     "@workday/canvas-colors-web": "^0.17.13",
-    "@workday/canvas-kit-react-core": "^3.1.0"
+    "@workday/canvas-kit-react-core": "^3.1.0",
+    "chroma-js": "^1.4.0",
```
lychyi: Should we update `chroma-js` and types here? So we can also close #229 ?

anicholls: Are you wanting to re-write the color contrast function as part of this PR?

modules/_labs/core/react/lib/CanvasProvider.tsx
```diff
@@ -0,0 +1,25 @@
...
+  static defaultProps = {
+    theme: defaultCanvasTheme,
+  };
+
+  render() {
+    const {children, theme} = this.props;
+
+    return (
+      <ThemeProvider theme={theme}>
+        <InputProvider>{children}</InputProvider>
```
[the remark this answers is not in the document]
anicholls: Multiple instances of `InputProvider` will work fine from a consumer level, but I can't speak to the performance implications of doing that.

We could easily add a check in `InputProvider` to noop if one exists higher in the tree. This would be my vote rather than adding prop complexity to `CanvasProvider` or removing `CanvasProvider` completely.

I realize `CanvasProvider` is not doing much right now. However, there are two reasons I would like to keep it:
- Future proofing. We will generally always need some sort of top-level component. I assume in the future we will need to add more to it. When we do, it's more likely that consumers will get that future functionality for free. 
- Reduces the complexity from the consumers end. They don't really need to know about the theme and input provider if they're not using it.
```tsx
<CanvasProvider>
  // My App
</CanvasProvider>
```
is much simpler and hides our noise, compared to:
```tsx
<ThemeProvider>
  <InputProvider>
    <FutureStuff?>
      // My App
    </FutureStuff?>
  </InputProvider>
</ThemeProvider>
```

Anyway, let me know if you agree. I will probably add that functionality to `InputProvider` either way.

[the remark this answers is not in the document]
anicholls: @lychyi noop added to `InputProvider` in https://github.com/Workday/canvas-kit/pull/272/commits/6c2d52513309feb1fdcadff9548d814a6faa52e1

PR #273, opened 2019-10-22
ci: Update to use chroma/action

NicholasBoll:
The Chromatic team recently updated their Github Action based on feedback we provided for forked repos. This change should reduce the complexity in our code.

I have to test this PR to make sure Chromatic can still be registered correctly. After that I have to introduce a visual change to make sure the correct baseline is chosen.

## Additional References

https://github.com/chromaui/action


NicholasBoll:
It looks like the correct baseline (branch) is chosen in the action. I haven't received a response on the configuration for the master branch yet (auto-accepting changes) so I'm leaving that one alone for now. The check seems to be working.

.github/workflows/chromaticqa.yml
```diff
@@ -16,14 +16,7 @@ jobs:
...
-      run: yarn chromatic --exit-zero-on-changes
-      env:
-        CI: true
-        CHROMATIC_BRANCH: ${{ github.event.pull_request.head.ref }}
-        CHROMATIC_SHA: ${{ github.event.pull_request.head.sha }}
-        CHROMATIC_APP_CODE: dlpro96xybh
+    - uses: chromaui/action@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        appCode: dlpro96xybh
```
anicholls: Could we not use `${{ secrets.CHROMATIC_APP_CODE }}` here too?

Issue #266, opened 2019-10-15
Drawer Component


mannycarrera4:
## 🚀 Feature Proposal

Develop a drawer / panel component.

## Motivation

We currently have a `Side Panel` component, but there are multiple use cases where the current configuration isn't flexible enough. The goal for the drawer/ side panel is to have different "flavors" in which the consumer can pick and choose the component that best suits them.

lychyi:
Done via #277

PR #277, opened 2019-10-24
feat(labs): Add Drawer component

mannycarrera4:
## Summary

Addresses: https://github.com/Workday/canvas-kit/issues/266

## Checklist

- [x] `yarn test` passes
      applicable


modules/_labs/panel/react/README.md
```diff
@@ -0,0 +1,42 @@
+# Canvas Kit React Panel
+
+<a href="https://github.com/Workday/canvas-kit/tree/master/modules/_labs/README.md">
+  <img src="https://img.shields.io/badge/LABS-alpha-orange" alt="LABS: Alpha" />
+</a>  This component is work in progress and currently in pre-release.
+
+
+
+## Installation
```
anicholls: You can only install by adding `@workday/canvas-kit-labs-react-panel`.

I'll add an issue to track the fact that the create-component script isn't doing this for us.

modules/_labs/panel/react/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```sh
+yarn add @workday/canvas-kit-react-panel
+```
+
+## Usage
+
+```tsx
+import * as React from 'react';
+import Panel from '@workday/canvas-kit-react-panel';
```
anicholls: labs

modules/_labs/panel/react/lib/Panel.tsx
```diff
@@ -0,0 +1,91 @@
+import * as React from 'react';
+import styled from '@emotion/styled';
+import {colors, spacing, CanvasSpacingValue} from '@workday/canvas-kit-react-core';
+
+export interface PanelProps extends React.HTMLAttributes<HTMLDivElement> {
+  padding?: CanvasSpacingValue;
```
anicholls: Can you document these props?

modules/_labs/panel/react/lib/Panel.tsx
```diff
@@ -0,0 +1,91 @@
...
+    } = this.props;
+
+    return (
+      <PanelContainer
+        {...elemProps}
+        showDropShadow={showDropShadow}
+        width={width}
+        openDirection={openDirection}
+      >
+        {header ? header : undefined}
```
anicholls: Nit: `{header}` will return undefined if it's not defined haha

modules/_labs/panel/react/README.md
anicholls: Created #279 to track.

modules/_labs/panel/react/lib/PanelHeader.tsx
```diff
@@ -0,0 +1,58 @@
...
+import {colors, spacing, H4, CanvasColor} from '@workday/canvas-kit-react-core';
+import {IconButton} from '@workday/canvas-kit-react-button';
+import {CanvasSystemIcon} from '@workday/design-assets-types';
+import {xIcon} from '@workday/canvas-system-icons-web';
+
+export interface PanelProps extends React.HTMLAttributes<HTMLDivElement> {
+  headerTitle?: string;
+  handleClose?: () => void;
+  iconLabel: string;
+  icon: CanvasSystemIcon;
```
anicholls: What are other icon use cases? How do they close/dismiss if it's not an X?

modules/_labs/panel/react/lib/PanelHeader.tsx
```diff
@@ -0,0 +1,58 @@
...
+const headerHeight = 56;
+
+const HeaderContainer = styled('div')<Pick<PanelProps, 'headerColor'>>(
+  {
+    height: headerHeight,
+    display: 'flex',
+    alignItems: 'center',
+    padding: `0 ${spacing.s}`,
+    justifyContent: 'space-between',
+    borderBottom: `1px solid ${colors.soap500}`,
```
anicholls: Design thing, but this border may look weird with some colors.

modules/_labs/panel/react/package.json
```diff
@@ -0,0 +1,40 @@
+
+{
+  "name": "@workday/canvas-kit-labs-react-panel",
+  "version": "0.0.0",
+  "description": "",
```
anicholls: Can you add a description?

modules/_labs/panel/react/lib/Panel.tsx
```diff
@@ -0,0 +1,91 @@
+import * as React from 'react';
+import styled from '@emotion/styled';
+import {colors, spacing, CanvasSpacingValue} from '@workday/canvas-kit-react-core';
+
+export interface PanelProps extends React.HTMLAttributes<HTMLDivElement> {
+  padding?: CanvasSpacingValue;
+  openDirection?: PanelDirection;
```
anicholls: What are your thoughts on `SidePanel` being controllable but this one not?

Also, do you plan to update `SidePanel` to use this eventually?

modules/_labs/panel/react/lib/Panel.tsx
mannycarrera4: yep yep!

modules/_labs/panel/react/lib/PanelHeader.tsx
mannycarrera4: I think I was trying to account for another icon/link up there, but I should probably keep it simple and provide the `xIcon` and if we need to iterate over then we can consider adding this capability!

modules/_labs/panel/react/lib/PanelHeader.tsx
mannycarrera4: I'll talk to Michael about this and consider a different approach!

modules/_labs/panel/react/lib/Panel.tsx
mannycarrera4: Considering that this component will probably we contained within another element, I'm ok with removing this prop.

Also, I'd love to create a a component that can be added to and build upon. So yes, I'd love for side panel to be build on the panel component. I was talking to @NicholasBoll and we discussed separating out the pieces of it, which is why I moved `PanelHeader` into its own component instead of having a super flexible single component.

anicholls: Oh I wasn't saying we should remove `openDirection`. I think this is a good way to call out which side the border, etc. should be on. I was just pointing out the fact that you can actually control whether `SidePanel` is open/closed where as this one you have to wrap with a component that does that for you. I wanted to get your thoughts on why we have them structured that way.

modules/_labs/panel/react/lib/Panel.tsx
anicholls: I meant in the readme haha. However, it sounds like we'll be dropping props in the readme with the new storybook format, so this is probably fine for now.

mannycarrera4: Yeah, I was talking to @NicholasBoll  about this, and since we're moving in this direction, I added them in the file :)

modules/_labs/panel/react/lib/Panel.tsx
mannycarrera4: Ah sorry, yeah misread that. If this component where manage being open/closed I'd say then yes, but my thought was, by leaving it out, it allows the consumer to manage if they want to put in something like layout, or control from where it opens, i suppose a little more flexible. As a base component i chose this approach, but something like side panel, it made sense to include that functionality. I'm open to adding it, especially maybe some animation? Thoughts?

anicholls: Makes sense - I just wanted to hear your thoughts!

modules/_labs/panel/react/stories/stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+      </div>
+    </div>
+  ))
+  .add('With Header', () => (
+    <div className="story">
+      <div style={{height: '80vh', position: 'relative'}}>
+        <Panel
+          header={
+            <PanelHeader
+              onClose={() => console.warn('close clicked')}
```
anicholls: Nit: Can you use storybook actions for this instead?

modules/_labs/panel/react/lib/Panel.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const PanelContainer = styled('div')<
+  Pick<PanelProps, 'width' | 'showDropShadow' | 'openDirection'>
+>(
+  {
+    height: '100%',
+    backgroundColor: 'white',
+    display: 'flex',
+    flexDirection: 'column',
+    position: 'absolute',
```
anicholls: The design specs show this with rounded borders if it is "popped out" from the side of the page. How do we intend to support that use case? Is that future functionality to be added?

modules/_labs/panel/react/spec/Panel.spec.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import Panel, {PanelHeader} from '../index';
+import {mount} from 'enzyme';
+// import ReactDOMServer from 'react-dom/server';
```
anicholls: Nit: Let's remove these.

modules/_labs/panel/react/spec/Panel.spec.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import Panel, {PanelHeader} from '../index';
+import {mount} from 'enzyme';
```
anicholls: Next time let's use RTL!

mkellman-workday:
@mannycarrera4 This looks great from a design perspective.  It's exactly what we spec'd out.

modules/_labs/drawer/react/lib/DrawerHeader.tsx
```diff
@@ -0,0 +1,107 @@
...
+   */
+  headerColor: CanvasColor | string;
+  /**
+   * Changes the border color to match something close to the header background color
+   */
+  borderColor: CanvasColor | string;
+  /**
+   * If the background of the header is dark, this will switch the icon so that it is white
+   */
+  showInverseButton?: boolean;
```
anicholls: Nit: Both this and `titleColor` are impacted by the same thing (headerColor). Why not just have an `inverse?: boolean` flag that changes the button type and changes the text color to white?

modules/_labs/drawer/react/spec/Drawer.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+        Hello World
+      </Drawer>
+    );
+
+    fireEvent.click(await findByLabelText('Close'));
+
+    expect(cb).toHaveBeenCalledTimes(1);
+  });
+
+  test('Modal should spread extra props', async () => {
```
anicholls: Remove the word "Modal" here (it's already in a "Drawer" describe block)

modules/_labs/drawer/react/lib/DrawerHeader.tsx
mannycarrera4: ah yep, that makes sense!

modules/_labs/drawer/react/spec/Drawer.spec.tsx
mannycarrera4: bad copy pasta

modules/_labs/drawer/react/spec/Drawer.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import {Drawer, DrawerHeader, DrawerDirection} from '../index';
+
+describe('Drawer', () => {
+  test('should call a callback function', async () => {
+    const cb = jest.fn();
+    const {findByLabelText} = render(
+      <Drawer
+        header={
+          <DrawerHeader iconLabel={'Close'} title={'Header Title'} onClose={cb}></DrawerHeader>
```
NicholasBoll: Is this a test of the `DrawerHeader` component? Is the `Drawer` component necessary?

modules/_labs/drawer/react/lib/Drawer.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <DrawerContainer
+        aria-labelledby={header ? ariaLabeledBy : undefined}
+        role={role}
+        aria-label={!header ? ariaLabel : undefined}
+        {...elemProps}
+        showDropShadow={showDropShadow}
+        width={width}
+        openDirection={openDirection}
+      >
+        {header && header}
```
NicholasBoll: Not sure why this code... The following should work fine
```suggestion
        {header}
```

modules/_labs/drawer/react/lib/Drawer.tsx
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Optional to pass a DrawerHeader component shows an optional string and close button
+   */
+  header?: React.ReactElement;
+  /**
+   * aria-labeledby when there's a header for accessibility
+   */
+  'aria-labelledby'?: string;
+  /**
+   * aria-label when there is NO header for accesssibility
```
NicholasBoll:
```suggestion
   * aria-label when there is NO header for accessibility
```

modules/_labs/drawer/react/lib/Drawer.tsx
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Allows to either show or hide a drop shadow on the Drawer. Default: `false`
+   */
+  showDropShadow?: boolean;
+  /**
+   * Optional to pass a DrawerHeader component shows an optional string and close button
+   */
+  header?: React.ReactElement;
+  /**
+   * aria-labeledby when there's a header for accessibility
```
NicholasBoll:
```suggestion
   * Use when there's a header for accessibility. The `role` attribute should also be used when this attribute is present.
```

modules/_labs/drawer/react/lib/Drawer.tsx
```diff
@@ -0,0 +1,120 @@
...
+   */
+  'aria-labelledby'?: string;
+  /**
+   * aria-label when there is NO header for accesssibility
+   */
+  'aria-label'?: string;
+  /**
+   * Optional role for the drawer
```
NicholasBoll:
```suggestion
   * Optional role for the drawer. If this attribute is used, also provide an `aria-labelledby` attribute to link the `header` element to this role
```

modules/_labs/drawer/react/lib/DrawerHeader.tsx
```diff
@@ -0,0 +1,108 @@
+import * as React from 'react';
+import styled from '@emotion/styled';
+import {typeColors} from '@workday/canvas-colors-web';
+import {colors, spacing, H4, CanvasColor} from '@workday/canvas-kit-react-core';
+import {IconButton, IconButtonVariant} from '@workday/canvas-kit-react-button';
+import {xIcon} from '@workday/canvas-system-icons-web';
+
+export interface DrawerHeaderProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * String to add a title to a Drawer
```
NicholasBoll:
```suggestion
   * Optional title for the Drawer header. Useful if the header element gets truncated.
```

modules/_labs/drawer/react/lib/DrawerHeader.tsx
```diff
@@ -0,0 +1,108 @@
...
+   */
+  title?: string;
+  /**
+   * Callback to handle closing the Drawer
+   */
+  onClose: () => void;
+  /**
+   * String to add aria-label to the icon button
```
NicholasBoll:
```suggestion
   * Add an `aria-label` to the close icon button. Default 'Close'. Useful for i18n
```

modules/_labs/drawer/react/lib/DrawerHeader.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export interface DrawerHeaderProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * String to add a title to a Drawer
+   */
+  title?: string;
+  /**
+   * Callback to handle closing the Drawer
+   */
+  onClose: () => void;
```
NicholasBoll:
```suggestion
  onClose: IconButtonProps['onClick']
```

modules/_labs/drawer/react/lib/DrawerHeader.tsx
```diff
@@ -0,0 +1,108 @@
...
+import {typeColors} from '@workday/canvas-colors-web';
+import {colors, spacing, H4, CanvasColor} from '@workday/canvas-kit-react-core';
+import {IconButton, IconButtonVariant} from '@workday/canvas-kit-react-button';
+import {xIcon} from '@workday/canvas-system-icons-web';
+
+export interface DrawerHeaderProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * String to add a title to a Drawer
+   */
+  title?: string;
```
NicholasBoll: This seems more like `headerText` and shouldn't be optional

modules/_labs/drawer/react/lib/DrawerHeader.tsx
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Callback to handle closing the Drawer
```
NicholasBoll:
```suggestion
   * Called when the user clicks the close icon in the header
```

NicholasBoll:
Accessibility might want a story that works within a context valid for accessibility (Drawer has a role and the close button works)

PR #288, opened 2019-10-31
feat: Add bidirectionality support

### LINKED ISSUE #278: Support Bidirectionality
> ✍️🐧 Feel free to edit this issue

## 🚀 Feature Proposal

Canvas Kit components need to support both Right-to-Left and Left-to-Right writing systems. This issue aims to centralize the work we will need to do to support bidirectionality.

For now, we're only looking at React components that are not in Labs, but we should discuss about the way we handle CSS before we start the work on React components.

### Questions:
- What font should we use? Potentially https://www.google.com/get/noto/
- What prop should we use? We're already using `isRTL`, so probably keep doing this
- Should isRTL be propagated to children components? Let's say we use a `IconButton` with a `Back` Icon, should we do:
```tsx
<IconButton title="Back">
   <SystemIcon icon={leftArrow} />
</IconButton>

<IconButton isRTL={true} title="Back">
   <SystemIcon icon={rightArrow} />
</IconButton>
```
or (my personal preference)
```tsx
<IconButton title="Back">
   <SystemIcon icon={leftArrow} iconLTR={rightArrow}/>
</IconButton>
<IconButton isRTL={true} title="Back">
   <SystemIcon icon={leftArrow} iconLTR={rightArrow}/>
</IconButton>
```

or
```tsx
<IconButton title="Back">
   <SystemIcon icon={leftArrow} iconLTR={rightArrow}/>
</IconButton>
<IconButton isRTL={true} title="Back">
   <SystemIcon icon={leftArrow} iconLTR={rightArrow} isRTL={true}/>
</IconButton>
```

Facts:
- A search icon is keep the same orientation because we hold with the right hand
- Numbers are still LTR in RTL languages.
- Some icons need to flip when in RTL like bullet lists, back button, etc... Some don't like the Play or Skip Track.
- Time goes the same direction, so loadings are also in the same direction. Same goes for Undo and Redo, icons should not flip.

#### Action Bar
In RTL, the elements should be align to the right. The order of the elements in the bar should be reversed.

#### Avatar
This component supports bidirectionality out of the box.

#### Banner
In RTL, the order of the elements in the banner should be reversed. We should update to documentation to also recommend flipping the placement of the `sticky` Banner to the other side of the screen when in RTL.

> ❓Should we manage the placement of the `sticky` Banner?

#### Button
- Text: Icons placement should be reversed.

> ❓In RTL, if a button has a directional arrow, for example `← Home`, it should be `الرئيسيّة →`, so not only the icon/text order is reversed, but the direction of the arrow is also different. We should update the documentation to recommend a isRTL icon when this occurs.
- Outline: This component supports bidirectionality out of the box.
- Dropdown: The arrow should be on the left.
- Icon Button
> ❓In RTL, if an Icon Button has a directional icon, for example a bullet list, it should be reversed. But for Button without a directional icon, they shouldn't. We should allow an isRTL icon. 
image
- Icon Button Toggle Group:  This component supports bidirectionality out of the box.

#### Card
In RTL, the alignement should change.

#### Common
This component supports bidirectionality out of the box.

#### Cookie Banner
In RTL, the order of the elements in the banner should be reversed. 

#### Core
This component supports bidirectionality out of the box.

#### Form Field
- Label: Text should be aligned to the right
- Hint: Text should be aligned to the right
- Checkbox, Radio: Component input and text order should be reversed
> ❓In RTL, it seems the direction of the Checkbox mark should stay the same than in LTR, but to be confirmed.
- Color Input/Preview: Component input and text order should be reversed, but the # should still be kept to the left of the hex code.
- Select: Arrow and text order should be reversed
- Switch: Component input and text order should be reversed. The selected state should place the dot to the left, and the unselected state should put it to the the right (reversed to what it is currently).
- TextArea: Component should resize to the left, text should be aligned to the right
- TextInput: Text should be aligned to the right

#### Icon
We should allow an isRTL icon, for when the Icon direction should change depending on the text direction.

#### Layout
This component supports bidirectionality out of the box.

#### Loading Animation
This component supports bidirectionality out of the box.

Note: because a loading animation is showing basically showing progress over time and time is going forward, progress bars and loading animation are not reversed (same if we were building a Podcast player app, the progress would go the same direction, and the Play icon would also be the same direction).

Therefore, it makes sense to keep the dots animating at they are for all languages.

#### Modal & Popup
In RTL, the order of the elements in the main container should be reversed. The CTA should be on the right, the close icon should be on the left.

#### Page Header
In RTL, the order of the elements in the main container should be reversed.

#### Side Panel
In RTL, the Expand/Collapse icon should be on the left and flipped.

#### Sketelon
This component supports bidirectionality out of the box, but `Complete` should be flipped.
> ❓In RTL, should we also reverse the direction of the animation?

#### Status Indicator
In RTL, the order of the elements in the main container should be reversed.

#### Table
In RTL, the order of the columns should be reversed.
> ❓Number are tricky, roman number should be displayed in LTR but arabic numbers should not for example. 

#### Toast
In RTL, the order of the elements in the main container should be reversed. 

#### Tooltip
This component supports bidirectionality out of the box.

stephanerangaya:
Closes #278

# Support Bidirectionality

We're building support for bidirectionality into our `theme` context.

[View `theme.ts` →](https://github.com/stephanerangaya/canvas-kit/blob/bidirectionality/modules/_labs/core/react/lib/theming/theme.ts)
```
export const defaultCanvasTheme: CanvasTheme = {
  palette: { ... },
  breakpoints: { ... },
  direction: ContentDirection.LTR,
};
```
As you can see, the direction of the `defaultCanvasTheme` is `ltr`.

In order to set a different direction for a set of components, developers using Canvas Kit React (we're not supporting Canvas Kit CSS at the moment), will have to create a theme and set the `direction` to the desired value, and then wrap the components into a `CanvasProvider`:

[View Example for Radio `stories_Radio.tsx` →](https://github.com/stephanerangaya/canvas-kit/blob/bidirectionality/modules/form-field/react/stories/stories_Radio.tsx)
```
const rtlCanvasTheme = createCanvasTheme({ direction: ContentDirection.RTL })
```

```
<CanvasProvider theme={rtlCanvasTheme}>
  <ControlledComponentWrapper>
    <RadioGroup name="contact">
      <Radio id="1" value="email" label="E-mail" />
      <Radio id="2" value="phone" label="Phone" />
      <Radio id="3" value="fax" label="Fax (disabled)" disabled={true} />
      <Radio id="4" value="mail" label="Mail" />
    </RadioGroup>
  </ControlledComponentWrapper>
</CanvasProvider>
```

It is also possible to nest `CanvasProvider` components to build UIs following different directions.

## Implementation

We're using [`rtlCSSJS`](https://www.npmjs.com/package/rtl-css-js), a library that takes a CSS Object and flips attributes that have a horizontal direction. 

The direction is passed as part of `CanvasTheme` using contexts, which can be accessed and used by any component extending `Themeable`.

Because we don't want to use a boolean check on each `styled` function to know wether or not to flip the styles using `rtlCSSJS`, we're using the new `styled` function, part of Theming, which takes a CSS Object and applies `rtlCSSJS` based on the theme direction.


.storybook/config.js
```diff
@@ -13,6 +13,7 @@ const req = require.context('../modules', true, /stories.*\.tsx?$/);
 
 function loadStories() {
   req.keys().forEach(req);
+  document.body.dir = 'rtl';
```
stephanerangaya: This sets `dir='rtl'` to the body element, triggering the browser's algorithm to flip most elements to `RTL`. We should use a context to do that at some point.

modules/checkbox/react/lib/Checkbox.tsx
```diff
@@ -31,6 +33,8 @@ const checkboxLabelDistance = spacing.xs;
 const checkboxWidth = 18;
 const rippleRadius = (spacing.l - checkboxWidth) / 2;
 
+const direction = Direction.RTL;
```
stephanerangaya: This value should come from `CanvasProvider`

modules/core/react/lib/DirectionProvider.tsx
```diff
@@ -0,0 +1,16 @@
+import rtlCSSJS from 'rtl-css-js';
+import {CSSObject} from '@emotion/css';
+
+export enum Direction {
+  LTR,
+  RTL,
+}
+
+function DirectionProvider(dir: Direction, styles: CSSObject) {
```
stephanerangaya: I'm not sure we should call this a `Provider`, open to suggestions

modules/popup/react/lib/Popup.tsx
```diff
@@ -29,6 +34,7 @@ export interface PopupProps extends React.HTMLAttributes<HTMLDivElement> {
 
 const closeIconSpacing = spacing.xs;
 const closeIconSpacingSmall = 10;
+const direction = Direction.RTL;
```
stephanerangaya: This value should come from `CanvasProvider`

modules/status-indicator/react/lib/StatusIndicator.tsx
```diff
@@ -149,6 +156,8 @@ export default class StatusIndicator extends React.Component<StatusIndicatorProp
     emphasis: StatusIndicatorEmphasis.High,
   };
 
+  direction = Direction.RTL;
```
stephanerangaya: This value should come from `CanvasProvider`

modules/toast/react/lib/Toast.tsx
```diff
@@ -18,37 +25,45 @@ export interface ToastProps {
   transformOrigin?: TransformOrigin;
 }
 const toastWidth = 360;
+const direction = Direction.RTL;
```
stephanerangaya: This value should come from `CanvasProvider`

modules/checkbox/react/lib/Checkbox.tsx
```diff
@@ -179,24 +183,27 @@ const CheckboxInput = styled('input')<CheckboxProps>(
...
+  DirectionProvider(direction, {
+    alignItems: 'center',
+    backgroundColor: colors.frenchVanilla100,
+    borderRadius: borderRadius.s,
+    border: '1px solid ' + inputColors.border,
+    boxSizing: 'border-box',
+    display: 'flex',
+    height: checkboxHeight,
+    justifyContent: 'center',
+    marginTop: direction === Direction.RTL ? '-22px' : 0,
```
stephanerangaya: This is to show how we could specify styles to a direction. This is to fix the checkbox in RTL but unfortunately it only works in Chrome (Safari works without the need for this fix), so Checkbox will need some kind of refactoring I think.

.storybook/config.js
NicholasBoll: There is a storybook addon: https://www.npmjs.com/package/storybook-addon-rtl

But from talking to the ChromaticQA team, they suggest stories that enable a RTL mode so that we can get RTL snapshots as regular stories rather than trying to globally flip and take screenshots for stories in LTR mode and RTL mode. They said other Design Systems of our scale have done the same thing.

modules/checkbox/react/lib/Checkbox.tsx
NicholasBoll: Assuming that RTL exists in the `CanvasProvider`, will we be able to change it from a story?

modules/checkbox/react/lib/Checkbox.tsx
```diff
@@ -44,7 +48,7 @@ const CheckboxContainer = styled('div')({
  * This stops the ripple from showing when you hover on the label.
  */
 const CheckboxInputWrapper = styled('div')<Pick<CheckboxProps, 'disabled'>>(
-  {
+  DirectionProvider(direction, {
```
NicholasBoll: If `direction` came from a React context, how would this API change?

.storybook/config.js
anicholls: @NicholasBoll Our plan was to also add optional rtl props to everything so we could test individually with those couldn't we?

modules/core/react/lib/DirectionProvider.tsx
anicholls: You could mimic the `useTheme` hook and end up wtih something like:

```tsx
styled('div')(
  useDirection(direction)({
    //styles
  })
)
```

maybe?

anicholls: However, this doesn't account for an overriding `isRTL` prop (or something similar).

modules/checkbox/react/lib/Checkbox.tsx
anicholls: We won't unfortunately, but that's what the optional prop is for

modules/checkbox/react/lib/Checkbox.tsx
stephanerangaya: I don't think this API would change, because we would get the `direction` as a prop and use it here.

modules/core/react/lib/DirectionProvider.tsx
stephanerangaya: Why wouldn't it? direction could be set at the component level, where we look at HTML, Context and then prop?

.storybook/config.js
NicholasBoll: Adding `dir="rtl"` should work on any level of the DOM. The story could add this attribute to the story container. The alternative is a decorator that adds the `document.body.dir` change. And a prop for `RTL` for components that need it. Will all components need a RTL prop?

modules/_labs/core/react/lib/theming/applyDirection.ts
```diff
@@ -0,0 +1,47 @@
...
+    return React.cloneElement(child as React.ReactElement, {
+      direction: direction,
+    });
+  });
+}
+
+/**
+ * This is used to apply `dir={direction}` to the top level element
+ */
+function applyDirection(
```
anicholls: @stephanerangaya This addresses the issue you were running into (where the `direction` prop was not being passed to all styled components from the parent component.

I'm unsure about the performance implications on this, but it may be worth it for the better devX.

On one hand, we could do it manually:
- It's a huge pain to pass `direction={direction}` to every single styled component. This is just asking for developer error.
- We have to manually ensure `dir={direction}` is added to the top level container of every component
- This is the fastest performance wise

On the other hand, we could use `applyDirection` function:
- This barely adds any extra markup in components
- Automatically adds `dir` attribute and passes `direction` to all children, reducing the likelihood of dev errors
- It's iterating over every child of a component and creating a clone (double the memory) on render. Not sure how quick this is and what garbage collection is like on the clones, but it isn't ideal. Would love some input from anyone who has more experience with the perf impacts of this.

lychyi:
@stephanerangaya @anicholls Can we do a quick write-up in the PR comments to outline the approach and how we're solving bidirectionality in this PR? I would love to take a look at it but it would help if some context was added so I can more quickly figure out what's going on.

modules/_labs/core/react/lib/theming/applyDirection.ts
NicholasBoll: The comment doesn't say _why_ this function is required. I'm unsure as well. Aren't we defining `direction` on the theme object which is provided through React context? Why do we need to add `dir` and `direction` to every child component? Doesn't the `styled` helper provide it? Is it for projected content we don't control?

modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -1,16 +1,19 @@
 import * as React from 'react';
-import styled from '@emotion/styled';
 import {CSSObject, jsx} from '@emotion/core';
 import {accessibleHide, GrowthBehavior} from '@workday/canvas-kit-react-common';
 import {depth, spacing, commonColors, borderRadius} from '@workday/canvas-kit-react-core';
 import {MenuItemProps} from '@workday/canvas-kit-labs-react-menu';
+import {styled, Themeable, ContentDirection} from '@workday/canvas-kit-labs-react-core';
```
NicholasBoll: This import violates the rule where labs components can import from production or other labs components, but production can't import from labs. This might be necessary in this case, especially if the prod APIs don't change with the changes to the labs components/utilities that are relied on. In this case, `theme` might have to be an unstable API inside a stable API.

modules/_labs/core/react/lib/theming/applyDirection.ts
anicholls: Yes, the `styled` helper providers the direction from the context. However, we also want to support a prop so users can change the direction explicitly for specific components. When they pass the prop to the component, we then need to share that with all of the styled components. That's what this is achieving.

Alternatively, @stephanerangaya and I have been talking about requiring the user to wrap the component with another `CanvasProvider` with `direction` changed if they need to swap the direction for a specific component.

modules/_labs/combobox/react/lib/Combobox.tsx
anicholls: Yep - if we want to support it for testing purposes (or to quietly add support to all of our components while we finalize the theme object/API), we need to break this rule in this case.

modules/form-field/react/stories/stories_Radio.tsx
```diff
@@ -7,14 +7,23 @@ import {ControlledComponentWrapper} from '../../../../utils/storybook';
...
+import {
+  defaultCanvasTheme,
+  ContentDirection,
+  CanvasProvider
+} from '@workday/canvas-kit-labs-react-core';
 
 const hintText = 'Helpful text goes here.';
 const hintId = 'error-desc-id';
 
+const rtlCanvasTheme = { ...defaultCanvasTheme};
```
anicholls: This works, but does not follow the docs. Based on these, you would use:
```tsx
const theme = createCanvasTheme({ direction: ContentDirection.RTL })
```

modules/_labs/core/react/lib/theming/applyDirection.ts
stephanerangaya: @anicholls Do we still need this if we're wrapping components in another CanvasProvider?

modules/_labs/core/react/lib/theming/styled.ts
```diff
@@ -0,0 +1,36 @@
...
+import {CanvasTheme, ContentDirection} from './types';
+import {useTheme} from './theme';
+import rtlCSSJS from 'rtl-css-js';
+
+const noop = (styles: any) => styles;
+
+// Pulled from https://github.com/emotion-js/emotion/blob/master/packages/styled-base/src/utils.js#L6 (not exported)
+type Interpolations = Array<any>;
+
+function styled<Props>(node: any) {
```
anicholls: Note: I'm moving this over to #272. You'll have to rebase once that's there and make any changes necessary.

modules/_labs/core/react/lib/theming/styled.ts
```diff
@@ -0,0 +1,36 @@
...
+      interpolation => (props: Props & {theme: CanvasTheme; direction: ContentDirection}) => {
+        const theme = useTheme(props.theme);
+        const direction = props.direction || theme.direction;
+
+        // TODO: Add support for RTL prop.
+        const maybeFlip = direction === ContentDirection.RTL ? rtlCSSJS : noop;
+
+        try {
+          if (typeof interpolation === 'function') {
+            return maybeFlip(interpolation(props) as CSSObject);
```
anicholls: We need to change `props.theme` to `theme` before this is called.

lychyi:
@stephanerangaya Same thing here as #337, it may be better if the base branch was #272 so that the changes on top would be a bit better to read

anicholls:
@lychyi He couldn't do that without closing this PR and opening one against my fork. I pushed a duplicate branch to the workday/canvas-kit remote temporarily for the sake of these two PRs. I will keep this up to date with anicholls/canvas-kit

@stephanerangaya There have been some changes to theming since this so you'll have to rebase onto upstream/theming

modules/_labs/core/react/index.ts
```diff
@@ -16,8 +17,8 @@ export {
   CanvasProvider,
   createCanvasTheme,
   styled,
-  useTheme,
```
anicholls: Was this just an issue with the rebase? I think we still need to export this.

modules/_labs/core/react/lib/theming/applyDirection.ts
anicholls: @stephanerangaya If we want to use `CanvasProvider` to locally override the direction, then we need to change our logic for setting the body `dir` attribute. Right now, the last rendered `CanvasProvider` will take precedence, so trying to override locally would set the attribute on the entire page.

anicholls: After discussing it with @stephanerangaya, we decided to just use a nested `CanvasProvider` to change the direction of a nested component. Since we recently made the change to only set the `dir` attribute for the children of a `CanvasProvider`, we no longer need to worry about overriding the body `dir` (we're going to have consumers control this themselves). 

This may be a bit more perf overhead than using a `direction` prop, but it's not a very common occurrence. Also, it avoids the high likelihood of human error when passing `direction` to all styled components, and alternatively, avoids using an unconventional render wrapper like the `applyDirection` function I wrote. Since we added a no-op for `InputProvider` in #272, we shouldn't need to worry about nested providers.

Lastly, this method enforces the fact that direction is just part of the theme object, so the API for setting a local theme is the same as setting a local content direction.

Example:
```tsx
<CanvasProvider theme={createCanvasTheme({direction: ContentDirection.RTL})}>

  // A bunch of components

  <CanvasProvider theme={createCanvasTheme({direction: ContentDirection.LTR})}>
    <Slider /> // Sliders should always be LTR
  </CanvasProvider>

  // Some more components

</CanvasProvider>
```

modules/_labs/core/react/lib/theming/README.md
```diff
@@ -142,16 +145,31 @@ const theme: PartialCanvasTheme = {
...
+  createCanvasTheme,
+  ContentDirection,
+} from '@workday/canvas-kit-labs-react-core';
+
+<CanvasProvider theme={createCanvasTheme({direction: ContentDirection.RTL})}>
+  {/* Your app with Canvas components */}
+</CanvasProvider>;
+```
+
 ### Nesting Theme Providers
```
anicholls: Can you provide an example here for reversing the top level content direction?

modules/_labs/core/react/lib/theming/README.md
```diff
@@ -142,16 +145,31 @@ const theme: PartialCanvasTheme = {
 </CanvasProvider>;
 ```
 
+Example changing the direction:
```
anicholls: Can you convert this to an H3 so we have something to link to? i.e.

```
### Bidirectionality/RTL

The theming API is how you can change the content direction of your canvas components. You simply need to change the `direction` field of the theme object. `CanvasProvider` wraps your components with a `div` that has the `dir` attribute set to the value of the theme direction. Styled components using the [Canvas `styled` function](https://github.com/Workday/canvas-kit/blob/master/modules/_labs/core/react/lib/theming/styled.ts) will have their styles automatically flipped if dictated by the closest theme object.

Example:
...
```

modules/_labs/core/react/lib/theming/styled.ts
```diff
@@ -1,20 +1,31 @@
...
 // Pulled from https://github.com/emotion-js/emotion/blob/master/packages/styled-base/src/utils.js#L6 (not exported)
 type Interpolations = Array<any>;
 
 function styled<Props>(node: any) {
   return (...args: Interpolation<Props>[]) => {
     const newArgs: Interpolations = args.map(
-      interpolation => (props: Props & {theme: CanvasTheme}) => {
+      interpolation => (props: Props & {theme: CanvasTheme; direction: ContentDirection}) => {
         props.theme = useTheme(props.theme);
+        const direction = props.direction || props.theme.direction;
```
anicholls: We can drop `props.direction` now.

modules/_labs/core/react/lib/theming/styled.ts
```diff
@@ -1,20 +1,31 @@
...
+
+        // TODO: Add support for RTL prop.
```
anicholls: Remove TODO

modules/banner/react/lib/Banner.tsx
```diff
@@ -67,7 +67,7 @@ const BannerWrapper = styled('button')<BannerProps>(
   })
 );
 
-const BannerIcon = styled(SystemIcon)({
+const BannerIcon = styled(SystemIcon)<BannerProps>({
```
anicholls: What was the reason for adding `<BannerProps>` here?

modules/_labs/core/react/lib/theming/applyDirection.ts
NicholasBoll: 👍

modules/_labs/core/react/lib/CanvasProvider.tsx
```diff
@@ -16,8 +17,10 @@ export default class CanvasProvider extends React.Component<CanvasProviderProps>
     const {children, theme} = this.props;
 
     return (
-      <ThemeProvider value={theme}>
-        <InputProvider>{children}</InputProvider>
+      <ThemeProvider theme={theme}>
+        <div dir={theme.direction}>
```
anicholls: Can we use a [`bdo` element](https://css-tricks.com/almanac/properties/d/direction/) instead of a `div`? Also, I think we should include css `direction` as a backup since it has better support.

```tsx
const DirectionContainer = styled('bdo')(({theme}) => ({direction: theme.direction}))
```

modules/_labs/core/react/lib/theming/README.md
stephanerangaya: I've done that below, in `### Nesting CanvasProvider components`

modules/banner/react/lib/Banner.tsx
stephanerangaya: It was needed before so styled had access to the direction (when we passed it manually) but not anymore. I'm removing this.

modules/_labs/core/react/stories/stories_direction.tsx
```diff
@@ -0,0 +1,56 @@
+/// <reference path="../../../../../typings.d.ts" />
+import * as React from 'react';
+import {storiesOf} from '@storybook/react';
+import withReadme from 'storybook-readme/with-readme';
+
+import {IconButton, IconButtonProps} from '../../../../button/react/index';
```
anicholls: Can you change this to `@workday/canvas-kit-react-button` and add it as a devDependency?

modules/_labs/core/react/lib/CanvasProvider.tsx
stephanerangaya: 👍

modules/_labs/core/react/lib/CanvasProvider.tsx
```diff
@@ -16,8 +24,10 @@ export default class CanvasProvider extends React.Component<CanvasProviderProps>
     const {children, theme} = this.props;
 
     return (
-      <ThemeProvider value={theme}>
-        <InputProvider>{children}</InputProvider>
+      <ThemeProvider theme={theme}>
+        <DirectionContainer theme={theme} dir={theme.direction}>
```
anicholls: `dir` will set the attribute _and_ pass the `dir` prop through. You can drop `theme={theme}` and just do 

```tsx
const DirectionContainer = styled('bdo')<{dir: ContentDirection}>(
  ({dir}) => ({
    direction: dir,
  })
); 
```

modules/_labs/core/react/lib/CanvasProvider.tsx
```diff
@@ -1,12 +1,20 @@
...
 import {CanvasTheme} from './theming/types';
-import {ThemeProvider, defaultCanvasTheme} from './theming/theme';
+import {defaultCanvasTheme} from './theming/theme';
+import styled from '@emotion/styled';
 
 export interface CanvasProviderProps {
   theme: CanvasTheme;
 }
 
+const DirectionContainer = styled('bdo')<Pick<CanvasProviderProps, 'theme'>>(
```
anicholls: Nit: We shouldn't pull from `CanvasProviderProps`. Let's use `<{dir: ContentDirection}>`.

modules/_labs/core/react/lib/theming/README.md
```diff
@@ -142,16 +138,38 @@ const theme: PartialCanvasTheme = {
 </CanvasProvider>;
 ```
 
-### Nesting Theme Providers
+### Bidirectionality
+
+The `CanvasProvider` also provide support for bidirectionality, useful for RTL languages. The direction, part of the theme, is set using `ContentDirection.LTR` or  `ContentDirection.RTL`.
```
anicholls: Nit: `provides`

anicholls: I still think calling out how this works is important here.

`CanvasProvider` wraps your components with a `bdo` element that has the `dir` attribute set to the value of the theme direction. Styled components using the [Canvas `styled` function](https://github.com/Workday/canvas-kit/blob/master/modules/_labs/core/react/lib/theming/styled.ts) will have their styles automatically flipped if dictated by the closest theme object.

modules/_labs/core/react/lib/theming/README.md
```diff
@@ -160,13 +178,15 @@ const theme: PartialCanvasTheme = {
       main: colors.greenApple400,
     },
   },
+  direction: ContentDirection.RTL
```
anicholls: Both providers are getting RTL. Should this be LTR?

modules/_labs/core/react/README.md
```diff
@@ -141,3 +141,8 @@ See the [@workday/canvas-kit-react-core docs](../../../core/react/README.md#inpu
 ## Theming
 
 Theming documentation has its own README. You can find it [here](./lib/theming/README.md)
+
+## Bidirectionality
+
+Bidirectionality is provided by Theming. You can find Theming documentation
```
anicholls: Might as well link to the bidirectionality header in the other readme now that we have one 🙂

modules/_labs/core/react/lib/theming/README.md
stephanerangaya: yes

PR #289, opened 2019-11-01
fix: Lower is-prop-valid dep into components that need it

anicholls:
## Summary

The `@emotion/is-prop-valid` dependency was mistakenly added to the workspace package.json rather than the components' that explicitly need it. This causes "Module not found" errors when using the reference outside of the Canvas Kit repo. This PR removes the workspace dep (it's not needed there), and adds it to the components that use `isValidProp`.

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/avatar/react/package.json
+++ b/modules/avatar/react/package.json
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "@emotion/core": "^10.0.21",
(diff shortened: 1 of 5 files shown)


lychyi:
@anicholls Did a quick global search. Missing in avatar still right?

PR #292, opened 2019-11-01
test: Add dependencies sync check

NicholasBoll:
## Summary

Module dependency versions are sometimes out of sync with other module dependency versions. This additional check prevents it. The output is meant to match that of `eslint` to trigger the Github Actions automatic annotation detection.

## Additional References

Screen Shot 2019-11-01 at 2 04 20 PM

Screen Shot 2019-11-04 at 9 57 05 AM
Screen Shot 2019-11-01 at 5 54 13 PM

## Suggested commit message:
test: Add dependencies sync check (#292)
* Added check for mismatched dependencies across all `package.json` files. Will throw an ESLint-style error that should be picked up as an annotation by Github Actions that will put an error annotation on the line of the error


.storybook/config.js
```diff
@@ -5,7 +5,6 @@ import addons from '@storybook/addons';
 import Events from '@storybook/core-events';
 import {toId} from '@storybook/router';
 import ReactDOM from 'react-dom';
-import 'storybook-chromatic';
```
NicholasBoll: There's a warning that says this should no longer be imported, so it is removed

modules/toast/react/lib/Toast.tsx
```diff
@@ -52,7 +52,7 @@ const Message = styled('div')({
 
 export default class Toast extends React.Component<ToastProps> {
   static defaultProps = {
-    icon: checkIcon,
+    icon: checkIcon as CanvasSystemIcon, // needed for TS2742 - https://github.com/microsoft/TypeScript/issues/29808
```
NicholasBoll: There might now be a type error with the latest version of `@workday/design-assets-types`. It was bumped from `0.2.4` to `0.2.6`. There is an open issue with no resolution for when it happens. It is not fully understood why.

PR #293, opened 2019-11-01
DNM - feat: Add check for missing or unused dependencies

anicholls:
## Summary

We occasionally don't catch missing dependencies in our package.json files. Due to how the mono repo works we do not notice this locally and we end up publishing packages that come with a `Module not found` error. This adds a utility script to check for missing and unused dependencies as well as a Github Action CI check that uses this script.

Uses `eslint-stylish` format so it gets picked up as annotations in Github Actions

## Checklist

- [ ] `yarn test` passes
      applicable

## Additional Resources

Screen Shot 2019-11-04 at 2 08 55 PM

Screen Shot 2019-11-04 at 2 07 55 PM


.github/workflows/ci.yml
```diff
@@ -40,3 +40,18 @@ jobs:
         TRAVIS: true # Pretend to be Travis. More info: https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/util/ci_provider.js
         TRAVIS_PULL_REQUEST_BRANCH: ${{ github.head_ref }} # Tell Cypress what the PR branch name is
         CYPRESS_RECORD_KEY: 3a9347b6-36ab-4a36-823d-709f4078b148
+
+	dependency_test:
```
NicholasBoll: @anicholls I'm not sure how this happened, but these are tab characters (`\t`) instead of space characters (`\s`). I'm not sure why this isn't failing a yaml validation on Github's part, but the significant whitespace issue is preventing this workflow file from running. Github is conveniently highlighting in red.

utils/check-dependencies-exist.js
```diff
@@ -0,0 +1,47 @@
...
+  for (let file of files) {
+    checks.push(
+      new Promise((resolve, reject) => {
+        /* Note: dependency-check can accept multiple files, but due to how it caches imported modules,
+         * it needs to be run separately for each module. */
+        exec(`yarn run dependency-check ${file} --no-dev`, (err, stdout, stderr) => {
+          if (err) {
+            pass = false;
+            const match = file.match(moduleRegex);
+            console.log(match);
```
NicholasBoll: If you want to have an annotation added to PRs (to prevent trying to understand the CLI output message), you could match the ESLint "stylish" output (not all the double-spaces):

```
${fullFilePath}
  ${line}:${column}  error  ${message}  ${type}
```

An example might look like this:
```
/home/runner/work/canvas-kit/canvas-kit/modules/button/react/package.json
  1:1  error  Modules in package.json not used in code: @workday/canvas-kit-react-common  check-dependencies-exist
```

This isn't required, but the annotations are nicer. The error messages appear inline on the `files` tab of a PR rather than digging through the CI logs.

anicholls:
There are lots of checks failing so it's a good thing we added this! There was some confusion around a few of these so I would like to clarify:

- `@emotion/core` isn’t needed in most modules because we’re only using styled components (none of `jsx, css, Global, ClassNames`)
- `react` isn’t used at all in labs core (we’re only using plain js)
- All types dependencies need to move to dev deps or be hoisted.

The only thing that’s weird is how babel is converting `@emotion/styled` imports into `@emotion/styled-base` in some cases. Still need to figure out how to address this. Maybe we just depend on `@emotion/styled-base` in these cases.

anicholls:
The latest commit avoids the need for compiled JS and addresses the second concern (missing imports specifically for type usage) by checking `.ts` and `.tsx` files. However, I ran into a blocker because these file types are not specified automatically within `dependency-checker` and `precinct`. I have it working locally, but it requires changes to `dependency-checker`. I opened an issue here: https://github.com/dependency-check-team/dependency-check/issues/111. We'll see what happens. 

This is blocked for now.

anicholls:
Closing until we're back on Github Actions

PR #296, opened 2019-11-04
build: Add Mdx support to webpack config

sahlhoff:
## Summary

This adds an `mdx` rule to our webpack config. 

I've also included the welcome section for storybook, this includes: `Api Pattern Guidelines`, `Changelog`, `Component Status`, `Contributing`, and `Getting Started`. 

I'm currently working out the emoji rendering issue. I'm also not 100% if we want to include the welcome section in this PR but it was helpful for me to test against so I kept it in. LMK if I should remove it. Including it now has `Other` as the title for components, if we do keep the welcome section we should change the stories base to `Components`.

I've also added an example `Card` using the new mdx format. You'll see we now can directly access story blocks like `Preview`, `Story`, `Props`. In this example we wouldn't need each preview to also include a canvas tab (this also removes it from the sidenav).

## Checklist

- [x] `yarn test` passes
      applicable


NicholasBoll:
Very interesting. I guess the question is: Do we like the MDX format or Storybook in general for documentation? Storybook still feels weird as a documentation front-end. I think it is great for stories - it is just weird for a primary form of documentation. It just isn't flexible enough.

I do like the ` modules/card/react/stories/stories.mdx` file where the whole documentation page is defined there. It is kind of silly that the non-story doesn't have a preview.

modules/card/react/stories/stories.mdx
```diff
@@ -0,0 +1,27 @@
...
+      <Card heading="Title">Card contents</Card>
+    </div>
+  </Story>
+</Preview>
+
+<Props of={Card} />
+
+## No Padding
+
+<Preview>
```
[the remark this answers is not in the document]
sahlhoff: Yeah, this was an example...you'll see that without the story tab these will be excluded from the navigation.

modules/card/react/stories/stories.mdx
```diff
@@ -0,0 +1,27 @@
...
+import README from '../README.md';
+import Card from '../index';
+
+<Meta title="Card" component={Card} />
+
+# Card
+
+<Preview>
+  <Story name="Default">
+    <div className="story">
```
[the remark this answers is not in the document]
sahlhoff: Makes sense the less overhead we have in the examples probably the better.

.storybook/stories/stories_api_pattern_guidelines.mdx
```diff
@@ -0,0 +1,7 @@
+import {Meta} from '@storybook/addon-docs/blocks';
+import Markdown from 'markdown-to-jsx';
+import ApiPatternGuidelines from '../../API_PATTERN_GUIDELINES.md';
+
+<Meta title="Welcome|API Pattern Guidelines" />
+
+<Markdown>{ApiPatternGuidelines}</Markdown>
```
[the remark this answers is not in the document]
sahlhoff: Yeah it would be...I looked into it for a bit did not see any options for it: https://github.com/probablyup/markdown-to-jsx#parsing-options. They do have a few `GFM` references in their docs so _not sure_ if this is using `GFM` internally? I can look into it more. 

@anicholls An alternative could be to convert these markdown files to `MDX`. Importing them would be more stream lined then. LMK if you would be ok with this. I haven't actually tested how github renders `MDX` so we could potentially run into rendering issues when you view these files on `github.com`

sahlhoff:
@anicholls @NicholasBoll `MDX` support would be really helpful just from the docs perspective. If everything "worked" performance wise I think it would be without a question to merge it in. I guess the question is how much of a performance trade are we prepared to take?

.storybook/stories/stories_api_pattern_guidelines.mdx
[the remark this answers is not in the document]
NicholasBoll: @chad Here's a preview: https://github.com/Workday/canvas-kit/blob/c71f0b5ddb648f237718f8e3243d23f2313dcc1b/modules/card/react/stories/stories.mdx

modules/card/react/stories/stories.mdx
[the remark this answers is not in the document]
NicholasBoll: 👍 for getting rid of `.story` divs. If we really want, you can add it as a decorator to all stories so it doesn't show up in source, but @anicholls I've noticed odd indentation with all Storybook source panels. It seems like a bug on their end.

modules/card/react/stories/stories.mdx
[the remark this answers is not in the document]
NicholasBoll: I've been thinking about this. @sahlhoff and I talked about the possibility of a story without showing up in the documentation for ChromaticQA or Cypress. I think if we stick with Storybook as the documentation frontend, we can just call these out as sections. Our Accessibility team runs audits on our stories. I think it would be good to call out sections. Like this section shows all the valid states of a component for manual and automated verification and this section shows a contextual example with approved accessibility features. Not always can we control accessibility, but our examples should show an implementation that matches a11y specs so application developers have a working example they can follow.

NicholasBoll:
> @anicholls @NicholasBoll `MDX` support would be really helpful just from the docs perspective. If everything "worked" performance wise I think it would be without a question to merge it in. I guess the question is how much of a performance trade are we prepared to take?

I'm willing to take a performance for a better DX. We may be able to mitigate build times.

utils/storybook/MarkdownWrapper.tsx
```diff
@@ -0,0 +1,18 @@
...
+import EmojiConvertor from 'emoji-js';
+import Markdown from 'markdown-to-jsx';
+
+const emoji = new EmojiConvertor();
+
+interface MarkdownWrapperProps {
+  content: string;
+}
+
+export default class MarkdownWrapper extends React.Component<MarkdownWrapperProps> {
```
anicholls: Can we just call this `Markdown`? The end "user" doesn't really need to know it's a wrapper.

utils/storybook/MarkdownWrapper.tsx
```diff
@@ -0,0 +1,18 @@
...
+  render() {
+    const ConvertedContent = emoji.replace_colons(this.props.content);
+
+    return <Markdown class="markdown-body">{ConvertedContent}</Markdown>;
```
anicholls: `className`

utils/storybook/MarkdownWrapper.tsx
sahlhoff: Updated 👍

utils/storybook/MarkdownWrapper.tsx
sahlhoff: Updated 👍

modules/card/react/stories/stories.mdx
```diff
@@ -0,0 +1,27 @@
...
+import {spacing} from '../../../core/react/index';
+import README from '../README.md';
+import Card from '../index';
+
+<Meta title="Card" component={Card} />
+
+# Card
+
+<Preview>
+  <Story name="Default">
```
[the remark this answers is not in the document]
NicholasBoll: @sahlhoff and I were talking about this. MDX seems to break the current belief that the order stories are loaded effects the order in the navigation - even though that advice comes from the Storybook core team. It could change later

sahlhoff:
@anicholls @NicholasBoll should be good to go after CI checks 🤞 – let me know if there is anything else.

utils/storybook/Markdown.tsx
```diff
@@ -0,0 +1,18 @@
...
+import EmojiConvertor from 'emoji-js';
+import MarkdownToJSX from 'markdown-to-jsx';
+
+const emoji = new EmojiConvertor();
+
+interface MarkdownProps {
+  content: string;
+}
+
+export default class Markdown extends React.Component<MarkdownProps> {
```
anicholls: No need to refactor this, but in the future, let's use function components for this - it's just way simpler:
```jsx
export default function(props: { content: string }) {
  const ConvertedContent = emoji.replace_colons(props.content);
  return <MarkdownToJSX className="markdown-body">{ConvertedContent}</MarkdownToJSX>;
}
```

sahlhoff: Good to know. I was wondering if there was a better way to write this 😅

NicholasBoll: Agreed

NicholasBoll: Although probably better to name the function, otherwise dev tools show this component as anonymous

NicholasBoll:
This is a documentation-only change. Not sure if it should invoke a minor or patch. Purely speaking, it isn't a feature that a downstream consumer of the published artifacts can take advantage of.

PR #300, opened 2019-11-06
docs: Update Storybook IA

sahlhoff:
## Summary

Re-organize Storybook to make use of sections and more categorization so that it matches our internal IA.

## Checklist

- [x] `yarn test` passes
- [ ] prep organizational changes on design site


modules/form-field/react/stories/stories.tsx
```diff
@@ -6,7 +6,7 @@ import withReadme from 'storybook-readme/with-readme';
 import {Label, Hint} from '../index';
 import README from '../README.md';
 
-storiesOf('Form Field/Label', module)
+storiesOf('Components|Inputs/Label', module)
```
anicholls: It struck me as odd that this is just under `Label`, when `Hint` is under `Form Field`

modules/common/react/stories/stories.tsx
```diff
@@ -49,7 +49,7 @@ class PopperPopup extends React.Component<{}, PopperPopupState> {
   };
 }
 
-storiesOf('Common', module)
+storiesOf('Tokens|Common', module)
```
anicholls: I feel like this should go under "Other" rather than "Tokens"

sahlhoff:
@anicholls awesome! Yeah those labels really add cohesion. 

This was a first take and was really hoping to get a discussion going of how we should handle some of these items 😄 

For `CSS` I'm leaning towards either:

 1) Embedding them along side the components level– so `Components|Containers/Card` would have both `css` and `react` stories.

2) Create a new `CSS` section so it is a sibling with `Components`, `Tokens`, `Labs`.

We could create a new storybook instance for CSS but I would be afraid that would just turn into a second class citizen and potentially confuse developers which framework they are looking at.

For `Common` stories:

1) Right now popper is closely related to `Popups` so we could really add it into that section under something like `Utils` or leave it as `Common`.

2) Move it to `Others` ... really don't like that name for a section though. We could come up with a better name like `Utils` or leave it as `Common`.

Let me know what you think on these two items or any other thoughts you have!

modules/form-field/react/stories/stories.tsx
sahlhoff: Hmm yeah I'll move this into `Form Field`

anicholls:
@sahlhoff Let's just make CSS it's own section for now. I think we need to consider how to differentiate them before combining them or it will be confusing.

Re. common, the `Popper` story could fit within popups, but you'd have to load the common readme.

sahlhoff:
@anicholls should be good for another look when you get a chance 👍

modules/avatar/react/stories/stories.tsx
```diff
@@ -13,7 +13,7 @@ const handleAvatarButtonClick = (e: React.SyntheticEvent) => {
   console.log('AvatarButton clicked');
 };
 
-storiesOf('Avatar/Default', module)
+storiesOf('Tokens|Avatar/Default', module)
```
anicholls: Are we sure this should be in tokens? I would call this a component.

modules/form-field/react/stories/stories_Radio.tsx
```diff
@@ -11,7 +11,7 @@ import README from '../../../radio/react/README.md';
 const hintText = 'Helpful text goes here.';
 const hintId = 'error-desc-id';
 
-storiesOf('Form Field/Radio/Top Label/Radio Group', module)
+storiesOf('Components|Inputs/Radio/Top Label/Radio Group', module)
```
anicholls: Do you think anyone will have a hard time finding these stories? They're no longer categorized under form-field in storybook and the readme in storybook shows the checkbox module. I think the organization is in the right place, but it's a concern. Not sure of the best way to solve this 🤔

modules/avatar/react/stories/stories.tsx
sahlhoff: @anicholls Ok, I'm ok with moving these into `Components` I think especially buttons would make sense...We may likely revisit this down the road but I think it's ok for now.

modules/form-field/react/stories/stories_Radio.tsx
sahlhoff: I think it's ok for now. The most important thing is alignment with design side of the house...If we get feedback on category name/find-ability we can always update.

modules/avatar/react/stories/stories.tsx
sahlhoff: Okay, going to ping and see if he has any thoughts...Not really sure where we would want to put it in the categories that we have currently.

carlosyllobre: Hey everyone, this is an interesting one, if we see, both of them, 'Avatar' and 'Avatar Button' are basically the same from a design pov, the specs for both will be the same since there are no states applied to this button. The only difference is that one is treated as a button since a link is added while the other one's primary function is just to display the avatar. If it makes sense for you we can have 'Avatar Button' under 'Buttons'. Just need to update the taxonomy.

modules/form-field/react/stories/stories_Radio.tsx
carlosyllobre: A valid concern @anicholls, this is something that always is going to happen until certain degree when introducing a new ordering but as @sahlhoff said with the new change we are getting alignment and from now one will makes sense, since we have a clear structure and naming convention in place. I don't think this is something we will be changing very often anyways 😊

modules/avatar/react/stories/stories.tsx
anicholls: I still think `Avatar` and `AvatarButton` should be grouped together, I just think "Avatar" could be its own section under components.

sahlhoff:
Note: when we merge this in we will need to merge in the parallel PR on `design-site-content`. Thanks guys! 👋

anicholls:
Would like some feedback on combining React + CSS components.

**NOTE: We need to prep this change for our design site before this is merged**

cypress/integration/Button.spec.ts
```diff
@@ -7,7 +7,7 @@ describe('Button', () => {
 
   context('given primary buttons are rendered', () => {
     beforeEach(() => {
-      h.stories.load('button', 'primary');
+      h.stories.load('components-buttons-button-react', 'primary');
```
NicholasBoll: These should have been capitalized before. The intent was to match the exact string located in the `storiesOf` and `add` functions to make it easier without having to translate manually the URL.

I'll adjust and see what we think.

NicholasBoll: Now this line reads:
```ts
h.stories.load('Components|Buttons/Button/React', 'Primary');
```

NicholasBoll:
As far as the React vs CSS, the current style of `Component -> React/CSS -> story` does help see parity of the 2 kits. The downside is it does expose the user to choose a kit every time when navigating.

PR #282, opened 2019-10-28
feat(pagination): Add pagination component

lychyi:
## Summary

WIP do not address yet.

## Checklist

- [ ] branch has been rebased on the latest master commit
- [ ] tests are changed or added
- [ ] `yarn test` passes
- [ ] all (dev)dependencies that the module needs is added to its `package.json`
- [ ] code has been documented and, if applicable, usage described in README.md
- [ ] module has been added to `canvas-kit-react` and/or `canvas-kit-css` universal modules, if
      applicable
- [ ] design approved final implementation
- [ ] a11y approved final implementation
- [ ] code adheres to the [API & Pattern guidelines](../API_PATTERN_GUIDELINES.md)

PR #301, opened 2019-11-07
feat(pagination): add pagination bar component

nikolasjchaconas:
## Summary

This PR includes a new pagination bar component. This can be used to manage paginated data in tables. The component includes a `onPageChange` prop/dispatch so that users can be notified of page changes. It also includes an optional GoTo box for navigating quickly between pages, and an optional customizable label below the pagination bar.

## Checklist

- [X] branch has been rebased on the latest master commit
- [X] tests are changed or added
- [X] `yarn test` passes
- [X] all (dev)dependencies that the module needs is added to its `package.json`
- [X] code has been documented and, if applicable, usage described in README.md
- [X] module has been added to `canvas-kit-react` and/or `canvas-kit-css` universal modules, if
      applicable
- [X] design approved final implementation
- [X] code adheres to the [API & Pattern guidelines](../API_PATTERN_GUIDELINES.md)

## Usage

```tsx
import Pagination from '@workday/canvas-kit-react-pagination';

const [currentPage, setCurrentPage] = React.useState(1);

return (
    <Pagination
      total={100}
      pageSize={10}
      currentPage={currentPage}
      onPageChange={p => setCurrentPage(p)}
      showLabel
      showGoTo
      customLabel={(from: number, to: number, items: number, item: string) =>
          `${from.toLocaleString()} to ${to.toLocaleString()} of ${items.toLocaleString()} ${
            items > 1 ? 'companies' : 'company'
          }`
      }
    />
);
```

#### total: number

> The total number of items

#### pageSize: number

> The number of items to display per page

#### currentPage: number

> The current page being displayed

#### onPageChange: (page: number) => void

> Dispatch which is invoked when the page is changed

#### showLabel?: boolean

> Shows a label below the pagination bar describing the items currently being viewed

#### showGoTo?: boolean

> Shows a box adjacent to the pagination bar where a page can be entered and is submitted when
> 'Enter' key is pressed

#### customLabel?: (from: number, to: number, items: number, itemLabel: string) => string

> A function to build a custom label below the pagination bar.

### Screenshots

image
image
image
image
image


modules/button/react/lib/ButtonStyles.ts
```diff
@@ -311,7 +312,8 @@ export const iconButtonStyles: ButtonGenericStyle = {
   variants: {
     sizes: {
       [ButtonSize.Small]: {
-        width: canvas.spacing.l,
+        minWidth: canvas.spacing.l,
+        width: 'auto',
```
nikolasjchaconas: The small buttons are normally of fixed size. This gives them a minimum width and allows their width to grow based on the content. In my case, this was very helpful for numbers like `10,000`

NicholasBoll:
This looks similar to the work done in #282. @nikolasjchaconas would you mind collaborating with @lychyi on this change? I know @lychyi is collaborating with the Canvas UX team.

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,71 @@
...
+    if (textAsInteger < 1) {
+      return 0;
+    }
+    if (textAsInteger > max) {
+      return 0;
+    }
+    return textAsInteger;
+  };
+  return (
+    <>
```
drschulz: We can get rid of the fragment here

lychyi:
> This looks similar to the work done in #282. @nikolasjchaconas would you mind collaborating with @lychyi on this change? I know @lychyi is collaborating with the Canvas UX team.

This was something that @nikolasjchaconas talked about behind-the-scenes, let's carry on with the review of this PR. I've closed #282 to avoid any confusion.

We had also talked about reviewing this component's API first as a "first pass through" so that reviewers don't have to get too caught up in the code. #282 didn't have a design review of its API either so now is a great time to do it before we spend time churning on the finer details.

For next steps, I recommend that we ensure the component API, its props, and prescribed usage are spot on before diving any deeper into the codebase. @nikolasjchaconas Can you update the PR with the proposed props and an example usage so we can all agree on the API? @NicholasBoll and @anicholls when you don't have any further comments or additions about the API, let's drop this PR out of draft mode and then pick up where we left off.

nikolasjchaconas:
@lychyi good point, I went ahead and updated my comment to include example usage and props 👍

modules/text-input/react/lib/TextInput.tsx
```diff
@@ -46,6 +44,13 @@ const Input = styled('input')<Pick<TextInputProps, 'error' | 'grow'>>(
   ({error}) => ({
     ...errorRing(error),
   }),
+  ({width}) => ({
+    minWidth: width || 280,
+    width: width || 280,
+  }),
+  ({height}) => ({
+    height: height || 40,
+  }),
```
nikolasjchaconas: I wanted to be able to customize the width/height of my text input box, so I added an optional height/width to Input

lychyi: Not sure if this will be approved by the design system team, this isn't something we're customizing at all since it the height of the inputs are what help the forms align together.

lychyi:
Some feedback on the API:

`dataLabel?: string` may benefit from being a formatter instead of a string: https://github.com/Workday/canvas-kit/pull/282/files#diff-70d947a4ff76cce616efca6cba576125R100-R113

That way the label string can be anything. It may be worth reaching out to the DS team to see if the language of this label has to be fixed. Either way, it has to be flexible because of localization so we may as well make it completely customizable.

modules/text-input/react/lib/TextInput.tsx
nikolasjchaconas: how should I go about making the `goTo` input match the height/width of the spec?

nikolasjchaconas:
> That way the label string can be anything. It may be worth reaching out to the DS team to see if the language of this label has to be fixed. Either way, it has to be flexible because of localization so we may as well make it completely customizable.

Makes sense, I'll go ahead and make this change

lychyi:
@nikolasjchaconas We had some build issues last week so we're trying to get through that before we come back to review certain PRs. Just letting you know we're still alive! Thanks for being patient.

NicholasBoll:
I miss having Github Actions - failures were annotations on exact lines that fail. Currently CI is failing on:

```
 FAIL  modules/_labs/pagination/react/spec/Pagination.spec.tsx
  ● Pagination Component › Test pagination label can have a custom data item
    expect(element).toHaveTextContent()
    Expected element to have text content:
      11–20 of 100 users
    Received:
      11–20 of 100 items
      35 |     );
      36 | 
    > 37 |     expect(getByTestId('paginationLabel')).toHaveTextContent('11–20 of 100 users');
         |                                            ^
      38 |   });
      39 | 
      40 |   it(`Left arrow should be disabled`, () => {
      at Object.toHaveTextContent (modules/_labs/pagination/react/spec/Pagination.spec.tsx:37:44)
```

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,70 @@
...
+      return 0;
+    }
+    return textAsInteger;
+  };
+  return (
+    <GoToWrapper>
+      <GoToLabel>Go To</GoToLabel>
+      <InputWrapper>
+        <TextInput
+          data-testid="goToPage"
```
NicholasBoll: `data-testid`? I'm assuming for some automation outside this repository? We haven't decided to add test ids directly to components (they can be added through `elemProps` to parent containers and further selected from there.

How do others view generic test ids for certain components?

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,70 @@
...
+          width={53}
+          height={32}
+          min={1}
+          max={max}
+          value={value}
+          type="number"
+          aria-label="Go to"
```
NicholasBoll: I think the `GoToLabel` should be a `label` element with a `for`/`htmlFor` pointing to this `input`'s `id`. An ID can be computed or passed. We have examples where it can be passed in or defaults to a `uuid`. This has the added benefit of clicking the label text focuses the input. This isn't required for the PR to be merged, but needs to be done before this component moves out of labs. It is also a good practice. Connecting the input to the label will remove the need for the `aria-label` attribute.

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,70 @@
...
+      return 0;
+    }
+    if (textAsInteger > max) {
+      return 0;
+    }
+    return textAsInteger;
+  };
+  return (
+    <GoToWrapper>
+      <GoToLabel>Go To</GoToLabel>
```
NicholasBoll: For internationalization, this string should be provided rather than hard-coded. It is fine if "Go To" is the default.

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,70 @@
...
+          min={1}
+          max={max}
+          value={value}
+          type="number"
+          aria-label="Go to"
+          onChange={e => {
+            setValue(e.target.value);
+          }}
+          onKeyDown={e => {
+            if (e.keyCode === 13) {
```
NicholasBoll: Every time I see a `onKeyDown` with a keyCode detection of `13` I think we should use a `form` somewhere instead. Not a requirement of a labs component, but forms have more accessibility more baked in than key detection.

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,70 @@
...
+  paddingLeft: '1px',
+});
+
+const InputWrapper = styled('div')({
+  minWidth: '10px',
+  maxWidth: '45px',
+});
+
+const GoTo: React.FC<{
+  onSubmit: (page: number, e: React.SyntheticEvent) => void;
```
NicholasBoll: I think the property interface should be an interface like the other components.

Also, I thought it was odd to have the event as the second parameter. Other components that have handlers list the event as the first parameter. Also we prefer `event` to `e`. This is part of the external API of the `GoTo` component.

modules/_labs/pagination/react/lib/Pages.tsx
```diff
@@ -0,0 +1,138 @@
...
+  alignItems: 'center',
+  justifyContent: 'center',
+  fontSize: 13,
+  fontWeight: 700,
+  height: 32,
+  width: 32,
+  backgroundColor: canvas.colors.blueberry400,
+  color: canvas.colors.frenchVanilla100,
+  borderRadius: borderRadius.m,
+  ':focus': {outline: 'none'},
```
NicholasBoll: Why is focus being removed from the active page?

modules/_labs/pagination/react/lib/Pages.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+const Pages: React.FC<{
+  numPages: number;
+  currentPage: number;
+  clickHandler: (page: number) => void;
+}> = props => {
+  const {numPages, currentPage, clickHandler} = props;
+  const [width, setWidth] = React.useState(window.innerWidth);
+
+  React.useEffect(() => {
```
NicholasBoll: Perhaps this could be extracted out to its own hook? Mobile detection?

modules/_labs/pagination/react/lib/Pages.tsx
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  const end = Math.min(start + pagesToDisplay, numPages + 1);
+  const pages = _.range(start, end);
+
+  const PaginationButton: React.FC<{
+    page: number;
+  }> = props => (
+    <IconButton
+      data-testid={`paginationButton${props.page}`}
```
NicholasBoll: Should we have these data-testids hard-coded in the components? The other idea is automation helpers since DOM is usually an implementation detail of components.

modules/_labs/pagination/react/lib/Pages.tsx
```diff
@@ -0,0 +1,138 @@
...
+    >
+      ...
+    </IconButton>
+  </NoPointerEvents>
+);
+
+const Pages: React.FC<{
+  numPages: number;
+  currentPage: number;
+  clickHandler: (page: number) => void;
```
NicholasBoll: Other components have an external API of `on*` vs `*Handler` Also props should be an interface.

modules/_labs/pagination/react/lib/Pages.tsx
```diff
@@ -0,0 +1,138 @@
...
+              </>
+            )}
+
+            {active ? (
+              <ActivePage
+                data-testid="paginationIconButtonActive"
+                key={page}
+                aria-current={true}
+                aria-label={`Page ${page}, Selected`}
+                autoFocus
```
NicholasBoll: I noticed some strange behavior of this component and this is the cause. When I click on a page, it comes into focus (even though the focus styling is removed). But when I click on the "Next" or "Back" buttons, they visibly and actually lose focus because this component keeps taking it again.

All of the branching logic that's happening here causes the React reconciler to destroy and recreate DOM since we're using style components instead of `div` components with a `css` property that changes depending on the type. This reconciler action normally isn't too bad (other than increased GC and overall less performant), but it does get rid of DOM state like focus and selection.

This should be fixed before this component leaves Labs

modules/_labs/pagination/react/lib/Pages.tsx
```diff
@@ -0,0 +1,138 @@
...
+import {IconButton} from '@workday/canvas-kit-react-button';
+import canvas, {borderRadius} from '@workday/canvas-kit-react-core';
+import _ from 'lodash';
+import React from 'react';
+
+const ActivePage = styled('button')({
+  display: 'inline-flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  fontSize: 13,
```
NicholasBoll: Is it possible to use canvas variables? I think even in labs, we should use canvas variables to prevent us from accidentally slipping from design tokens and introducing design drift

modules/_labs/pagination/react/lib/Pagination.tsx
```diff
@@ -0,0 +1,99 @@
...
+    ? customLabel(labelFrom, labelTo, labelItems, 'item')
+    : `${labelFrom.toLocaleString()}\u2013${labelTo.toLocaleString()} of ${labelItems.toLocaleString()} ${item}`;
+
+  return (
+    <>
+      <Container aria-label={`Pagination Navigation`} {...elemProps}>
+        <IconButton
+          data-testid="leftPaginationArrow"
+          disabled={currentPage - 1 <= 0}
+          aria-disabled={currentPage - 1 <= 0}
```
NicholasBoll: Nit, but `aria-disabled` is not needed if there is a `disabled`

modules/_labs/pagination/react/lib/Pagination.tsx
```diff
@@ -0,0 +1,99 @@
...
+          aria-label={'Previous Page'}
```
NicholasBoll: This should be a defaulted, but editable prop

modules/_labs/pagination/react/lib/GoTo.tsx
nikolasjchaconas: I'm using `data-testid` throughout my tests in this repo ( for example `getByTestId('paginationLabel')`). Is adding them to `elemProps` to parent containers the preferred way? or do you have another standard for selecting elements in your unit tests?

modules/_labs/pagination/react/lib/GoTo.tsx
nikolasjchaconas: definitely a good point. I'll go ahead and make it a `<form>` with an `onSubmit` handler

modules/_labs/pagination/react/lib/GoTo.tsx
nikolasjchaconas: I went ahead and made the props an interface. I actually don't really see a need for passing event so I removed it. If you'd like me to add it back, I can.

modules/_labs/pagination/react/lib/Pages.tsx
nikolasjchaconas: there was no focus state in the figma, I went ahead and added it back for accessibility purposes.

modules/_labs/pagination/react/lib/Pages.tsx
nikolasjchaconas: The reason I always give focus to the active page on render is because I was told by design that it should always hold focus. I agree it's strange, so I'm removing the autofocus logic.

Regarding your second point, I can refactor to avoid this.

modules/_labs/pagination/react/lib/Pages.tsx
nikolasjchaconas: @NicholasBoll I went ahead and remove all usages of data-testid

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <GoToLabel htmlFor={goToId}>{goToLabel || 'Go To'}</GoToLabel>
+      <InputWrapper onSubmit={formSubmit}>
+        <TextInput
+          width={53}
+          height={32}
+          min={1}
+          max={max}
+          value={value}
+          id={goToId}
+          type="number"
```
NicholasBoll: I remember being told `type="number"` was a problem. @Parker-Ledoux do you know if a number input is an okay thing here?

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,72 @@
...
+          onChange={(e: any) => {
```
NicholasBoll: We lose a bit of type checking here. The type is `React.SyntheticEvent`. If you need to be more accurate, it is `React.SyntheticEvent<HTMLInputElement>`

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,72 @@
...
+      return 0;
+    }
+    if (textAsInteger > max) {
+      return 0;
+    }
+    return textAsInteger;
+  };
+
+  const goToId = uuid();
+  const formSubmit = (e: any) => {
```
NicholasBoll:
```suggestion
  const formSubmit = (e: React.SyntheticEvent) => {
```

modules/_labs/pagination/react/lib/GoTo.tsx
NicholasBoll:
```suggestion
          onChange={(e: React.SyntheticEvent) => {
```

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <TextInput
+          width={53}
+          height={32}
+          min={1}
+          max={max}
+          value={value}
+          id={goToId}
+          type="number"
+          onChange={(e: any) => {
+            setValue(e.target.value);
```
NicholasBoll: It doesn't matter in this case, but the following is more semantically accurate (https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget):

```suggestion
            setValue(e.currentTarget.value);
```

For inputs, the value should be the same (there are no descendants to an input), but I think it is a good practice to use `currentTarget` for all events that aren't delegation events.

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,72 @@
...
+    if (textAsInteger < 1) {
+      return 0;
+    }
+    if (textAsInteger > max) {
+      return 0;
+    }
+    return textAsInteger;
+  };
+
+  const goToId = uuid();
```
NicholasBoll: I'm not sure the performance or screen reader implications by this, but every render this id will change. You can use a Ref instead:

https://codesandbox.io/s/p2ndq

```suggestion
  const goToId = React.useRef(uuid()).current;
```

modules/_labs/pagination/react/lib/useIsMobile.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const useIsMobile = (container?: React.RefObject<HTMLElement>) => {
+  const [mobile, setMobile] = useState(false);
+
+  const widthIsMobile = () =>
+    container && container.current && container.current.clientWidth
+      ? container.current.clientWidth
+      : window.innerWidth;
+
+  useEffect(() => {
```
NicholasBoll: Did you experience a rapid flash using `useEffect` vs `useLayoutEffect`? `useEffect` will happen after a frame has been flushed to the DOM and could cause flashing and more layout thrashing when mobile is true (quite possibly worst-case for lower-power mobile devices.

`useEffect` is a good default, especially given that using `useLayoutEffect` to change DOM can lead to cascading read/write DOM cycles which makes frames more expensive.

This whole hook looks like it might be a problem... First, `mobile` defaults to `false`. When the container ref changes (usually always after a render has happened), the real calculation happens after another frame has rendered. At that point a resize handler is used against the window object, but the calculation is against the container clientWidth...

So the implementation is mixed. It will default to the `window.innerWidth` if the ref hasn't been set by React yet. If it has, it will use the `clientWidth` of the container. After that, it no longer cares about the container width anymore and will re-evaluate when the `window` size is changed. So is `isMobile` determined by the width of the browser or the width of the container? Should we just pick one? @lychyi

modules/_labs/pagination/react/lib/Pages.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+import {PaginationAriaLabelsDefault} from './PaginationAriaLabels';
+
+/** @jsx jsx */
+interface PagesProps {
+  numPages: number;
+  currentPage: number;
+  onPageClick: (page: number) => void;
+  mobile: boolean;
+  ariaLabels: PaginationAriaLabelsDefault;
```
NicholasBoll: I don't like prop objects like this. @anicholls @lychyi what do you think?

Pros:
- Namespacing similar properties
- Less props overall
- More processing - often confusing processing like the key filtering used somewhere

Cons:
- Documenting is harder. Documentation in Storybook will show a named interface as the type. Interplay will show a whole object (I think a JSON object)
- More props
- More clear what sub-properties are being used where

For the Tables API, we requested a flattening of the event handlers.

modules/_labs/pagination/react/lib/Pages.tsx
```diff
@@ -0,0 +1,124 @@
...
+const noPointerEvents = css({
+  pointerEvents: 'none',
+});
+
+const activeStyling = css(noPointerEvents, {
+  color: canvas.colors.frenchVanilla100,
+});
+
+const Ellipse = ({ariaLabel}: {ariaLabel: string}) => (
+  <IconButton
```
NicholasBoll: I'm guessing `IconButton` is being used for the styling? We're disabling everything that makes it a button. Is there a better way to do this?

modules/_labs/pagination/react/lib/Pages.tsx
```diff
@@ -0,0 +1,124 @@
+import {css, jsx} from '@emotion/core';
+import {IconButton} from '@workday/canvas-kit-react-button';
+import canvas from '@workday/canvas-kit-react-core';
+import _ from 'lodash';
+import React, {useRef} from 'react';
+
+import {PaginationAriaLabelsDefault} from './PaginationAriaLabels';
+
+/** @jsx jsx */
```
NicholasBoll: This might have to be moved to the top of the file to avoid the Typescript error that's preventing Chromatic and builds from running

nikolasjchaconas: done

modules/_labs/pagination/react/lib/Pages.tsx
nikolasjchaconas: okay I replaced it a span

nikolasjchaconas:
Made some changes based on your comments @NicholasBoll. I also found out why the buttons appear to flash when I navigate through pages. in `modules/button/react/lib/ButtonStyles.ts`, there is a transition on all button styles that it making the IconButton appear to flash to the new active button. When I remove the transition, the flashing stops.

What do you recommend I do? I can try to make the transition customizable, or I can move off of using IconButtons.

nikolasjchaconas:
I went ahead and removed transitions on my IconButtons (except on hover) to remove the flickering

modules/_labs/pagination/react/lib/useIsMobile.tsx
nikolasjchaconas: I went ahead and cleaned up this hook to only look at the window inner width. The customization by passing your own container was confusing and the logic wasn't completely correct.

NicholasBoll: This is better. As it is right now, `setState` is called on _every_ resize, not just when `mobile` switches from `false` to `true` or vice versa. Better performance would be to call `setMobile` only when `mobile` changes for its previous value.

I do wonder though if we could actually handle window resizing without using en event listener. If the `Pages` component renders both wide and compressed DOM and then uses a media query to set `display: none` on the non-matching size. The trade-off is not having an additional event listener. but gaining extra VDOM comparisons.

I'm not sure we decided how to handle parent container sizes. So far we're avoiding them. There are many libraries (both React-based and non-react) and most of them don't scale well and get stuck in DOM read/write layout thrashing.

I remember doing a custom solution based on [fastdom](https://github.com/wilsonpage/fastdom) to avoid multiple forced reflows. No matter what, it is hard to scale non-native element resize events and should only be used when absolutely necessary and that should probably be controlled by the application.

I think flexbox and css grid are the best way we can make components responsive and resize tricks should be avoided.

modules/_labs/pagination/react/lib/Pages.tsx
NicholasBoll: The consensus was flattened attributes

nikolasjchaconas: latest commit flattens the aria label props

modules/_labs/pagination/react/lib/GoTo.tsx
nikolasjchaconas: @NicholasBoll was there resolution on this?

modules/button/react/lib/ButtonStyles.ts
nikolasjchaconas: @lychyi @NicholasBoll just want to confirm you are okay with this change

modules/text-input/react/lib/TextInput.tsx
nikolasjchaconas: @lychyi @NicholasBoll just want to confirm you are okay with this change

modules/_labs/pagination/react/lib/GoTo.tsx
NicholasBoll: I think the main problem they had with number was currencies and such. I think we're fine here. It is easy enough to change if necessary.

modules/button/react/lib/ButtonStyles.ts
NicholasBoll: It's okay with me

modules/_labs/pagination/react/lib/useIsMobile.tsx
NicholasBoll: It seems hooks have a reference-based check when notifying render functions, so calling `setMobile` seems to have no performance issue

NicholasBoll:
> I went ahead and removed transitions on my IconButtons (except on hover) to remove the flickering

https://github.com/Workday/canvas-kit/pull/460

modules/text-input/react/lib/TextInput.tsx
NicholasBoll: I'll have to try it out and involve Design System designers if necessary. I don't understand the context yet

modules/_labs/pagination/react/lib/Pages.tsx
NicholasBoll: We're having talks with accessibility, design and UX about how and what interactive states should be in designs. Focus rings should never be removed though.

modules/_labs/pagination/react/lib/Pages.tsx
NicholasBoll: k. I'll update one of the labels. I noticed a misspelling of "ellipsis", but I'll resolve that conflict

modules/button/react/lib/ButtonStyles.ts
lychyi: It may be a good idea to add some visual stories to ensure that they work properly in all browsers. I know `width` can be tricky across browsers.

modules/text-input/react/lib/TextInput.tsx
lychyi: Figma has been updated with the 40px height `TextInput`s. Let's revert these changes for now while the team figures out what we want to do with "compact" style inputs across the board.

I don't think it's a good idea to augment the API for `TextInput` without having a holistic version across all input components.

NicholasBoll: I agree

modules/button/react/lib/ButtonStyles.ts
NicholasBoll: Each story handles resize overrides. Can we add chromatic directives instead?

modules/_labs/pagination/react/lib/GoTo.tsx
```diff
@@ -0,0 +1,72 @@
...
+  ...type.body,
+  ...type.variant.label,
+});
+
+const GoToWrapper = styled('div')({
+  display: 'flex',
+  paddingLeft: '12px',
+});
+
+const InputWrapper = styled('form')({
```
NicholasBoll: This probably won't be part of another form, right?

commitlint.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
 };
 
 // Extra scopes supported outside the `modules` folder
-const scopes = ['labs'];
+const scopes = ['labs', 'pagination'];
```
NicholasBoll: Remove this

NicholasBoll: Possibly change the scoping script to pick up modules/packages within "_labs" instead

modules/_labs/pagination/react/lib/Pagination.tsx
```diff
@@ -0,0 +1,142 @@
...
+          <Pages
+            isMobile={isMobile}
+            total={numPages}
+            current={currentPage}
+            onPageClick={onPageChange}
+            pageButtonAriaLabel={pageButtonAriaLabel}
+          />
+          <IconButton
+            disabled={currentPage + 1 > numPages}
+            aria-disabled={currentPage + 1 > numPages}
```
NicholasBoll: Remove this

modules/_labs/pagination/react/lib/useIsMobile.tsx
```diff
@@ -0,0 +1,17 @@
+import {useEffect, useState} from 'react';
+
+const useIsMobile = (threshold: number) => {
+  const [isMobile, setMobile] = useState(window.innerWidth < 500);
```
NicholasBoll:
```suggestion
  const [isMobile, setMobile] = useState(window.innerWidth < threshold);
```

modules/_labs/pagination/react/lib/GoTo.tsx
lychyi: I say we should just ship with this as a form in Labs and see what folks say about it (since this can still be used without a "Go To" form)

modules/_labs/pagination/react/lib/Pages.tsx
```diff
@@ -0,0 +1,98 @@
...
+import {css, jsx} from '@emotion/core';
+import type from '@workday/canvas-kit-labs-react-core';
+import {IconButton} from '@workday/canvas-kit-react-button';
+import canvas from '@workday/canvas-kit-react-core';
+import _ from 'lodash';
```
lychyi:
```suggestion
import range from 'lodash/range';
```

This necessitates a change to the usage of lodash in this file as well.

modules/_labs/pagination/react/lib/useIsMobile.tsx
```diff
@@ -0,0 +1,17 @@
+import {useEffect, useState} from 'react';
+
+const useIsMobile = (threshold: number) => {
```
lychyi: Should we remove this hook? I feel like this is a very odd hook given that it is detecting whether or not something is below a certain threshold. This can probably be accomplished by using something more generic like: 

```tsx
import {useWindowSize} from 'react-use';

// In Pagination.tsx, instead of:
// const isMobile = useIsMobile(500);
const isMobile = useWindowSize().width < 500;
```

or I'm thinking we can use the matchMedia API (compatibility seems good):

```tsx
import {useMedia} from 'react-use';

// In Pagination.tsx, instead of:
// const isMobile = useIsMobile(500);
const isMobile = useMedia('(max-width: 500px)');
```

I would like to check with the design team as well to see if 500 absolutely needs to be hardcoded, I think it makes this component less useful when it is.

cypress/integration/Pagination.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+    beforeEach(() => {
+      cy.viewport(414, 736); // iPhone 6/7/8 Plus
+      h.stories.load('Labs/Pagination', 'Default');
+    });
+
+    it('should show 2 page buttons', () => {
+      cy.findAllByLabelText(/Page \d+$/).should('have.length', 2);
+    });
+
+    context('when the last page is clicked', () => {
```
lychyi: I don't think we've tested `should disable next page button` yet. This spot may be a good place to add it.

modules/_labs/pagination/react/stories/stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+      <div>
+        <br />
+        <WithGoToExample />
+      </div>
+    </div>
+  ))
+  .addParameters({
+    chromatic: {
+      disable: false,
+      viewports: [500, 1000],
```
lychyi: I believe this viewport needs to be 499 for it to trigger the mobile layout

modules/_labs/pagination/react/lib/useIsMobile.tsx
NicholasBoll: I agree 500 is an odd number. The Modal specs call out `768` as the threshold for "mobile" - basically tablet or phone.

I don't know if there is official guidelines on how we detect "mobile"

modules/_labs/pagination/react/stories/stories.tsx
NicholasBoll: I opted to make `width` an input. We don't need these anymore. I just made a state table of Desktop and Mobile Pagination with a few edge cases I wanted to cover visuall.

modules/_labs/pagination/react/lib/useIsMobile.tsx
NicholasBoll: I opted to make Pagination take an optional width property. It makes more sense for Pagination to respond to a width than some random 'isMobile' input. The application can always pass in this width based on window innerWidth. This makes testing easier and makes the component more flexible. @nikolasjchaconas what do you think?

modules/text-input/react/lib/TextInput.tsx
NicholasBoll: Figma's updated. I'm reverting height changes.

package.json
```diff
@@ -63,7 +63,7 @@
     "cypress-axe": "^0.5.1",
     "cypress-pipe": "^1.5.0",
     "cypress-plugin-tab": "^1.0.3",
-    "cypress-storybook": "^0.0.1",
+    "cypress-storybook": "^0.1.1",
```
NicholasBoll: This update adds support for knobs

modules/_labs/pagination/react/stories/stories.tsx
```diff
@@ -0,0 +1,133 @@
...
+          pageSize={number('pageSize', 10) || 10}
+          showLabel={boolean('showLabel', false)}
+          showGoTo={boolean('showGoTo', false)}
+          goToLabel={text('goToLabel', 'Go To')}
+          currentPage={currentPage}
+          onPageChange={p => setCurrentPage(p)}
+          {...getAriaLabels()}
+        />
+      </Wrapper>
+    );
```
NicholasBoll: These stories no longer handle auto-resize by the browser. Perhaps we should use `react-use` to add functionality back? I think `width` should be provided for greater flexibility, but our stories should show how to pass it in.

lychyi:
https://github.com/Workday/canvas-kit/pull/301/commits/1912cd6b65ffa99ee6a2d00b1800361ec37b4cbf#diff-6d9fd0b9c3ff083fe440fe499966caafR243

This is bothering the heck out of me. Having your builds be non-deterministic in this manner makes me nervous. Like it's a sign something is off but it's not pointing us in the right direction.

However, it shouldn't gate merging into Labs. We should get this branch updated and merge. 🎉

NicholasBoll:
Yeah. I'm not happy with that diff. I wasn't able to figure it out. My editor didn't show an issue, but I could reproduce by running `yarn build-storybook` locally. So it wasn't just CI.

Perhaps upgrading Typescript would fix the issue? I gave up on trying to figure out why I was getting the error.

PR #310, opened 2019-11-08
fix(common): Port accessible hide styles from JS to CSS

vibdev:
## Summary

Code to make text screen reader only was added to JS common repo but was not mirrored in the css. This adds a new class to the common css

## Checklist

- [x] `yarn test` passes

## Additional References

This PR also changes Change word-wrap: normal to white-space, see:

https://medium.com/@jessebeach/beware-smushed-off-screen-accessible-text-5952a4c2cbfe

### DIFF
--- a/modules/common/css/lib/accessibility.scss
+++ b/modules/common/css/lib/accessibility.scss
@@ -21,3 +21,16 @@
     outline: none;
   }
 }
+
+.wdc-accessible-hide {
+  clip: 'rect(1px, 1px, 1px, 1px)';
+  clip-path: 'polygon(0px 0px, 0px 0px, 0px 0px, 0px 0px)';
+  position: 'absolute';
+  overflow: 'hidden';
+  word-wrap: 'normal';
+  height: '1px';
+  width: '1px';
+  margin: '-1px';
+  padding: 0;
+  border: 0;
+}
--- a/modules/common/css/lib/accessibility.scss
+++ b/modules/common/css/lib/accessibility.scss
@@ -22,7 +22,7 @@
   }
 }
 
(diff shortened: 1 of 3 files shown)


modules/common/css/lib/accessibility.scss
```diff
@@ -21,3 +21,16 @@
     outline: none;
   }
 }
+
+.wdc-accessible-hide {
```
anicholls: Can you turn this into a mixin and call it from within `.wdc-accessible-hide` so it's available as both a mixin and a class?

modules/common/react/lib/styles/accessibleHide.ts
```diff
@@ -9,7 +9,7 @@ const accessibleHide: CSSObject = {
   clipPath: 'polygon(0px 0px, 0px 0px, 0px 0px, 0px 0px)',
   position: 'absolute',
   overflow: 'hidden',
-  wordWrap: 'normal',
+  whiteSpace: 'nowrap',
```
anicholls: What was the reason for this change?

vibdev: linked a article at top from facebook engineer. 
TLDR: Some browser / screen reader combos don't read the space between words hidden with clip. Basically the words get newline separated and they don't read new lines as a space -> Hello World -> Hello{LineFeed}World
I read a few blogs today to make sure this is still the best method and this seems like the most bullet proof.

modules/common/css/lib/accessibility.scss
vibdev: updated!

PR #318, opened 2019-11-15
ci: Temporarily convert to TravisCI while billing is being figured out

NicholasBoll:
## Summary

Github shut off our access to Github Actions. Hopefully it can be resolved, but in the meantime we'll fall back to TravisCI


NicholasBoll:
I left the Github Action defined, so Github is failing a check instantly. I removed it from a required check, so it won't block the build. We can remove the GH Action files entirely if we don't want that red `x` on PRs, but my hope is this is temporary.

package.json
```diff
@@ -102,6 +102,7 @@
     "rollup-plugin-node-resolve": "^5.1.0",
     "rollup-plugin-terser": "^5.0.0",
     "sass-loader": "^8.0.0",
+    "storybook-chromatic": "^3.1.0",
```
NicholasBoll: The `chromaui/actions` Github Action normally takes care of this dependency. Github Actions can provide their own `node_modules` dependencies which is cool. This will be required until we get Github Actions back.

.travis.yml
```diff
@@ -1,42 +1,79 @@
...
       on:
         tags: true
-  - stage: master check and build storybook
+
+  - stage: master
     script:
-    - yarn test
-    - yarn build-storybook
+      - yarn chromatic --auto-accept-changes
+      - yarn build-storybook
```
NicholasBoll: I'm unclear if both `yarn chromatic` and `yarn build-storybook` are required. Chromatic builds storybook, but I'm not sure if the directories are the same. I'm leaving both for now.

PR #326, opened 2019-11-21
Docs: Audit prop descriptions

sahlhoff:
## Summary

This is updating our prop descriptions in components. I mostly copied over the descriptions from the component `Readme`. 

## Checklist

- [x] `yarn test` passes
      applicable


modules/tooltip/react/lib/Tooltip.tsx
```diff
@@ -5,8 +5,19 @@ import {TransformOrigin, getTranslateFromOrigin} from '@workday/canvas-kit-react
...
+   */
   transformOrigin: TransformOrigin;
+  /**
+   * Unique id of the Tooltip.
+   */
   id?: string;
+  /**
+   * The content for the Tooltip.
+   */
+  children?: string;
```
sahlhoff: Should we include this prop?

modules/_labs/header/react/lib/parts/WorkdayLogoTitle.tsx
```diff
@@ -14,6 +14,9 @@ export type WorkdayLogoTitleProps = {
    * The title to display in the header. Not used if `brand` is provided
```
anicholls: defaults?

modules/_labs/menu/react/lib/Menu.tsx
```diff
@@ -7,13 +7,37 @@ import {commonColors, spacing, borderRadius} from '@workday/canvas-kit-react-cor
 import {hideMouseFocus, GrowthBehavior} from '@workday/canvas-kit-react-common';
 
 export interface MenuProps extends GrowthBehavior, React.HTMLAttributes<HTMLUListElement> {
+  /**
+   * One or more MenuItem components.
```
anicholls: Or a component that has the same interface as `MenuItem`

modules/_labs/menu/react/lib/Menu.tsx
```diff
@@ -7,13 +7,37 @@ import {commonColors, spacing, borderRadius} from '@workday/canvas-kit-react-cor
...
+   */
   children?: React.ReactElement<MenuItemProps> | React.ReactElement<MenuItemProps>[];
+  /**
+   * Allows you to show and hide the menu from a parent component. Usefully for things like menu buttons.
```
anicholls: Missing default. Shouldn't be optional actually

modules/banner/react/lib/Banner.tsx
```diff
@@ -16,19 +16,19 @@ export interface BannerProps extends React.ButtonHTMLAttributes<HTMLButtonElemen
    */
   onClick?: (e: React.SyntheticEvent) => void;
   /**
-   * Text inside the banner
+   * Label of the banner
    */
   label?: React.ReactNode;
   /**
-   * Size and position of the banner
+   * Set the banner variant as `full` or `sticky`
```
anicholls: Nit: Can we standardize on tenses and consistency here? i.e.
```
Label of the banner
```
vs
```
The label of the banner
```
vs
```
Set the label of the banner
```

sahlhoff: @anicholls Yup, let's do it! Leaning towards any of these 3?

sahlhoff:
@anicholls Re:defaults, I simply copied over what was in the `Readme` – i'll take a closer look to include these defaults.

modules/banner/react/lib/Banner.tsx
jamesfan: My vote would be a slightly modified version of the third option with `Sets` instead of `Set` (and with a period at the end):
```
Sets the label of the banner.
```

I like having the sentence work grammatically if you read the prop name immediately followed by the description (e.g., "Label sets the label of the banner." works but not "Label set the label of the banner.")

modules/tooltip/react/lib/Tooltip.tsx
anicholls: What was the intent behind adding it? We have quite a few places like this where we are omitting inherited props that are critical to functionality (`children` is very common, `value` for inputs, etc.). We should probably have a consistent stance.

modules/banner/react/lib/Banner.tsx
anicholls: While I agree with @jamesfan on it being nice when it works grammatically, I prefer the second option due to it's consistency across different types of props. Going with "Sets" forces us to pick different verbs for different types of props. For example, "Sets an action when interacting with the banner" is kind of a weird description for `onClick`.

I like to think about each prop as a noun, so IMO starting with "The [noun]" makes the most sense. "The action performed when...", "The text inside...", "The color of...", "The function called when...", etc.

anicholls: One exception to this actually, would be `boolean` props. I think they need to start with a verb like @jamesfan says. e.g. `Specifies/determines whether X happens`

anicholls: If we agree on some rules, it would be great to add them to https://github.com/Workday/canvas-kit/blob/master/API_PATTERN_GUIDELINES.md#props 🙂

modules/status-indicator/react/lib/StatusIndicator.tsx
```diff
@@ -109,19 +109,20 @@ export const statusIndicatorStyles: StatusIndicatorGenericStyle = {
...
 export interface StatusIndicatorProps extends React.HTMLAttributes<HTMLSpanElement> {
   /**
-   * Type of status indicator (e.g. Gray, Orange etc.).
+   * Type of status indicator. The Canvas system has predefined types that are meant to indicate various kinds of statuses.
+   * @default StatusIndicatorEmphasis.High
    */
   type: StatusIndicatorType;
   /**
-   * Emphasis of status indicator (e.g. High v Low).
+   * Emphasis of status indicator. Currently supports High(default) and Low emphasis.
```
anicholls: Let's use `@default` instead of `(default)` in the comment

modules/tooltip/react/lib/Tooltip.tsx
sahlhoff: Yeah I wasn't 100% if this should be included. Hence the comment. 

My perspective was we document it for `Toast` which is `The message text for the Toast.`. 

Now thinking about it more I think we should omit `children` from the docs unless there are restrictions on the `children` type.

modules/banner/react/lib/Banner.tsx
jamesfan: Sounds fair 👍, let's go with option 2 then? @sahlhoff let's chat later about how to divvy up the work for standardizing the descriptions and updating the guidelines doc.

modules/_labs/drawer/react/lib/Drawer.tsx
```diff
@@ -4,27 +4,31 @@ import {colors, spacing, CanvasSpacingValue} from '@workday/canvas-kit-react-cor
...
 export interface DrawerProps extends React.HTMLAttributes<HTMLDivElement> {
   /**
-   * Adjust padding for the contents of the Drawer. Default: `spacing.s`
+   * Adjust padding for the contents of the Drawer.
+   * @default spacing.s
    */
   padding?: CanvasSpacingValue;
   /**
-   * Determines the side in which the Drawer can open from. Default: `DrawerDirection.Right`
+   * Determines the side in which the Drawer can open from.
```
sahlhoff:
```suggestion
   * Determines test
```

jamesfan:
This may be outside the scope of this PR, but do we have a convention for when we define `defaultProps` for boolean props? For example, in `Checkbox.tsx`, `checked` is present in `defaultProps`, but `disabled` isn't. One side-effect of this inconsistency is a bit of weirdness in the Storybook docs now that we've established a convention of tagging all two-state boolean props with `@default false` in the documentation:

image

Note how the default for `disabled` is enclosed in quotes. It seems that if the default is pulled from `defaultProps`, the docs format it as `false`. However, if the default is pulled from the `@default` tag (i.e., the docs parser was unable to locate a `defaultProp` and looked for a `@default` tag as a fallback), the docs format it as `"false"`.

Regardless of this formatting quirk, if we're tagging all booleans with `@default false` in the docs, wouldn't it make sense to set all booleans to `false` in `defaultProps` to keep things consistent? The exception would be three-state booleans defaulting to `undefined` -- we would omit these from `defaultProps`.

jamesfan:
@sahlhoff A couple questions:

1. Any idea why the docs for `Radio Group` are showing the props for `Radio`?
2. Also, the docs for `Select` are showing the wrong prop table (only shows `disabled` and its not even pulling that from the description in `Select.tsx`).

NicholasBoll:
> This may be outside the scope of this PR, but do we have a convention for when we define `defaultProps` for boolean props? For example, in `Checkbox.tsx`, `checked` is present in `defaultProps`, but `disabled` isn't. One side-effect of this inconsistency is a bit of weirdness in the Storybook docs now that we've established a convention of tagging all two-state boolean props with `@default false` in the documentation:
> 
> image
> 
> Note how the default for `disabled` is enclosed in quotes. It seems that if the default is pulled from `defaultProps`, the docs format it as `false`. However, if the default is pulled from the `@default` tag (i.e., the docs parser was unable to locate a `defaultProp` and looked for a `@default` tag as a fallback), the docs format it as `"false"`.
> 
> Regardless of this formatting quirk, if we're tagging all booleans with `@default false` in the docs, wouldn't it make sense to set all booleans to `false` in `defaultProps` to keep things consistent? The exception would be three-state booleans defaulting to `undefined` -- we would omit these from `defaultProps`.

I think the string thing is a problem with how Storybook interprets the `@default` JSDoc parameter. VSCode seems to understand it correctly. I think we should default booleans to `false` in the `defaultProps` for class components and destructured defaults in functional components except for intentional 3-state booleans (although I don't generally like 3-state booleans)

sahlhoff:
@jamesfan good, Q's. For the first question:

Storybook generates the prop table from the `component` parameter, see: https://github.com/storybookjs/storybook/blob/master/addons/docs/docs/docspage.md#component-parameter for more information.

I'm not seeing the `Select` issue you are. I'll sync up with you offline and confirm we're looking at the same thing.

modules/_labs/header/react/lib/parts/SearchBar.tsx
```diff
@@ -16,61 +16,69 @@ import uuid from 'uuid/v4';
...
   openButtonLabel: string;
   /**
-   * Screenreader Label to close collapsed search bar
+   * The screenreader Label to close collapsed search bar.
+   * @default Cancel
    */
   closeButtonLabel: string;
   /**
-   * Show button to clear input field
+   * It true, show button to clear input field.
```
NicholasBoll:
```suggestion
   * If true, show button to clear input field.
```

modules/_labs/menu/react/lib/Menu.tsx
```diff
@@ -7,13 +7,38 @@ import {commonColors, spacing, borderRadius} from '@workday/canvas-kit-react-cor
 import {hideMouseFocus, GrowthBehavior} from '@workday/canvas-kit-react-common';
 
 export interface MenuProps extends GrowthBehavior, React.HTMLAttributes<HTMLUListElement> {
+  /**
+   * One or more MenuItem components or a component that has the same interface as `MenuItem`.
+   */
   children?: React.ReactElement<MenuItemProps> | React.ReactElement<MenuItemProps>[];
+  /**
+   * It true, sets the menu to open from a parent component. Usefully for things like menu buttons.
```
NicholasBoll:
```suggestion
   * If true, sets the menu to open from a parent component. Usefully for things like menu buttons.
```

modules/_labs/header/react/lib/parts/SearchBar.tsx
```diff
@@ -16,61 +16,69 @@ import uuid from 'uuid/v4';
...
    */
-  inputLabel: string;
+  inputLabel?: string;
   /**
-   * Screenreader label for submit button
+   * The screenreader label for submit button
+   * @default Search
    */
-  submitLabel: string;
+  submitLabel?: string;
```
sahlhoff: We're failing CI because of this.

We pass `submitLabel` to `aria-label` which has a required type `string`. I'm not sure why this is failing though because we set defaultProps for these optional props. Is there another way to have optional `defaultProps` that are required for render without CI failing?

modules/_labs/menu/react/lib/Menu.tsx
```diff
@@ -7,13 +7,38 @@ import {commonColors, spacing, borderRadius} from '@workday/canvas-kit-react-cor
...
+  /**
+   * Zero based index of which menu item should initially receive focus.
+   */
   initialSelectedItem?: number;
+  /**
+   * A unique id for the menu that is used for aria and html id attributes.
+   */
   id?: string;
+  /**
+   * An html id of the element that labels the menu. Often used with menu buttons.
```
lychyi:
```suggestion
   * An HTML id of the element that labels the menu. Often used with menu buttons.
```

jamesfan:
@sahlhoff Check out the updates to the guidelines from my latest commit. I added some additional guidelines based on issues I encountered while updating the prop descriptions for the Input components. Happy to make adjustments, let me know if you have any feedback.

Issue #320, opened 2019-11-18
Incompatible role and aria attribute defined on SidePanel


yingzuo:
## 🐛 Bug Report

I ran the Lighthouse accessibility audit on a page with SidePanel. It found an issue with the incompatible role and aria attribute defined on SidePanel. Basically, aria-orientation="vertical" can not be defined with role="region".

Each ARIA `role` supports a specific subset of `aria-*` attributes. Mismatching these invalidates the `aria-*` attributes.

## To Reproduce

Run the lighthouse accessibility audit with Chrome devtool on SidePanel.

## Expected Behavior

No issue is found by Lighthouse accessibility audit 

## Actual Results

The report shows an issue found on SidePanel: [aria-*] attributes do not match their roles

PR #327, opened 2019-11-21
fix(side-panel): Add flexibility for aria attributes

mannycarrera4:
Fixes: https://github.com/Workday/canvas-kit/issues/320

We were forcing aria attributes so the side panel and it wasn't accessible. We still kept the role as default, but removed `aria-orientation`. The consumer can now pass any other aria attribute if they need to override anything.

### DIFF
--- a/modules/side-panel/react/lib/SidePanel.tsx
+++ b/modules/side-panel/react/lib/SidePanel.tsx
@@ -19,6 +19,7 @@ export interface SidePanelProps extends React.HTMLAttributes<HTMLDivElement> {
   backgroundColor?: SidePanelBackgroundColor;
   closeNavigationLabel: string;
   openNavigationLabel: string;
+  role?: string;
 }
 
 export interface SidePanelState {
@@ -136,6 +137,7 @@ export default class SidePanel extends React.Component<SidePanelProps, SidePanel
     backgroundColor: SidePanelBackgroundColor.White,
     closeNavigationLabel: 'close navigation',
     openNavigationLabel: 'open navigation',
+    role: 'region',
   };
 
   constructor(props: SidePanelProps) {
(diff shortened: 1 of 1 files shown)


modules/side-panel/react/lib/SidePanel.tsx
```diff
@@ -166,13 +168,13 @@ export default class SidePanel extends React.Component<SidePanelProps, SidePanel
...
       closeNavigationLabel,
+      role,
       ...elemProps
     } = this.props;
 
     return (
       <SidePanelContainer
-        role="region"
-        aria-orientation="vertical"
+        role={role}
```
anicholls: Shouldn't this have been possible before because we extend `React.HTMLAttributes<HTMLDivElement>`?

It would have been spread with `{...elemProps}` and if we end up with 
```tsx
<SidePanelContainer
  role="region"
  // ...
  role={this.props.role}
>
```
Where the second role prop would override the first

mannycarrera4: oh herm.. I think either way we shouldn't enforce the aria attributes, and these two together doesn't pass Lighthouse as per the issue. Should we still change it?

anicholls: So the problem was more around having `role="region"` and `aria-orientation="vertical"` by default?

mannycarrera4: correct! @anicholls

anicholls: So why not just remove `aria-orientation="vertical"`? `role="region"` is still the correct default, right? Consumers still have the ability to override it without explicitly setting it as a prop. I just question it because we don't have the `role` exposed in any other component interfaces.

mannycarrera4: the drawer component in Labs has a role as well but no default, so I was going off of what I did there for this use case. I'm ok with adding it back in and setting it as a default!

anicholls:
Can you update the summary to explain a bit more about where we ended up? Thanks!

PR #328, opened 2019-11-22
feat(modal): Update focus-trap library

### LINKED ISSUE #340: Modal: focus-trap-react performance issues
## 🐛 Bug Report

If the popup has many tabbable nodes, tabbing blocks the main thread. We had this issue on the main codebase using this library. That's why we had to create our own version of focus trap.

## To Reproduce

Add 400 buttons in the popup, throttle the CPU, open the popup, and hold tab key.

https://codesandbox.io/s/canvas-kit-modal-focus-trap-performance-cfk2x

alexandrzavalii:
## Summary

The Modal component uses `focus-trap-react` which uses [tabbable](https://github.com/davidtheclark/tabbable) which doesn't scale with many focusable elements on the page. The main performance issue in `tabbable` is detection around `display: none` which requires `getComputedStyle` be run on every potentially focusable element on the page.

This change switches focus trap library to [focus-trap-js](https://github.com/alexandrzavalii/focus-trap-js) which has a simpler and faster next/previous tab detection than the version of `tabbable` used by `focus-trap-react`.

Fixes #340

## Checklist

- [x] `yarn test` passes

## Additional References

Before (43ms handler for tab key):
Before

After (1.5ms handler for tab key):
Screen Shot 2019-12-03 at 1 50 44 PM


NicholasBoll:
This PR contains a lot of unrelated code changes.

Also, how is this library different from [tabbable](https://github.com/davidtheclark/tabbable) + [focus-trap](https://github.com/davidtheclark/focus-trap) (which `focus-trap-react` depends on)? `focus-trap-react` gives us some extras like automatically handling focus transfer when the focus-trap is enabled and disabled (which is why the tests are failing).

The only benefit I see from using the base tab library is having finer-grain control on _when_ focus order is computed/recomputed.

It looks like `tabbable` does more checks for if something is able to receive focus. Are we concerned about a loss of accuracy? Neither library (nor any library for that matter) is 100% accurate.

If `tabbable` is performant enough, can we use that library directly when we want finer-grained control? Or does this library offer significant performance?

modules/_labs/header/react/lib/Header.tsx
```diff
@@ -129,7 +129,7 @@ const navStyle = ({themeColor}: Pick<HeaderProps, 'themeColor'>) => {
           height: 'inherit',
           transition: `color 150ms ease-out 0s`,
           cursor: 'pointer',
-          '&:first-child > *': {
+          '&:first-of-type > *': {
```
NicholasBoll: What does this change have to do with changing focus trap libraries? Is it a necessary change?

modules/modal/react/lib/Modal.tsx
```diff
@@ -69,125 +69,124 @@ const transformOrigin = {
...
+      const elem =
+        (firstFocusableRef && firstFocusableRef.current) ||
+        getFirstElementToFocus(modalRef.current);
+      elem.focus();
+    }
+  }, [modalRef]);
+};
+
+const Modal = ({
+  open,
```
NicholasBoll: This PR removes default props, but doesn't set them back up again here. Adding `open = false` would replace existing functionality of the default props.

modules/modal/react/lib/Modal.tsx
```diff
@@ -69,125 +69,124 @@ const transformOrigin = {
...
 
-  componentDidMount() {
-    document.addEventListener('keydown', this.handleKeydown);
-  }
-
-  componentWillUnmount() {
-    document.removeEventListener('keydown', this.handleKeydown);
-  }
+  useKeyDownListener(handleKeydown);
+  useInitialFocus(modalRef, firstFocusRef);
```
NicholasBoll: We haven't released a component that depends on React hooks yet. Not all our dependents are using React 16.8+ yet. Hooks in this file have been thus far optional. Are we ready to break compatibility @anicholls?

modules/modal/react/lib/Modal.tsx
```diff
@@ -69,125 +69,124 @@ const transformOrigin = {
...
     );
   }
-}
+  return null;
+};
+
+Modal.Padding = PopupPadding;
+Modal.Width = ModalWidth;
+
+Modal.propTypes = {
```
NicholasBoll: Did you mean `defaultProps`? That might be why the interface had to be changed.

lychyi:
I agree with @NicholasBoll here. This PR engages in more than just changing a library.

To adequately review this PR, it would help the reviewers if you walked through what was changed and why in the PR description or inline comments. Otherwise, this code change is really difficult to comprehend (especially considering the diff).

modules/_labs/header/react/lib/Header.tsx
alexandrzavalii: @NicholasBoll you are right its not related to that PR, https://github.com/storybookjs/storybook/issues/6998
I am reverting this

modules/modal/react/lib/Modal.tsx
alexandrzavalii: @NicholasBoll thanks! that was very silly mistake :)

alexandrzavalii:
@lychyi I have reverted changes related to first-of-type server side render issue, should be easier to review now.

modules/modal/react/lib/Modal.tsx
alexandrzavalii: hmm okay, I just saw that canvas is using 16.8.6. What dependencies are blocking us to use hooks?

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -0,0 +1,178 @@
...
+    }
+    return () => {
+      if (handlerRef) {
+        handlerRef.focus();
+      }
+    };
+  }, [modalRef, firstFocusRef]);
+};
+
+const ModalContent = ({
```
NicholasBoll: The ModalContent was moved out of the `Modal` component to prevent processing if `open` was false.

cypress/support/commands.ts
```diff
@@ -14,3 +15,28 @@ Cypress.Commands.add('injectAxe', () => {
     window.axe = axe;
   });
 });
+
+// Add better logging to cy.tab
+Cypress.Commands.overwrite('tab', (originalFn, subject) => {
```
NicholasBoll: This change isn't strictly necessary, but I wanted better logging of cy.tab. I opened an issue: https://github.com/Bkucera/cypress-plugin-tab/issues/12

PR #334, opened 2019-11-26
feat(menu): Fix import of React dependency.

jsievenpiper:
fix: Change React dependency import for Menu to fix type declarations for consumers.

## Summary
React was being imported the more ES6-y way, rather than the TypeScript method. This causes some linters and IDEs to not resolve the depedency correctly, and therefore not see the Menu as a valid react component.

## Checklist

- [x] `yarn test` passes


jsievenpiper:
Yay 🎉  my first canvas-kit PR!

PR #336, opened 2019-11-27
chore: Point npm script to use proper Prettier configuration

Parker-Ledoux:
The `format:pretty` command would fail because we were pointing to a `.prettierrc` file but we actually store the prettier configurations in `.prettierrc.js`.
I then ran the command to update any files that needed formatting.
Finally, I added `format:pretty` as a linting hook because we say in our contributing guide that we run both prettier and eslint upon commit but we've actually only been running the linter. 
https://github.com/Workday/canvas-kit/blame/master/CONTRIBUTING.md#L54

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,7 @@
     "commitmsg": "commitlint --edit $HUSKY_GIT_PARAMS",
     "build-storybook": "build-storybook -c .storybook -o docs",
     "lint": "node utils/check-lockfile-links.js && node utils/check-mismatched-dependencies.js && eslint -c ./eslintrc.js --ext=jsx,ts,tsx .",
-    "format:prettier": "prettier --config=./.prettierrc --ignore-path=./.prettierignore \"**/*.{js,jsx,json,ts,tsx,md}\" --write",
+    "format:prettier": "prettier --config=./.prettierrc.js --ignore-path=./.prettierignore \"**/*.{js,jsx,json,ts,tsx,md}\" --write",
     "format": "yarn lint --fix",
     "test": "TZ=UTC jest",
     "bump": "lerna version",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -130,7 +130,7 @@
     "commitmsg": "commitlint --edit $HUSKY_GIT_PARAMS",
     "build-storybook": "build-storybook -c .storybook -o docs",
     "lint": "node utils/check-lockfile-links.js && node utils/check-mismatched-dependencies.js && eslint -c ./eslintrc.js --ext=jsx,ts,tsx .",
-    "format:prettier": "prettier --config=./.prettierrc --ignore-path=./.prettierignore \"**/*.{js,jsx,json,ts,tsx,md}\" --write",
+    "format:prettier": "prettier --config=./.prettierrc.js --ignore-path=./.prettierignore \"**/*.{js,jsx,json,ts,tsx,md}\" --write",
```
anicholls: Good catch! 👍 I renamed this file so it could be imported into `eslintrc.js` and missed updating it here.

package.json
```diff
@@ -145,6 +145,7 @@
   "lint-staged": {
     "linters": {
       "*.{ts,tsx}": [
+        "yarn format:prettier",
```
anicholls: So [prettier is run as part of ESLint](https://github.com/Workday/canvas-kit/blob/master/eslintrc.js#L82), which is why this was left out. I would expect this to be running prettier twice. However, there do seem to be some formatting changes that have been picked up, so I'm not quite sure why there would be a difference. Any ideas?

anicholls:
@Parker-Ledoux For future PRs, do you mind [using a fork](https://github.com/Workday/canvas-kit/blob/master/CONTRIBUTING.md#pull-requests)? Thanks!

package.json
Parker-Ledoux: Probably this: https://github.com/Workday/canvas-kit/blob/master/eslintrc.js#L1 😄 
I'll fork and put up a new PR.

Parker-Ledoux:
Closing so I can put up a PR from my fork.

PR #337, opened 2019-11-27
feat: Add bidirectionality support to input components

stephanerangaya:
Blocked on #288

# Add support for Bidirectionality to Form Field components

This PR update our Form Field components to use the new `styled` function introduced by bidirectionality


stephanerangaya:
All `form-field` components support Bidirectionality except for `ColorInput` and `ColorPreview`.

image

As you can see here, in an ideal scenario, we should show this: 
image

This is a bit more complex than what `rtlCSSJS` allows us to do, we might need to add some logic inside the component to extract the current theming direction. I'm looking into it, but if you have ideas, please share them!

Temporarily, we could also recommend to not change the direction of this component but this is not an optimal user experience.

rickshaffer:
I looked into this a bit more and reviewed with Sam. Your ideal scenario is.... ideal. 
So we'd want to do something like this:
image

stephanerangaya:
@rickshaffer @anicholls Ideally, we would like the hex code in the input to be at 8px from the swatch, but at the moment it is difficult to position because the hex can be a combination of letters and numbers. Look at this:

image
image
image

To solve this issue, we could:
1. Use a monospace font for this input. 
2. Align the `#` and `hex code` to the left of the Input instead of 8px from the swatch.

What are your thoughts?

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -44,23 +49,37 @@ const CustomHexInput = styled(TextInput)<Pick<ColorInputProps, 'disabled' | 'gro
     },
   ({disabled}) => ({
     backgroundColor: disabled ? colors.soap200 : '',
+  }),
+  ({theme}) => ({
+    paddingLeft: theme.direction === ContentDirection.LTR ? '46px' : undefined,
+    paddingRight: theme.direction === ContentDirection.LTR ? undefined : 'calc(100% - 86px)',
```
anicholls: Not sure if you realize this, but this is getting flipped by rtl-css-js resulting in `padding-left`

anicholls:
Personally, I would say we should go with a monospace font. Then if the field grows, the value is in the right place rather than way off on the left.

Even if we could get it to align properly on the right without a monospace font, the hashtag would be jumping left and write as you enter a hex depending on whether you enter narrow vs. wide letters.

modules/color-picker/react/lib/ColorInput.tsx
stephanerangaya: Yep, this is on purpose. Basically, I want `padding-left` to be `46px` on LTR but `calc(100% - 86px)` on RTL. If I do something like this:
`paddingLeft: theme.direction === C...n.LTR ? '46px' : 'calc(100% - 86px)',`

The result will be  `paddingRight : 'calc(100% - 86px)',`, which is the wrong side (because as you mentioned, it is getting flipped)

stephanerangaya: I refactored this to use `/* @noflip */` that tells rtl-css-js to not flip.

lychyi:
@stephanerangaya Can we change the target branch so that the diff is a bit cleaner? Right now it looks like it includes everything from the two blocking PRs.

anicholls:
I love how simple this branch is 👌. Speaks to the elegance of #272 and #288!

modules/checkbox/react/lib/Checkbox.tsx
```diff
@@ -12,7 +12,7 @@ import {SystemIcon} from '@workday/canvas-kit-react-icon';
 import {checkSmallIcon} from '@workday/canvas-system-icons-web';
 import uuid from 'uuid/v4';
 
-export interface CheckboxProps extends React.InputHTMLAttributes<HTMLInputElement> {
+export interface CheckboxProps extends Themeable, React.InputHTMLAttributes<HTMLInputElement> {
   checked: boolean;
```
anicholls: Shouldn't the checkmark icon flip?

modules/checkbox/react/lib/Checkbox.tsx
```diff
@@ -57,7 +58,7 @@ const CheckboxInputWrapper = styled('div')<Pick<CheckboxProps, 'disabled'>>(
       height: checkboxHeight,
       transition: 'box-shadow 150ms ease-out',
       width: checkboxWidth,
-      zIndex: 1,
+      pointerEvents: 'none',
```
anicholls: What is this for?

modules/form-field/react/lib/Hint.tsx
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react';
-import styled from '@emotion/styled';
+import {styled, Themeable} from '@workday/canvas-kit-labs-react-core';
 import {ErrorType} from '@workday/canvas-kit-react-common';
 import {spacing, type} from '@workday/canvas-kit-react-core';
 
-export interface HintProps extends React.HTMLAttributes<HTMLParagraphElement> {
+export interface HintProps extends Themeable, React.HTMLAttributes<HTMLParagraphElement> {
```
anicholls: `Label` isn't flipping to the right side

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -30,9 +35,9 @@ const colorInputWidth = 116;
 const CustomHexInput = styled(TextInput)<Pick<ColorInputProps, 'disabled' | 'grow'>>(
   {
     boxSizing: 'border-box',
-    paddingLeft: '46px',
     minWidth: colorInputWidth,
     width: colorInputWidth,
+    ...type.variant.mono,
```
anicholls: The `#` should be mono too. Now they don't align vertically.

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -13,7 +13,12 @@ import {checkSmallIcon} from '@workday/canvas-system-icons-web';
 import {SystemIcon} from '@workday/canvas-kit-react-icon';
 import TextInput, {TextInputProps} from '@workday/canvas-kit-react-text-input';
 
-export interface ColorInputProps extends TextInputProps, GrowthBehavior {
+
```
anicholls: Nit: can you remove the extra new line here?
```suggestion
```

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -44,23 +49,36 @@ const CustomHexInput = styled(TextInput)<Pick<ColorInputProps, 'disabled' | 'gro
     },
   ({disabled}) => ({
     backgroundColor: disabled ? colors.soap200 : '',
+  }),
+  ({theme}) => ({
+    paddingLeft: theme.direction === ContentDirection.LTR ? '46px' : 'calc(100% - 86px) /* @noflip */' ,
```
anicholls: Can you add a comment here that this is because the input is always forced to LTR? I would say it's good practice to always explain when/why we need to override `rtl-css-js` (whether it's with `@noflip`, or just based on `theme.direction`).

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -44,23 +49,36 @@ const CustomHexInput = styled(TextInput)<Pick<ColorInputProps, 'disabled' | 'gro
...
-    left: 36,
     top: 10,
     ...type.body,
     ...type.variant.hint,
   },
   ({disabled}) => ({
     color: disabled ? inputColors.disabled.text : undefined,
+  }),
+  ({theme}) => ({
+    left: theme.direction === ContentDirection.LTR ? '36px' : '88px',
```
anicholls: A comment explaining this would be nice too.

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -107,8 +125,10 @@ export default class ColorInput extends React.Component<ColorInputProps> {
     const formattedValue = this.formatValue(value);
 
     return (
-      <ColorInputContainer>
+      <ColorInputContainer 
+      {...elemProps}>
```
anicholls: Why were `elemProps` added here? This is a potentially breaking change

modules/form-field/react/stories/stories_ColorInput.tsx
```diff
@@ -48,7 +48,7 @@ storiesOf('Components|Inputs/Color Picker/Color Input/React/Top Label', module)
   ))
   .add('Grow', () => (
     <FormField label="Label" inputId="input-grow" grow={true}>
-      {controlComponent(<ColorInput placeholder="Placeholder" />)}
+      {controlComponent(<ColorInput placeholder="??????" />)}
```
anicholls: Can we remove `placeholder` and just leave it as the default?

modules/radio/react/lib/Radio.tsx
```diff
@@ -43,10 +43,9 @@ const RadioContainer = styled('div')({
  */
 const RadioInputWrapper = styled('div')<Pick<RadioProps, 'disabled'>>(
   {
+    display: 'flex',
     height: radioHeight,
     width: radioWidth,
-    marginTop: '3px',
```
anicholls: This was added in https://github.com/Workday/canvas-kit/commit/1981fc802b30e0b03ef70ae8a64e37f7e30d933a. Do you know what it was for?

modules/text-input/react/lib/TextInput.tsx
```diff
@@ -1,10 +1,11 @@
 import * as React from 'react';
-import styled from '@emotion/styled';
+import {styled, Themeable} from '@workday/canvas-kit-labs-react-core';
```
anicholls: There's some weird cursor and positioning going on for `TextInput` when RTL. 

First, the cursor stays on the left as you type. If you move the cursor to the end and start typing, the text you type appears at the beginning (far left) of the string. It's like the text is appearing in RTL, but the cursor behavior is still LTR.

Second, the text is sliding slightly as you type. You only notice it on long strings. In one of the `grow` stories, type "asdf" 12-14 times and you will see that the last "f" is partially cut off on the right hand side of the field.

The same behavior is happening on `TextArea`

modules/checkbox/react/lib/Checkbox.tsx
stephanerangaya: I did some research and it seems it's not necessary to flip checkmark in RTL.

modules/checkbox/react/lib/Checkbox.tsx
stephanerangaya: When I added `display: 'flex',`, it created an issue where the checkbox wouldn't register clicks and this seemed to fix the issue but I removed it and it still work 🤷‍♂. So I'm going to remove it.

modules/form-field/react/lib/Hint.tsx
stephanerangaya: It won't flip if your text is in english but if you use a RTL language it should. The `Error:` and `Alert:` is not translatable at this point, I created an issue: https://github.com/Workday/canvas-kit/issues/385

image

modules/color-picker/react/lib/ColorInput.tsx
stephanerangaya: Looks like it is not useful, I might have done that to debug something.

modules/text-input/react/lib/TextInput.tsx
stephanerangaya: This only happens if you write in LTR language, if you write in an RTL language such as arabic, these issues are not occurring.

modules/radio/react/lib/Radio.tsx
stephanerangaya: I think it's related to `flex-start`, which we don't use anymore.

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -44,26 +49,56 @@ const CustomHexInput = styled(TextInput)<Pick<ColorInputProps, 'disabled' | 'gro
...
 
-const ColorInputContainer = styled('div')({
-  position: 'relative',
-});
+// Note regarding:
+//   ({theme}) => ({
+//     paddingLeft: theme.direction === ContentDirection.LTR ? '46px' : 'calc(100% - 86px) /* @noflip */' ,
+//   })
+//
+// We're using @noflip because ColorInput should stay LTR, therefore, we need to adjust the padding using ContentDirection, not using rtl-css-js.
```
anicholls: Can you just put this inline on line 54/55 and remove the stuff above?

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -44,26 +49,56 @@ const CustomHexInput = styled(TextInput)<Pick<ColorInputProps, 'disabled' | 'gro
...
 );
 
+// Note regarding:
+//   left: theme.direction === ContentDirection.LTR ? '36px' : '88px',
+//
+// This returns:
+//    - LTR -> left: 36px;
+//    - RTL -> right: 88px;
+//
+// It's because we're changing the value of this attribute based on ContentDirection but the attribute itself is getting flipped by rtl-css-js.
```
anicholls: Can you just put this inline on line 89/90? Also you're explaining what's going on here, but not _why_ we needed to do it.

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -44,23 +44,44 @@ const CustomHexInput = styled(TextInput)<Pick<ColorInputProps, 'disabled' | 'gro
...
 
 const PoundSignPrefix = styled('span')<Pick<ColorInputProps, 'disabled'>>(
   {
     position: 'absolute',
-    left: 36,
     top: 10,
     ...type.body,
-    ...type.variant.hint,
+    ...type.variant.mono,
+    color: colors.blackPepper300,
```
anicholls: Nit: This is part of `type.body`. Can we remove this?

```suggestion
```

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -44,23 +44,44 @@ const CustomHexInput = styled(TextInput)<Pick<ColorInputProps, 'disabled' | 'gro
...
 
-const ColorInputContainer = styled('div')({
-  position: 'relative',
-});
+const ColorInputContainer = styled('div')<Pick<ColorInputProps, 'grow'>>(
+  {
+    position: 'relative',
+    width: colorInputWidth,
+  },
+  ({grow}) =>
```
anicholls: Nit: This isn't hurting anything, but why is this necessary? Block level elements are naturally 100% width. If it's not needed, you can remove it from the render function as well.

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -44,23 +44,44 @@ const CustomHexInput = styled(TextInput)<Pick<ColorInputProps, 'disabled' | 'gro
...
     ...type.body,
-    ...type.variant.hint,
+    ...type.variant.mono,
+    color: colors.blackPepper300,
   },
   ({disabled}) => ({
     color: disabled ? inputColors.disabled.text : undefined,
+  }),
+  ({theme}) => ({
+    left: theme.direction === ContentDirection.LTR ? '36px' : '88px',
```
anicholls: Re. https://github.com/Workday/canvas-kit/pull/337#pullrequestreview-333591564, any time we do a special check like this we're going to have issues if the theme is undefined. For example, if I go into the storybook knob and set `direction` to an empty string (easy to do without typescript), this will turn into `left: 88px`.

Not sure of the best way to fix this. Maybe we explicitly set theme.direction to `ContentDirection.LTR` in [createCanvasTheme](https://github.com/Workday/canvas-kit/blob/master/modules/_labs/core/react/lib/theming/createCanvasTheme.ts#L80) if `partialTheme.direction` is not `ContentDirection.LTR` or `ContentDirection.RTL`

modules/color-picker/react/lib/ColorInput.tsx
stephanerangaya: It is necessary because the width is `colorInputWidth` otherwise (line 58)

anicholls: Ah - I missed that you added that as well, sorry. That way it behaves like the other input elements 👍

modules/color-picker/react/lib/ColorInput.tsx
stephanerangaya: Ok, should we do that in another branch?

anicholls: This is the only code affected by the RTL stuff so it's probably fine to do here

PR #345, opened 2019-12-02
fix: Allow icon injecting in shadowDOM

### LINKED ISSUE #338: initializeIcons function from @workday/canvas-kit-css-icon does not support the Shadow DOM
## 🐛 Bug Report

The functions in [canvas-kit-css-icon.js](https://github.com/Workday/canvas-kit/blob/master/modules/icon/css/lib/canvas-kit-css-icon.js) make use of the document.querySelectorAll() function to find all icon place holder elements and inject the SVG icon into them. However document.querySelectorAll() only queries the light DOM and misses any elements attached to a shadow DOM. 

The injectIcons function should be updated to take a root element parameter which would default to document. This way initializeIcons(this.shadowRoot) could be invoked and the attached icons could be discovered.

## To Reproduce

This is a LitElement example:

```
import { LitElement, html, css, customElement, property } from 'lit-element';
// @ts-ignore
import initializeIcons from '@workday/canvas-kit-css-icon';

const workdayStyle = require('@workday/canvas-kit-css/index.scss');
const litWorkdayStype = css(<any>[workdayStyle]);

@customElement('wd-css-example')
export class WDCSSElement extends LitElement {
 
  firstUpdated() {
    initializeIcons();
  }

  static get styles() {
    return [litWorkdayStype];
  }

  render() {

    return html`
     <div class="wdc-type">
        <div class="wdc-menu" role="menu" aria-label="Menu">
          <ul>
            <li role="menuitem" class="wdc-menu-item-focused">
              <i class="wdc-icon" data-icon="uploadCloud" data-category="system" />
              <span class="wdc-menu-item-label">First Item</span>
            </li>
          </ul>
        </div>
      </div>
    `;
  }
}


```

## Expected Behavior

injectIcons should take a base element parameter so that there is an option to not scan the whole DOM for icons or to scan a shadow DOM instance.

## Actual Results

document.querySelectorAll() does not inspect the shadow DOM by design.

## Browser (if applicable)

Chrome

vibdev:
Fixes #338 

## Summary

Allows you to pass a DOM Element to scope where to look for icons. You can use this to narrow the search instead of looking at the whole document or to target a shadowDOM

## Checklist

- [ ] `yarn test` passes
      applicable


vibdev:
@aaronanderson would this fix the issue for you?

aaronanderson:
Thanks for looking into this! There are other invocations of document.querySelectorAll() in canvas-kit-css-icon.js. Could the iconRoot reference be passed into the other functions?

vibdev:
@aaronanderson good call, updated

PR #346, opened 2019-12-02
fix(text-input): Update type of inputRef

davvidbaker:
Typing the ref as `React.Ref` instead of `React.RefObject` allows the prop to also accept [callback refs](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs).

Example use:
[Edit fancy-sea-ryh19](https://codesandbox.io/s/fancy-sea-ryh19?fontsize=14&hidenavigation=1&theme=dark)

### DIFF
--- a/modules/text-input/react/lib/TextInput.tsx
+++ b/modules/text-input/react/lib/TextInput.tsx
@@ -7,7 +7,7 @@ export interface TextInputProps
   extends GrowthBehavior,
     React.InputHTMLAttributes<HTMLInputElement> {
   error?: ErrorType;
-  inputRef?: React.RefObject<HTMLInputElement>;
+  inputRef?: React.Ref<HTMLInputElement>;
 }
 
 const Input = styled('input')<Pick<TextInputProps, 'error' | 'grow'>>(


modules/text-input/react/lib/TextInput.tsx
```diff
@@ -7,7 +7,7 @@ export interface TextInputProps
   extends GrowthBehavior,
     React.InputHTMLAttributes<HTMLInputElement> {
   error?: ErrorType;
-  inputRef?: React.RefObject<HTMLInputElement>;
+  inputRef?: React.Ref<HTMLInputElement>;
```
anicholls: Good catch! Was there a reason you only made this change to `TextInput`? There are several other components that use the `React.Ref` type.

davvidbaker: That's the only place it affected us 🤷‍♂

lychyi: Cool. Let's update the rest and merge this in for the next release

davvidbaker: I only found one other place where a ref that is a **prop exposed in the API** was typed `RefObject`. Added a commit that updates that instance.

anicholls: @davvidbaker good call on only worrying about the ones that are exposed. For Modal, a [PR recently when in that requires it to be a `RefObject`](https://github.com/Workday/canvas-kit/pull/328) . I've removed that commit and rebased, so we should be good to go now 👍

PR #370, opened 2019-12-13
fix:  Build Storybook for IE11

### LINKED ISSUE #352: Storybook not working in IE11
Right now, Storybook is not rendering in IE11 (probably a polyfill issue). So we'll need to get that up and running before vis reg can be run in ChromaticQA.

lychyi:
## Summary

Fixes #352 

There were two problems with #352 that broke Storybook for IE11.

1) An indirect dependency of `storybook-readme` had an incorrect distribution of their CJS modules. That issue is documented here: https://github.com/algolia/react-element-to-jsx-string/issues/387 and was fixed in 14.1.0. Unfortunately when we first installed `storybook-readme` we got unlucky with the broken version. This was fixed by doing a `yarn remove -W storybook-readme` followed by a `yarn add -W storybook-readme`. Unfortunately, there's no clean way to upgrade indirect dependencies (see: https://github.com/yarnpkg/yarn/issues/4986). Since this is a devDep I think it's okay to do it this way. But if it introduces any bugs in the future, another fix is to add it to `resolutions` in `package.json` (I'm not a huge fan of since it becomes hard to remember why we added it in the first place).
2) We added MDX 🎉to our Storybook builds. This has a new loader for `.mdx` files in our `webpack.config.js` so that just needed to respect our `browserslist`. Fix was pretty simple here and `@babel/preset-env` was added to the `babel-loader` options.

Now we can finally turn on vis reg on ChromaticQA with IE11 snapshots 👍


jamesfan:
Looks like we'll need to add a polyfill for `closest`.

image

sahlhoff:
@lychyi thanks for looking into this 🎉

lychyi:
@jamesfan I got bamboozled by docs. I'll need to add polyfills. Apparently I'm not the only one: https://github.com/babel/website/issues/1660

jamesfan:
> @jamesfan I got bamboozled by docs. I'll need to add polyfills. Apparently I'm not the only one: [babel/website#1660](https://github.com/babel/website/issues/1660)

@lychyi Also, in case you were planning on using `babel-polyfill` (or anything based on `core-js`), it doesn't look like it includes a polyfill for `closest`:

https://github.com/zloirock/core-js/issues/317 (admittedly, this is from over two years ago)
https://dev.to/michi/what-babel-polyfill-doesnt-include-31bi (this is more recent)

jamesfan:
@anicholls See above for context re: `closest` polyfill. Looks like `closest` is just used in one place in `InputProvider`.

NicholasBoll:
Polyfilling is tricky for a library. On one hand, it is convenient to polyfill automatically. On the other, if every lib polyfills, our dependencies become bloated where multiple fixes exist.

Applications already need to polyfill for features like `new Promise`.

We can either not use functions that are not available in IE11 or add documentation about what polyfills are needed.

I've probably added stuff like `str.includes('foo')` vs `str.indexOf('foo') > -1`

anicholls:
@NicholasBoll that's a very good point. I'd rather not need a note that consumers have to manually add a polyfill. Maybe we just add it inline since `closest` is only used in one place. Though we do use `str.includes()` in a few places too.

.storybook/webpack.config.js
```diff
@@ -7,7 +7,19 @@ const welcomeSectionPath = path.resolve(__dirname, './');
 const utilsPath = path.resolve(__dirname, '../utils');
 const postcssConfigPath = path.resolve(__dirname, './postcss.config');
 
-module.exports = ({config, mode}) => {
+module.exports = ({ config, mode }) => {
+  // This is so we get consistent results when loading .ts/tsx and .mdx files
+  const babelPresetEnvConfig = [
```
lychyi: We use this in a couple places in the config (.ts/.tsx and .mdx files) so I stored it here.

.storybook/webpack.config.js
```diff
@@ -39,8 +51,14 @@ module.exports = ({config, mode}) => {
     include: [modulesPath, utilsPath],
     loader: require.resolve('babel-loader'),
     options: {
-      presets: [['react-app', {flow: false, typescript: true}]],
+      presets: [
+        '@babel/preset-typescript',
```
lychyi: Changed up the loader here, I'm not really sure of what all `react-app` does behind-the-scenes so I'm manually adding the presets/plugins needed for TS here.

.storybook/webpack.config.js
```diff
@@ -60,11 +78,14 @@ module.exports = ({config, mode}) => {
...
-        options: {plugins: ['@babel/plugin-transform-react-jsx']},
+        options: {
+          presets: [babelPresetEnvConfig],
+          plugins: ['@babel/plugin-transform-react-jsx']
+        },
       },
       {
         loader: '@mdx-js/loader',
-        options: {compilers: [createCompiler()]},
+        options: { compilers: [createCompiler()] },
```
lychyi: Ran "format document" in VS Code, don't hate me on this one 😬

modules/core/react/lib/InputProvider.tsx
```diff
@@ -1,4 +1,6 @@
 import * as React from 'react';
+// @ts-ignore
```
lychyi: There are no definition files for `element-closest`. We could technically add a `.d.ts` file and exclude it from compilation so we don't ship it. But honestly it's a function that takes a `window`-like object and returns void so I made the tradeoff to ignore it here.

Let me know if you think this should be different.

package.json
```diff
@@ -53,6 +53,7 @@
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2",
     "colors": "^1.4.0",
     "commitizen": "^3.1.1",
+    "core-js": "^3.5.0",
```
lychyi: This should be a dependency even though it's just part of a config for a preset in `babel-loader`.

https://babeljs.io/docs/en/babel-preset-env#usebuiltins

modules/core/react/lib/InputProvider.tsx
NicholasBoll: I'm fine with this

.storybook/webpack.config.js
NicholasBoll: I'm for decreasing the magic. Magic is useful for getting started, but as things get more complex, it makes sense to decrease magic.

Ideally our transforms are pretty simple/standard.

Issue #363, opened 2019-12-10
Fix: a11y - Lack of association for left label and checkbox 544


jpante:
## Bug Report

 Label is not announced in SR announcement. Reads "Checkbox option
check box not checked".

## Expected Behavior

 Label is announced. Reads "Label Checkbox option
check box not checked"

Issue #286, opened 2019-10-29
❗Component Testing Checklist


lychyi:
# Test Coverage Checklist

Reference to PR that removed snapshots (#259) for reference when re-writing tests that we blew away from this.

## Blockers
- ~~Storybook does not load into IE11~~
- ~~Storybook 5.2 gets merged (#267) for MDX - _Format for visual tests will be reconsidered after Storybook 5.2 gets merged_~~

## Guides
[Testing philosophy and guidelines](https://github.com/Workday/canvas-kit/blob/master/TESTING.md)

## React
For these components, we'll have to ensure we have adequate coverage for unit tests (Jest, React Testing Library) and specification/e2e tests (Cypress).

### Canvas Kit React
- [x] @workday/canvas-kit-react-action-bar (#1396)
- [x] @workday/canvas-kit-react-avatar (#614)
- [ ] @workday/canvas-kit-react-banner
- [x] @workday/canvas-kit-react-button (#468)
- [x] @workday/canvas-kit-react-card (#708)
- [x] @workday/canvas-kit-react-checkbox (#372)
- [x] @workday/canvas-kit-react-color-picker (#407)
- [ ] @workday/canvas-kit-react-common
- [ ] @workday/canvas-kit-react-core
- [ ] @workday/canvas-kit-react-fonts
- [x] @workday/canvas-kit-react-form-field (#482)
- [ ] @workday/canvas-kit-react-icon
- [x] @workday/canvas-kit-react-layout (#1027)
- [ ] @workday/canvas-kit-react-loading-animation
- [x] @workday/canvas-kit-react-modal (specs #184)
- [x] @workday/canvas-kit-react-popup (#600)
- [x] @workday/canvas-kit-react-radio (#381)
- [x] @workday/canvas-kit-react-select (#412)
- [x] @workday/canvas-kit-react-skeleton (#720)
- [x] @workday/canvas-kit-react-status-indicator (#692)
- [x] @workday/canvas-kit-react-switch (#386)
- [ ] @workday/canvas-kit-react-table
- [x] @workday/canvas-kit-react-text-area (#394)
- [x] @workday/canvas-kit-react-text-input (#390)
- [x] @workday/canvas-kit-react-toast (#584 )
- [x] @workday/canvas-kit-react-tooltip (#528)

### Canvas Kit Labs React
- [x] @workday/canvas-kit-labs-react-combobox (#535)
- [x] @workday/canvas-kit-labs-react-menu (#812)

[...]

PR #372, opened 2019-12-13
test(checkbox): Change tests to react-testing-library and improve coverage

anicholls:
## Summary

Closes https://github.com/Workday/canvas-kit/issues/363

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/cypress/integration/Checkbox.spec.ts
+++ b/cypress/integration/Checkbox.spec.ts
@@ -35,7 +35,7 @@ describe('Checkbox', () => {
       h.stories.load('Components|Inputs/Checkbox/React/Top Label', 'Disabled');
     });
 
-    it('should not have any axe errors', () => {
+    it('should pass accessibility checks', () => {
       cy.checkA11y();
     });
 
--- a/cypress/integration/Checkbox.spec.ts
+++ b/cypress/integration/Checkbox.spec.ts
@@ -1,0 +1,46 @@
+import * as h from '../helpers';
+
+const getCheckbox = () => {
+  return cy.get(`[type="checkbox"]`);
+};
+
+describe('Checkbox', () => {
+  before(() => {
+    h.stories.visit();
+  });
+  ['Default', 'Alert', 'Error'].forEach(story => {
+    context(`given the '${story}' story is rendered`, () => {
+      beforeEach(() => {
+        h.stories.load('Components|Inputs/Checkbox/React/Top Label', story);
+      });
+
+      it('should not have any axe errors', () => {
+        cy.checkA11y();
+      });
+
+      context('when clicked', () => {
+        beforeEach(() => {
+          getCheckbox().click();
+        });
+
+        it('should be checked', () => {
+          getCheckbox().should('be.checked');
+        });
+      });
+    });
+  });
+
+  context(`given the 'Disabled' story is rendered`, () => {
+    beforeEach(() => {
+      h.stories.load('Components|Inputs/Checkbox/React/Top Label', 'Disabled');
+    });
+
+    it('should not have any axe errors', () => {
+      cy.checkA11y();
+    });
+
+    it('should be disabled', () => {
+      getCheckbox().should('be.disabled');
+    });
+  });
+});
--- a/cypress/integration/Checkbox.spec.ts
+++ b/cypress/integration/Checkbox.spec.ts
@@ -14,7 +14,7 @@ describe('Checkbox', () => {
         h.stories.load('Components|Inputs/Checkbox/React/Top Label', story);
       });
 
-      it('should not have any axe errors', () => {
+      it('should pass accessibility checks', () => {
         cy.checkA11y();
       });
 
--- a/modules/checkbox/react/lib/Checkbox.tsx
+++ b/modules/checkbox/react/lib/Checkbox.tsx
@@ -43,29 +43,13 @@ const CheckboxContainer = styled('div')({
  * :hover on the checkbox when you hover on it's corresponding label.
  * This stops the ripple from showing when you hover on the label.
  */
-const CheckboxInputWrapper = styled('div')<Pick<CheckboxProps, 'disabled'>>(
-  {
-    display: 'flex',
-    height: checkboxHeight,
-    width: checkboxWidth,
-    marginTop: '3px',
-    alignSelf: 'flex-start',
-    '&::after': {
-      borderRadius: borderRadius.circle,
-      boxShadow: '0 0 0 0 ' + colors.soap200,
-      content: '""',
-      display: 'inline-block',
-      height: checkboxHeight,
-      transition: 'box-shadow 150ms ease-out',
-      width: checkboxWidth,
-    },
-  },
-  ({disabled}) => ({
-    '&:hover::after': {
-      boxShadow: disabled ? undefined : '0 0 0 ' + rippleRadius + 'px ' + colors.soap200,
-    },
-  })
-);
+const CheckboxInputWrapper = styled('div')<Pick<CheckboxProps, 'disabled'>>({
+  display: 'flex',
+  height: checkboxHeight,
+  width: checkboxWidth,
+  marginTop: '3px',
+  alignSelf: 'flex-start',
+});
 
 /**
  * Note: `~ div:first-of-type` refers to `CheckboxBackground`
@@ -110,7 +94,6 @@ const CheckboxInput = styled('input')<CheckboxProps>(
     '&:focus ~ div:first-of-type': {
       borderColor: colors.blueberry400,
       borderWidth: '2px',
-      zIndex: 2,
       boxShadow: 'none',
     },
     '&:checked:focus ~ div:first-of-type': {
@@ -136,6 +119,26 @@ const CheckboxInput = styled('input')<CheckboxProps>(
       },
     }),
   },
+
+  // Ripple
+  {
+    '& ~ div:first-of-type::after': {
+      borderRadius: borderRadius.circle,
+      boxShadow: `0 0 0 0 ${colors.soap200}`,
+      content: '""',
+      display: 'inline-block',
+      height: checkboxHeight,
+      transition: 'box-shadow 150ms ease-out',
+      width: checkboxWidth,
+      position: 'absolute',
+      zIndex: -1,
+    },
+  },
+  ({disabled}) => ({
+    '&:hover ~ div:first-of-type::after': {
+      boxShadow: disabled ? undefined : `0 0 0 ${rippleRadius}px ${colors.soap200}`,
+    },
+  }),
   ({error}) => {
     let errorRingColor;
     let errorRingBorderColor = 'transparent';
--- a/modules/checkbox/react/lib/Checkbox.tsx
+++ b/modules/checkbox/react/lib/Checkbox.tsx
@@ -13,7 +13,7 @@ import {checkSmallIcon} from '@workday/canvas-system-icons-web';
 import uuid from 'uuid/v4';
 
 export interface CheckboxProps extends Themeable, React.InputHTMLAttributes<HTMLInputElement> {
-  checked: boolean;
+  checked?: boolean;
   disabled?: boolean;
   id?: string;
   inputRef?: React.Ref<HTMLInputElement>;
@@ -237,7 +237,6 @@ export default class Checkbox extends React.Component<CheckboxProps> {
   static ErrorType = ErrorType;
 
   public static defaultProps = {
-    checked: false,
     label: '',
   };
 
@@ -247,6 +246,7 @@ export default class Checkbox extends React.Component<CheckboxProps> {
     // TODO: Standardize on prop spread location (see #150)
     const {
       checked,
+      defaultChecked,
       disabled,
       id = this.id,
       inputRef,
@@ -258,11 +258,13 @@ export default class Checkbox extends React.Component<CheckboxProps> {
       ...elemProps
     } = this.props;
 
+    const isChecked = checked || (checked === undefined && defaultChecked) || false;
+
     return (
       <CheckboxContainer>
         <CheckboxInputWrapper disabled={disabled}>
           <CheckboxInput
-            checked={checked}
+            checked={isChecked}
             disabled={disabled}
             id={id}
             ref={inputRef}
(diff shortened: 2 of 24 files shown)


modules/checkbox/react/lib/Checkbox.tsx
```diff
@@ -237,7 +237,6 @@ export default class Checkbox extends React.Component<CheckboxProps> {
   static ErrorType = ErrorType;
 
   public static defaultProps = {
-    checked: false,
```
anicholls: Interested in opinions on this. To properly support `defaultChecked`, I had to remove `checked` as a default prop so that it is initially `undefined`. Otherwise, `defaultChecked` would always override `checked` whenever it is false.

modules/checkbox/react/spec/Checkbox.spec.tsx
```diff
@@ -1,102 +1,134 @@
...
 
-  test('render an checked checkbox input', () => {
-    const component = mount(<Checkbox checked={true} onChange={cb} />);
-    expect(component.find('input').props().checked).toBe(true);
-    component.unmount();
-  });
+    describe('with checked=true', () => {
+      test('should render a checked checkbox input', () => {
+        const {getByLabelText} = render(<Checkbox checked={true} onChange={cb} label={label} />);
+        expect(getByLabelText(label)).toHaveProperty('checked', true);
```
anicholls: Would love to use the new `toHaveChecked` matcher from `@testing-library/jest-dom`, but there seem to be type issues within Emotion. We can update these when we update to Emotion 11 (see https://github.com/emotion-js/emotion/issues/1689 for more info)

modules/checkbox/react/spec/Checkbox.spec.tsx
```diff
@@ -1,102 +1,134 @@
...
   });
 
-  test('render an checkbox input with id', () => {
-    const component = mount(<Checkbox id="myCheckbox" onChange={cb} />);
-    expect(component.find('input').props().id).toBe('myCheckbox');
-    component.unmount();
-  });
+  describe('when rendered', () => {
+    describe('with an id', () => {
+      test('should render a checkbox input with id', () => {
```
anicholls: I restructured these `describe`/`test` functions to be behavior based and read more like specs (similar to how we do our Cypress tests). Hopefully everyone likes this!

modules/checkbox/react/spec/Checkbox.spec.tsx
```diff
@@ -1,102 +1,134 @@
...
+          <FormField label="My Field" inputId="my-checkbox-field">
+            <Checkbox disabled={false} checked={true} id="my-checkbox-field" onChange={jest.fn()} />
+          </FormField>
+        );
+        expect(await axe(html)).toHaveNoViolations();
+      });
+    });
+
+    describe('without an id', () => {
+      test('should create a unique id for each instance', async () => {
```
anicholls: Check for the same unique ID if rerendered (https://github.com/Workday/canvas-kit/pull/386/files#diff-c9977ec716fc27e28a4ea0ff50add633R66)

modules/checkbox/react/lib/Checkbox.tsx
mannycarrera4: @anicholls do we want to have the same behavior for radio as well?

anicholls: Yeah - we should. Maybe this should be done in a separate PR. Thoughts @lychyi?

modules/checkbox/react/spec/Checkbox.spec.tsx
```diff
@@ -11,92 +11,118 @@ describe('Checkbox', () => {
...
-  test('Checkbox should spread extra props', () => {
-    const component = mount(<Checkbox data-propspread="test" onChange={cb} />);
-    const input = component
-      .find('input') // TODO: Standardize on prop spread location (see #150)
-      .getDOMNode();
-    expect(input.getAttribute('data-propspread')).toBe('test');
-    component.unmount();
-  });
-});
+    test('should pass axe DOM accessibility guidelines', async () => {
```
anicholls: Move a11y tests to Cypress!

modules/checkbox/react/spec/Checkbox.spec.tsx
NicholasBoll: 👍 My only nit is using `it` instead of `test` for specifications

modules/checkbox/react/lib/Checkbox.tsx
anicholls: Ended up removing this! Ignore 🙂

cypress/integration/Checkbox.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+      });
+    });
+  });
+
+  context(`given the 'Disabled' story is rendered`, () => {
+    beforeEach(() => {
+      h.stories.load('Components|Inputs/Checkbox/React/Top Label', 'Disabled');
+    });
+
+    it('should not have any axe errors', () => {
```
stephanerangaya: Why not write something like `it('should pass accessibility tests')`?

modules/form-field/react/stories/stories_Checkbox.tsx
```diff
@@ -109,6 +125,69 @@ storiesOf('Components|Inputs/Checkbox/React/Left Label', module)
...
     >
-      {control(<Checkbox id="1" label="Checkbox option" />)}
+      {control(<Checkbox label="Checkbox option" />)}
     </FormField>
   ));
+
+storiesOf('Components|Inputs/Checkbox/React', module)
+  .addParameters({component: Checkbox})
+  .addDecorator(withReadme(README))
+  .add('States', () => (
```
stephanerangaya: We're not putting states in a visual testing folder?

modules/form-field/react/stories/stories_Checkbox.tsx
```diff
@@ -160,24 +160,26 @@ storiesOf('Components|Inputs/Checkbox/React', module)
                     return;
                   }
 
+                  const key = `${checked ? 'checked' : 'unchecked'}-${
```
anicholls: Having some problems with key errors in the story - can't figure out why. Getting `react.development.js:188 Warning: Each child in a list should have a unique "key" prop.`. However, I've programmatically checked all of the keys to ensure there are no duplicates. Anyone see something I'm missing?

stephanerangaya: It looks like each children in the return needs to have a key. Something like this removes the error:
```
return (
  <tr key={`${key}-tr`}>
    <td key={`${key}-td`}>{`${disabled ? 'Disabled ' : ''}${variant} (${
      checked ? (indeterminate ? 'indeterminate' : 'checked') : 'unchecked'
      })`}</td>

    {['', 'hover', 'focus', 'active'].map(className => (
      <td
        key={`${key}-${className}-td`}>
        <Checkbox
          checked={checked}
          disabled={disabled}
          indeterminate={indeterminate}
          error={type}
          className={className}
          key={`${key}-${className}-checkbox`}
          onChange={() => { }} // eslint-disable-line no-empty-function
        />
      </td>
    ))}
  </tr>
);
```

anicholls: Interesting... I thought it was only the top level one for some reason.

modules/checkbox/react/spec/Checkbox.spec.tsx
```diff
@@ -1,102 +1,104 @@
...
+      expect(getByRole('checkbox')).toHaveProperty('checked');
+      expect(getByRole('checkbox')).toHaveProperty('id', uniqueId);
+    });
   });
 
-  test('Checkbox without a defined id should pass axe DOM accessibility guidelines', async () => {
-    const html = ReactDOMServer.renderToString(<Checkbox label={'Label'} onChange={jest.fn()} />);
-    expect(await axe(html)).toHaveNoViolations();
+  describe('when rendered with extra, arbitrary props', () => {
+    it('should spread extra props', () => {
```
lychyi: Thinking about when it fails, it may help to indicate where it spreads to
```suggestion
    it('should spread extra props onto the checkbox', () => {
```

modules/form-field/react/stories/stories_Checkbox.tsx
anicholls: @stephanerangaya It is indeed only needed at the top level - I was just silly and had the `key` on the `Checkbox` instead of its containing `td`.

stephanerangaya: Oh, I see!

modules/checkbox/react/spec/Checkbox.spec.tsx
```diff
@@ -1,102 +1,104 @@
...
-      <FormField label="My Field" inputId="my-checkbox-field">
-        <Checkbox disabled={false} checked={true} id="my-checkbox-field" onChange={jest.fn()} />;
-      </FormField>
-    );
-    expect(await axe(html)).toHaveNoViolations();
+  describe('when rendered with an input ref', () => {
+    it('input ref should be defined', () => {
+      const ref = React.createRef<HTMLInputElement>();
+      render(<Checkbox inputRef={ref} onChange={cb} />);
+      expect(ref.current).toBeDefined();
```
lychyi: While, this test does verify the spec `'input ref should be defined'`, shouldn't there be another spec (or adjust the wording on this one) that verifies whether or not `current` refers to the checkbox input itself? Something like `expect(inputRef.current).toHaveAttribute('role', 'checkbox');`

lychyi: After going through this same thing in #386 this will always pass since React sets the value to `null`. It may be good to have two assertions: 1) the current property is not null and 2) the current property refers to the checkbox

modules/form-field/react/stories/stories_Checkbox.tsx
```diff
@@ -109,6 +125,70 @@ storiesOf('Components|Inputs/Checkbox/React/Left Label', module)
...
+          <tr>
+            <th></th>
+            <th>No state</th>
+            <th>Hover</th>
+            <th>Focused</th>
+            <th>Active</th>
+          </tr>
+        </thead>
+        <tbody>
+          {[false, true].map(checked => {
```
NicholasBoll: Should we align on what states we show? For instance, “disabled” is actually a state that is mutually exclusive to “hover” and “focus”. You cannot focus or hover over a disabled checkbox, so we shouldn’t have a visual state that shows an invalid state in reality.

NicholasBoll:
To make the visual work with the className trick would require changing the Checkbox.tsx file a bit to apply the className to both the wrapper and the container. My main question there is “should we really prevent the hover circle from showing if we are actively fighting a browser standard to do so?”

If yes, we could allow the hover circle when the label is hovered.

If no, maybe we should adjust the implementation of the Checkbox.tsx file slightly to allow the "hover" state to show correctly on our visual testing

anicholls:
@NicholasBoll I've addressed both of your comments and am waiting to hear back from the design team on whether tying the ripple to the label is okay 👍

utils/storybook/StaticStatesTable.tsx
```diff
@@ -0,0 +1,143 @@
...
+    paddingRight: 16,
+    textAlign: 'left',
+  },
+});
+
+export default class StaticStatesTable extends React.Component<StaticStatesTableProps> {
+  public static defaultProps = {
+    componentProps: {},
+    states: [
+      '',
```
mannycarrera4: should we just add in default? or do we want empty here?

anicholls: Good point - I'm rewriting this to `Default` anyways. Will update.

utils/storybook/StaticStatesTable.tsx
```diff
@@ -0,0 +1,151 @@
...
+   * the ability to add extra styling or markup (a blue background for an inverse variant, for example).
+   */
+  renderComponent(props: ComponentProps, disabled: boolean, className: string): React.ReactNode;
+  /**
+   * The validation function called to check the validity of the current prop combination.
+   * Return false if you would like to skip this prop combination (table row).
+   * Return true otherwise.
+   * @default (props) => true
+   */
+  verifyProps(props: ComponentProps): boolean;
```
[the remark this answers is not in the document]
NicholasBoll: Perhaps `shouldRenderRow` (or perhaps `shouldRenderCell`)?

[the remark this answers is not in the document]
anicholls: This does the whole row. Any checks for the specific cell would happen within `renderComponent`. Happy to add `Row` to the end 👍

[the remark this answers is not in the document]
NicholasBoll: Ah. I see now. It is creating permutations of prop combinations which are rows.

utils/storybook/StaticStatesTable.tsx
```diff
@@ -0,0 +1,151 @@
...
+  [key: string]: any;
+}
+
+/**
+ * A helper to generate a table of all possible states for component visual testing.
+ *
+ * This uses our `StaticStates` system. Note: This system requires the use of Canvas'
+ * `styled` wrapper within the component, instead of the default emotion one.
+ */
+export interface StaticStatesTableProps {
```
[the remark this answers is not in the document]
NicholasBoll: I think more generically it would be a matrix of x props and y props. As it is now, "Props" represent a row and "state" (the className for the psuedo-selector state) are the columns.

More generically it is a matrix of prop combinations since `className` is really a prop.

I'm not sure I like the implementation as a permutation of props that has to be filtered. What about this instead?

```tsx
<ComponentPropsTable
  rowProps={[
    { label: 'Checked', checked: true, indeterminate: false },
    { label: 'Indeterminate', checked: false, indeterminate: true },
    // etc
  ]}
  columnProps={[
    { label: 'Focus', className: 'focus' },
    { label: 'Hover', className: 'hover' },
    // etc
  ]}
>
  (props) => <Checkbox {...props} />
</ComponentPropsTable>
```

You can always make some utility functions that will give you the `rowProps` array from permutations

anicholls:
Based on conversation, we're going to address the follow up changes to `StaticStatesTable` in another PR!

PR #374, opened 2019-12-13
test: Fix modal specs

NicholasBoll:
## Summary

Tests are failing on CI, but not locally. I'm making a PR to see what it takes to fix on the CI.

There is some strange issue where the "tab" key is hit before the event listener is added. There are some issues on `cypress-tab-plugin` that causes errors. On investigating, I discovered this issue: https://github.com/Bkucera/cypress-plugin-tab/issues/13

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -18,7 +18,7 @@ Cypress.Commands.add('injectAxe', () => {
 
 // Add better logging to cy.tab
 Cypress.Commands.overwrite('tab', (originalFn, subject) => {
-  const prevSubject = cy.$$(subject || cy.state('window').document.activeElement);
+  const prevSubject = cy.$$(subject || (cy as any).state('window').document.activeElement);
 
   const log = Cypress.log({
     $el: prevSubject,
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -34,7 +34,7 @@ Cypress.Commands.overwrite('tab', (originalFn, subject) => {
   return Cypress.Promise.try(() => originalFn(subject))
     .then(value => {
       log.set('$el', value).snapshot();
-      return value;
+      return Cypress.$(value);
     })
     .finally(() => {
       log.end();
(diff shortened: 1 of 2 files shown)


cypress/support/commands.ts
```diff
@@ -34,7 +34,7 @@ Cypress.Commands.overwrite('tab', (originalFn, subject) => {
   return Cypress.Promise.try(() => originalFn(subject))
     .then(value => {
       log.set('$el', value).snapshot();
-      return value;
+      return Cypress.$(value);
```
NicholasBoll: This isn't a strictly necessary change, but sometimes the value that comes back is a DOM element instead of a jQuery-wrapped element. This happens when the focus trap cancels the event. The plugin returns `document.activeElement` instead of `$(document.activeElement)`. This line normalizes the return value.

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -98,7 +98,8 @@ function getFirstElementToFocus(modalEl: HTMLElement): HTMLElement {
 }
 
 const useKeyDownListener = (handleKeydown: EventListenerOrEventListenerObject) => {
-  React.useEffect(() => {
+  // `useLayoutEffect` for automation
+  React.useLayoutEffect(() => {
```
NicholasBoll: This was difficult to reproduce. I had to add a delay to when the event listener was added. Running from the command line saw failures sooner than running Cypress through the GUI runner. Since Travis is so slow, it happened earlier. My guess is Travis is getting slower.

`useLayoutEffect` isn't strictly necessary for production because it probably isn't possible for a user to hit a key fast enough and if they did, they could just hit the key again. Automation is fast enough though. We could either slow down the automation or not allow a frame to render without a keybinding.

`useLayoutEffect` is there to prevent frames from rendering between side effects. The documentation suggests not having any UI-blocking work here. Adding an event listener is very fast in browsers and fast-enough in IE11.

cypress/support/commands.ts
```diff
@@ -18,7 +18,7 @@ Cypress.Commands.add('injectAxe', () => {
 
 // Add better logging to cy.tab
 Cypress.Commands.overwrite('tab', (originalFn, subject) => {
-  const prevSubject = cy.$$(subject || cy.state('window').document.activeElement);
+  const prevSubject = cy.$$(subject || (cy as any).state('window').document.activeElement);
```
NicholasBoll: I never got around to adding `state` to the Cypress interface

modules/modal/react/lib/ModalContent.tsx
mannycarrera4: I like this solution rather than implicitly slowing down the automation.

Issue #375, opened 2019-12-16
Improve CanvasProvider to support optional InputProvider and add new utility to styled function


mannycarrera4:
## 🚀 Feature Proposal
While working on focus states for icon button, we realized it was valuable to see all states in a table rather than manually testing each individual button. A few things came out of that work that I think would provide support for other components.

1. There might be some cases where we want to disable InputProvider for a story, so rewriting CanvasProvider so that we can optionally disable InputProvider will help with this

2. Extract out a utils function that will allow us to run through a styled function and be able to replace any pseudo selectors with classes so that we can force a state on an element to be able to see it's different states such as: Focus, Hover, Active...

By adding these two things, we'll be able to test visual regressions on different states without worrying about InputProvider and not affecting any components to the consumer.

PR #377, opened 2019-12-16
chore: Add StaticStates wrapper

mannycarrera4:
Fixes: https://github.com/Workday/canvas-kit/issues/375

This is separating the work in https://github.com/Workday/canvas-kit/pull/271

This creates a wrapper for stories to allow for treating pseudo-selectors like `:hover` or `:focus` as static states that are controllable as classes rather than the default pseudo-selectors.

For example, this is now a valid story:

```tsx
.add('States', () => {
  <StaticStates>
    <Button className="focus">Focus</Button>
  </StaticStates>
});
```

This code will render a button with focus styling always applied. This is useful for enumerating states for design or validating states in visual regression tests.

To enable this functionality, Components must use the Canvas-kit `styled` function instead of the one from Emotion. The canvas-kit `styled` function also seamlessly handles a fallback to default theme and RTL support out-of-the-box.


modules/button/react/stories/stories_iconButton.tsx
```diff
@@ -108,6 +108,7 @@ storiesOf('Components|Buttons/Button/React/Icon Button', module)
   .add('Circle', () => (
     <div className="story">
       <h3>Medium Default</h3>
+
```
mannycarrera4: remove this

modules/common/react/lib/utils/changeToStaticClass.ts
```diff
@@ -0,0 +1,28 @@
...
+      .replace(/^:/, '&:')
+      .replace(/:/g, '.')
+      .replace(/\.not/g, ':not')
+      .replace(/\.disabled/, ':disabled');
+    const value =
+      typeof obj[key] === 'object' ? changeToStaticClass(true, obj[key] as CSSObject) : obj[key];
+    const newObj = {...result, [newKey]: value};
+    return newObj;
+  }, {});
+};
```
mannycarrera4: Looking for feedback @NicholasBoll @anicholls, Nicholas was mentioning that we can maybe use context when we refactor this. Currently with this implementation we need to add a hidden prop to the component to handle the `shouldChangePseudoSelectors`

NicholasBoll: It is unfortunate the current implementation requires wrapping of each styled object/function. This is a bigger problem on button due to all the objects/functions defined in the `styled` factory function.

utils/storybook/CanvasProviderDecorator.tsx
```diff
@@ -5,11 +5,21 @@ import {
...
 
 const label = 'theme';
 
-export default (storyFn: () => React.ReactNode) => (
-  <CanvasProvider theme={createCanvasTheme(object(label, defaultCanvasTheme))}>
-    {storyFn()}
-  </CanvasProvider>
-);
+export default makeDecorator({
+  name: 'withCanvasProviderDecorator',
```
anicholls: Nit: What is the `with` here for?

utils/storybook/CanvasProviderDecorator.tsx
```diff
@@ -5,11 +5,21 @@ import {
...
+  parameterName: 'canvasProviderDecorator',
+  wrapper: (storyFn, context, {parameters}) => {
+    return (
+      <CanvasProvider
+        includeInputProvider={parameters ? parameters.disabled : undefined}
```
anicholls: We should change the name of this parameter to match the name of the `includeInputProvider` prop. If I use `.addParameters({ canvasProviderDecorator: { disabled: true }})`, I wouldn't expect the `CanvasProvider` to render at all.

You can also simplify this check:
```tsx
  wrapper: (storyFn, context, {parameters = {}}) => {
);	    return (
      <CanvasProvider
        includeInputProvider={parameters.includeInputProvider}
```
(note the default in the `parameters` instantiation)

modules/_labs/core/react/lib/CanvasProvider.tsx
```diff
@@ -7,6 +7,7 @@ import styled from '@emotion/styled';
 
 export interface CanvasProviderProps {
   theme: CanvasTheme;
+  includeInputProvider: boolean;
```
anicholls: Nit: I would maybe consider changing this to `disableInputProvider`, since that's generally how it's going to be used. For example, I wouldn't set the `disabled` attribute on an input component using an `enabled={false}` prop (that normally defaults to true)

modules/common/react/lib/utils/changeToStaticClass.ts
```diff
@@ -0,0 +1,28 @@
+import {CSSObject} from '@emotion/core';
+
+/**
+ * Allows the ability to force a state by a pseudo selector by changing the selector into a class creating a new styled object with the class names
+ * @param shouldChangePseudoSelectors If true, it will replace pseudo selectors from the styled object and change them to classes
+ * @param obj The styled object passed to the function to replace any pseudo selectors with a class name
+ */
+
+export const changeToStaticClass = (
```
anicholls: Can you add a test for this?

modules/common/react/lib/utils/changeToStaticClass.ts
anicholls: @NicholasBoll @mannycarrera4 Wouldn't working it into https://github.com/Workday/canvas-kit/blob/master/modules/_labs/core/react/lib/theming/styled.ts address both of these concerns?

I realize we have the issue with infinitely nested style objects, but that's easily addressable in `styled.

modules/_labs/core/react/lib/CanvasProvider.tsx
anicholls: Can you add this prop to the readme?

utils/storybook/CanvasProviderDecorator.tsx
mannycarrera4: the documentation suggesting naming things like, will fix!

utils/storybook/CanvasProviderDecorator.tsx
```diff
@@ -5,11 +5,21 @@ import {
...
-    {storyFn()}
-  </CanvasProvider>
-);
+export default makeDecorator({
+  name: 'canvasProviderDecorator',
+  parameterName: 'canvasProviderDecorator',
+  wrapper: (storyFn, context, {parameters}) => {
+    return (
+      <CanvasProvider
+        disableInputProvider={parameters ? parameters.disableInputProvider : undefined}
```
mannycarrera4: @anicholls I have to check for `parameters` or else every other story fails, the only way to by pass this is if we add the option of `skipIfNoParametersOrOptions` per: https://storybook.js.org/docs/addons/api/ However, if we add this option, canvas provider would never run, this was the only work around I could come up with. I'm open to other options too!

anicholls:
Thoughts on loading `Default` stories first and `States` stories last by updating [`storySort`](https://github.com/Workday/canvas-kit/blob/master/.storybook/config.js#L41) as part of this PR?

mannycarrera4:
@anicholls that makes sense to me. In the icon button PR, i had the states as a last story after seeing all the defaults!

modules/common/react/lib/utils/changeToStaticStates.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Allows the ability to force a state by a pseudo selector by changing the selector into a class creating a new styled object with the class names
+ * @param shouldChangePseudoSelectors If true, it will replace pseudo selectors from the styled object and change them to classes
+ * @param obj The styled object passed to the function to replace any pseudo selectors with a class name
+ */
+
+export interface ChangeStaticStateProps {
+  /** @ignore */
+  shouldChangeStyleToStaticStates?: boolean;
```
anicholls: I wonder if this should be prefixed so that if anyone sees it in the codebase, it's immediately clear that it's an internal prop. I feel less bad if we make any breaking changes to it in that case haha.

Maybe something as simple as an underscore (`_shouldConvertPseudoStatesToClasses`)?

modules/common/react/lib/utils/changeToStaticStates.ts
```diff
@@ -0,0 +1,62 @@
...
+import {WithTheme} from '@emotion/styled';
+import {Overwrapped} from '@emotion/styled-base/types/helper';
+
+/**
+ * Allows the ability to force a state by a pseudo selector by changing the selector into a class creating a new styled object with the class names
+ * @param shouldChangePseudoSelectors If true, it will replace pseudo selectors from the styled object and change them to classes
+ * @param obj The styled object passed to the function to replace any pseudo selectors with a class name
+ */
+
+export interface ChangeStaticStateProps {
```
anicholls: I kind of like the "behavior" naming approach of `GrowthBehavior`. What if we want with `ConvertPseudoStatesBehavior`?

modules/button/react/stories/stories_iconButton.tsx
```diff
@@ -103,7 +103,9 @@ export class IconButtonToggleGroupWrapper extends React.Component<
 }
 
 storiesOf('Components|Buttons/Button/React/Icon Button', module)
-  .addParameters({component: IconButton})
+  .addParameters({
```
anicholls: Can we revert this?

modules/common/react/lib/utils/changeToStaticStates.ts
```diff
@@ -0,0 +1,62 @@
...
+export const convertToStaticStates = <
+  InnerProps,
+  ExtraProps extends ChangeStaticStateProps,
+  StyleProps extends Omit<
+    Overwrapped<InnerProps, StyleProps>,
+    keyof React.ComponentClass<any>
+  > = Omit<InnerProps & ExtraProps, keyof React.ComponentClass<any>>,
+  Theme extends object = object
+>(
+  ...styles: Array<Interpolation<WithTheme<StyleProps, Theme>>>
```
anicholls: Nit: Can we call these `interpolations` since that's what they are?

modules/common/react/lib/utils/changeToStaticStates.ts
```diff
@@ -0,0 +1,62 @@
...
+    const value =
+      typeof obj[key] === 'object'
+        ? changeStyleToStaticStates(true, obj[key] as CSSObject)
+        : obj[key];
+    const newObj = {...result, [newKey]: value};
+    return newObj;
+  }, {});
+};
+
+export const convertToStaticStates = <
```
anicholls: What do you think about the name `convertPseudoStatesToClasses`? I'm not sure about the term "static states"

modules/_labs/core/react/README.md
```diff
@@ -111,6 +111,16 @@ import {CanvasProvider} from '@workday/canvas-kit-react';
 <CanvasProvider>{/* All your components containing any Canvas components */}</CanvasProvider>;
 ```
 
+If you'd like to disable `InputProvider` you can pass `disableInputProvider` which will disable InputProvider for all your components that are being wrapped by CanvasProvider
+
+```tsx
+import {CanvasProvider} from '@workday/canvas-kit-react';
+
+<CanvasProvider disabledInputProvider={true}>{/* All your components containing any Canvas components */}</CanvasProvider>;
```
anicholls:
```suggestion
<CanvasProvider disableInputProvider={true}>{/* All your components containing any Canvas components */}</CanvasProvider>;
```

modules/button/react/lib/Button.tsx
```diff
@@ -4,7 +4,7 @@ import * as React from 'react';
 import {ButtonBaseCon, ButtonBaseLabel, ButtonLabelData, ButtonLabelIcon} from './ButtonBase';
 import {DeprecatedButtonVariant, ButtonSize, ButtonVariant} from './types';
 import {CanvasSystemIcon} from '@workday/design-assets-types';
-import {GrowthBehavior} from '@workday/canvas-kit-react-common';
+import {GrowthBehavior, ChangeStaticStateProps} from '@workday/canvas-kit-react-common';
```
anicholls: Can we revert this file if we're removing the tests?

modules/button/react/stories/stories.tsx
```diff
@@ -46,63 +46,6 @@ storiesOf('Components|Buttons/Button/React', module)
       <Button size={Button.Size.Large} variant={Button.Variant.Primary}>
         Primary
       </Button>
-      <Button size={Button.Size.Large} variant={Button.Variant.Primary} icon={activityStreamIcon}>
```
anicholls: How come all of these examples were removed?

modules/common/react/spec/changeToStaticStates.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  it('should convert ":active" to "&.active"', () => {
+    const input = {':active': {display: 'none'}};
+    const expected = {
+      '&.active': {
+        display: 'none',
+      },
+    };
+    expect(changeStyleToStaticStates(true, input)).toEqual(expected);
+  });
+  it('should convert ":focus" to "&.focus"', () => {
```
anicholls: Can you add tests for converting from `&:state` as well?

modules/common/react/lib/utils/changeToStaticStates.ts
mannycarrera4: since it's such an internal thing, I'm ok with this!

.storybook/config.js
```diff
@@ -39,13 +39,16 @@ const prefix = (phrase, prefix) => value => (value.indexOf(phrase) > -1 ? prefix
...
 
 function storySort(a, b) {
+  // console.warn('b', b);
   const prefixFn = pipe(
     prefix('welcome-', '0'),
     prefix('getting-started', '0'),
     prefix('tokens-', '1'),
-    prefix('components-', '2'),
+    prefix('components-', '3'),
+    prefix('states', '4'),
```
anicholls: This works, but I think we may want to do this separately from this calculation. This is specifically meant to target the prefix from the categories at the top level of storybook. I believe this could result in some unexpected edge cases for something like a story named `Tokens|Foo/Bar/States`. I would suggest doing a separate check after the number prefix has already been added.

.storybook/config.js
```diff
@@ -39,13 +39,16 @@ const prefix = (phrase, prefix) => value => (value.indexOf(phrase) > -1 ? prefix
 const pipe = (...fns) => value => fns.reduce((result, fn) => fn(result), value);
 
 function storySort(a, b) {
+  // console.warn('b', b);
   const prefixFn = pipe(
     prefix('welcome-', '0'),
     prefix('getting-started', '0'),
     prefix('tokens-', '1'),
-    prefix('components-', '2'),
+    prefix('components-', '3'),
```
anicholls: Is it necessary to change this one? This may be related to the edge cases mentioned in my other comment

.storybook/config.js
```diff
@@ -39,13 +39,16 @@ const prefix = (phrase, prefix) => value => (value.indexOf(phrase) > -1 ? prefix
 const pipe = (...fns) => value => fns.reduce((result, fn) => fn(result), value);
 
 function storySort(a, b) {
+  // console.warn('b', b);
```
anicholls:
```suggestion
```

modules/common/react/lib/utils/convertPseudoStatesBehavior.ts
```diff
@@ -0,0 +1,62 @@
...
+import {WithTheme} from '@emotion/styled';
+import {Overwrapped} from '@emotion/styled-base/types/helper';
+
+/**
+ * Allows the ability to force a state by a pseudo selector by changing the selector into a class creating a new styled object with the class names
+ * @param shouldChangePseudoSelectors If true, it will replace pseudo selectors from the styled object and change them to classes
+ * @param obj The styled object passed to the function to replace any pseudo selectors with a class name
+ */
+
+export interface ConvertPseudoStatesBehaviorProps {
```
anicholls: Sorry, I meant without the `Props` at the end. I think we generally only add `Props` when it's an interface for a components whole set of props (i.e. `Themeable` and `GrowthBehavior` don't have this)

mannycarrera4: ahh i si, i think i just got use to naming them props, will rename!

anicholls:
I just realized that we have the same problem with this that we did when we tried to add an `dir` prop to components for RTL. It worked in the example case you tried (buttons) because we only have states on the button element and we spread extra props on that element. But for components with pseudo states in multiple nested styled components, we need to ensure the `_shouldConvertPseudoStatesToClasses` prop is provided for every styled component.

@NicholasBoll I think this is probably okay to go in as is, but we really need to look at re-writing the `styled` factory so we can inject our own props like this. If we don't, it's going to add a lot of noise/complexity to our components very quickly.

.storybook/config.js
NicholasBoll: This function replaced another function that only handled top-level categories. The input of that function was something like `['Welcome', 'Tokens', 'Components']`, but I also needed to make sure "Getting Started" was the first item under "Welcome" so that Storybook picked it as the default page. Numbers were chosen because sorting alphabetically ensures numbers rank higher than letters, so '0first' will come before 'first'. Also numbers are easy for us to understand in terms of "ranking".

This function is definitely a hack, but the intent is to be powerful enough to handle sorting on a global level while still being easy enough to add new sorting rules. I actually think adding `prefix('states', '3')` _before_ the `prefix('components-', '3')` would accomplish the same goal without introducing a new sorting hierarchy ('3' would be prefixed for having 'states' in the storyname first, then '3' would be prefixed first for being in the "Components" category).

NicholasBoll: Actually, I retract the `3`. It has to be a `4` to not be the first story in `Components`... Perhaps not very intuitive

NicholasBoll:
> I just realized that we have the same problem with this that we did when we tried to add a `dir` prop to components for RTL. It worked in the example case you tried (buttons) because we only have states on the button element and we spread extra props on that element. But for components with pseudo states in multiple nested styled components, we need to ensure the `_shouldConvertPseudoStatesToClasses` prop is provided for every styled component.
> 
> @NicholasBoll I think this is probably okay to go in as is, but we really need to look at re-writing the `styled` factory so we can inject our own props like this. If we don't, it's going to add a lot of noise/complexity to our components very quickly.

Agreed. I think the ideal implementation is using a Context where a Provider (we originally called it `StaticState`) when wrapping a component will provide a context value that initiates the static state functionality and put it in the `StyledComponent` component. This way, the functionality is easy to initiate from the documentation in a way that each component doesn't have to know anything about.

We tried it in the `styled.ts` factory function, but Context is only available to components. We'd either have to create a new wrapper for every component or rewrite the `styled` function from `emotion` to add this functionality to the StyledComponent component. I didn't like the idea of creating an extra component wrapper around every styled component and it also sounds like more maintenance for our own version of `styled`. 

For a better developer experience, I think we'd take on the maintenance burden of maintaining our own styled wrapper to make it seamless.

NicholasBoll:
After talking with @anicholls, it is possible to use our `styled` from the labs/themes as long as we use the `theme` context which is provided to our interpolations. We are already doing this with RTL, so we're just tacking on.

I opted out of having the boolean parameter part of `CanvasTheme`, so it is a private boolean.

.storybook/config.js
```diff
@@ -35,19 +35,26 @@ addDecorator(FontsDecorator);
 addDecorator(CanvasProviderDecorator);
 
 /** If the string contains a phrase, prefix it. This is useful for making ordering sections */
-const prefix = (phrase, prefix) => value => (value.indexOf(phrase) > -1 ? prefix + value : value);
+const prefix = (phrase, prefix) => (/** @type {string} */ value) => {
+  const index = value.indexOf(phrase);
+  return index > -1 ? value.substr(0, index) + prefix + value.substr(index) : value;
```
NicholasBoll: This change allows for multi-tier sorting instead of a global sorting. For example, "Welcome" is the category and "getting-started" is the item. The below will produce a string like:

"0welcome--agetting-started"

Since the "a" is now added before "getting-started" instead of before "welcome", it allows sorting to be local to other items. Lowercase letters should be used in multi-tier sorting so it flows naturally with other items without disrupting them. For example, numbers are always before lowercase letters in ASCII. So even if you add a '9', it will sort first rather than the desired last which is why `z` is used below

modules/_labs/core/react/lib/theming/styled.ts
```diff
@@ -7,21 +7,40 @@ const noop = (styles: any) => styles;
...
+  if (!obj) {
+    return obj;
+  }
+
+  return Object.keys(obj).reduce((result, key) => {
+    const newKey = key
+      .replace(/^:/, '&:') // handle shorthand like ":focus"
+      .replace(/,(\s+):/g, ',$1&:') // handle selectors like ":focus, :hover"
+      .replace(/:(focus|hover|active)/g, '.$1')
+      .replace(/\[data\-whatinput="?(mouse|touch|keyboard|pointer)"?]/g, '[data-whatinput="noop"]');
```
NicholasBoll: Disables the `InputProvider` style overrides

modules/_labs/core/react/lib/theming/styled.ts
```diff
@@ -7,21 +7,40 @@ const noop = (styles: any) => styles;
...
+    const newObj = {...result, [newKey]: value};
+    return newObj;
+  }, {});
+};
+
 function styled<Props>(node: any) {
   return (...args: Interpolation<Props>[]) => {
     const newArgs: Interpolations = args.map(
-      interpolation => (props: Props & {theme: CanvasTheme; direction: ContentDirection}) => {
+      interpolation => (props: Props & {theme: CanvasTheme & {_staticStates?: boolean}}) => {
```
NicholasBoll: Not adding `_staticStates` to `CanvasTheme` keeps it private

Issue #366, opened 2019-12-11
ARIA 1.0 Combobox support


vibdev:
## 🐛 Bug Report

The combobox component supports ARIA spec 1.1 however browsers have not added full support yet.

## To Reproduce

Try to use the combobox component with a screenreader

## Expected Behavior

All roles and aria attributes should be read out

## Actual Results

The role of combobox is not read out.

## Browser (if applicable)

I think issue is still in most browser and screenreader combos, e.g. chrome and voiceover

------

I am proposing to make a new prop that can toggle between the specs. It should default to using v1.0

NicholasBoll:
Thanks @vibdev, I'm looking into the following links for more context:
* https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.0pattern/combobox-autocomplete-list.html
* https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
* https://www.levelaccess.com/differences-aria-1-0-1-1-changes-rolecombobox/

NicholasBoll:
After some digging, our internal accessibility team knows about this. This proposal came at the request of the accessibility team. Supporting 1.1 but defaulting to 1.0 is to be forward-thinking so that if implementation do catch up to the spec, we can flip a switch. Support for 1.0 isn't meant to be dropped by any assistive technology.

Accessibility also mentioned issues when navigating to the combobox with the arrow keys vs tabbing.

anicholls:
Closed by #380

virtualmason:
Is the Workday Canvas Combo Box / autocomplete currently working? I know it's in beta but does it work for anyone after adding the library? It seems like Chrom does not recognize it?

lychyi:
@virtualmason I'm not sure what you mean by this. Can you open an issue with the appropriate details so we can take a look? Combobox is in our "Labs" state but it doesn't mean it should ship broken.

virtualmason:
If you look at the docs it mentions ComboBox with no way to use it. I
assume to get the ComboBox I'm to "import {ComboBox} from ''
@workday/canvas-kit-labs-react-combo-box' so I can use Combobox but the
browser does not recognize it. I also tried doing it just like in the read
me below without importing "ComboBox" only npm installing the library. Is
there some missing information for it to work that is not mentioned below,
on the npm page or workday labs page
https://github.com/Workday/canvas-kit/tree/master/modules/_labs/combobox/react
?

[...]

PR #380, opened 2019-12-18
fix(combobox): default to using ARIA 1.0 spec

vibdev:
Adds a prop to allow manual toggling of ARIA 1.1 if needed

## Summary

Fixes issue #366 

The accessibility team found spotty support for ARIA 1.1 and suggested we default to ARIA 1.0 right now. ~In case the user need ARIA 1.1 support and prop toggle is provided.~

To test we should verify that the combobox element now reads out that it is a combobox on common screen reader browser combinations.

## Checklist

- [x] `yarn test` passes

## Additional References

The difference is what attributes are added to text input and containing element.
A full write up can be found https://www.levelaccess.com/differences-aria-1-0-1-1-changes-rolecombobox/


vibdev:
a11y team confirmed this fixes the issues they were seeing 👍

modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -125,6 +129,7 @@ export default class Combobox extends React.Component<ComboboxProps, ComboboxSta
   static defaultProps = {
     clearButtonLabel: `Reset Search Input`,
     clearButtonVariant: IconButton.Variant.Plain,
+    useLegacyAriaAttributes: true,
```
NicholasBoll: I have an open question with our accessibility team asking if we should support both ARIA 1.0 and 1.1 for combobox. There are issues with both. More context to the issue with the 1.1 spec: https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues

Pending the answer to the question, this might be a good time to add accessibility specifications to the combobox to ensure the correct behavior of this component.

modules/_labs/combobox/react/stories/stories.tsx
```diff
@@ -40,6 +48,7 @@ class Autocomplete extends React.Component<
           .splice(0, 5)}
         onChange={this.autocompleteCallback}
         showClearButton={this.props.showClearButton == null ? true : this.props.showClearButton}
+        useLegacyAriaAttributes={boolean('useLegacyAriaAttributes', true)}
```
NicholasBoll: If we want to support both ARIA 1.0 and 1.1 specs, I think we should have 2 different stories that state the difference vs one story with a knob.

modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -51,6 +51,10 @@ export interface ComboboxProps extends GrowthBehavior, React.HTMLAttributes<HTML
    * The id of the form field.
    */
   labelId?: string;
+  /**
+   * Toggle the use of ARIA 1.0 or ARIA 1.1
+   */
+  useLegacyAriaAttributes: boolean;
```
NicholasBoll: I'm not sure what to call this prop. It defaults to `true`. 1.0 isn't strictly legacy and there are problems already with the 1.1 spec... What if there's a 1.2 spec and 1.1 support is simply skipped?

What about:
```ts
useAriaSpec: '1.0' | '1.1';
```
with a default to `'1.0'`?

lychyi:
@NicholasBoll @vibdev 

Before we increase the API footprint for this component can we be absolutely sure that it is worth it?

Combobox is currently in `Labs` and I think it would be fine to change from ARIA 1.1 to 1.0 without triggering a breaking change. Supporting both specs with the rationale that "someone _may_ need it" isn't a good enough reason in my opinion and makes this component unnecessarily complicated.

vibdev:
@lychyi I am leaning towards removing the option. Since 1.0 has such good support, and the change is relatively easy, we could always look at 1.1 again when the need arises.

NicholasBoll:
@vibdev Agreed. Let's take out the 1.1 code for now. I don't think the maintenance cost is justified without a solid use-case where only 1.1 will work. We can always add 1.1 support or switch to 1.1 in the future when/if such a case exists.

NicholasBoll:
@vibdev @lychyi Our accessibility team confirmed the only reason to support the 1.1 spec is for forward compatibility. We should remove it for now when/if support for 1.1 gets better.

It is possible 1.1 never catches on. If it does, we can add it easily enough in the future.

modules/_labs/combobox/react/README.md
```diff
@@ -5,7 +5,12 @@
 </a>  This component is work in progress and currently in pre-release.
 
 A specialized input wrapper that adds an autocomplete list. Based on W3 spec for
-[List Autocomplete with Manual Selection](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html)
+[List Autocomplete with Manual Selection Aria 1.1](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html)
```
NicholasBoll: Remove this line and the next. Maybe add a note as to why we're supporting the 1.0 spec and not the 1.1. I think it is fine to leave a link for the difference

modules/_labs/combobox/react/lib/Combobox.tsx
NicholasBoll: @vibdev Remove this line now as it is no longer being used

vibdev:
@lychyi @NicholasBoll talked with the accessibility team, they didn't find any more changes 👍

PR #381, opened 2019-12-18
test(radio): Change tests to react-testing-library and improve coverage

mannycarrera4:
Fixes adding test for radio: https://github.com/Workday/canvas-kit/issues/286

## Checklist

- [x] `yarn test` passes
      applicable


package.json
```diff
@@ -24,7 +24,7 @@
     "@storybook/components": "^5.2.3",
     "@storybook/react": "^5.2.5",
     "@storybook/source-loader": "^5.2.4",
-    "@testing-library/jest-dom": "^4.1.0",
+    "@testing-library/jest-dom": "^4.2.4",
```
mannycarrera4: This can probably be removed after Alex's PR gets merged

cypress/integration/Radio.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const getRadio = () => {
+  return cy.get(`[type="radio"]`);
+};
+
+describe('Radio', () => {
+  before(() => {
+    h.stories.visit();
+  });
+  ['Default'].forEach(story => {
```
anicholls: No tests for alert and error?

cypress/integration/Radio.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    context(`given the '${story}' story is rendered`, () => {
+      beforeEach(() => {
+        h.stories.load('Components|Inputs/Radio/React/Left Label/Radio', story);
```
anicholls: No Radio Group tests?

modules/form-field/react/stories/stories_Radio.tsx
```diff
@@ -143,4 +159,53 @@ storiesOf('Components|Inputs/Radio/React/Left Label/Radio', module)
...
+          <tr>
+            <th></th>
+            <th>No state</th>
+            <th>Hover</th>
+            <th>Focused</th>
+            <th>Active</th>
+          </tr>
+        </thead>
+        <tbody>
+          {[false, true].map(checked => {
```
anicholls: disabled? error states?

modules/form-field/react/stories/stories_Radio.tsx
```diff
@@ -143,4 +159,53 @@ storiesOf('Components|Inputs/Radio/React/Left Label/Radio', module)
...
+            return (
+              <tr key={key}>
+                <td>{`${checked ? 'checked' : 'unchecked'}`}</td>
+
+                {['', 'hover', 'focus', 'active'].map(className => (
+                  <td key={`${key}-${className}`}>
+                    <Radio
+                      checked={checked}
+                      className={className}
+                      onChange={() => {}} // eslint-disable-line no-empty-function
```
anicholls: We should add a label to track the disabled coloring

modules/radio/react/spec/RadioGroup.spec.tsx
```diff
@@ -1,151 +1,39 @@
...
-        <Radio id="2" value="phone" label="Phone" />
-      </RadioGroup>
-    );
-    const container = component.at(0).getDOMNode();
-    expect(container.getAttribute('data-propspread')).toBe('test');
-    component.unmount();
+  describe('when rendered', () => {
+    describe('with an id', () => {
+      test('should render a radio group with a name', () => {
+        const id = 'myRadio';
```
anicholls: Unused

modules/radio/react/spec/RadioGroup.spec.tsx
```diff
@@ -1,151 +1,39 @@
...
-    const component = mount(
-      <RadioGroup data-propspread="test">
-        <Radio id="1" value="email" label="E-mail" />
-        <Radio id="2" value="phone" label="Phone" />
-      </RadioGroup>
-    );
-    const container = component.at(0).getDOMNode();
-    expect(container.getAttribute('data-propspread')).toBe('test');
-    component.unmount();
+  describe('when rendered', () => {
```
anicholls: Let's combine these:
```tsx
describe('when rendered with a name', () => {
  it('should render a radio group with a name', () => {
```

modules/radio/react/spec/RadioGroup.spec.tsx
```diff
@@ -1,151 +1,39 @@
...
+          <RadioGroup name="contact" data-testid="radiogroup">
+            <Radio id="1" value="email" label="E-mail" />
+            <Radio id="2" value="phone" label="Phone" />
+          </RadioGroup>
+        );
+        expect(getByTestId('radiogroup')).toHaveAttribute('name');
+      });
+    });
+
+    describe('with a value', () => {
```
anicholls: Same here

modules/radio/react/spec/RadioGroup.spec.tsx
```diff
@@ -1,151 +1,39 @@
...
+      test('should render a selected radio that matches that value of a string', async () => {
```
anicholls: Can you also test setting a value by the index?

modules/radio/react/spec/RadioGroup.spec.tsx
```diff
@@ -1,151 +1,39 @@
...
+          <RadioGroup name="contact" value="phone">
+            <Radio id="1" value="email" label="E-mail" />
+            <Radio id="2" value="phone" label="Phone" />
+          </RadioGroup>
+        );
+
+        const phoneRadio = await getByLabelText('Phone'); /*? */
+        expect(phoneRadio).toHaveProperty('checked', true);
+      });
+    });
```
anicholls: Need to add tests for prop spreading and onChange

cypress/integration/Radio.spec.ts
mannycarrera4: I might be wrong, but individual  radios don't have error or alert states, i think that's only for radio group?

cypress/integration/Radio.spec.ts
mannycarrera4: Yep yep, was waiting for your checkbox PR to go in before adding some!

modules/form-field/react/stories/stories_Radio.tsx
mannycarrera4: I was wondering if we wanted to break them out differently like Nicholas had suggested, will fix!

cypress/integration/Radio.spec.ts
anicholls: You are correct, haha sorry. Will wait until you get the RadioGroup cypress tests up.

cypress/integration/RadioGroup.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+const getInidivualRadio = (index: number) => {
+  return cy
+    .get('div[name="contact"]')
+    .children()
+    .eq(index)
+    .children()
+    .eq(0)
+    .children()
+    .eq(0);
+};
```
mannycarrera4: if there's a better way to do this, I'm open to suggestions!

modules/form-field/react/stories/stories_Radio.tsx
```diff
@@ -11,6 +13,20 @@ import README from '../../../radio/react/README.md';
...
+    paddingBottom: 16,
+  },
+  'td, th': {
+    minWidth: 100,
+    paddingBottom: 16,
+    paddingRight: 16,
+    textAlign: 'left',
+  },
+});
+
```
mannycarrera4: waiting for @anicholls PR to go int for checkbox to re write some of the static states stories

modules/radio/react/spec/Radio.spec.tsx
```diff
@@ -1,108 +1,104 @@
 import * as React from 'react';
-import {mount} from 'enzyme';
+import {render, fireEvent} from '@testing-library/react';
 import Radio from '../lib/Radio';
-import ReactDOMServer from 'react-dom/server';
-import {axe} from 'jest-axe';
-import FormField from '@workday/canvas-kit-react-form-field';
 
-describe('Radio Input', () => {
```
mannycarrera4: matches our `checkbox` spec

cypress/integration/RadioGroup.spec.ts
```diff
@@ -0,0 +1,40 @@
+import * as h from '../helpers';
+
+const getInidivualRadio = (index: number) => {
```
lychyi: There's a typo here but I think you can simplify this and let the params do the talking.

```suggestion
const getRadio = (index: number) => {
```

cypress/integration/RadioGroup.spec.ts
lychyi: This can probably be simplified, I believe jQuery has the ability to return a collection of `radio`s using a selector. Or there's always just `'div[name="contact"] input[type="radio"]'` which should get you all the radios in a collection.

Thoughts on making the `name` configurable too or add a parameter that takes a custom selector string since `name` is technically optional. I could easily see us renaming the `name` in the story and breaking the helper for everyone

modules/form-field/react/stories/stories_Radio.tsx
```diff
@@ -140,7 +145,109 @@ storiesOf('Components|Inputs/Radio/React/Left Label/Radio', module)
...
+            <FormField
+              useFieldset={true}
+              hintText={hintText}
+              hintId={hintId}
+              labelPosition={FormField.LabelPosition.Left}
+              {...props}
+            >
+              <ControlledComponentWrapper>
+                <RadioGroup name="contact">
+                  <Radio id="1" value="email" label="E-mail" />
```
anicholls: Nit: Could probably limit this to 2 `Radio`s (one enabled, one disabled) for less complexity.

modules/radio/react/spec/Radio.spec.tsx
```diff
@@ -1,108 +1,104 @@
...
+      const {getByRole} = render(<Radio data-propspread={attr} onChange={cb} />);
+      expect(getByRole('radio')).toHaveAttribute('data-propspread', attr);
+    });
   });
 
-  test('Radio without a defined id should pass axe DOM accessibility guidelines', async () => {
-    const html = ReactDOMServer.renderToString(<Radio label={'Label'} onChange={jest.fn()} />);
-    expect(await axe(html)).toHaveNoViolations();
+  describe('when rendered with an input ref', () => {
+    it('input ref should be defined', () => {
```
anicholls: @lychyi Found a bug with this logic and the input ref is always defined even if it's not pointing to the correct element. Can you check for non-null and look for `type="radio"`? (you can reference the checkbox tests)

modules/radio/react/spec/RadioGroup.spec.tsx
```diff
@@ -1,151 +1,72 @@
...
-        <Radio id="2" value="phone" label="Phone" />
-      </RadioGroup>
-    );
-    const container = component.at(0).getDOMNode();
-    expect(container.getAttribute('data-propspread')).toBe('test');
-    component.unmount();
+  describe('when clicked', () => {
+    it('should call a callback function', async () => {
+      const mockFunction = jest.fn((value: number) => value);
+      mockFunction(1);
```
anicholls: Obviously this is going to pass when you call it in the test! haha. I think this needs to be removed. Right now you're clicking on the radio group container and not one of the radios.

modules/form-field/react/stories/stories_Radio.tsx
```diff
@@ -140,7 +145,109 @@ storiesOf('Components|Inputs/Radio/React/Left Label/Radio', module)
...
+              }
+              return true;
+            }
+          )}
+        >
+          {props => (
+            <Radio
+              {...props}
+              onChange={() => {}} // eslint-disable-line no-empty-function
+              label="Checkbox"
```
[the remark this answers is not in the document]
mannycarrera4: si

cypress/integration/RadioGroup.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    h.stories.visit();
+  });
+  ['Default', 'Alert', 'Error with Grow'].forEach(story => {
+    context(`given the '${story}' story is rendered`, () => {
+      beforeEach(() => {
+        h.stories.load('Components|Inputs/Radio/React/Top Label/Radio Group', story);
+      });
+
+      it('should pass accessibility checks', () => {
+        cy.checkA11y();
```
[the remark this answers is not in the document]
NicholasBoll: I asked Accessibility this question a few months ago. They estimate `axe-core` catches around 13% of accessibility issues. They think we should run it because it is so cheap to do so, but not treat it as a replacement for accessibility specifications.

What you describe should be specifications in our Cypress files.

```ts
describe('given the Error story is rendered', () => {
  it('should have a legend element', () => {
    // get radio input
    // .parents('legend').should('exist')
  })

  it('should have an aria-describedby linking to the input element', () => {
    // find radio input and get id to check against the value of the `aria-describedby`
  })
})
```

I think the test block description is misleading. It should say something like:

```ts
it('should pass axe-core checks', () => {
```

To clarify:
`cy.checkA11y()` is _not_ a replacement for accessibility specifications. We simply treat it as a black box and not care if we duplicate accessibility checks.

cypress/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "compilerOptions": {
-    "types": ["cypress"]
+    "types": ["cypress", "@types/testing-library__cypress"]
```
mannycarrera4: remove

modules/_labs/header/react/package.json
```diff
@@ -42,7 +42,6 @@
     "@emotion/core": "^10.0.21",
     "@emotion/styled": "^10.0.17",
     "@workday/canvas-kit-labs-react-combobox": "^3.3.2",
-    "@workday/canvas-kit-labs-react-menu": "^3.3.0",
```
mannycarrera4: add back in

PR #382, opened 2019-12-18
docs: Add test documentation

NicholasBoll:
Official documentation on how we think about testing - the different levels and writing successful and maintainable tests.

## Checklist


TESTING.md
```diff
@@ -0,0 +1,10 @@
+## Testing
+
+`canvas-kit` has 3 levels of testing: unit, functional and visual. Each area has special use-cases even though what they test might overlap. Here's a simplified explaination of what each level is meant to test:
+
+- *unit*: This level is meant to test the expected outputs of developer APIs (ex: When a developer passes in X prop, then expect a specific output
+- *functional*: This level is browser-based testing and is where we specify UI behavior and accessibility requirements
```
anicholls: Also where we test combinations of components (i.e. `FormField` + `Checkbox`)

NicholasBoll: Right. I've heard "Patterns" and "Use Cases". I'm not sure what to call them. Basically they are stories that should use-cases where a component to encapsulate the use-case would be too complicated and rigid or contain too much customization and so we leave them to composition patterns.

TESTING.md
```diff
@@ -0,0 +1,69 @@
+## Testing
+
+`canvas-kit` has 3 levels of testing: unit, functional and visual. Each area has special use-cases even though tests at various levelsmay overlap. Here's a simplified explaination of what each level is meant to test:
+
+- *unit*: This level is meant to test the expected outputs of developer APIs (ex: When a developer passes in X prop, then expect a specific output
+- *functional*: This level is browser-based testing and is where UI behavior and accessibility requirements are specified
```
mannycarrera4: should we mention what we use for each level, example ```unit: (Using RTL and Jest): .....```

TESTING.md
```diff
@@ -0,0 +1,69 @@
...
+- *visual*: This level is where all styling is tested. Some props have only visual changes and it is difficult to reason about a in any other level
+
+Tests are supposed to be a source of feedback and confidence. To acheive that, a test should be clear about intent and failures. Developers tend to ignore passing tests and focus only on failing tests. If a the intent of the test cannot be determined, it is difficult to fix the test. If the failure isn't immediately obvious, too much time is spent fixing the test. Canvas-kit adopted a BDD (behavior-driven development) approach of testing. The most common way to think is about behavior is "given/when/then". These behaviors are a seriees of specifications and not necessarily "tests". Testing in this repo is not just testing the correctness of our applications, but running a series of specifications. Specifications tend to be more granular to give context of what code is expected to do. A "test" simply has to pass. A specification requires meaning.
```
mannycarrera4: `If a the intent` -> `If the intent`

mannycarrera4:
I talked with Nicholas about this, but it would be nice if we have an overview of each testing strategy, and then potentially have a more in depth article/link to how we test with cypress for example.

TESTING.md
```diff
@@ -0,0 +1,69 @@
+## Testing
+
+`canvas-kit` has 3 levels of testing: unit, functional and visual. Each area has special use-cases even though tests at various levelsmay overlap. Here's a simplified explaination of what each level is meant to test:
```
lychyi:
```suggestion
`canvas-kit` has 3 levels of testing: unit, functional and visual. Each area has special use-cases even though tests at various levels may overlap. Here's a simplified explanation of what each level is meant to test:
```

TESTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+- *unit*: This level is meant to test the expected outputs of developer APIs (ex: When a developer passes in X prop, then expect a specific output
```
lychyi:
```suggestion
- *Unit*: This level is meant to test the expected outputs of developer APIs (e.g., When a developer passes in a set of props, then expect a specific output).
```

TESTING.md
lychyi:
```suggestion
- *Functional*: This level is browser-based testing and is where UI behavior and accessibility requirements are specified.
```

lychyi: An example would be nice here too

TESTING.md
```diff
@@ -0,0 +1,69 @@
...
+- *functional*: This level is browser-based testing and is where UI behavior and accessibility requirements are specified
+- *visual*: This level is where all styling is tested. Some props have only visual changes and it is difficult to reason about a in any other level
```
lychyi:
```suggestion
- *Visual*: This level is where all styling is tested. Some props have only visual effects and it is difficult to reason about at any other level.
```

TESTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+Tests are supposed to be a source of feedback and confidence. To acheive that, a test should be clear about intent and failures. Developers tend to ignore passing tests and focus only on failing tests. If the intent of the test cannot be determined, it is difficult to fix the test. If the failure isn't immediately obvious, too much time is spent fixing the test. Canvas-kit adopted a BDD (behavior-driven development) approach of testing. The most common way to think is about behavior is "given/when/then". These behaviors are a seriees of specifications and not necessarily "tests". Testing in this repo is not just testing the correctness of our applications, but running a series of specifications. Specifications tend to be more granular to give context of what code is expected to do. A "test" simply has to pass. A specification requires meaning.
```
lychyi:
```suggestion
Tests are supposed to be a source of feedback and confidence. To achieve that, a test should be clear about intent and failures. Developers tend to ignore passing tests and focus only on failing tests. If the intent of the test cannot be determined, it is difficult to fix the test. If the failure isn't immediately obvious, too much time is spent fixing the test. Canvas Kit adopts a BDD (behavior-driven development) approach of testing. The most common way to think is about behavior is "given/when/then". These behaviors are a series of specifications and not necessarily "tests". Testing in this repo is not just testing the correctness of our components, but running a series of specifications. Specifications tend to be more granular to give context of what code is expected to do. A "test" simply has to pass. A specification requires meaning.
```

lychyi: Yikes, diffs don't work here. I just made a couple of typo changes ("acheive" -> "achieve" and "seriees" -> "series"), changed "Canvas-kit" to "Canvas Kit" (which is our preferred proper noun for the repo), and changed the tense of "adopted" to "adopts" since it sounds more authoritative.

lychyi: And... Amen, this is a great paragraph. ✊

TESTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Unit tests
+"Unit" test is not actually an accurate name since most canvas-kit unit tests include testing a DOM implementation. Canvas-kit uses [Jest](https://jestjs.io/), [React Testing Library](https://testing-library.com/docs/react-testing-library/intro) and [jest-dom](https://github.com/testing-library/jest-dom) which have a nice API for testing React components. "Unit" tests should test a React component's props. For example, if documentation says that a component will pass extra props to a container component, there should be a specification that verifies this claim. If a component's documentation states that children will be rendered, a test should verify the children exist within the containing component.
```
lychyi:
```suggestion
"Unit" test is not actually an accurate name since most Canvas Kit unit tests include testing a DOM implementation. Canvas Kit uses [Jest](https://jestjs.io/), [React Testing Library](https://testing-library.com/docs/react-testing-library/intro), and [jest-dom](https://github.com/testing-library/jest-dom) which have a nice API for testing React components. "Unit" tests should test a React component's props. For example, if documentation says that a component will pass extra props to a container component, there should be a specification that verifies this claim. If a component's documentation states that children will be rendered, a test should verify the children exist within the containing component.
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+
+  const component = await getByTestId('test');
+
+  expect(component.textContent).toEqual('foo');
+  expect(component.getAttribute('aria-label')).toEqual('foo');
+});
+```
+
+This is a "test" and not a specification, but a test that asserts multiple outputs. If any of the
+expectations fail, the CI will output something like "Failed: SomeComponent should render
```
lychyi: Nit of all nits:

I like the back-tick approach here when referring to system messaging since it tends to imply code or output.

> will output something like `Failed: SomeComponent should render...`

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+correctly". The message doesn't give any indication of why any of the expectations would fail.
+Further more, the failures don't give much indication as to what is failing. Without looking at the
```
lychyi:
```suggestion
Furthermore, the failures don't give much indication as to what is failing. Without looking at the
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+  });
+});
+```
+
+This series of specifications give more context to the test. It is more code, but it is more clear
+what the intent of each specification is. If one fails, it will be easier to diagnose what and why.
+The assertions also make use of `jest-dom` which gives more context into _how_ it fails. Instead of
+a generic assertion, the CLI will output something like
+`Expected <div data-testid="test" aria-label="bar">bar</div> to have an "aria-label" of "foo" but received "bar"`.
+The specification should tell us which spec failed and the assertion gives insight into how.
```
lychyi: This is money, we should bold (or emphasize).

```suggestion
**The specification should tell us which spec failed and the assertion gives insight into how.**
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+  </SomeComponent>
+);
+
+const component = await getByTestId('container');
+const child = await getByTestId('test');
+
+expect(component).toContainElement(child);
+```
+
+The first test has the DOM structure encoded in it. The second
```
lychyi: Looks like you sneezed and hit delete in your editor here. Bless you.

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+specification is met). Also this example will have a more useful failure message since
+`.toContainElement` has the context that it is expecting an element in another element vs a match of
+a string.
+
+## Functional tests
+
+Canvas Kit uses [Cypress][https://cypress.io] for browser-based behavior testing (additional info:
+[Why Cypress?](./cypress/WHY_CYPRESS.md)). Functional tests ensure the code meets functional
+specifications from a user's perspective. Specifications can come from many different stakeholders
+including product management, user research and accessibility. A functional test should read in the
```
lychyi: oxford comma

```suggestion
including product management, user research, and accessibility. A functional test should read in the
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+Given/When/Then format and fits in well as acceptance criteria.
```
lychyi:
```suggestion
Given/When/Then format and fit in well as acceptance criteria.
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+
+Example of a usability specification: "Given the modal example, when a user clicks the 'Delete Item'
+button, then the modal opens".
+
+Example of an accessibility specification: "Given the modal example, when a user clicks the 'Delete
+Item' button, then the focus should be transferred to the first focusable element in the dialog"
+
+The BDD-style (behavior-driven development) syntax lends well to the nesting of actions performed on
+a example (also called "story" borrowed from Storybook nomenclature). The wording used inside the
+BDD blocks should read in English and be understandable by non-developers. The implementation should
```
lychyi: It sounds like you're specifying the English language rather than what I think you're trying to say:

```suggestion
BDD blocks should read in plain English and be understandable by non-developers. The implementation should
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+be as readable as possible by non-developers for trust and verification. The intent of the wording
+of the BDD-style syntax is to parse into an outline form to go into documentation as a verification
```
lychyi:
```suggestion
of the BDD-style syntax is to parse into an outline form and go into documentation as a verification
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+});
+```
+
+The outline would look like:
+
+- Given the Default Modal is rendered
+  - when the target button is clicked
+    - then it should open the modal
+
+Tests should rely on aria attributes as much as possible for selecting and asserting. Test IDs
```
lychyi:
```suggestion
Tests should rely on ARIA attributes as much as possible for selecting and asserting. Test IDs
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+should be used only to disambiguate DOM elements.
+
+Read [Intro to Cypress](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html)
+for a good idea of how Cypress works. The core concept to Cypress is enqueued command chains that
```
lychyi:
```suggestion
for a good idea on how Cypress works. The core concept to Cypress is enqueued command chains that
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+typically wrap jQuery collections. Cypress uses jQuery, Mocha, Chai, Sinon, Lodash and Moment. All
+are exposed to the developer of the test.
+
+Let's break that down. Cypress has a jQuery-like chainable API to select and interact with DOM
+nodes. Cypress control when and how often a command is run to ensure assertions either pass or time
```
lychyi:
```suggestion
nodes. Cypress controls when and how often a command is run to ensure assertions either pass or time
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+cy.get('body')
+  .contains('button', 'Delete Item')
+  .click();
+
+cy.get('[data-testid="MyResult"]').should('contain', 'Success!');
+```
+
+The Cypress Command queue will show `['get', 'contains', 'click', 'get', 'should']`, but it will not
+run these commands immediately. There is also no need to use `.then` to wait for any previous
+condition to be met first. The Cypress run time will do the following:
```
lychyi:
```suggestion
condition to be met first. The Cypress run-time will do the following:
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+
+- Run `$('body')` every 50ms until `$('body')` returns a non-empty jQuery collection
+- Run a function that gets all `button` elements inside of the `body` element that have text content
+  that contains 'Delete Item' and that jQuery collection is not empty
+- Ensure the previous element is clickable (visible, not covered, has actual width/height), waiting
+  50ms to detect animations (waiting for animations to complete), etc. Basically make sure a user
```
lychyi:
```suggestion
  50ms to detect animations (waiting for animations to complete), etc. Basically, make sure a user
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+  .click();
+
+cy.get('[data-testid="MyResult"]').should('contain', 'Success!');
+```
+
+The Cypress Command queue will show `['get', 'contains', 'click', 'get', 'should']`, but it will not
+run these commands immediately. There is also no need to use `.then` to wait for any previous
+condition to be met first. The Cypress run time will do the following:
+
+- Run `$('body')` every 50ms until `$('body')` returns a non-empty jQuery collection
```
lychyi: I would make bullets either end in full stops or not, but not both

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+- Run a function that gets all `button` elements inside of the `body` element that have text content
+  that contains 'Delete Item' and that jQuery collection is not empty
+- Ensure the previous element is clickable (visible, not covered, has actual width/height), waiting
+  50ms to detect animations (waiting for animations to complete), etc. Basically make sure a user
+  could actually click on it.
+- Run `$('[data-testid="MyResult"]')` every 50ms until it returns a non-empty jQuery collection.
+  Usually clicking has side-effects like network requests to the server before the side-effect is
+  reflected in the DOM. Since Cypress implicitly retries, this command will _eventually_ return a
+  non-empty jQuery collection and the author doesn't need to explicitly do anything. Note this
```
lychyi:
```suggestion
  non-empty jQuery collection and the author doesn't need to explicitly do anything. Note: this
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+  selector _could_ return a non-empty jQuery collection instantly, but it might not contain
+  "Success!" yet...
+- Run internal `chai` + `chai-jquery` matchers (`expect($el).to.contain('Success!')`) until an error
+  is no longer thrown. `.should` will rerun the previous command _and_ the assertion until the
```
lychyi:
```suggestion
  is no longer thrown. `.should` will re-run the previous command _and_ the assertion until the
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+  condition is met or a timeout occurs. This is the secret sauce to the stability of Cypress tests
```
lychyi: I think you use "time out" and "timeout" interchangeably in this piece, probably better to just pick one.

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+  thrown. If an error is never thrown, it will only be run once.
+
+Cypress is very easy to get started with, but there isn't enough documentation about how to scale a
+complicated suite of Cypress tests.
+
+### Component helpers
+
+Component helpers should be written to help with automation. These component helpers are useful for
+a more readable implementation as well as more maintainable test code. Component helpers can also be
+exported by this repository for use in downstream testing (Cypress tests in applications that use
```
lychyi:
```suggestion
exported by this repository for use in downstream testing (e.g., Cypress tests in applications that use
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+Canvas Kit). Helpers contain implementation details about how components are composed internally.
+The goal is that if the Component implementation changes, the component helpers should be the only
```
lychyi:
```suggestion
The goal is that if the component's implementation changes, the component helpers should be the only
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+code that needs to change to keep tests passing.
+
+Component helpers come in 3 flavors:
+
+- **Starter**: This helper types starts a Cypress chain. The return type is always
```
lychyi: This one always gets me, but "types starts" sounds weird but "types start" sounds about right?

```suggestion
- **Starter**: This helper types start a Cypress chain. The return type is always
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+  scrolling the virtualized menu until the desired item is visible and clicking on the item. This
+  code would be repeated for every instance and has implementation knowledge of the component. A
+  high-level semantic action might be called "select" that wraps up this implementation detail to
+  free up automation and development engineers to think about the test case and not about
+  implementation details.
+
+### How to write a functional test
+
+A functional test describes the behavior and specifications of components. Functional tests should
+follow recommendations of behavior driven development where "specs" read like a specification and
```
lychyi:
```suggestion
follow recommendations of behavior-driven development where "specs" read like a specification and
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+  action is useful to give a higher level semantic meaning to a series of low-level commands. For
+  example, a Select component might compose a Popup and a virtualized menu. This action may include
+  opening the select, finding the popup element, seeing if the desired item is currently visible,
+  scrolling the virtualized menu until the desired item is visible and clicking on the item. This
+  code would be repeated for every instance and has implementation knowledge of the component. A
+  high-level semantic action might be called "select" that wraps up this implementation detail to
+  free up automation and development engineers to think about the test case and not about
+  implementation details.
+
+### How to write a functional test
```
lychyi: This sounds like it should go somewhere else in the document, some of the content below is repeated, is this intentional or still part of the draft process?

TESTING.md
anicholls: This is super helpful 👌

TESTING.md
anicholls: Should it be "This helper type starts..."?

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+- **Starter**: This helper types starts a Cypress chain. The return type is always
+  `Cypress.Chainable<JQuery>` and usually contains a `cy.get` and the selector usually contains
+  something that is unique to that component type. For example, the Modal helper uses
+  `[role=dialog]` with an option to disambiguate with a `data-testid`.
+- **Transform**: Imagine a web application is a collection of boxes that determine how components of
+  an application are laid out. A transform helper takes in a larger box and returns a smaller box,
+  further reducing the scope. Ideally these helpers contain no Cypress commands (which are
+  asynchronous), only jQuery calls (which are synchronous). Synchronous helper functions can be run
+  more than once at the discretion of the Cypress runtime. A good use-case for this type is to
+  replace CSS selectors
```
anicholls:
```suggestion
  replace CSS selectors.
```

TESTING.md
```diff
@@ -0,0 +1,274 @@
...
+actions on a component (a set of "when"s) and some type of expected outcome (a "then").
+
+For example:
+
+- Given the Default Modal story is rendered
+- When I click on the "Delete Item" button
+- Then the Modal should be open
+
+Put together, a sentence is formed: "Given the Default Modal story is rendered when I click on the
+"Delete Item" button then the Modal should be open"
```
anicholls: This is such a comprehensive and lengthy writeup I almost feel like we need a summary/conclusion paragraph. Your call though

TESTING.md
NicholasBoll: lol. I'll finish my thought

NicholasBoll: This line seems redundant. The first sentence exists under the first example already

TESTING.md
NicholasBoll: "for an intro to how Cypress works."?

NicholasBoll: Hmm. Maybe that's redundant if the link text is "Intro to Cypress"

TESTING.md
NicholasBoll: Interesting. The only other place I could find "time out" was "...assertions either pass or time out." "timeout" is used as a verb here. Is that okay?

TESTING.md
NicholasBoll: I think this should go somewhere else... This section isn't related to the 3 levels of testing and what should go where... I think we could remove it from this document and it should be a different one

NicholasBoll: The only content I think is relevant to this document is that both visual regression and functional tests use stories as a unit to test against.

NicholasBoll: I added info about how Storybook stories relate to functional and visual tests in the intro to the levels. I think I can remove this last section and save it for another document.

Functional testing is probably the broadest topic with the most nuances. It is also generally less understood than unit testing.

cypress/helpers/modal.ts
```diff
@@ -6,7 +6,7 @@
  *   .should('have.attr', 'role', 'dialog')
  */
 export function get(testId?: string): Cypress.Chainable<JQuery> {
-  const selector = testId ? `[data-testid='${testId}'] [role-dialog]` : `[role=dialog]`;
+  const selector = testId ? `[data-testid='${testId}'] [role=dialog]` : `[role=dialog]`;
```
NicholasBoll: This is a fix for the helper although no test currently passes `testId`

TESTING.md
NicholasBoll: I added a "Summary" section. It is a bulleted section on what the different test types are. Let me know if this is what you were thinking or what other kind of wrap up we should do

TESTING.md
```diff
@@ -0,0 +1,293 @@
...
+[knobs](https://github.com/storybookjs/storybook/tree/next/addons/knobs).
+
+Not all visual states are application states (E.g. `focus`, `hover`, and `active` on a button
+element). For these pseudo-selectors, the `StaticStates` wrapper can be used where pseudo-selectors
+are turned into CSS class names instead (E.g. `:focus` turns into `.focus`) and a class name can be
+added to the component. Note: this turns off the psuedo-selectors entirely.
+
+The following will render a button where the focus styling is always applied:
+
+```ts
```
NicholasBoll: This should fix the syntax highlighting

```suggestion
```tsx
```

TESTING.md
```diff
@@ -0,0 +1,293 @@
...
+"given/when/then". These behaviors are a series of specifications and not necessarily "tests".
+Testing in this repo is not just testing the correctness of our components, but running a series of
+specifications. Specifications tend to be more granular to give context of what code is expected to
+do. A "test" simply has to pass. A specification requires meaning.
+
+**Bad**:
+
+```tsx
+test('SomeComponent should render correctly', async () => {
+  const {getByTestId} = render(<SomeComponent data-testid="test" text="foo" />);
```
sahlhoff: @NicholasBoll originally after reading this I assumed it was best practice to use `data-testid` as a query since it is used throughout this document. However after looking at the RTL docs looks like this should be used as a last choice for a query: https://testing-library.com/docs/dom-testing-library/api-queries#bytestid

Should we also adopt this convention? Or am I ok using `data-testid` for majority of my selectors?

NicholasBoll: I advise using `data-testid` as a last resort. In Cypress, component helpers can be used to help with selecting sub-components and can be selected usually by accessibility attributes. Helpers should be the only part of the test that has implementation details about how the DOM is structured. This allows updates to the Component + component helpers without changes to the test file. In the past, testing helpers is part of the CI process guaranteeing they work for downstream consumers of Canvas Kit. If done right, this means applications can be tested using Cypress without worry of implementation changes breaking tests.

In React Testing Library, we're often testing for specific attributes and not using helpers. Helpers are additional logic that should be tested to avoid confusing failure messages. Often in React Testing Library, I'm only testing implementation details, but I want to avoid flaky tests that are difficult to debug. Most of our components allow for additional attributes to be applied to container components. This allows for easy selecting. The alternative is often `container.children[0].children[1]` or something like that. A failure due to changes in DOM structure are unhelpful.

That being said, I don't recommend using `data-testid` inside the Component source code. I don't have much issue adding them only to the tests.

The guiding principals say the same thing from the link you sent:
> In the spirit of the guiding principles, it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible. That said, they are way better than querying based on DOM structure or styling css class names. Learn more about data-testids from the blog post "Making your UI tests resilient to change"

You should use other selectors when available. The best are `getByLabelText` which will return the input/textarea/select/etc that is associated with the label by the label's text. It is easy to understand and assumes that all form fields are accessible (which they should be). There are other helpers that work well for data entry and accessible roles.

Kent C. Dodds seems to be saying much the same thing here: https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change

As for low-level component unit tests? Often the selectors provided by React Testing Library just don't give me what I need and I fall back to test ids.

NicholasBoll: Perhaps we should link to the article to help clear things up

NicholasBoll: I added a section about selectors here: https://github.com/Workday/canvas-kit/pull/382/files#diff-84889d0af78b35050da8d12673769ebaR76-R85

@sahlhoff @lychyi, what do you think?

PR #386, opened 2019-12-27
test(switch): Redo switch tests in react-testing-library

lychyi:
## Summary

Alright, lots of changes here:
- Migrated to react testing library
- Rearranged and added a bunch of tests (both API and styles)
- Extended Jest matchers for `jest-emotion` in `setupTests` instead of locally in `PageHeader.spec.tsx`
- Added functional tests in Cypress
- Added a visual state table for ChromaticQA

Doing these tests revealed a few changes that we'll need to make (tracked in #387):
- I don't think we need to specify `tabIndex={0}`, this is the default already.
- We should map a static member `Switch.ErrorType.Alert` to `ErrorType` from `canvas-kit-react-common`. It's kind of a pain to import that also just to use `Switch`.
- Per some conversations that @NicholasBoll had with a11y in the past, we should probably make the `role` attribute a configurable prop. Instead of `checkbox` it could be something like `switch`.
- We should probably rename the elements `SwitchBackground` and `SwitchCircle` to something more semantically correct like `SwitchTrack` and `SwitchThumb`.


modules/form-field/react/stories/stories_Switch.tsx
```diff
@@ -90,3 +107,42 @@ storiesOf('Components|Inputs/Switch/React/Left Label', module)
       {control(<Switch />)}
     </FormField>
   ));
+
+storiesOf('Components|Inputs/Switch/React/Visual Testing', module)
```
lychyi: "Visual Testing" seems to make sense for me or I could remove the grouping name and organize "States" under `Inputs/Switch/React` instead. Thoughts?

cypress/integration/Switch.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+        });
+      });
+
+      context('when the switch is clicked', () => {
+        it('should be toggled', () => {
+          getSwitch()
+            .invoke('prop', 'checked')
+            .as('checked');
+          getSwitch().click();
+          getSwitch().should(this.checked ? 'not.be.checked' : 'be.checked');
```
NicholasBoll: I don't think it is a good idea to have conditional testing. Either we expect a checkbox to be checked after a click or we expect it to not be checked after a click

cypress/integration/Switch.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      context('when the tab key is pressed', () => {
+        it('should be the focused item on the page', () => {
+          cy.tab();
+          getSwitch().should('have.focus');
+        });
+      });
+
+      context('when the switch is clicked', () => {
+        it('should be toggled', () => {
+          getSwitch()
```
NicholasBoll: I've been putting the actions of the `context` in a `beforeEach` of that context. This keeps the "when {something}" out of the spec itself. It allows for additional tests. So the spec is only an assertion of what we expect an outcome to be.

This block would look like:
```ts
      context('when the switch is clicked', () => {
        beforeEach(() => {
          getSwitch().click();
        });

        it('should be toggled', () => {
          getSwitch().should('be.checked');
        });
      });
```

cypress/integration/Switch.spec.ts
NicholasBoll: If you expect the default loaded state to be unchecked, have a spec for that. Then this spec clicks the switch and verifies it is now checked

modules/form-field/react/stories/stories_Switch.tsx
NicholasBoll: I kind of like a visual testing group. It might be too noisy to have every state of complex components represented in one story.

modules/switch/react/spec/Switch.spec.tsx
```diff
@@ -1,44 +1,182 @@
...
+        const {getByRole} = render(<Switch checked onChange={cb} error={ErrorType.Alert} />);
+        const bs = expectedBoxShadow(
+          canvas.colors.frenchVanilla100,
+          canvas.inputColors.warning.border,
+          canvas.colors.cantaloupe600
+        );
 
-    expect(id1).not.toEqual(id2);
-    fragment.unmount();
+        expect(getByRole('checkbox')).toHaveStyleRule('box-shadow', bs, {
```
NicholasBoll: If this test fails, what is actionable about it? Does it mean the input does not have an alert-colored border?

If the DOM changes so that the `target` doesn't match the selector, this test will fail without an obvious path to fix.

modules/switch/react/spec/Switch.spec.tsx
```diff
@@ -1,44 +1,182 @@
...
+        expect(backgroundElement).toHaveStyleRule('background-color', canvas.colors.licorice200);
+      });
+
+      test('the circle element should be to the very left (`translateX(0)`) when unchecked', () => {
+        const {container} = render(<Switch checked={false} disabled={true} onChange={cb} />);
+        const circleElement = getCircle(container);
+        expect(circleElement).toHaveStyleRule('transform', 'translateX(0)');
+      });
+
+      test('the circle element should be to the very right (`translateX(spacing.s)`) when unchecked', () => {
```
NicholasBoll: Test like this are hard to reason about when they fail. What does it mean if the `transform` rule is different? What if we accomplish the same visual output using different styling rules? Is that still okay?

modules/switch/react/spec/Switch.spec.tsx
```diff
@@ -1,44 +1,182 @@
...
+      test('the background element should have a neutral background color when not disabled and unhecked', () => {
+        const {container} = render(<Switch checked={false} onChange={cb} />);
+        const backgroundElement = getBackground(container);
+        expect(backgroundElement).toHaveStyleRule('background-color', canvas.colors.licorice200);
+      });
+
+      test('the circle element should be to the very left (`translateX(0)`) when unchecked', () => {
+        const {container} = render(<Switch checked={false} disabled={true} onChange={cb} />);
+        const circleElement = getCircle(container);
+        expect(circleElement).toHaveStyleRule('transform', 'translateX(0)');
```
NicholasBoll: If we accomplish the visual specification without using `transform: translateX(0)`, does it make sense for this test to fail?

modules/switch/react/spec/Switch.spec.tsx
lychyi: Yup, you're right, the specific style rule should be removed here

modules/switch/react/spec/Switch.spec.tsx
lychyi: All of these style ones can be kind of fragile, I was on the fence about including this one in a unit test, but this was also before I worked on the state table which would probably catch this case now.

modules/switch/react/spec/Switch.spec.tsx
lychyi: This is covered in viz reg so it does make sense to remove all the style tests except the ones about the `cursor` style rule.

modules/form-field/react/stories/stories_Switch.tsx
```diff
@@ -90,3 +107,66 @@ storiesOf('Components|Inputs/Switch/React/Left Label', module)
...
+                return (
+                  <tr
+                    key={`${checked ? 'checked' : 'unchecked'} ${
+                      disabled ? 'disabled' : 'enabled'
+                    } ${variant}`}
+                  >
+                    <td>{`${disabled ? 'Disabled ' : ''}${variant} (${
+                      checked ? 'checked' : 'unchecked'
+                    })`}</td>
+                    <td>
```
anicholls: Could simplify this with
```tsx
{['', 'focus', 'hover', 'active'].map(className => 
  <td>
    <Switch checked={checked} disabled={disabled} error={type} className={className} />
  </td>
)}
```

modules/switch/react/spec/Switch.spec.tsx
```diff
@@ -1,44 +1,123 @@
...
+
 describe('Switch', () => {
   const cb = jest.fn();
+
   afterEach(() => {
-    cb.mockReset();
+    cb.mockClear();
+  });
+
+  it('should be unchecked by default', () => {
```
anicholls: What do you think about comparing and aligning on all of these string "specs" for all the inputs so the tests are more consistent? You took a different approach [than I did in #372](https://github.com/Workday/canvas-kit/pull/372/files#diff-8e4752febc0225be3fe65f35a4b6072eR16)

modules/switch/react/spec/Switch.spec.tsx
```diff
@@ -1,44 +1,123 @@
...
+    const {getByRole} = render(<Switch onChange={cb} />);
+    expect(getByRole('checkbox')).toHaveProperty('checked', false);
+  });
+
+  it('should apply extra props to the input element', () => {
+    const testId = 'test';
+    const {container} = render(<Switch data-testid="test" onChange={cb} />);
+    expect(container.querySelector('input')).toHaveAttribute('data-testid', testId);
```
anicholls: Why use `getByRole` above and `container.querySelector` here?

modules/switch/react/spec/Switch.spec.tsx
```diff
@@ -1,44 +1,123 @@
...
+  it('should have a reference to the input', () => {
+    const inputRef = React.createRef<HTMLInputElement>();
+
+    render(<Switch inputRef={inputRef} onChange={cb} />);
+    expect(inputRef.current!.tagName.toLowerCase()).toBe('input');
+  });
+
+  it('should have a unique id by default', () => {
+    const {getByRole} = render(<Switch disabled={true} onChange={cb} />);
+    expect(getByRole('checkbox')).toHaveProperty('disabled', true);
```
anicholls: Not testing what you wanted to here!

modules/switch/react/spec/Switch.spec.tsx
```diff
@@ -1,44 +1,123 @@
...
-      .find('input')
-      .at(1)
-      .getDOMNode()
-      .getAttribute('id');
+  it('should be a `type` of "checkbox"', () => {
+    const {getByRole} = render(<Switch onChange={cb} />);
+    expect(getByRole('checkbox')).toHaveProperty('type', 'checkbox');
+  });
+
+  it('should have a `tabIndex` of 0', () => {
```
anicholls: isn't this just testing browser functionality?

modules/switch/react/spec/Switch.spec.tsx
```diff
@@ -1,44 +1,123 @@
...
 
+// Implementation details of the switch. These are meant to be helpers.
+
+// The switch track
+const getBackground = (container: HTMLElement) => {
+  return container.querySelector('div')!.querySelector('div');
+};
+
+// The switch thumb
+const getCircle = (container: HTMLElement) => {
```
anicholls: These helpers aren't used anymore

NicholasBoll:
Dang, TravisCI is slow! 10s isn't enough to verify a Storybook doc is loaded.

cypress/integration/Switch.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+      it('should not have any axe errors', () => {
+        cy.checkA11y();
+      });
+
+      context('when the tab key is pressed', () => {
+        beforeEach(() => {
+          cy.tab();
+        });
+        it('should be the focused item on the page', () => {
+          getSwitch().should('have.focus');
```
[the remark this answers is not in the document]
NicholasBoll: This specification is a little weird because hitting the tab key in this instance will give focus to the first focusable element on the page. I think we are saying the Switch should be focusable as per accessibility

cypress/integration/Switch.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+        });
+      });
+
+      context('when the switch is clicked', () => {
+        beforeEach(() => {
+          getSwitch().click();
+        });
+        it('should be checked', () => {
+          getSwitch().should('be.checked');
```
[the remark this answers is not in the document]
NicholasBoll: Cypress tests are about specifications from a user's perspective. This spec reads in full "Non-disabled Switch when the switch is clicked should be checked". Linking the given of the story being rendered, the when of the switch being clicked, and the then of the side effect of that action where the switch is now checked.

I might argue that "checked" is an implementation detail of a switch being "on", but there is a bit of controversy around how a "switch" should be labelled/read by screen readers from an accessibility standpoint. I don't remember the final say the `role` attribute of a `Switch` which effects how it will be read by a screen reader.

I personally avoid actions in unit-level tests including clicks. Most of the time firing a `click` event is enough for the semantic action we'd call "click", but some components will expect other browser events in a specific order and some events aren't as simple as a "click". I leave that all for Cypress which has wiring for all that. In Cypress, a "click" includes other events normally associated with "click" like "mouseover". I've seen unit tests that try to recreate all that. Yuck.

I wouldn't worry about possible duplication in testing levels. Each serve a specific purpose.

cypress/integration/Switch.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  context(`Given the 'Disabled' story is rendered`, () => {
+    beforeEach(() => {
+      h.stories.load('Components|Inputs/Switch/React/Top Label', 'Disabled');
+    });
+
+    it('should not have any axe errors', () => {
+      cy.checkA11y();
+    });
+    it('should be disabled', () => {
+      getSwitch().should('be.disabled');
```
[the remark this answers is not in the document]
NicholasBoll: Documentation on specifications vs tests:

> Specifications tend to be more granular to give context of what code is expected to
do. A "test" simply has to pass. A specification requires meaning.

I'm not sure the only specifications for a disabled switch are to ensure the switch is disabled, although that is a specification of the Disabled Story. Specifications are written against stories and not just components. There are probably other specifications like clicking on a disabled switch does not change its state.

We could probably argue that this is a specification guaranteed by the browser, but that guarantee is on a specific DOM element and not our component. It is an implementation detail that our component uses a checkbox. Specifications for the Switch dictate how it should behave from a user's perspective, not dictate the specific implementation we use.

modules/form-field/react/stories/stories_Switch.tsx
```diff
@@ -90,3 +107,66 @@ storiesOf('Components|Inputs/Switch/React/Left Label', module)
       {control(<Switch />)}
     </FormField>
   ));
+
+storiesOf('Components|Inputs/Switch/React/Visual Testing', module)
```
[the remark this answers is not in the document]
NicholasBoll: Agreed. We haven't come up with a set way of describing what stories are intended to show a working example with some type of accessibility context (which should go through an accessibility audit) and which stories show off states for visual verification.

I attempted to use MDX to do this, but Storybook has fought me every step of the way.

modules/switch/react/spec/Switch.spec.tsx
lychyi: removed, as discussed this was testing something we explicitly added to the checkbox `tabIndex=0`

cypress/integration/Switch.spec.ts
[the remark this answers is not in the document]
lychyi: Removed

modules/form-field/react/stories/stories_Switch.tsx
```diff
@@ -90,3 +107,66 @@ storiesOf('Components|Inputs/Switch/React/Left Label', module)
...
+  .addParameters({component: Switch})
+  .addDecorator(withReadme(README))
+  .add('States', () => (
+    <StaticStates>
+      <Table>
+        <thead>
+          <tr>
+            <th></th>
+            <th>No state</th>
+            <th>Focused</th>
```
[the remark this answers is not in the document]
lychyi: There's no hover state here in the Switch, but I guess I could cover it to ensure that doesn't change in the future.

modules/switch/react/spec/Switch.spec.tsx
```diff
@@ -1,44 +1,131 @@
...
+    it('should apply those extra props to the checkbox', () => {
+      const testValue = 'test';
+
+      const {getByRole} = render(<Switch data-propspread={testValue} onChange={cb} />);
+      expect(getByRole(role)).toHaveAttribute('data-propspread', testValue);
+    });
+  });
+
+  describe('when rendered with an input ref', () => {
+    it('the current property should not be null', () => {
```
anicholls: I find it odd when these don't read as a sentence (i.e. "when rendered with an input ref, it the current...")

What if we changed this to 
```suggestion
    it('should set the ref current property', () => {
```

I also think we could combine these two tests and just have two `expect`s, with a name like `should set the ref to the checkbox input element`

lychyi: Yea I like that, it was just reading funny from a test output. I'll make one minor adjustment to make the `ref` possessive`

modules/form-field/react/stories/stories_Switch.tsx
```diff
@@ -90,3 +107,59 @@ storiesOf('Components|Inputs/Switch/React/Left Label', module)
...
+                  variant === 'Alert'
+                    ? ErrorType.Alert
+                    : variant === 'Error'
+                    ? ErrorType.Error
+                    : undefined;
+
+                const key = `${checked ? 'checked' : 'unchecked'} ${
+                  disabled ? 'disabled' : 'enabled'
+                } ${variant}`;
+
```
anicholls: @NicholasBoll made a [good point](https://github.com/Workday/canvas-kit/pull/372#discussion_r367481573) that we probably don't want to show the combination of disabled and focus/active as it may be confusing. I addressed it [here](https://github.com/Workday/canvas-kit/pull/372/commits/96551b85982dc1910136f6a893c1dbcdde69fbae) if you want to reference

modules/form-field/react/stories/stories_Switch.tsx
```diff
@@ -90,3 +107,59 @@ storiesOf('Components|Inputs/Switch/React/Left Label', module)
...
+                      checked ? 'checked' : 'unchecked'
+                    })`}</td>
+                    {['', 'hover', 'focus', 'active'].map(className => (
+                      <td key={`${key} + ${className}`}>
+                        <Switch
+                          checked={checked}
+                          onChange={() => {}} // eslint-disable-line no-empty-function
+                          disabled={disabled}
+                          error={type}
+                          className={className}
```
anicholls: Let's add a label so we're testing the color when disabled as well.

modules/switch/react/spec/Switch.spec.tsx
```diff
@@ -1,44 +1,124 @@
...
+    it('should apply those extra props to the checkbox input', () => {
+      const testValue = 'test';
+
+      const {getByRole} = render(<Switch data-propspread={testValue} onChange={cb} />);
+      expect(getByRole(role)).toHaveAttribute('data-propspread', testValue);
+    });
+  });
+
+  describe('when rendered with an input ref', () => {
+    it("it should set the ref's current property to the checkbox input element", () => {
```
anicholls:
```suggestion
    it('should set the ref's current property to the checkbox input element', () => {
```

modules/form-field/react/stories/stories_Switch.tsx
lychyi: Can you elaborate on what adding a label will do here? I'm not sure I understand the ask

anicholls: Nevermind! I thought switch ahd the same type of label that checkbox and radio do 😂

PR #390, opened 2020-01-02
test(text-input): Change tests to react-testing-library and improve coverage

stephanerangaya:
## Summary

## Checklist

- [x] `yarn test` passes


modules/text-input/react/spec/TextInput.spec.tsx
```diff
@@ -1,50 +1,72 @@
...
-        error={FormField.ErrorType.Error}
-      >
-        <TextInput placeholder="Placeholder" value={'Hello'} onChange={jest.fn()} />;
-      </FormField>
-    );
-    expect(await axe(html)).toHaveNoViolations();
+  describe('when rendered', () => {
+    describe('with an placeholder', () => {
+      it('should render a text input with placeholder', () => {
+        const {getByPlaceholderText} = render(
```
lychyi: I think getting by the placeholder text and then asserting it has a placeholder probably won't work here. You'll never get a proper failure since you wouldn't be able to assert on anything if the placeholder fails.

Maybe a `getByTestId` instead?

modules/text-input/react/spec/TextInput.spec.tsx
```diff
@@ -1,50 +1,72 @@
...
-        />
-        ;
-      </>
-    );
-    expect(await axe(html)).toHaveNoViolations();
+  describe('when provided an input ref', () => {
+    it('should render a text input with ref defined', () => {
+      const ref: React.RefObject<HTMLInputElement> = React.createRef();
+      render(<TextInput inputRef={ref} />);
+      expect(ref.current).toBeDefined();
```
lychyi: See this comment https://github.com/Workday/canvas-kit/pull/372#discussion_r366109323

According to React, `ref.current` will always be defined, we need to check if it's `null` or not. And if it's not, we should be checking if it is an input element.

modules/text-input/react/spec/TextInput.spec.tsx
```diff
@@ -1,50 +1,72 @@
 import * as React from 'react';
-import {mount} from 'enzyme';
-import ReactDOMServer from 'react-dom/server';
+import {render, fireEvent} from '@testing-library/react';
 import TextInput from '../lib/TextInput';
+import ReactDOMServer from 'react-dom/server';
```
lychyi: I don't think this import is needed anymore

lychyi: ... or `axe` on the line below

modules/_labs/core/react/index.ts
```diff
@@ -5,4 +5,5 @@ import CanvasProvider from './lib/CanvasProvider';
 export default type;
 export {type, space, CanvasProvider};
 export * from './lib/type';
+export * from './lib/StaticStates';
```
mannycarrera4: do we still want to export this?

anicholls: I think it just makes our imports cleaner. Plus it's labs, we can always remove it if we want.

mannycarrera4: ok sounds good to me!

Issue #379, opened 2019-12-17
CSS for selected Radio doesn't maintain centered dot at non-100% screen zoom


elliot-at-workday:
## 🐛 Bug Report

When viewing the CSS styling for Radio at non-100% zoom the white dot is not centered

## To Reproduce

- Go to the story: https://workday.github.io/canvas-kit/?path=/story/components-inputs-radio-css-left-label-radio--default

- Change browser zoom to a non-100% value. The following ones are pretty noticeable: 90%, 110%, 150%

- Fun fact, it looks fine at 125%

## Expected Behavior

White dot remains in the center of the select radio at all zooms.

## Actual Results

White dot moves around within the blue dot at different zooms

## Browser (if applicable)

Google Chrome: 79.0.3945.79 (Official Build) (64-bit)
Have not tested on other browsers.

## Link to repl or repo (highly encouraged)

Can be repro'd in this repro as of revision https://github.com/Workday/canvas-kit/commit/3c083cea09e685c4231c95d89160b4985f377a5c
I have not tested previous versions.

### Error Output

<img width="92" alt="110%" src="https://user-images.githubusercontent.com/55417192/71034474-651d6b00-20ce-11ea-8461-d0eebddcbc02.png">

PR #393, opened 2020-01-05
fix(radio): Update checked dot position to scale properly

erik-vanlankvelt:
#308  Summary
Closes: #379
CSS for selected Radio doesn't maintain centered dot at non-100% screen zoom

## Checklist
- [x] `yarn test` passes

## Additional References
Screen Shot 2020-01-05 at 2 22 55 PM
Screen Shot 2020-01-05 at 2 23 05 PM
Screen Shot 2020-01-05 at 2 23 12 PM
Screen Shot 2020-01-05 at 2 23 22 PM
Screen Shot 2020-01-05 at 2 23 40 PM


lychyi:
Hi there @erik-vanlankvelt, welcome to the party and thanks for the time and contribution. We'll take a look at it as soon as we can!

modules/radio/css/lib/radio.scss
```diff
@@ -3,13 +3,13 @@
 
   input + label::before {
     border-radius: $wdc-border-radius-circle;
-    background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg'%3E%3Cg%3E%3Ccircle fill='%23FFFFFF' cx='12' cy='12' r='4'%3E%3C/circle%3E%3C/g%3E%3C/svg%3E");
+    background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg'%3E%3Cg%3E%3Ccircle fill='%23FFFFFF' cx='9' cy='9' r='4.5'%3E%3C/circle%3E%3C/g%3E%3C/svg%3E");
```
stephanerangaya: Thank you for your contribution! Is there any reasons why you increase the radio of the white dot from 4 to 4.5? It looks like 4 is matching the react component whereas 4.5 is a tiny bit bigger.

NicholasBoll: I tested it and it is a bit bigger. We don't have a visual regression with the radio activated, so it wasn't caught. My guess is `4.5` was chosen because it is half of `9`?

erik-vanlankvelt: @stephanerangaya @NicholasBoll 
4.5 was chosen because it is half of 9.  When I tested in various browsers with it at 4, it still appeared to be slightly off.  Let me know if you feel the size should remain at 4 and the slight difference is marginal enough to resolve this issue.

anicholls:
I noticed another issue when testing this in Firefox. The radio shape is offset upwards, which means it doesn't align with the ripple in the background or the label text. Should we address this in this branch as well?
image

modules/radio/css/lib/radio.scss
anicholls: @erik-vanlankvelt Which browser(s) did you notice an issue with 4px? I think we would prefer to keep it at 4px since 4.5px would put make it subpixel.

erik-vanlankvelt: @anicholls 
I've updated the size to be 4px again.  I noticed a slight issue with the positioning of the white dot in Chrome and Safari at 75%.
Chrome 75%
Safari 75%

erik-vanlankvelt:
@anicholls 
Regarding the ripple in the background, why not address it too :)
I think I've done so and included screenshots in Firefox.
Firefox hover state
Firefox hover state selected

modules/radio/css/lib/radio.scss
erik-vanlankvelt: @anicholls please let me know if you feel there is more to be done here, thanks!

lychyi:
Thank you @erik-vanlankvelt 

Deeply appreciate the contribution 🙏

PR #394, opened 2020-01-06
test(text-area): Change tests to react-testing-library and improve coverage

stephanerangaya:
## Summary

## Checklist

- [x] `yarn test` passes


modules/text-area/react/spec/TextArea.spec.tsx
```diff
@@ -1,44 +1,81 @@
...
+    });
+
+    describe('wrapped in a FormField', () => {
+      test('should pass axe DOM accessibility guidelines', async () => {
+        const html = ReactDOMServer.renderToString(
+          <FormField label="My Field" inputId="my-text-area-field">
+            <TextArea id="my-text-area-field" onChange={cb} />
+          </FormField>
+        );
+        expect(await axe(html)).toHaveNoViolations();
```
NicholasBoll: I think this type of test should be moved to Cypress, right? Accessibility runs audits on our stories and not our unit tests.

This request can be part of our broader refactor of testing where component specifications are written and not part of this PR

anicholls: This may be good to include in the Cypress section of the testing docs

stephanerangaya:
> I think for StaticStates to work, we need to import the styled function from labs core, i forgot about this!

They work I think, aren't they?

modules/text-area/react/spec/TextArea.spec.tsx
```diff
@@ -1,44 +1,67 @@
...
+      expect(getByRole('textbox')).toHaveAttribute('data-propspread', attr);
+    });
+  });
+
+  describe('when provided an input ref', () => {
+    it('should set the ref to the input element', () => {
+      const ref: React.RefObject<HTMLInputElement> = React.createRef();
+      render(<TextArea inputRef={ref} id={id} />);
+      expect(ref.current).not.toBeNull();
+      expect(ref.current).toHaveAttribute('id', id);
```
NicholasBoll: I think the following should work:
```ts
expect(await getByRole('textbox')).toEqual(ref.current)
```

Also, as a side-note, I think it is safer for Testing Library APIs to be used with the promise API since rendering can be asynchronous. The synchronous API could break if we slightly change the implementation of components or React slightly changes timing (e.g. `useEffect` vs `componentDidMount` or concurrent mode)

NicholasBoll:
@mannycarrera4 Do you want to approve the IconButton differences? I think they are from the IconButton changes, but I wanted to be sure.

The other change is new visual states for the text area. I'm looping in UX to make sure we got it right and that it makes sense.

modules/text-area/react/spec/TextArea.spec.tsx
anicholls: @NicholasBoll Are you saying that we should aways be using `expect(await getBy*())` instead of `expect(getBy*())`?

NicholasBoll: I know a bit more about the React Testing Library APIs. There are `get*`, `query*`, and `find*` queries.

`find*` is most likely what we want for React because of timing. `find*` returns a promise that resolves when the element is found (retrying until it is found). If there is a delay in rendering something (hooks, concurrent mode, etc), `find*` will eventually find the element and return it for testing.

So I should update my previous code snippet:

```ts
expect(await findByRole('textbox')).toEqual(ref.current)
```

NicholasBoll: I'm not sure if `await find*` vs `get*` is strictly necessary. React Testing Library hooks into some testing internals to fix timing issues introduced with hooks. I just consider timing to be an implementation detail of the component that should not introduce failures or flakiness of tests.

anicholls: As discussed, we agreed to leave this change to a later PR because it will be an all-or-nothing change to our TL/R tests.

PR #396, opened 2020-01-07
fix(text-input): Removed the "clear" button on text input in IE 11

Parker-Ledoux:
## Summary

In IE 11, an `<input type="text"/>` natively displays an "X" icon to clear the input. This is undesired so using a pseudo selector to remove the icon.

## Additional References

Before:
Screen Shot 2020-01-07 at 9 50 31 AM

After:
Screen Shot 2020-01-07 at 10 05 26 AM


modules/text-input/react/lib/TextInput.tsx
```diff
@@ -42,6 +42,9 @@ const Input = styled('input')<Pick<TextInputProps, 'error' | 'grow'>>(
         color: inputColors.disabled.text,
       },
     },
+    '::-ms-clear': {
+      display: 'none',
+    },
```
mannycarrera4: just curious if any other of our inputs need this? good find!

Parker-Ledoux: @mannycarrera4 This is when the "X" icon will appear: https://developer.mozilla.org/en-US/docs/Web/CSS/::-ms-clear. I took a look at color input because it also has `type="text"` however the icon does not show up in IE 11. I didn't do too deep of a dive but I think it's because of how we offset the text in the input via styling.

PR #397, opened 2020-01-07
chore: Update Storybook Chromatic

NicholasBoll:
## Summary

Update to the latest Storybook Chromatic. This should fix the asset issue we've had with Firefox

### DIFF
--- a/modules/_labs/menu/react/package.json
+++ b/modules/_labs/menu/react/package.json
@@ -35,7 +35,7 @@
     "menu"
   ],
   "devDependencies": {
-    "@material-ui/core": "^4.1.0",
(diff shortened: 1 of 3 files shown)


modules/_labs/menu/react/package.json
```diff
@@ -35,7 +35,7 @@
     "menu"
   ],
   "devDependencies": {
-    "@material-ui/core": "^4.1.0",
+    "@material-ui/core": "^3.9.2",
```
NicholasBoll: This dependency was never updated in the `yarn.lock` file, so I reverted to a version used by other dependencies. Ideally this is not a dependency at all

PR #399, opened 2020-01-07
fix: Add missing rtl-css-js dependency

josephnle:
## Summary

`rtl-css-js` is available as a workspace dev dependency but `canvas-kit-labs-react-core` depends on `rtl-css-js` as a production dependency.

Thus, I get this error:
```
Can't resolve 'rtl-css-js' in '/node_modules/@workday/canvas-kit-labs-react-core/dist/es6/lib/theming'
```

## Checklist

- [x] `yarn test` passes


modules/_labs/core/react/package.json
```diff
@@ -41,7 +41,9 @@
     "@workday/canvas-colors-web": "^0.17.13",
     "@workday/canvas-kit-react-core": "^3.2.0",
     "chroma-js": "^2.1.0",
-    "lodash": "^4.17.14"
+    "emotion-theming": "^10.0.10",
+    "lodash": "^4.17.14",
+    "rtl-css-js": "^1.13.1"
```
anicholls: Thanks for catching this 👍. Can you remove `emotion-theming` and `rtl-css-js` in the root package.json since we've lowered it to this module?

NicholasBoll: Removing `emotion-theming` prevents stories from rendering

PR #404, opened 2020-01-09
fix: Update Storybook to fix Firefox asset issue

NicholasBoll:
## Summary

According to the Chromatic team, the Firefox asset issue should be fixed in Storybook 5.3.0-beta31. I'm upgrading to the latest 5.3.0 RC version. I've run various tests against this update. The CI should handle the rest.


lychyi:
@NicholasBoll This is pretty complicated with the webpack changes that may be involved. I move that we close this PR in favor of another when we have some time. #416

NicholasBoll:
I'm good with closing.

PR #407, opened 2020-01-09
test(colorpicker): Add tests and convert them to TL/R

sahlhoff:
Adds TL/R to `ColorInput` and `ColorPreview`.

#286 

It could be worth it to converge these into a single component and use the `readonly` prop for the variant `ColorPreview`. Could be a reason why we have two components but wanted to leave a note here.


modules/color-picker/react/spec/ColorInput.spec.tsx
```diff
@@ -1,115 +1,173 @@
 import * as React from 'react';
 import ColorInput from '../lib/ColorInput';
-import {mount} from 'enzyme';
+import {render, fireEvent} from '@testing-library/react';
 import ReactDOMServer from 'react-dom/server';
 import {axe} from 'jest-axe';
```
sahlhoff: Going to move these tests to cypress.

modules/color-picker/react/spec/ColorInput.spec.tsx
```diff
@@ -1,115 +1,150 @@
...
+    });
+
+    describe('with a value', () => {
+      test('should render a ColorInput with a value', () => {
+        const {getByTestId} = render(<ColorInput value={value} data-testid={id} />);
+        expect(getByTestId(id).value).toBe(value);
+      });
+    });
+
+    describe('with a value and check', () => {
```
lychyi: Not a comment on this test but I guess while you're in here with the stories, it may be good to add a story, example, or knob for the checked state, I realized that isn't demonstrated at all in our Storybook. It could also be a separate PR as well.

modules/color-picker/react/spec/ColorInput.spec.tsx
```diff
@@ -1,115 +1,150 @@
...
+
+    describe('when provided a value with a hashtag', () => {
+      test('the value is stripped', () => {
+        const {getByTestId} = render(<ColorInput value={'#eee'} data-testid={id} />);
+        expect(getByTestId(id).value).toBe(value);
+      });
+    });
+
+    describe('when provided more than 6 characters as the value', () => {
+      test('the value is stripped', () => {
```
lychyi: This looks like a copy and paste error from above? Something like `'the value is truncated to a length of 6'` or something like that

modules/color-picker/react/spec/ColorInput.spec.tsx
```diff
@@ -1,115 +1,150 @@
...
+    describe('when provided an input ref', () => {
+      test('input ref should be defined', () => {
+        const ref: React.RefObject<HTMLInputElement> = React.createRef();
+        render(<ColorInput inputRef={ref} />);
+        expect(ref.current).toBeDefined();
+      });
+    });
+
+    describe('when provided a value with a hashtag', () => {
+      test('the value is stripped', () => {
```
lychyi: From a failure perspective, it may not be clear what fails. Maybe something more specific like `'the hash is stripped from the value'`.

modules/color-picker/react/spec/ColorInput.spec.tsx
```diff
@@ -1,115 +1,150 @@
...
+
+    describe('when provided an input ref', () => {
+      test('input ref should be defined', () => {
+        const ref: React.RefObject<HTMLInputElement> = React.createRef();
+        render(<ColorInput inputRef={ref} />);
+        expect(ref.current).toBeDefined();
+      });
+    });
+
+    describe('when provided a value with a hashtag', () => {
```
lychyi: I think `#` is just 'hash' haha.

modules/color-picker/react/spec/ColorInput.spec.tsx
```diff
@@ -1,115 +1,150 @@
...
-describe('ColorInput Accessibility', () => {
-  test('ColorInput should pass axe DOM accessibility guidelines', async () => {
-    const html = ReactDOMServer.renderToString(
-      <FormField label="Label" inputId="input-plain">
-        <ColorInput />
-      </FormField>
-    );
-    expect(await axe(html)).toHaveNoViolations();
+  describe('when changed', () => {
+    test('should update value', () => {
```
lychyi: I don't think this test is doing what you think it's doing. You cover setting the value [here](https://github.com/Workday/canvas-kit/pull/407/files#diff-8a0bed2634f532ea7599ae6a06bed58cR35).

Since this is a controlled component, we probably don't need to test for how it updates the value, that is beyond the scope of this component.

modules/color-picker/react/spec/ColorInput.spec.tsx
```diff
@@ -1,115 +1,150 @@
...
+      const {getByTestId} = render(<ColorInput value="" data-testid={id} />);
+
+      expect(getByTestId(id).value).toBe('');
+      getByTestId(id).value = value;
+      expect(getByTestId(id).value).toBe(value);
+    });
+
+    test('should call onChange callback on value change', () => {
```
lychyi: Curious what others think here, I think most of these tests in `'when changed'` could just be functional tests in Cypress. They're not really testing API contracts at this point but rather functionality of the input.

modules/color-picker/react/spec/ColorInput.spec.tsx
sahlhoff: Yup, added a `Checked` story 😎

modules/color-picker/react/spec/ColorInput.spec.tsx
sahlhoff: I removed this test since it was redundant.

cypress/integration/ColorPicker.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    ['Default', 'Alert', 'Error', 'Checked', 'Grow', 'Grow with Error'].forEach(story => {
+      context(`given the '${story}' story is rendered`, () => {
+        beforeEach(() => {
+          h.stories.load(colorInputStory, story);
+        });
+
+        it('should pass accessibility checks', () => {
+          cy.checkA11y();
+        });
+      });
```
anicholls: Can you test inputting text into the input and checking if the color changes? We should be testing all interaction with the story here. Things like:

- Input a correct hex (e.g. `000000`)
- input a 3 digit hex (e.g. `000`)
- Input an invalid hex (e.g. `zzzzzz`)
- Input a string that's too long (e.g. `0000000`)
etc.

modules/color-picker/react/spec/ColorInput.spec.tsx
```diff
@@ -1,115 +1,140 @@
...
-  });
-  test('should set inputRef as reference to native input', () => {
-    const ref = React.createRef<HTMLInputElement>();
-    mount(<ColorInput inputRef={ref} value={''} />);
-    expect(ref.current && ref.current.tagName && ref.current.tagName.toUpperCase()).toBe('INPUT');
-  });
+  describe('when rendered', () => {
+    describe('with a placeholder', () => {
+      test('should render a ColorInput with a placeholder', () => {
+        const {getByTestId} = render(
```
anicholls: Based on the [recommendations in the docs](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid), we should avoid `getByTestId` wherever possible. I think you'll be most successful with `getByRole('textbox')` here so it doesn't require any extra props.

modules/color-picker/react/spec/ColorInput.spec.tsx
```diff
@@ -1,115 +1,140 @@
...
+    describe('with a value and check', () => {
+      test('should render a ColorInput with a value and check', () => {
+        const {container} = render(<ColorInput value={value} showCheck={true} />);
+
+        expect(container.querySelector('svg')).toHaveClass('wd-icon-check-small');
+      });
+    });
+
+    describe('with a value and background', () => {
+      test('should render a ColorInput with a value and the value as a background', () => {
```
anicholls: This could be combined with the value test on line 32.

modules/color-picker/react/spec/ColorInput.spec.tsx
anicholls: Agreed! 👍 I think it's good to check that `onChange` and `onValidColorChange` is called, but a lot of the specific behavioral functionality (hash stripping, too long of input, etc.) could/should be checked in cypress

modules/form-field/react/stories/stories_ColorInput.tsx
```diff
@@ -124,3 +150,61 @@ storiesOf('Components|Inputs/Color Picker/Color Input/React/Left Label', module)
       {controlComponent(<ColorInput />)}
     </FormField>
   ));
+
+storiesOf('Components|Inputs/Color Picker/Color Input/React/Visual Testing', module)
+  .addParameters({component: ColorInput})
+  .addDecorator(withReadme(README))
+  .add('States', () => (
+    <StaticStates>
```
anicholls: Do you mind updating this with the changes in #418?

modules/color-picker/react/spec/ColorInput.spec.tsx
sahlhoff: Will do!

modules/color-picker/react/spec/ColorInput.spec.tsx
sahlhoff: Updated.

modules/form-field/react/stories/stories_ColorInput.tsx
sahlhoff: Updated.

Issue #358, opened 2019-12-09
feat: Add theming support to input components


anicholls:
## 🚀 Feature Proposal

The framework for theming recently got merged (#272). We need to add support for it within our components. As a first step, we should use the theme object color values within every component. We can look at breakpoints as the followup.

## Implementation

Simply extend the `Themeable` interface, and use the `theme` prop provided to our custom `styled` function to style things

```tsx
import {styled, Themeable} from '@workday/canvas-kit-labs-react-core';

export interface MyComponentProps extends Themeable, React.HTMLAttributes<HTMLDivElement> {
  // Props
}

// ...

const Container = styled('div')<MyComponentProps>(
  ({theme}) => ({
    background: theme.palette.primary.main
  })
);

// ...
```

Note: We're going to start with just strong colors (focus outline, blue, red, green, yellow, etc.). We will tackle backgrounds, default borders, etc. in a future update.

## Components
- [x] Checkbox (#456)
- [x] Color Picker (#411)
- [x] Radio (#457)
- [ ] Select (#570)
- [x] Switch (#465)
- [x] TextArea (#459)
- [x] TextInput (#411)

anicholls:
Seems like #416 is required to [remove the docs page for our theming and visual stories](https://github.com/Workday/canvas-kit/pull/411#discussion_r373733032)

jpante:
See  #570 for final themed component

PR #411, opened 2020-01-16
feat(text-input): Add theming to Text Input

mannycarrera4:
We're starting to add theming to our inputs. This is a draft using the the `theme` prop to change colors. The consumer must wrap their components in a `CanvasProvider` and then they can pass their custom theme to their child components through the `theme` prop.

Partially addresses https://github.com/Workday/canvas-kit/issues/358


modules/form-field/react/stories/stories_TextInput.tsx
```diff
@@ -152,3 +170,14 @@ storiesOf('Components|Inputs/Text Input/React/Left Label', module)
...
   ));
+
+storiesOf('Components|Inputs/Text Input/React/Visual', module)
+  .addParameters({component: TextInput})
+  .addDecorator(withReadme(README))
+  .add('Theming', () => (
+    <div>
+      <CanvasProvider theme={customTheme}>
+        <TextInput error={ErrorType.Alert} placeholder="Placeholder" />
+      </CanvasProvider>
```
mannycarrera4: Add more examples with different error states and one without a error state

modules/common/react/lib/styles/errorRing.ts
```diff
@@ -1,17 +1,22 @@
...
+    errorBoxShadow = `inset 0 0 0 1px ${
+      theme ? theme.palette.error.main : inputColors.error.border
+    }`;
   } else if (error === ErrorType.Alert) {
-    errorBorderColor = colors.cantaloupe600;
-    errorBoxShadow = `inset 0 0 0 2px ${inputColors.warning.border}`;
+    errorBorderColor = theme ? theme.palette.alert.darkest : colors.cantaloupe600;
+    errorBoxShadow = `inset 0 0 0 2px ${
+      theme ? theme.palette.alert.main : inputColors.warning.border
+    }`;
```
mannycarrera4: this bring up an interesting problem. For alert, we have two different colors for accessibility reasons, how do we communicate or handle something like this? Open to suggestions

modules/form-field/react/stories/stories_TextInput.tsx
```diff
@@ -2,15 +2,23 @@
 import * as React from 'react';
 import {storiesOf} from '@storybook/react';
 import withReadme from 'storybook-readme/with-readme';
-import {controlComponent} from '../../../../utils/storybook';
+import {controlComponent, customThemePallete} from '../../../../utils/storybook';
```
mannycarrera4: type

utils/storybook/CanvasProviderDecorator.tsx
```diff
@@ -3,13 +3,16 @@ import {
...
 } from '@workday/canvas-kit-labs-react-core';
 import {object} from '@storybook/addon-knobs';
 
 const label = 'theme';
 
-export default (storyFn: () => React.ReactNode) => (
-  <CanvasProvider theme={createCanvasTheme(object(label, defaultCanvasTheme))}>
+export default (storyFn: () => React.ReactNode, customTheme?: CanvasTheme) => (
+  <CanvasProvider
+    theme={customTheme ? customTheme : createCanvasTheme(object(label, defaultCanvasTheme))}
```
mannycarrera4: also open to this approach. This allows for the custom theme to override the default theme we have for each component. I added an extra param so that we can pass whatever custom theme object we'd like

utils/storybook/CustomTheme.ts
```diff
@@ -0,0 +1,22 @@
...
+    success: {
+      main: 'aquamarine',
+    },
+    neutral: {
+      main: 'gray',
+    },
+    common: {
+      focusOutline: 'purple',
+    },
+  },
```
mannycarrera4: This file can grow and be edited, rather than having multiple custom theme's per story.

modules/form-field/react/stories/stories_TextInput.tsx
```diff
@@ -206,4 +206,59 @@ storiesOf('Components|Inputs/Text Input/React/Visual Testing', module)
         )}
       </ComponentStatesTable>
     </StaticStates>
+  ))
+  .addParameters({
```
mannycarrera4: Something I noticed is that this gets overridden in `Docs` and not quite sure how to solve for that yet..

modules/text-input/react/lib/TextInput.tsx
```diff
@@ -59,7 +59,28 @@ const Input = styled('input')<Pick<TextInputProps, 'error' | 'grow'>>(
...
+      '&:hover': {
+        borderColor: theme.palette.neutral.main,
+      },
+      '&:focus:not([disabled])': {
+        borderColor: theme.palette.common.focusOutline,
+        boxShadow: `inset 0 0 0 1px ${theme.palette.common.focusOutline}`,
+        outline: 'none',
+      },
+      '&:disabled': {
+        backgroundColor: theme.palette.neutral.lightest,
```
mannycarrera4: Another thought about this is how do we communicate to the consumer when they're using a custom theme, that they need to use this specific property `theme.palette.neutral.lightest` to be able to override this

modules/form-field/react/stories/stories_TextInput.tsx
anicholls: 🤔 How about we just [remove the docs tab](https://github.com/storybookjs/storybook/blob/1e53a3ae7ab3d1b95fcc191087811f6a8e8a455c/examples/official-storybook/stories/addon-docs/addon-docs.stories.js#L51) for these stories? We could probably do it for all visual testing stories.

utils/storybook/customThemes.ts
```diff
@@ -0,0 +1,22 @@
+export const customColorTheme = {
```
NicholasBoll: Perhaps we want this to be a partial of the `CanvasTheme` object? This way we could get autocomplete.

anicholls: We have a `PartialCanvasTheme` type 😉 . `Partial<>` doesn't work because it's not recursive.

modules/form-field/react/stories/stories_TextInput.tsx
mannycarrera4: So i tried this approach by this might be in 5.3?

modules/text-input/react/lib/TextInput.tsx
```diff
@@ -59,7 +59,17 @@ const Input = styled('input')<Pick<TextInputProps, 'error' | 'grow'>>(
...
-    }
+    },
+  ({theme, error}) => {
+    return {
+      '&:focus:not([disabled])': {
+        borderColor: theme.palette.common.focusOutline,
+        boxShadow: `inset 0 0 0 1px ${theme.palette.common.focusOutline}`,
+        outline: 'none',
+      },
+      ...errorRing(error, theme),
```
anicholls: The error function on line 56 is no longer needed.

modules/common/react/lib/styles/errorRing.ts
```diff
@@ -1,17 +1,35 @@
...
 import {colors, inputColors} from '@workday/canvas-kit-react-core';
+import chroma from 'chroma-js';
 
-export default function errorRing(error?: ErrorType): CSSObject {
+export default function errorRing(error?: ErrorType, theme?: CanvasTheme): CSSObject {
   let errorBorderColor;
   let errorBoxShadow;
 
+  const isAccessibleThemeColor = (themeColor: CanvasTheme): boolean => {
+    return chroma.contrast(themeColor, colors.frenchVanilla100) < 3;
```
mannycarrera4: if it's above 3, it's accessible? i was trying to understand https://www.w3.org/TR/WCAG20-TECHS/G18.html
so should the number be 4? @anicholls

anicholls: Not sure you need to factor this into a separate function just for one line, but with the way you've written it, it would be `>= 3`. Anything 3 or above passes. You can say do nothing if it's 3 or above, or make a special case if it's less than 3.

anicholls:
Since we use a `TextInput` for the `ColorInput`, can you add a theming story for that as well?

modules/form-field/react/stories/stories_TextInput.tsx
anicholls: Okay let's just wait until #416 goes in 👍

Issue #354, opened 2019-12-06
Add Slider Component 


devonsoto77:
## 🚀 Feature Proposal

A slider component, there will be two different variations. One with a direct input and one without. 

## Example

Used to set the height and width for legends within charts. 

## Designs

<img width="534" alt="Screen Shot 2019-12-06 at 11 12 31 AM" src="https://user-images.githubusercontent.com/17392486/70345589-a59cff00-1819-11ea-8674-8a82407490ad.png">

<img width="490" alt="Screen Shot 2019-12-06 at 11 16 06 AM" src="https://user-images.githubusercontent.com/17392486/70345685-d3824380-1819-11ea-83c2-fdf2d5ec7803.png">

alanbsmith:
We closed this PR (#413) a while back but forgot to close the issue.

PR #367, opened 2019-12-11
DNM feat(labs): Adds a slider component

devonsoto77:
Currently I am waiting for additional designs from mele. 

 
Introduces a slider component with two different variations. 
1. With a input box
2. Without an input box

## Summary

The motivation for this component is to use it for the aurora support (charts ) team. 

## Checklist

- [x] branch has been rebased on the latest master commit
- [x] tests are changed or added
- [x] `yarn test` passes
- [x] all (dev)dependencies that the module needs is added to its `package.json`
- [x] code has been documented and, if applicable, usage described in README.md
- [ ] module has been added to `canvas-kit-react` and/or `canvas-kit-css` universal modules, if
      applicable
- [ ] design approved final implementation
- [ ] a11y approved final implementation
- [ ] code adheres to the [API & Pattern guidelines](../API_PATTERN_GUIDELINES.md)

PR #413, opened 2020-01-16
feat(labs): Add slider component

laurenraddatz:
## Summary

this is a continuation of #367. the main difference is some styling changes, including styling for chrome/safari/opera, firefox, and ie11. there is no way to style with css for all browsers, hence the amount of css in this. you can read more about this problem [here](https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/).

## Checklist

- [x] `yarn test` passes
      applicable

## Additional References

slider

Fixes https://github.com/Workday/canvas-kit/issues/354

### DIFF
--- a/modules/_labs/slider/react/README.md
+++ b/modules/_labs/slider/react/README.md
@@ -4,7 +4,7 @@
   <img src="https://img.shields.io/badge/LABS-alpha-orange" alt="LABS: Alpha" />
 </a>  This component is work in progress and currently in pre-release.
 
-slider component 
+slider component
 
 ## Installation
 
@@ -24,19 +24,78 @@ yarn add @workday/canvas-kit-react-slider
 import * as React from 'react';
 import Slider from '@workday/canvas-kit-react-slider';
 
-<Slider />;
+const sliderProps: SliderProps = {
+  max: 100,
+  min: 0,
+  startValue: 27,
+  useInputRange: true,
+};
+
+<Slider {...sliderProps} />;
 ```
-
-## Static Properties
-
-> None
 
 ## Component Props
 
 ### Required
 
-> None
+#### `max: number`
+
+> Max number for the slider
+
+---
+
+#### `min: number`
+
+> Min number for the slider
+
+---
+
+#### `startValue: number`
+
+> The starting value for the slider
+
+---
 
 ### Optional
 
-> None
+#### step: number
+
+> How much to increase or decrease the value by
+
+---
+
+#### useInputRange: boolean
+
+> If specified, the input box for the value will show up
+
+---
+
+#### onChange?: (newValue: number) => void
+
+> If specified, this callback is executed when the value is changed
+
+---
+
+#### onSliderStartDrag?: (newValue: number) => void
+
+> If specified, this callback is executed when the mouse is pressed
+
+---
+
+#### onKeyDown?: (newValue: number) => void
+
+> If specified, this callback is executed when a key is pressed
+
+---
+
+#### onKeyUp?: (newValue: number) => void
+
+> If specified, this callback is executed when a key is released
+
+---
+
+#### onSliderEndDrag?: (newValue: number) => void
+
+> If specified, this callback is executed when the mouse is released
+
+---
--- a/modules/_labs/slider/react/README.md
+++ b/modules/_labs/slider/react/README.md
@@ -1,0 +1,42 @@
+# Canvas Kit React Slider
+
+<a href="https://github.com/Workday/canvas-kit/tree/master/modules/_labs/README.md">
+  <img src="https://img.shields.io/badge/LABS-alpha-orange" alt="LABS: Alpha" />
+</a>  This component is work in progress and currently in pre-release.
+
+slider component 
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-react
+```
+
+or
+
+```sh
+yarn add @workday/canvas-kit-react-slider
+```
+
+## Usage
+
+```tsx
+import * as React from 'react';
+import Slider from '@workday/canvas-kit-react-slider';
+
+<Slider />;
+```
+
+## Static Properties
+
+> None
+
+## Component Props
+
+### Required
+
+> None
+
+### Optional
+
+> None
--- a/modules/_labs/slider/react/README.md
+++ b/modules/_labs/slider/react/README.md
@@ -9,13 +9,13 @@ slider component
 ## Installation
 
 ```sh
-yarn add @workday/canvas-kit-react
+yarn add @workday/canvas-kit-labs-react
 ```
 
 or
 
 ```sh
-yarn add @workday/canvas-kit-react-slider
+yarn add @workday/canvas-kit-labs-react-slider
 ```
 
 ## Usage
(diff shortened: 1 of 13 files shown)


modules/_labs/slider/react/lib/Slider.tsx
```diff
@@ -0,0 +1,269 @@
...
+  border: 'none',
+  width: sliderContainerWidth,
+  minWidth: sliderContainerMinWidth,
+  height: '100%',
+  cursor: 'pointer',
+  WebkitAppearance: 'none',
+  MozAppearance: 'none',
+  background: 'transparent',
+  position: 'absolute',
+  zIndex: 1,
```
NicholasBoll: From what I can tell, `z-index` is needed because the `SliderInput` is defined before the `ProgressBarContainer` in the DOM. Is it possible to swap the definition of the 2 and remove this line?

I'm always cautious of `z-index` changes. There is often unintended side effects of changing it.

modules/_labs/slider/react/README.md
```diff
@@ -0,0 +1,101 @@
...
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-react
+```
+
+or
+
+```sh
+yarn add @workday/canvas-kit-react-slider
```
mannycarrera4: this is an error in our script, but this should read `@workday/canvas-kit-labs-react-slider`

modules/_labs/slider/react/README.md
```diff
@@ -0,0 +1,101 @@
...
+<a href="https://github.com/Workday/canvas-kit/tree/master/modules/_labs/README.md">
+  <img src="https://img.shields.io/badge/LABS-alpha-orange" alt="LABS: Alpha" />
+</a>  This component is work in progress and currently in pre-release.
+
+slider component
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-react
```
mannycarrera4: `@workday/canvas-kit-labs-react`

modules/_labs/slider/react/spec/Slider.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+    max: 100,
+    min: 1,
+    startValue: 50,
+  };
+  test('render with required props', () => {
+    const wrapper = mount(<Slider {...defaultSliderProps} />);
+    expect(wrapper.props().max).toEqual(100);
+    expect(wrapper.props().min).toEqual(1);
+    expect(wrapper.props().startValue).toEqual(50);
+    wrapper.unmount();
```
NicholasBoll: React Testing Library tests the observable output of a component rather than the implementation details. For instance, if you pass in `min` and `max`, the observable output of the DOM is that there is a `min` and `max` set on the `input[type=range]` element.

```tsx
const { findByRole } = render(<Slider min={1} max={100} value={50} />)

const rangeInput = await findByRole('slider')
expect(rangeInput).toHaveAttribute('min', '1')
expect(rangeInput).toHaveAttribute('max', '100')
expect(rangeInput).toHaveValue(50)
```

If the `toHaveAttribute` fails, it will say something like "Expected element to have attribute: min=1" vs "Expected x to equal 1" which is more descriptive. Same with `toHaveValue`

modules/_labs/slider/react/lib/Slider.tsx
```diff
@@ -0,0 +1,264 @@
...
+  type,
+} from '@workday/canvas-kit-react-core';
+import {focusRing} from '@workday/canvas-kit-react-common';
+
+import styled from '@emotion/styled';
+
+export interface SliderProps {
+  max: number;
+  min: number;
+  startValue: number;
```
NicholasBoll: I see this uses `startValue` instead of `value`. I think this should be a [controlled component](https://reactjs.org/docs/forms.html#controlled-components) like other form fields

modules/_labs/slider/react/lib/Slider.tsx
```diff
@@ -0,0 +1,264 @@
...
+  borderRadius,
+  colors,
+  spacingNumbers as spacing,
+  type,
+} from '@workday/canvas-kit-react-core';
+import {focusRing} from '@workday/canvas-kit-react-common';
+
+import styled from '@emotion/styled';
+
+export interface SliderProps {
```
NicholasBoll: Most components merge with a container. This allows applications to send extra attributes to something. I don't know what that should be in this case? The `input[type=range]`? The container?

modules/_labs/slider/react/lib/Slider.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+const ProgressBarContainer = styled('div')({
+  position: 'absolute',
+  height: sliderTrackHeight,
+  top: '50%',
+  width: '100%',
+  transform: 'translateY(-50%)',
+});
+
+const ProgressBar = styled('progress')<{value: number; max: number}>({
```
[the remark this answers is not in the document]
laurenraddatz: webkit doesn't support a way to style the range track like a progress bar, unfortunately. firefox and ie do.

modules/_labs/slider/react/lib/Slider.tsx
```diff
@@ -0,0 +1,264 @@
...
+        <SliderInput
+          type="range"
+          ref={inputRef}
+          aria-valuemax={max}
+          max={max}
+          aria-valuemin={min}
+          min={min}
+          step={step}
+          aria-valuenow={value}
+          value={value}
```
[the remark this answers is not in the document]
laurenraddatz: nope! will fix

modules/_labs/slider/react/lib/Slider.tsx
```diff
@@ -0,0 +1,243 @@
...
+          id={id}
+          type="range"
+          ref={inputRef}
+          max={max}
+          min={min}
+          step={step}
+          value={value}
+          onChange={onChangeHandler}
+          onMouseDown={onSliderDragStart}
+          onMouseUp={onSliderDragEnd}
```
vibdev: I think there are a couple aria attributes that can improve the accessibility of this:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role

laurenraddatz:
can no longer work on this, will pick it back up or allow someone else to uptake 🙂

modules/_labs/slider/react/lib/Slider.tsx
NicholasBoll: Are you talking about these?:
* aria-valuemin
* aria-valuemax
* aria-valuenow

`role=slider` is a superclass of `input[type=range]`. Do you know if these attributes are redundant given the `input` is a range already?

vibdev: A quick test with Voice Over seems like they are redundant, but might want to double check with Sam for other systems

alanbsmith:
We very much appreciate all the work that went into this PR. ❤️ Unfortunately, it's become stale from sitting for so long. 😞 I'm going to close it for now until we have a clear direction on adding Slider to CK. We can certainly reopen it later when we have the work prioritized. 👍🏻 Thanks again for the contribution and apologies we dropped the ball on this.

PR #419, opened 2020-01-23
fix(modal): Use React portals for accessibility fixes

### LINKED ISSUE #330: Modal: Component is constrained by overflow content
## 🐛 Bug Report

Modal is constrained by overflow content when conflicting stacking contexts are created.

The `Modal` uses `position: fixed` on the containing element which normally takes it out of the flow of the document and ignores overflow constraints (like `overflow: hidden` or `overflow: scroll`). But when combined with certain style properties like `transform`, `perspective` and `filter`, the new stacking context created is constrained by the parent component's overflow container.

The fix would be to use a Portal to take the modal out of the standard layout of the containing element and thus not be restricted by the stacking context of the parent element. This could be a breaking change if `z-index` is used on parent elements. For example if a parent element of a Modal uses `position` and `z-index` outside defaults, the Modal could show behind these elements unless a `z-index` was also passed to the Modal component. It is possible to have a very large `z-index` applied to the Modal container. As long as all Portalled content has the same z-index values, the browser should stack them based on order in the DOM. That would mean all content that uses a Portal would need to have the same z-index value (tooltips and other Popups). Many applications use a Popup Manager to manage z-index values (even though they probably shouldn't - Portals can be used instead).

## To Reproduce

Put the Modal into an overflow constrained container with a `transform` style applied to it. When the Modal is opened, it will be constrained by that overflowed container.

## Expected Behavior

The Modal should not be constrained by the overflowed parent element

## Actual Results

The Modal is constrained by the overflowed parent element

## Browser (if applicable)

This applies to browsers differently. It was found in Chrome. More info here: https://developer.mozilla.org/en-US/docs/Web/CSS/position#fixed

https://codesandbox.io/s/n32vx

### Error Output

Screen Shot 2019-11-25 at 12 36 43 PM

### LINKED ISSUE #414: Modal is not fully accessible
## 🐛 Bug Report

The modal component traps keyboard focus and hides content behind an overlay, but does not hide contents for assistive technologies. To hide contents for assistive technology, `aria-hidden` must be added to content that is not part of the modal. This will prevent users from accessing content outside the modal.

This work is related to #330 

## To Reproduce

Using VoiceOver in MacOS:
* Navigate to https://workday.github.io/canvas-kit/iframe.html?id=components-popups-modal-react--default
* Open the Modal
* Enable VoiceOver
* Open Web Rotor (default keyboard shortcut is Control+Option+U)
* Hit the right key until "Form Controls" is active

## Expected Behavior

"Delete Item" or "Delete Item button" should not appear on the list of controls (depending on VoiceOver verbosity)

## Actual Results

"Delete Item button" appears on the list of controls

## Browser (if applicable)

Doesn't matter

Screen Shot showing 'Delete Item Button' in the list of form controls

Additional internal info:
https://jira2.workday.com/browse/CANVAS-541

NicholasBoll:
## Summary

Modal now uses a portal for content to not be constrained by a parent container. This also allows for the hiding of non-modal content (visibly covered by modal overlay) to assistive technology.
Fixes #414
Fixes #330

## Checklist

- [x] `yarn test` passes


cypress/integration/Modal.spec.ts
```diff
@@ -6,7 +6,6 @@ function getModalTargetButton() {
 
 describe('Modal', () => {
   before(() => {
-    cy.viewport(500, 300);
```
NicholasBoll: Not actually sure why this is here.

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -172,7 +176,28 @@ const ModalContent = ({
     }
   };
 
-  return (
+  React.useEffect(() => {
+    const siblings = [...((document.body.children as any) as HTMLElement[])].filter(
+      el => el !== modalRef.current!.parentElement
```
alexandrzavalii: I wonder if we should actually check if current is not undefined?
```
  const observed = useRef(null);

  useEffect(() => {
     if(observed.current) console.log(observed.current);
  }, [observed]);
```

modules/modal/react/README.md
```diff
@@ -117,6 +117,18 @@ Default: `PopupPadding.l`
 
 Default: `true`
 
+### `shouldUsePortal: boolean`
```
anicholls: Since we've decided to put this in 4.0, can we remove this prop entirely and just make this the default?

Are there any cases where we wouldn't want someone to use a portaled modal?

modules/modal/react/stories/stories.tsx
```diff
@@ -25,7 +25,56 @@ const DefaultModalExample = () => {
       <Button variant={Button.Variant.Delete} buttonRef={buttonRef} onClick={openModal}>
         Delete Item
       </Button>
-      <Modal data-testid="TestModal" heading="Delete Item" open={open} handleClose={closeModal}>
+      <p>
+        <a href="#">Link</a>
+      </p>
+
+      <button type="button">Button</button>
```
anicholls: Why do we need all this extra content?

anicholls:
@NicholasBoll Should we be adding portals to all of our popups? If it happens for modals, I would also expect it to happen for Popup, Toast, etc.

modules/modal/react/lib/Modal.tsx
```diff
@@ -15,21 +15,8 @@ export interface ModalProps extends ModalContentProps {
   open: boolean;
 }
 
-const Modal = ({
-  open,
-  shouldUsePortal = false,
-  ...modalContentProps
-}: ModalProps): JSX.Element | null => {
-  if (process.env.NODE_ENV !== 'production') {
-    React.useEffect(() => {
```
mannycarrera4: @anicholls do we still need this whole block for non production?

anicholls:
Something worth noting that I just noticed. `InputProvider` will not work with any portalled components because they are no longer a child of the element that it controls. Maybe we need to consider making `InputProvider` set its data attributes on the body element.

modules/modal/react/lib/Modal.tsx
```diff
@@ -15,8 +15,9 @@ export interface ModalProps extends ModalContentProps {
   open: boolean;
 }
 
-const Modal = ({open, ...modalContentProps}: ModalProps): JSX.Element | null =>
-  open ? <ModalContent {...modalContentProps} /> : null;
+const Modal = ({open, ...modalContentProps}: ModalProps): JSX.Element | null => {
+  return open ? <ModalContent {...modalContentProps} /> : null;
```
anicholls: Nit: Can we revert this change?

NicholasBoll:
> Maybe we need to consider making `InputProvider` set its data attributes on the body element.

This is an open question for all things we do in Canvas Kit that effect element outside a component's render tree. This current change set `aria-hidden="true"` on other elements outside the control of the Modal, but that is required to work properly. I made a compromise by allowing the "container" to be set to anything and defaulting to `document.body`. Hopefully that is a good compromise. Documentation should state this is happening in case it causes issues.

NicholasBoll:
> @NicholasBoll Should we be adding portals to all of our popups? If it happens for modals, I would also expect it to happen for Popup, Toast, etc.

I think this should happen for all Popups. I think portalling should be the default unless we have a good reason not to. Portalling can complicate focus management if the portalled content has focusable elements inside. For the most part, focusable elements should be avoided due to complications with accessibility anyways. If a focusable element is not avoidable, have a conversation with accessibility about how to handle this. One compromise was capturing a `blur` event outside the portalled content and forcing focus back on the target element and closing the portalled popup. That way the next `tab` key is the native browser focus management.

If browsers came with a focus manager, things would be different: https://discourse.wicg.io/t/proposal-focus-traversal-api/3427

modules/modal/react/stories/stories.tsx
NicholasBoll: Accessibility requested this for testing if content was really being hidden from assistive technology

mannycarrera4: should i add this back in?

modules/modal/react/stories/stories.tsx
```diff
@@ -24,50 +24,7 @@ const DefaultModalExample = () => {
     <>
       <DeleteButton buttonRef={buttonRef} onClick={openModal}>
         Delete Item
-      </Button>
-      <p>
-        <a href="#">Link</a>
```
NicholasBoll: Ooo. Removed in a merge commit. Sneaky

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -172,7 +173,28 @@ const ModalContent = ({
     }
   };
 
-  return (
+  React.useEffect(() => {
+    const siblings = [...((document.body.children as any) as HTMLElement[])].filter(
```
NicholasBoll: I forgot to make this line and https://github.com/Workday/canvas-kit/pull/419/files#diff-3fe25d4d6af6314872e682fe13b7f7ccR212 use a configurable container instead of just `document.body`. Applications may need a different container than `document.body`

modules/modal/react/lib/ModalContent.tsx
NicholasBoll: Interesting... Will `useEffect` ever be called before the ref is applied? Technically this is an unsupported use-case. If this case happens, we cannot guarantee accessibility of the Modal

anicholls:
> I think this should happen for all Popups. I think portalling should be the default unless we have a good reason not to. 

@NicholasBoll are we going to look at that in this PR or in a followup? If the latter, we should create an issue for it

NicholasBoll:
> > I think this should happen for all Popups. I think portalling should be the default unless we have a good reason not to.
> 
> @NicholasBoll are we going to look at that in this PR or in a followup? If the latter, we should create an issue for it

I think the latter.

modules/modal/react/stories/stories.tsx
NicholasBoll: Maybe do that in an accessible story? Maybe behind a knob setting called `extraElements` that defaults to `false`? I don't think it is strictly necessary. We don't have a way to test this automatically, but it is nice for manual verification.

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -46,6 +47,11 @@ export interface ModalContentProps extends React.HTMLAttributes<HTMLDivElement>
    * it will make the modal heading focusable and focus on that instead.
    */
   firstFocusRef?: React.RefObject<HTMLElement>;
+  /**
+   * Is used to configure where the Modal will portal into.
```
NicholasBoll:
```suggestion
   * The containing element for the Modal elements. The Modal uses
   * {@link https://reactjs.org/docs/portals.html Portals} to place the DOM elements
   * of the Modal in a different place in the DOM to prevent issues with overflowed containers.
   * When the modal is opened, `aria-hidden` will be added to siblings to hide background
   * content from assistive technology like it is visibly hidden from sighted users. This property
   * should be set to the element that the application root goes - not containing element of content.
   * This should be a sibling or higher than the header and navigation elements of the application.
```

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -186,6 +220,8 @@ const ModalContent = ({
       </Popup>
     </Container>
   );
+
+  return ReactDOM.createPortal(content, document.body);
```
NicholasBoll:
```suggestion
  return ReactDOM.createPortal(content, container);
```

Issue #420, opened 2020-01-23
Header icons squished in "@workday/canvas-kit-labs-react-header": "3.2.0",


ginna-baker:
## 🐛 Bug Report

The header icons (notifications, user avatar, inbox, chat) on the right hand side are squished compared to the designs.

## To Reproduce

Load PEX in an SUV, observe icons.

You can also see visual regression snapshots taken by @NicholasBoll:

https://www.chromaticqa.com/component?appId=5d854c26ba934e0020f5e98a&name=Labs%7CHeader%2FReact&mode=snapshot&buildNumber=629&specName=Global%20Header&browser=ie

## Expected Behavior

The icons should match the designs.

## Actual Results

The icons are squished and some are not completely visible

## Browser (if applicable)
ie11.   Nicholas is also seeing it in Safari and Chrome.  (I think?  I couldn't reproduce)

### Error Output
![Screen Shot 2020-01-23 at 11 32 18 AM](https://user-images.githubusercontent.com/6732956/73023131-1c04b900-3de8-11ea-8828-ec297512e8a1.png)

Include a screenshot of a visual bug

PR #422, opened 2020-01-23
Fix(header) icon sizing and spacing

vibdev:
## Summary

This fixes issue #420 where header icons are squished in IE 11. Also turns out this also fixes an issue with the centeredNav prop that was masked by not having a centeredMenu ={true} with no leftSlot story. 

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/_labs/header/react/lib/Header.tsx
+++ b/modules/_labs/header/react/lib/Header.tsx
@@ -183,6 +183,7 @@ const ChildrenSlot = styled('div')<Pick<HeaderProps, 'centeredNav' | 'themeColor
       cursor: 'pointer',
     },
     display: 'flex',
+    flexGrow: 0,
     alignItems: 'center',
     justifyContent: 'flex-end',
     height: '100%',
--- a/modules/_labs/header/react/lib/Header.tsx
+++ b/modules/_labs/header/react/lib/Header.tsx
@@ -94,7 +94,7 @@ const BrandLink = styled('a')({
   },
 });
 
-const navStyle = ({themeColor}: Pick<HeaderProps, 'themeColor'>) => {
+const navStyle = ({ themeColor, centeredNav }: Pick<HeaderProps, 'themeColor' | 'centeredNav'>) => {
   const theme = themes[themeColor];
 
   return css({
@@ -103,7 +103,7 @@ const navStyle = ({themeColor}: Pick<HeaderProps, 'themeColor'>) => {
       flexGrow: 1,
       justifyContent: 'center',
       height: 'inherit',
-      marginLeft: spacing.xl,
+      marginLeft: !centeredNav ? spacing.xl : 0,
 
       '& ul': {
         color: theme.linkColor,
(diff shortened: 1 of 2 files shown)


modules/_labs/header/react/lib/Header.tsx
```diff
@@ -197,6 +197,7 @@ const ChildrenSlot = styled('div')<Pick<HeaderProps, 'centeredNav' | 'themeColor
     '> *:last-child': {
       marginRight: isCollapsed ? '' : 0,
     },
+    width: !isCollapsed ? '100%' : 'auto',
```
ginna-baker: should this be auto or just 'unset'?

vibdev:
Looks like there is an issue with the Full Header stories, nav is centered now

lychyi:
@vibdev Updated the branch again, trying to get these snapshots to pass on the right baselines

@ginna-baker thanks for the review!

modules/_labs/header/react/stories/stories.tsx
```diff
@@ -333,7 +333,7 @@ storiesOf('Labs|Header/React', module)
       <div css={containerStyle}>
         <Header
           variant={Header.Variant.Dub}
-          title="Centered Menu"
+          title="Centered Menu With Search"
```
ginna-baker: nb Looks like this added text is what's pushing the Logout button over the boundary on the screenshot https://www.chromaticqa.com/snapshot?appId=5d854c26ba934e0020f5e98a&id=5e2b41d359313d0022640f60

vibdev: since the responsive stuff is by prop the smaller screenshots can get squished or overflow

vibdev:
@lychyi Thanks for fixing the snaps. For the chromatic, looks like icons and centering is better now. The only thing that is standing out to me is the first dub header (without no icons or avatar) the search bar is now left aligned. I think that is more correct, but wanted to double check.

lychyi:
@vibdev @ginna-baker

I added an alternative approach in #427 if you want to take a look. I think it fixes it in the spirit that the header and the slots were supposed to work (fixed branding slot, then evenly flexed children). Let me know if the fix makes sense to you and if adding `overflow: hidden` or not affects your use cases.

vibdev:
@lychyi you method seems to work, but still have a question about the first dub header (without no icons or avatar); the search bar is now left aligned. Seems kinda weird that the left slot prop is on the right in just that one case. RE: overflow: hidden that breaks search and avatar dropdowns.

modules/_labs/header/react/stories/stories.tsx
NicholasBoll: IE11 handles this lack of space differently. IE11 is squishing the navigation tabs while Chrome and Firefox are pushing the "Logout" button to the right.

Should the Header or the navigation tabs handle not having enough space? Maybe by compressing into a dropdown menu instead? Or is the app supposed to detect this and do something different?

vibdev: I think user of the component user should decide and pass isCollapsed at the correct time. I think that allows for the most flexibility and makes it so the header doesn't have to do expensive measuring of all different parts anytime there is a change.

vibdev:
closing in favor of https://github.com/Workday/canvas-kit/pull/427

Issue #385, opened 2019-12-26
Error and Alert Hint Label are hard coded


stephanerangaya:
![image](https://user-images.githubusercontent.com/723109/71482769-9996d280-2804-11ea-8650-12cec27bd898.png)

At the moment, the text for the Alert and Error hint are hard coded, we should add support for a (optional?) prop to support translations.

```
let errorLabel: string | undefined;
switch (error) {
  case Hint.ErrorType.Error:
    errorLabel = 'Error';
    break;
  case Hint.ErrorType.Alert:
    errorLabel = 'Alert';
    break;
  default:
}
```

PR #423, opened 2020-01-24
fix: Remove hard coded strings for alert and error and added new props

mannycarrera4:
Fixes: https://github.com/Workday/canvas-kit/issues/385
With the translation efforts, we forgot to update the `error` and `alert` labels for the error messages. We also had to add a prop to `FormField` to allow customization.

### DIFF
--- a/modules/form-field/react/lib/FormField.tsx
+++ b/modules/form-field/react/lib/FormField.tsx
@@ -51,13 +51,13 @@ export interface FormFieldProps
    */
   children: React.ReactNode;
   /**
-   * The Label for the error messages. This props is only added for translation, otherwise it shouldn't change
-   * @default Error
+   * The label for the error message hint text if `hintText` and `error` are defined. This prop should only be used for translating the default string 'Error'.
+   * @default 'Error'
    */
   errorLabel?: string;
   /**
-   * The Label for the error messages. This props is only added for translation, otherwise it shouldn't change
-   * @default Alert
+   * The label for the alert message hint text if `hintText` and `error` are defined. This prop should only be used for translating the default string 'Alert'.
+   * @default 'Alert'
    */
   alertLabel?: string;
 }
--- a/modules/form-field/react/lib/FormField.tsx
+++ b/modules/form-field/react/lib/FormField.tsx
@@ -20,6 +20,8 @@ export interface FormFieldProps
   required?: boolean;
   useFieldset: boolean;
   children: React.ReactNode;
+  errorLabel?: string;
+  alertLabel?: string;
 }
 
 export interface FormFieldErrorBehavior {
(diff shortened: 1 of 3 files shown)


mannycarrera4:
I should probably add a test to FormField but i think we want to re write these test to use react testing library as well.

modules/form-field/react/spec/Hint.spec.tsx
```diff
@@ -1,28 +1,24 @@
...
-    const component = mount(<Hint data-propspread="test" />);
-    const container = component.at(0).getDOMNode();
-    expect(container.getAttribute('data-propspread')).toBe('test');
-    component.unmount();
+  describe('when rendered with a custom Error string', () => {
+    it('should render a custom label for type Error', async () => {
+      const {getByTestId} = render(
+        <Hint errorLabel={'Hay un Error'} data-testid="hintError" error={Hint.ErrorType.Error} />
+      );
+      expect(getByTestId('hintError').children[0].textContent).toEqual('Hay un Error: ');
```
mannycarrera4: I'm not sure this is the best way to get the text, I'm open to improving this! @anicholls @NicholasBoll

anicholls: Rather than finding the div and then checking its text content, I would just use `getByText`!

modules/form-field/react/lib/FormField.tsx
```diff
@@ -50,6 +50,8 @@ export interface FormFieldProps
    * The input component wrapped by the FormField.
    */
   children: React.ReactNode;
+  errorLabel?: string;
```
anicholls: Can you document that these props are specifically for translations and should not be used to customize the label? Also defaults in the jsdoc woudl be nice 👍

modules/form-field/react/lib/Hint.tsx
```diff
@@ -8,6 +8,8 @@ export interface HintProps extends Themeable, React.HTMLAttributes<HTMLParagraph
    * The type of error associated with the Hint (if applicable).
    */
   error?: ErrorType;
+  errorLabel?: string;
```
anicholls: Same here for docs

modules/form-field/react/lib/Hint.tsx
```diff
@@ -16,24 +18,27 @@ const Message = styled('p')(type.body2, {width: '100%', margin: `${spacing.xxs}
...
-        errorLabel = 'Alert';
+        hintLabel = alertLabel;
         break;
       default:
     }
 
     return (
       <Message {...this.props}>
-        {typeof error !== 'undefined' && <Label>{errorLabel && `${errorLabel}: `}</Label>}
+        {typeof error !== 'undefined' && <Label>{hintLabel && `${hintLabel}: `}</Label>}
```
anicholls:
```suggestion
        {typeof error !== 'undefined' && hintLabel && <Label>{hintLabel}: </Label>}
```
A little cleaner and no possibility of an empty `<Label>`

PR #427, opened 2020-01-26
fix(header): Allow elements to flex correctly in IE11

### LINKED ISSUE #420: Header icons squished in "@workday/canvas-kit-labs-react-header": "3.2.0",

lychyi:
## Summary

An alternative solution from the one proposed in #422

This fixes #420, a bug where the slot containing the nav and icon buttons were getting squished. This was because the slot containing the nav and icons shrunk past the default content size. In addition, icon buttons weren't sized properly in IE11, they need a min-width set, otherwise they collapse.

**Before**:
Screen Shot 2020-01-26 at 12 11 51 PM

[Flexbugs](https://github.com/philipwalton/flexbugs#flexbug-1) describes this bug: "Minimum content sizing of flex items not honored" in IE11.

> The flexbox spec defines an initial flex-shrink value of 1 but says items should not shrink below their default minimum content size. You can usually get this same behavior by setting a flex-shrink value of 0 (instead of the default 1) and a flex-basis value of auto. That will cause the flex item to be at least as big as its width or height (if declared) or its default content size.

By setting a min-width in the `Search` component and `flex: 1 0 auto` in the header (so it respects content sizing), the everything flexes properly in IE11 now.

This PR also adds: 
* a `min-width` equal to the `width` of an `IconButton` so it doesn't collapse if the container gets too small. 
* an `overflow: hidden` to the header shell so stuff doesn't leak out of the header, but I think it's debatable if this needs to be there or is considered a breaking change. @vibdev what do you think?
* an example to the story that shows a dub header without the search bar that #422 added
* a fix for vertical centering by fixing height values in the `Search` component

**After:**
Screen Shot 2020-01-27 at 4 57 01 AM

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/_labs/header/react/lib/Header.tsx
+++ b/modules/_labs/header/react/lib/Header.tsx
@@ -98,7 +98,7 @@ const BrandLink = styled('a')({
   },
 });
 
-const navStyle = ({themeColor, leftSlot}: Pick<HeaderProps, 'themeColor' | 'leftSlot'>) => {
+const navStyle = ({themeColor}: Pick<HeaderProps, 'themeColor'>) => {
   const theme = themes[themeColor];
 
   return css({
@@ -107,7 +107,7 @@ const navStyle = ({themeColor, leftSlot}: Pick<HeaderProps, 'themeColor' | 'left
       flex: `1 0 auto`, // Instead of just flex-grow: 1 for IE11, see https://github.com/philipwalton/flexbugs#flexbug-1
       justifyContent: 'center',
       height: 'inherit',
-      marginLeft: leftSlot ? spacing.xl : 0,
+      marginLeft: spacing.xl,
 
       '& ul': {
         color: theme.linkColor,
@@ -179,9 +179,7 @@ const navStyle = ({themeColor, leftSlot}: Pick<HeaderProps, 'themeColor' | 'left
   });
 };
 
-const ChildrenSlot = styled('div')<
-  Pick<HeaderProps, 'centeredNav' | 'themeColor' | 'isCollapsed' | 'leftSlot'>
->(
+const ChildrenSlot = styled('div')<Pick<HeaderProps, 'centeredNav' | 'themeColor' | 'isCollapsed'>>(
   {
     marginRight: spacing.m,
     // TODO: remove this when we get real icon buttons
@@ -391,7 +389,6 @@ export default class Header extends React.Component<HeaderProps, {}> {
           themeColor={themeColor}
           centeredNav={shouldCenteredNav}
           isCollapsed={isCollapsed}
-          leftSlot={leftSlot}
         >
           {isCollapsed ? (
             // Screen size is smaller than our largest breakpoint so turn nav into a hamburger
--- a/modules/_labs/header/react/lib/Header.tsx
+++ b/modules/_labs/header/react/lib/Header.tsx
@@ -180,7 +180,7 @@ const navStyle = ({themeColor, leftSlot}: Pick<HeaderProps, 'themeColor' | 'left
 };
 
 const ChildrenSlot = styled('div')<
-  Pick<HeaderProps, 'centeredNav' | 'themeColor' | 'isCollapsed' | 'leftslot'>
+  Pick<HeaderProps, 'centeredNav' | 'themeColor' | 'isCollapsed' | 'leftSlot'>
 >(
   {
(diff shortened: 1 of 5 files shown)


modules/_labs/header/react/lib/Header.tsx
```diff
@@ -67,6 +67,7 @@ const HeaderShell = styled('div')<Pick<HeaderProps, 'variant' | 'themeColor'>>(
     WebkitFontSmoothing: 'antialiased',
     MozOsxFontSmoothing: 'grayscale',
     position: 'relative',
+    overflow: 'hidden',
```
lychyi: This prevents stuff from flowing out of the header when the viewport shrinks

vibdev: I don't think we can have this. It block the dropdowns, search autocomplete and avatar menu

modules/_labs/header/react/lib/Header.tsx
```diff
@@ -201,7 +202,7 @@ const ChildrenSlot = styled('div')<Pick<HeaderProps, 'centeredNav' | 'themeColor
     '> *:last-child': {
       marginRight: isCollapsed ? '' : 0,
     },
-    flexGrow: !isCollapsed && centeredNav ? 1 : 'unset',
+    flex: !isCollapsed && centeredNav ? `1 0 auto` : 'unset',
```
vibdev: Is this line doing anything, when I was testing it didn't seem to effect the centering of the menu.

modules/_labs/header/react/lib/Header.tsx
```diff
@@ -104,7 +105,7 @@ const navStyle = ({themeColor}: Pick<HeaderProps, 'themeColor'>) => {
   return css({
     nav: {
       display: 'flex',
-      flexGrow: 1,
+      flex: `1 0 auto`, // Instead of just flex-grow: 1 for IE11, see https://github.com/philipwalton/flexbugs#flexbug-1
       justifyContent: 'center',
       height: 'inherit',
       marginLeft: spacing.xl,
```
vibdev: I think this margin left shouldn't be here when the menu is suppose to be centered, otherwise it is pushed a bit right

modules/_labs/header/react/lib/Header.tsx
lychyi: This allows the nav to be centered in your new example in #422, without it, the inner `nav` element would look like this:

Screen Shot 2020-01-27 at 8 21 23 AM

modules/_labs/header/react/lib/Header.tsx
lychyi: I can't find the specs anymore but at one time it was probably used to compensate for all the margins of the branding. I'll explicitly set it to 0 here (otherwise it sets to 16px somewhere else)

modules/_labs/header/react/lib/Header.tsx
lychyi: 🤦‍♂ That's right, totally forgot about that.

modules/_labs/header/react/lib/Header.tsx
lychyi: Oh actually it's for spacing against the `leftSlot`. I'll make the `ChildrenSlot` `leftSlot` aware so if it's been included, it'll have a 40px margin-left... if not it'll be flush (0px margin-left).

lychyi: nvm folks, this margin is for many things, we'll have to think of the header more holistically, I'll open an issue for this that we can track later.

PR #430, opened 2020-01-28
fix(labs): Remove global type from useTheme hook

lychyi:
## Summary

The global type declaration overrides an internal Workday window object. We are going to remove the type declaration but should also rethink where the fallback should be for an internal use case of `useTheme` (or if we should even include it here?)

## Checklist

- [ ] `yarn test` passes
      applicable


lychyi:
@anicholls newp, I resolved it. I'll merge after checks pass

Issue #324, opened 2019-11-19
Add a CI check to ensure all necessary modules are in the universal bundle modules


anicholls:
## 🚀 Feature Proposal

We should have a CI check to ensure all necessary modules are in `@workday/canvas-kit-css` and `@workday/canvas-kit-react`. We need to exclude all labs modules and the fonts module. Everything else should be included.

## Motivation

We've missed some of these in the past (e.g. https://github.com/Workday/canvas-kit/pull/322).

PR #437, opened 2020-01-30
ci: Add check for missing or unused dependencies

anicholls:
## Summary

This is a different take on https://github.com/Workday/canvas-kit/pull/293 using [depcheck](https://github.com/depcheck/depcheck).

Closes https://github.com/Workday/canvas-kit/issues/324

We occasionally don't catch missing dependencies in our package.json files. Due to how the mono repo works we do not notice this locally and we end up publishing packages that come with a `Module not found` error. This adds a utility script to check for missing and unused dependencies. I've added it as a script in our travis config so that it will run for PRs.

Note: I could not the scss checking to resolve properly so it thought all of our deps were unused. I think it may be something to do with the `~` imports. Will need to play with it a bit more.

## Additional References

image

### DIFF
--- a/modules/form-field/react/stories/stories_Select.tsx
+++ b/modules/form-field/react/stories/stories_Select.tsx
@@ -6,7 +6,7 @@ import {controlComponent} from '../../../../utils/storybook';
 
 import FormField from '..';
 import README from '../../../select/react/README.md';
-import {Select, SelectOption} from '../../../select/react/index';
+import {Select, SelectOption} from '../../../select/react';
 
 const hintText = 'Helpful text goes here.';
 const hintId = 'error-desc-id';
--- a/modules/form-field/react/stories/stories_Select.tsx
+++ b/modules/form-field/react/stories/stories_Select.tsx
@@ -6,7 +6,7 @@ import {controlComponent} from '../../../../utils/storybook';
 
 import FormField from '..';
 import README from '../../../select/react/README.md';
-import {Select, SelectOption} from '@workday/canvas-kit-react-select';
+import {Select, SelectOption} from '../../../select/react/index';
 
 const hintText = 'Helpful text goes here.';
 const hintId = 'error-desc-id';
--- a/package.json
+++ b/package.json
@@ -136,7 +136,8 @@
     "test": "TZ=UTC jest",
     "bump": "lerna version",
     "cypress:open": "cypress open",
-    "cypress:run": "cypress run"
+    "cypress:run": "cypress run",
+    "depcheck": "./utils/check-dependencies-exist.js"
   },
   "config": {
     "commitizen": {
@@ -161,5 +162,8 @@
   "workspaces": [
     "modules/**"
   ],
-  "browserslist": "> 5%, last 2 firefox versions, last 2 chrome versions, last 2 safari versions, last 2 edge versions, ie 11"
+  "browserslist": "> 5%, last 2 firefox versions, last 2 chrome versions, last 2 safari versions, last 2 edge versions, ie 11",
+  "dependencies": {
+    "depcheck": "^0.9.2"
+  }
 }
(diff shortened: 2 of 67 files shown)


modules/form-field/react/stories/stories_Select.tsx
```diff
@@ -6,7 +6,7 @@ import {controlComponent} from '../../../../utils/storybook';
 
 import FormField from '..';
 import README from '../../../select/react/README.md';
-import {Select, SelectOption} from '@workday/canvas-kit-react-select';
+import {Select, SelectOption} from '../../../select/react/index';
```
anicholls: Open question. We have a bit of a split on the methodology used when relying on other components for a story (dev dependency). 

The majority of places use relative imports like this. The plus side is that it won't cause errors with this script and we don't have to manually manage the dev deps. The downside is that it's less similar to how consumers would use our components.

There are three options. We should probably standardize.

1. Module imports (`import {Something} from '@workday/canvas-kit-react-something'`). Requires adding a dev dependency for each component
2. Relative imports (`import {Something} from '../../../something/react'`). Requires no dev dependency management
3. Universal module (`import {Something} from '@workday/canvas-kit-react'`). Requires only managing the one dev dependency (plus anything that's not exported as part of the universal module).

Thoughts?

lychyi: I believe this change was made so we don't have to build when developing. Otherwise this references the dist folder and a dev would have to build packages when working on a story.

anicholls: @lychyi Maybe originally, but if there is no dist folder (which we never have now), it resolves to the workspace npm link. So that's not really an issue anymore

package.json
```diff
@@ -136,7 +136,8 @@
     "test": "TZ=UTC jest",
     "bump": "lerna version",
     "cypress:open": "cypress open",
-    "cypress:run": "cypress run"
+    "cypress:run": "cypress run",
+    "depcheck": "./utils/check-dependencies-exist.js"
```
lychyi: Should we set this to prebuild too? Instead of running it explicitly in Travis?

modules/form-field/react/stories/stories_Select.tsx
lychyi: But if you have tested a build for some reason a dist folder exists and it will always default to that right? You could spend a bunch of time trying to fix something without knowing that's actually not the code you're testing. With relative imports there's no indirection involved.

utils/check-dependencies-exist.js
```diff
@@ -0,0 +1,132 @@
...
+            .join('\n');
+        }
+      })
+      .join('\n')
+  );
+}
+
+const searchPath = `${path.resolve(__dirname)}/../modules/**/react`;
+
+glob(searchPath, undefined, function(er, modules) {
```
lychyi: A lot of this logic is searching for the proper folders. Wouldn't it be simpler to update each package to run depcheck as a prebuild step in its own module instead of at the root? Much like we have builds in their respective modules?

modules/form-field/react/stories/stories_Select.tsx
anicholls: Good point. Okay I will standardize on relative imports

anicholls:
I think this is ready for another look @lychyi. 

I did some exploration into why SCSS isn't working. It seems our folder based imports (i.e. `@import '~@workday/canvas-kit-css-core/lib/mixins'`) are not expanded to pull in the individual .scss files. However, I'm not sure why this isn't working because they are using the [same method](https://github.com/depcheck/depcheck/blob/master/src/parser/sass.js#L10) as [we do](https://github.com/Workday/canvas-kit/blob/master/utils/css-build.js#L51) to get/render the sass.

Any thoughts on the above would be appreciated

package.json
anicholls: @lychyi Just realized we don't run `yarn build` as part of Travis PR builds (meaning `prebuild` doesn't get run). Do you think we should keep it in both places?

Issue #387, opened 2019-12-28
Switch Component Improvements


lychyi:
## 🚀 Feature Proposal

After doing a thorough test re-write (#386) of `Switch`, there are some improvements we can make to this component.

- I don't think we need to specify tabIndex={0}, this is the default already.
- We should map a static member Switch.ErrorType.Alert to ErrorType from canvas-kit-react-common. It's kind of a pain to import that also just to use Switch.
- Per some conversations that @NicholasBoll had with a11y in the past, we should probably make the role attribute a configurable prop. Instead of checkbox it could be something like switch.
- We should probably rename the elements SwitchBackground and SwitchCircle to something more semantically correct like SwitchTrack and SwitchThumb.

## Motivation

Our components can always be a bit better 🌞

jaclynjessup:
Closing to address in #1774

PR #442, opened 2020-02-05
test: Add @testing-library/cypress to the project

NicholasBoll:
## Summary

`@testing-library/cypress` is now in a state where we can recommend use in Canvas Kit.

## Checklist

- [x] `yarn test` passes

## Additional References

PR that updated the commands: https://github.com/testing-library/cypress-testing-library/pull/108


cypress/integration/Switch.spec.ts
```diff
@@ -1,7 +1,7 @@
 import * as h from '../helpers';
 
 const getSwitch = () => {
-  return cy.get(`[type="checkbox"]`);
+  return cy.findByRole('checkbox');
```
NicholasBoll: This change isn't strictly necessary, but does demonstrate it is working. I'm not sure `checkbox` is the correct role for this component yet, but I didn't update implementation.

lychyi: Yup, we should be addressing this in #387

PR #443, opened 2020-02-05
ci: Add canary build script on master commit/merge

anicholls:
## Summary

To ensure consumers can test out hot fixes as soon as possible, we would like to introduce Canary builds for each merge to master. This will make every PR available immediately after its merged rather than waiting for our ~2 week release cycle. Hopefully this will also reduce the number of ad hoc releases we need to do.

### Version Format:

Rather than using tags to indicate the version like we do for release (because this would tag every commit on master), we will tie canary builds to commit SHAs using lerna's [`--canary`](https://github.com/lerna/lerna/tree/master/commands/publish#--canary) flag
- The default behavior is `3.3.3-alpha.7+eddbcc7` (where `7` is the index of the commit from the last release (`3.3.2`)). 
- We can configure `alpha` here, but nothing else. I recommend we use `next` instead (e.g. `3.3.3-next.7+eddbcc7`).
- This means we will have 3 different npm [dist-tags](https://docs.npmjs.com/cli/dist-tag): `latest` (latest stable version), `prerelease` (alpha/beta versions), `next` (canary builds)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,7 @@ jobs:
       - yarn chromatic --auto-accept-changes --app-code="dlpro96xybh" # v1
       - yarn chromatic --auto-accept-changes # v2
       - yarn build-storybook
+      - npm config set //registry.npmjs.org/:_authToken=$NPM_PUBLISH_TOKEN && yarn lerna publish --yes --canary --preid next --pre-dist-tag next
     env:
       - CHROMATIC_APP_CODE="m1dh5kc7oj"
     deploy:


NicholasBoll:
Just trying to understand our Travis file.

Pre-release (on `tag` _and_ branch contains "v" + "n.n" or "n.n.n"):
https://github.com/Workday/canvas-kit/blob/58c854f3d6aa638a2c8b70aad0d438b9d5dcec86/.travis.yml#L58-L66

Canary (on commits to `master`):
https://github.com/Workday/canvas-kit/blob/58c854f3d6aa638a2c8b70aad0d438b9d5dcec86/.travis.yml#L68-L82

Do we manually run `learna publish` for "latest"?

anicholls:
@NicholasBoll the default dist-tag is `latest`. 

In the `tag` stage, we use the flag `--pre-dist-tag prerelease`. This means that if the latest unpublished version is a prerelease, use the dist-tag `prerelease`. Otherwise, use the default (`latest`)

In the `master` stage, because we're publishing with `--canary`, every publish will be a prerelease. So when we add ` --pre-dist-tag next`, it means every release published from master merges will use the `next` dist-tag.

Does that clarify it?

.travis.yml
```diff
@@ -70,6 +70,7 @@ jobs:
       - yarn chromatic --auto-accept-changes --app-code="dlpro96xybh" # v1
       - yarn chromatic --auto-accept-changes # v2
       - yarn build-storybook
+      - npm config set //registry.npmjs.org/:_authToken=$NPM_PUBLISH_TOKEN && yarn lerna publish --yes --canary --preid next --pre-dist-tag next
```
anicholls: I guess technically `--dist-tag next` applies to both normal and prerelease versions, so that could be used here instead.

NicholasBoll:
Hushing the bots: https://github.com/Workday/canvas-kit/pull/443#event-3011032004

PR #448, opened 2020-02-06
ci: Force publish all modules for canary builds

anicholls:
## Summary

After struggling to get the version checks to work properly with `lerna publish --canary` and discussing how we'd like to publish canary modules, we've decided to force publish everything rather than only doing canary builds of the modules that have changed since the last release. This PR utilizes the `--force-publish` flag to accomplish this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ cache:
     - ~/.cache
 
 git:
-  depth: 50
+  depth: 3
 
 stages:
 - name: pull_request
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,7 +17,7 @@ cache:
     - ~/.cache
 
 git:
-  depth: 50
+  depth: 3
```
anicholls: We can reset the git clone depth back to 3 since we're no longer looking at the history to determine which modules need to be published.

PR #453, opened 2020-02-10
chore: Update Storybook to 5.3

### LINKED ISSUE #416: Upgrade to Storybook ^5.3
To fix some Firefox issues for ChromaticQA we'll need to update to the latest version of Storybook. The issue was fixed in an 5.3RC in Storybook.

However, updating this will probably require us to rethink, reconfigure our custom webpack config. Probably for the better...

lychyi:
Fixes #416

## Summary

Update Storybook to ^5.3.x. This fixes a Firefox browser issue with ChromaticQA where a delay in loading fonts triggers visual regressions. Unfortunately we don't have an issue to track but this is what the ChromaticQA team had us do for a fix.

## Checklist

- [ ] `yarn test` passes
      applicable


NicholasBoll:
The CI failure seems to be related to an issue where a type definition file created by a newer version of Typescript fails older versions of the compiler: https://github.com/aws-amplify/amplify-js/issues/4389#issuecomment-557700295

It looks like upgrading to Typescript 3.7+ is how we would fix the issue.

lychyi:
We could upgrade or we could petition Storybook to create declarations using 3.6.x.

Upgrading to 3.7 could possibly put our library consumers in a similar predicament.

This is sort of like Storybook adopting TS 4 without signaling a breaking change. There is a fairly large TS issue about this since a few people have viewed this declaration output change as a breaking change (but only happened from 3.6 to 3.7).

NicholasBoll:
We could open an issue against Storybook. Is there an issue on Typescript? Seems like that type of output change should have been behind a flag.

I'm assuming 3.7 can read output from 3.6.

That's a lesson to all of us... Bumping a shared library a minor version and making use of that feature is a breaking change.

lychyi:
For posterity purposes:

This is the [Storybook issue](https://github.com/storybookjs/storybook/issues/9463).

This is largely due to TypeScript's update to 3.6 to 3.7 where type definitions for getters are generated differently. This is [TS's response](https://github.com/microsoft/TypeScript/issues/33939), saying that it's working as intended.

For a lot of library maintainers and Angular folks who are stuck on `typescript@~3.6.0` the jump from 3.6 to 3.7 was actually a breaking change. And now, since it wasn't signaled that way from TypeScript, it is a pain to do simple things like go from Storybook 5.2 to 5.3.

I'm working on a PR for Storybook, but in the meantime, I'll put a comment on there to see if they know of a quicker fix.

lychyi:
This is getting resolved in Storybook itself. We're close 🤞 

https://github.com/storybookjs/storybook/pull/9847

lychyi:
Someone give me an amen @NicholasBoll

NicholasBoll:
We may want a blurb in our documentation that mentions this issue. Someone using Storybook need to have their version of Storybook synced for our theming to work, right? It doesn't necessarily matter which patch version we require.

PR #454, opened 2020-02-10
feat(select): Implement Canvas menu

### LINKED ISSUE #294: Select Input with Canvas Menu
## 🚀 Feature Proposal

Currently the `Select` component uses the native drop down.

To align to Canvas specifications the drop down should be styled to match the canvas menu style.

Screen Shot 2019-11-03 at 4 53 11 PM

jamesfan:
## Summary

Closes #294.

Modifies the Select component to use the Canvas-designed menu instead of the native select menu.

Note this is a **Labs** component. It can be found on Storybook under **Labs > Select**. The original Select component (with the native menu) remains in Storybook under **Components > Inputs > Select**.

_Last updated: April 22, 2020_

## Completed

- [x] Fix TypeScript errors
- [x] Confirm background colors for hover and selected states
- [x] Replace text `input` with `button` for Accessibility
- [x] Add `Alert`, `Disabled`, `Error`, `Grow`, and `Grow - Error` stories
- [x] Finalize animation after selecting an option
  - ~Seeking guidance from Accessibility on current flashing animation~ Design and Accessibility agree no animation is necessary.
- [x] Handle use case where there are many options
  - ~Awaiting spec from Design, but current guidance is to set a `max-height` to the menu and make it scrollable~ No spec necessary. First pass with `max-height` of `200px` was acceptable.
- [x] Implement remaining keyboard shortcuts
  - [x] Home/End
  - [x] Type-ahead
- [x] Finalize focus styles and behavior
  - Component shifts focus from the button to the menu as per Accessibility's recommendation
- [x] Add ARIA attributes
  - [x] `aria-activedescendant`
- [x] Update documentation
  - [x] README
  - [x] Prop descriptions
- [x] Ensure proper visuals in IE11
- [x] Modify focus behavior to shift focus back to the button on Tab/Shift-Tab/Esc rather than when we blur off the button (too aggressive).
- [x] Update component API to use render props (https://github.com/Workday/canvas-kit/pull/454#discussion_r397446714).
  - [x] Modify the options API to accept custom key/value pairs as part of a `data` object rather than as siblings of the existing option keys to prevent possible collisions with reserved keys (like `focused` and `selected`).
- [x] Ask Design if we need a visually distinct and uniformly formatted option for the null (nothing selected) state. The current implementation supports a null state in that you can pass in an entry in the options array that represents the null state, but it'll be rendered like every other option in the menu.
  - Design agrees the current implementation is acceptable: no need for a baked-in null option. Best practices for how to write select options will be documented on the Design Site.
- [x] Handle lengthy options

## For a Later PR

* Render menu using portals (#571)
* Add support for theming (#570)
* Convert components to FC/Hooks for improved code readability and reusability (see scrolling code in `SelectBase`)
* Add more tests
  * Active/hover visual testing states for the menu (currently, there are no active/hover styles for the menu so these wouldn't be particularly interesting, but this could change)
  * Make sure state (selected item, focused item) is preserved if the options change.
  * Fill out unit tests


modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,342 @@
...
+  handleKeyboardShortcuts = (event: React.KeyboardEvent): void => {
+    // console.log(event.key);
+
+    const children = React.Children.toArray(this.props.children);
+    const itemCount = children.length;
+    let isShortcut = false;
+    let nextFocusedIndex = 0;
+
+    switch (event.key) {
+      case 'ArrowUp':
```
NicholasBoll: Please see this table for compatible key events for IE11: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Example

NicholasBoll:
There are several Typescript errors that are preventing a Storybook example from running in Chroma

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,342 @@
...
+  ): React.ReactNode => {
+    return React.cloneElement(child, {
+      focused: this.state.focusedItemIndex === index,
+      justSelected: this.state.justSelectedItemIndex === index,
+      onMouseDown: (event: React.MouseEvent) => {
+        event.preventDefault();
+        this.handleOptionClick(child.props);
+      },
+      selected: this.state.selectedItemIndex === index,
+    });
```
NicholasBoll: I remember @irobato using Context instead of `React.cloneElement` for various reasons. One of them is cloning elements limits how options can be passed to the Select.

The API as it is now makes the assumption that there will not be many options and that they will always be known. What about larger selects that might need dynamic options? Maybe the `Select` component isn't a good component for those use-cases, so maybe this is fine?

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,342 @@
...
+    // offer visual feedback (flashing animation) briefly
+    // before dismissing menu; set new label when menu has been
+    // dismissed
+    setTimeout(() => {
+      this.setState({
+        justSelectedItemIndex: null,
+        label: optionProps.label,
+      });
+      this.toggleMenu(false);
+    }, dismissMenuTimeout);
```
NicholasBoll: Anytime there is a `setTimeout`, `setInterval` or anything that has a callback that works outside the lifecycle of the component, it _must_ be removed when the component unmounts. Otherwise you'll get errors like "Cannot call `setState` on an unmounted component". React hooks make this easier by colocating setup and teardown code. In React component classes, setup and teardown are separated by lifecycle methods...

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,342 @@
...
+        // modern browsers
+        event = new Event('input', {bubbles: true});
+      } else {
+        // IE 11
+        event = document.createEvent('Event');
+        event.initEvent('input', true);
+      }
+
+      this.inputRef.current.dispatchEvent(event);
+    }
```
NicholasBoll: Maybe add a comment about why this code is necessary. It is not common to see code like this, so it is not immediately obvious why such interesting code is required.

My guess is we're trying to change the value of an input programmatically in a way that forces native `change` and/or `input` events. Perhaps so that `onChange` listeners will get a native event?

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,342 @@
...
+          onClick={this.handleSelectClick}
+          onKeyDown={this.handleKeyboardShortcuts}
+          readOnly
+          ref={this.inputRef}
+          type="text"
+          {...elemProps}
+        />
+        {this.state.showingMenu && (
+          <SelectMenu>{React.Children.map(children, this.renderChildren)}</SelectMenu>
+        )}
```
NicholasBoll: There are several instances where existential UI like this should always be classified as a `Popup` and be portalled to the bottom of the `body` tag. We can provide a prop to opt-out of this feature.

The common problem is popups like this are constrained by container elements. If the select menu shows below the Select input and it is he bottom of a container, it will increase the size of the container but not scroll, causing the menu to be invisible to the user. If the `Select` is inside a Modal with an overflow container, the select menu will not be able to show outside the borders of the Modal

jamesfan:
@NicholasBoll Thanks for the feedback! I'll start addressing these fixes/questions shortly.

modules/_labs/select/react/lib/SelectOption.tsx
```diff
@@ -0,0 +1,96 @@
...
+const flashAnimation = keyframes`
+  0%, 49% {
+    background-color: ${commonColors.focusBackground};
+    color: ${typeColors.inverse};
+  }
+  50%, 100% {
+    background-color: transparent;
+    color: inherit;
+  }
+`;
```
NicholasBoll: The flashing animation is a bit harsh:

<details>
    <summary>Warning: Flashing animation</summary>
    
select-flashing

</details>

jamesfan: Yep, I tried to approximate the behavior of the native select, but looking at it again it is a bit much :). Timings/colors probably need to be adjusted.

Either way, I'm coordinating with Design on what the selection animation should be (this was just a guess, the final animation may be quite different).

modules/_labs/select/react/lib/Select.tsx
jamesfan: That's correct, I'll add some comments.

For context: I wanted to trigger the `change` event for the `SelectInput` when an option from the menu is selected so that `Select` could be used as a controlled component. I found a solution on Stack Overflow (https://stackoverflow.com/questions/23892547/what-is-the-best-way-to-trigger-onchange-event-in-react-js) before I noticed some very similar code had already been implemented in `Combobox` and ended up copying the code from there.

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,787 @@
...
+    }
+  };
+
+  // TODO: Move code for scrollIntoViewIfNeeded to a centralized place.
+  // TODO: This code also has minor issues (sometimes it scrolls
+  // unnecessarily, the centerIfNeeded doesn't always center)
+  // Lifted from https://gist.github.com/hsablonniere/2581101
+  // This scrolling behavior is preferable even to the WebKit-proprietary
+  // scrollIntoViewIfNeeded method.
+  private scrollIntoViewIfNeeded = (elem: HTMLElement, centerIfNeeded = true): void => {
```
jamesfan: What would be the recommended location for this scrolling code? I've pasted it directly into the Select component for now, but it should probably be placed elsewhere for possible reuse. I can also look into finding a package which provides the same functionality if that's preferable.

FYI: this function is used to keep the focused option in view when you use the keyboard to navigate a scrollable menu, and to scroll the selected option into view when you activate a scrollable menu.

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,787 @@
...
+      event.stopPropagation();
+      event.preventDefault();
+    }
+  };
+
+  renderChildren = (
+    child: React.ReactElement<SelectOptionProps>,
+    index: number
+  ): React.ReactNode => {
+    return React.cloneElement(child, {
```
jamesfan: Currently, `Select` injects several props into its `SelectOption` children via this `React.cloneElement` call: `error`, `focused`, `selected`, etc. These props are meant to be managed by the `Select` component and aren't intended to be explicitly passed into `SelectOption` (i.e., you should never use `<SelectOption focused={true} selected={true} ...>`).

For example, `selected` is a boolean which tells the `SelectOption` whether or not its selected -- the parent `Select` compares the `selectedOptionIndex` in its internal state against the index of the `SelectOption` and sets `selected` to `true` if they match.

Is there a different preferred method for implementing "private" props? @NicholasBoll mentioned using Context in an earlier comment (still investigating). I've also heard suggestions of render props (https://medium.com/better-programming/passing-data-to-props-children-in-react-5399baea0356).

modules/_labs/select/react/lib/Select.tsx
NicholasBoll: If the component is a functional component, this logic could be done as a React hook

modules/_labs/select/react/lib/Select.tsx
NicholasBoll: Render props would be my suggestion. The Tables 2.0 is going that direction. But also the way data is defined is different. The current API looks like this:

```tsx
<Select>
  <SelectItem />
  <SelectItem />
</Select>
```

While that is a nice clear API, it does have some issues with scaling. If you have a lot of items, you'll have to define them all. Possibly with a `map` function. The component will have to map over children again to add the extra props necessary. You'll have to make props like `selected`, `focused` optional so that Typescript doesn't complain to you. If we go for virtualization later (which we probably should for performance), this will still be a very expensive render because the render method is so large. Also the API is limited to only showing strings.

Another API would be slightly less obvious, but more flexible using a render prop:

```tsx
const options = [ { value: "United States", flag: "US" }, { value: "Ireland", flag: "IE" } ]
<Select
  options={options}
  renderItem={( itemProps, data ) => {
    <SelectOption {...itemProps} value={data.value}>
      <img src={`https://www.countryflags.io/${data.flag}/flat/32.png`} aria-hidden="true" /> {data.value}
    </SelectOption>
  }}
/>
```

This is a more flexible API where the rendering is more efficient and more obviously controlled by the `Select` component. We only need to process rendering of `SelectItems` only when the Select is opened. The render prop is called by the `Select` at the right time. Rendering will be more efficient. If the `Select` component optimizes by utilizing virtual rendering of items, our external API doesn't need to change and we don't waste time with expensive render methods. It is also more obvious that `options` can be extracted out of the render method entirely, allowing further optimizations by the `Select` component.

This API also makes it more obvious how dynamic `options`. If options comes back from the server, it will probably be an array anyways. We would just hand it to the component.

cypress/integration/SelectLabs.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+  });
+
+  // TODO: Re-enable these tests
+  /*
+  context(`given the 'Disabled' story is rendered`, () => {
+    beforeEach(() => {
+      h.stories.load('Components|Inputs/Select/React/Top Label', 'Disabled');
+    });
+
+    it('should pass accessibility checks', () => {
```
NicholasBoll: We should be more clear about what is being checked in `checkA11y`. It isn't really "accessibility checks", it is "axe-core" checks. Axe core doesn't catch all or even most accessibility issues. Our A11y team figures it is less than 20%. That's why we have manually specified accessibility specifications. We figure axe-core checks are easy enough to just throw in there just in case we missed something, but the output is harder to understand and know what action needs to be taken.

jamesfan: Yep, I remember you mentioning this earlier. I'll update the wording.

modules/_labs/select/react/lib/SelectOption.tsx
```diff
@@ -0,0 +1,111 @@
...
+  ({disabled, focused, selected, suppressed}) => {
+    if (disabled) {
+      return {
+        color: colors.licorice100,
+      };
+    } else if (selected) {
+      return {
+        backgroundColor: colors.blueberry100,
+      };
+    } else if (focused) {
```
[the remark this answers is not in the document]
jamesfan: Agreed -- this is based on the requirements I received from Design, but it seems odd to me as well. I'll double check with them to make sure this is what they really want.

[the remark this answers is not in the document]
melehamasaki: Hey guys - sorry about the mix up. Let's make focus take precedent over selected. Confirmed that Prompts has the same behavior so we should be good to go 👍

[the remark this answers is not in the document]
jamesfan: Awesome! Double-checking with a11y before I make the change, but I don't anticipate any issues.

modules/_labs/select/react/stories/stories.tsx
```diff
@@ -0,0 +1,283 @@
...
+import {Select, SelectOption} from '../index';
+import README from '../README.md';
+
+const hintText = 'Helpful text goes here.';
+const hintId = 'error-desc-id';
+
+storiesOf('Labs|Select/React/Top Label', module)
+  .addParameters({component: Select})
+  .addDecorator(withReadme(README))
+  .add('Plain', () => (
```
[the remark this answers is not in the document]
jamesfan: Fixed

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,787 @@
...
+      isShortcut = true;
+      this.handleKeyboardTypeAhead(event.key, numOptions);
+    } else {
+      switch (event.key) {
+        case 'ArrowUp':
+        case 'Up': // IE/Edge specific value
+        case 'ArrowDown':
+        case 'Down': // IE/Edge specific value
+          if (this.state.isMenuHidden) {
+            this.toggleMenu(true);
```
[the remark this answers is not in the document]
jamesfan: Fixed

modules/_labs/select/react/lib/SelectOption.tsx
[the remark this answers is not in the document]
jamesfan: Fixed

cypress/integration/SelectLabs.spec.ts
jamesfan: Fixed

modules/_labs/select/react/lib/Select.tsx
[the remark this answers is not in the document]
jamesfan: I inadvertently introduced a bug where space triggered the menu when it should have been treated as part of a type-ahead sequence instead (if the select had focus while the menu was hidden). This should be fixed in the latest build.

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,787 @@
...
+  ({error}) => ({
+    ...menuBorderCSS(error),
+  }),
+  ({isMenuHiding}) =>
+    isMenuHiding && {
+      animationName: fadeOutAnimation,
+    },
+  ({grow}) =>
+    grow && {
+      width: '100%',
```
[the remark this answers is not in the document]
jamesfan: No good reason :) I've set the menu width to 100% at all times and removed the `grow` prop from the menu as well.

modules/_labs/select/react/README.md
```diff
@@ -0,0 +1,170 @@
...
+##### `value?: string`
+
+> The value of the SelectOption
+
+##### `label?: string;`
+
+> The label of the SelectOption
+
+
+```sh
```
anicholls, line 139: Everything below this line needs to be deleted

modules/_labs/select/react/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Optional
+
+#### `disabled: boolean`
+
+> Whether or not the select is disabled (not able to be selected)
+
+Default: `false`
+
+# Canvas Kit SelectOption
```
anicholls, line 118: Nit:
```suggestion
# SelectOption
```

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,681 @@
...
+
+export interface SelectState {
+  focusedOptionIndex: number;
+  isMenuHidden: boolean;
+  isMenuHiding: boolean;
+  label: string;
+  selectedOptionIndex: number;
+}
+
+const focusButtonCSS = (): CSSObject => ({
```
anicholls, line 46: Nit: This doesn't need to be a function, it can just be a regular object
```suggestion
const focusButtonCSS = {
```

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -0,0 +1,681 @@
...
+    // control of focus
+    event.preventDefault();
+
+    this.toggleMenu(this.state.isMenuHidden);
+  };
+
+  handleSelectMouseDown = (event: React.MouseEvent<HTMLButtonElement>): void => {
+    // Cancel default handling of the event to ensure we maintain
+    // control of focus
+    event.preventDefault();
```
anicholls, line 491: Is this needed for both click and mousedown?

jamesfan:
> This is looking pretty great to me! 👍 Let me know when you finish the hooks/FC and render props updates and I will have another look

@anicholls Awesome, will do!

modules/_labs/select/react/lib/Select.tsx
NicholasBoll: Preventing default is only needed on mouse down since that's the default action of a `mousedown` event. It shouldn't also be done on click

modules/_labs/select/react/README.md
jamesfan: Previous README was just a copy-and-paste job from the non-Labs Select (with some weird typos apparently). I've updated the document to accurately represent the Labs Select.

modules/_labs/select/react/README.md
jamesfan: No longer applies with the new render prop API.

modules/_labs/select/react/lib/Select.tsx
jamesfan: Fixed

modules/_labs/select/react/lib/Select.tsx
jamesfan: @NicholasBoll @anicholls I've implemented the render prop (`renderOption`) in the latest changes if you'd like to check it out. I've included example code in the README as well as a `Custom Options` story.

It's a slightly modified version of Nicholas's proposed API -- I opted to have the render prop only handle the content of the `<SelectOption>` and not the `<SelectOption>` itself for simplicity.

It would be nice if we could use `renderOption` to render the actual select (which is actually a button) itself, but since it's a `<button>` we're limited in what we can include in its content. Something for a later PR.

jamesfan:
@NicholasBoll @anicholls

_I'm adding this as a brand new comment rather than adding it to the original thread since it's getting buried in the hidden items. Here's the original thread for reference: https://github.com/Workday/canvas-kit/pull/454#discussion_r395942602_

I've implemented the render prop (`renderOption`) in the latest changes if you'd like to check it out. I've included example code in the README as well as a `Custom Options` story.

It's a slightly modified version of Nicholas's proposed API -- I opted to have `renderOption` only handle the content of the `<SelectOption>` and not the `<SelectOption>` itself for simplicity.

It would be nice if we could use `renderOption` to render the actual Select itself (which is actually a button), but since it's a `<button>` we're limited in what we can include in its content. Something for a later PR perhaps.

modules/_labs/select/react/lib/Select.tsx
jamesfan: Resolved now that we're using a render prop instead of cloneElement

modules/_labs/select/react/lib/Select.tsx
jamesfan: Removed `preventDefault` from click handler.

modules/_labs/select/react/lib/Select.tsx
jamesfan: Implementing this as a hook in a later PR.

NicholasBoll:
It's unfortunate how these stories show up in documentation: 
Screen Shot 2020-04-23 at 5 11 49 PM

jamesfan:
> It's unfortunate how these stories show up in documentation:
> Screen Shot 2020-04-23 at 5 11 49 PM

Agreed, do you think it would be preferable to remove `defaultWithSimpleOptionsComponent` and replace it with the actual component code (at the expense of adding duplicate code and making it a bit more tedious to update stories)?

NicholasBoll:
> Agreed, do you think it would be preferable to remove `defaultWithSimpleOptionsComponent` and replace it with the actual component code (at the expense of adding duplicate code and making it a bit more tedious to update stories)?

Yes. A story's implementation is meant to show an example use-case that is as copy/paste-friendly as possible. We want the source to be easy to understand.

Ideally someone could just copy/paste everything in the "source" tab of the component and make their modifications, but unfortunately Storybook's source plugin doesn't do that very well. Despite the plugin's shortcomings, I'd like to make story source as good as we can without modifying Storybook or the plugins directly. Perhaps we'll fix the source at some point in the future.

Related: #597

cypress/helpers/selectLabs.ts
```diff
@@ -0,0 +1,93 @@
...
+      });
+
+      context('the first option', () => {
+        it('should have an aria-selected attribute set to "true"', () => {
+          getOption(0).should('have.attr', 'aria-selected', 'true');
+        });
+      });
+    }
+  );
+};
```
NicholasBoll, line 93: I didn't notice before, but we shouldn't have test code like this in helper files. In Modal, I simply duplicated code. We could move this to the test file instead. I tend to minimize the use of behavior as much as possible. We make exceptions for helpers where they give us higher-level semantics.

Ideally these helper functions will get exported as part of CK to help aid in downstream Cypress testing. The helpers here are a bit generic and only work really well when there is only one of the page.

PR #456, opened 2020-02-11
feat(checkbox): Add theming support

anicholls:
## Summary

We're adding theming to our inputs. The consumer must wrap their components in a `CanvasProvider` and then they can pass their custom theme to any child components through the `theme` prop.

Partially addresses #358 

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/common/react/lib/styles/focusRing.ts
+++ b/modules/common/react/lib/styles/focusRing.ts
@@ -1,5 +1,6 @@
 import {keyframes, CSSObject} from '@emotion/core';
 import canvas from '@workday/canvas-kit-react-core';
+import {CanvasTheme} from '@workday/canvas-kit-labs-react-core';
 import memoize from 'lodash/memoize';
 
 type FocusRingParams = {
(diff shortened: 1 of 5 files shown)


modules/common/react/lib/styles/focusRing.ts
```diff
@@ -80,3 +81,50 @@ export default function focusRing(
...
+}
+
+/**
+ * A utility to create a canvas style focus ring around your widget.
+ * By default, this mixin will create a 2px focus ring tightly wrapped
+ * to the container (no whitespace).
+ *
+ * @returns {CSSObject} the css object for the focus ring style
+ */
+export function themedFocusRing(
```
anicholls: Interested to hear thoughts on this. I did this for a few reasons:

1. Unfortunately because we are usually calling this from within a class component, we can't use `React.useContext`. This means we need to pass the `theme` into the `focusRing` function. It would have been pretty nasty passing in every prop just to get `theme` at the end.

2. I'm often frustrated by the unclear nature of these props when you see them in use. The added complexity of an options object was worth it for the legibility and possibility for future changes. This now looks like
`themedFocusRing(theme, {separation: 2, animate: false})` instead of `focusRing(2, 2, false)`. I realize IDE hinting could help, but I still prefer the options object.

3. Rather than trying to hack the existing `focusRing` function to accept individual params vs. an options object without a breaking change, I figured a new function would be cleaner. We can consider deprecating `focusRing` in the future.

modules/form-field/react/stories/stories_Checkbox.tsx
```diff
@@ -118,56 +119,64 @@ storiesOf('Components|Inputs/Checkbox/React/Left Label', module)
...
-          <Checkbox
-            {...props}
-            onChange={() => {}} // eslint-disable-line no-empty-function
-            label="Checkbox"
-          />
-        )}
-      </ComponentStatesTable>
-    </StaticStates>
-  ));
+  .add('States', () => <CheckboxStates />)
```
NicholasBoll: Story source won't be very useful for these stories, but that is _probably_ okay because we don't expect people to use the source code for these visual stories.

mannycarrera4: yep yep, I think we pulled this out into a variable because `StaticStates` won't make much sense to someone viewing the source code.

modules/common/react/lib/styles/focusRing.ts
NicholasBoll: I'm fine with this change. The `theme` context can be passed either through the Hooks API or the Context API. I think this function shouldn't be burdened to care _how_ `theme` gets here.

PR #459, opened 2020-02-11
feat(text-area): Add theming support

mannycarrera4:
## Summary

We're adding theming to our inputs. The consumer must wrap their components in a CanvasProvider and then they can pass their custom theme to any child components through the theme prop.
## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/form-field/react/stories/stories_TextArea.tsx
+++ b/modules/form-field/react/stories/stories_TextArea.tsx
@@ -3,7 +3,12 @@ import * as React from 'react';
 import {storiesOf} from '@storybook/react';
 import withReadme from 'storybook-readme/with-readme';
 import {StaticStates} from '@workday/canvas-kit-labs-react-core';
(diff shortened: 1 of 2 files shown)


modules/form-field/react/stories/stories_TextArea.tsx
```diff
@@ -3,14 +3,69 @@ import * as React from 'react';
...
+  customColorTheme,
+} from '../../../../utils/storybook';
 
 import {TextArea} from '../../../text-area/react';
 import FormField from '../index';
 import README from '../../../text-area/react/README.md';
 
 const hintText = 'Helpful text goes here.';
 const hintId = 'error-desc-id';
+const TextAreaStates = () => (
```
anicholls: Nit: can you move this to the bottom of the file above the visual testing stories? That way it's not the first thing people see when they look for examples.

Issue #368, opened 2019-12-11
New Button spreads on Deprecated Button styles


emjabs:
## 🐛 Bug Report

There's a `transition: all` happening in the Deprecated Button styles, and we're getting that style even though we're importing the new Button, because the new button spreads on the Deprecated Button styles.

## To Reproduce

For my use case, in Workday PT's Drive, resize the window until the left sidebar snaps to a smaller width by media query.

## Expected Behavior

Either no animation, the ability the toggle the animation off, or a much smoother animation.

## Actual Results

A jagged animation that lags behind the rest of the page changing size.

## Browser (if applicable)

Chrome

PR #460, opened 2020-02-12
fix(button): Avoid using transition all for buttons

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/368
## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/button/react/lib/ButtonStyles.ts
+++ b/modules/button/react/lib/ButtonStyles.ts
@@ -142,7 +142,7 @@ export const deprecatedButtonStyles: ButtonGenericStyle = {
     position: 'relative',
     cursor: 'pointer',
     outline: 'none',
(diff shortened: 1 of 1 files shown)


NicholasBoll:
It is hard to say if this will be a breaking change for anybody. It won't show up on visual regressions since it is a transition...

We may want a note somewhere stating this might change how properties update.

mannycarrera4:
@NicholasBoll i agree, do you have any ideas how we can communicate this?

NicholasBoll:
I think just put the note in the squashed commit message. I think that gets used in changlogs, right? Or do we only use summaries? @anicholls ?

anicholls:
@NicholasBoll @mannycarrera4 We could add a note in the PR summary (merge commit) message

Issue #376, opened 2019-12-16
Add color picker component


mannycarrera4:
## Motivation
ColorInput was created but we never got to the Color Picker. The Color Picker provides the ability to select colors an easily show colors a user can choose from.

PR #462, opened 2020-02-12
feat(color-picker): Add color picker

laurenraddatz:
## Summary

this the **_first iteration_** of a color picker for canvas! it is only the "card" component and its contents, with callbacks to pass event handlers.

this is designed to be modular, but focusing on the use case of having a color picker in a popup/dropdown/popover component, specifically. the component minimally contains state, but the callbacks on the component must be used with application state/hooks for full functionality.

see the [README](https://github.com/Workday/canvas-kit/blob/8e1121181fa8cf270e9fcd3faee70252aeebe7fc/modules/_labs/color-picker/react/README.md) for details on how to implement. accessibility has not fully reviewed it, yet.

Closes https://github.com/Workday/canvas-kit/issues/376

## Checklist

- [x] `yarn test` passes
      applicable

## Additional References

see the storybook for more interactive examples.
Screen Shot 2020-03-04 at 1 46 11 PM

note: the `Selected Color` button is not shipped. but could be?

note 2: I chose to create a separate component than what is used in the `Color Input` component. This is mostly because of the unique way it is styled, it didn't seem worth it right now to consolidate them.


modules/_labs/color-picker/react/spec/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,170 @@
+import * as React from 'react';
+import {fireEvent, render} from '@testing-library/react';
+import {colors} from '@workday/canvas-kit-react-core';
+import {ColorPicker, ColorPickerProps} from '@workday/canvas-kit-labs-react-color-picker';
+
+describe('<ColorPicker /> component', () => {
+  const dataTestId = 'TESTY_TEST';
+
+  const renderColorPicker = (props?: Partial<ColorPickerProps>) => {
+    return render(<ColorPicker data-test-id={dataTestId} onChange={jest.fn()} {...props} />);
```
anicholls: We don't use the test ID for anything. Can we remove?

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -0,0 +1,174 @@
...
+import {ResetButton} from './parts/ColorReset';
+
+export interface ColorPickerProps {
+  ariaLabel?: string;
+
+  onChange: (color: string) => void;
+  selectedColor?: string;
+  customColors?: string[];
+
+  showCustomInput?: boolean;
```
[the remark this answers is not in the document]
laurenraddatz: > Nit: Anything with defaults should not be typed as optional

can you expand on this? it has a default because it is optional, so I'm not sure what you mean

modules/_labs/color-picker/react/lib/parts/Swatch.tsx
```diff
@@ -0,0 +1,52 @@
...
+      : {}
+);
+
+export interface SwatchProps extends React.HTMLAttributes<HTMLDivElement> {
+  color: string;
+  showCheck: boolean;
+  style?: React.CSSProperties;
+}
+
+export const Swatch: React.FunctionComponent<SwatchProps> = ({color, showCheck, style}) => (
```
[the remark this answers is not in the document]
laurenraddatz: good idea!!

modules/_labs/color-picker/react/lib/ColorPicker.tsx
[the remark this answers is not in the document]
anicholls: Even for booleans, we still put them in `defaultProps`. Technically `showCustomInput` has a default value of `undefined`, not `false`. It doesn't make a difference here since you're only checking if it's truthy, but in some cases it can make a difference ([example](https://github.com/Workday/canvas-kit/blob/master/modules/button/react/README.md#toggled-boolean--undefined)). Just good practice 🙂 

In terms of optional typing, [anything in defaultProps should be typed as required, but documented as optional](https://github.com/Workday/canvas-kit/blob/master/API_PATTERN_GUIDELINES.md#default-props). This is a typescript thing.

Let me know if that clarifies it!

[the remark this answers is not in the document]
laurenraddatz: I don't think making these components class-based just to be able to use `defaultProps` is necessary? I'll go ahead and assign defaults while destructuring the props instead.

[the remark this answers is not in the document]
anicholls: It's still possible to add `ColorPicker.defaultProps`, keeping it a functional component. Defaults during destructuring are also a reasonable alternative.

Referencing some research @NicholasBoll did:

```tsx
// Example 1
// * Typescript understands `optional` is optional externally, but always defined inside render function
// * Storybook Docs can understand the properties and will make a prop table
// * Storybook Docs can understand default values (only non-functions)
// * `Foo.defaultProps` cannot be type-safe without intermediate object
​
interface FooProps {
  optional: string
}
​
export function Foo({ optional }: FooProps) {
  // optional is `string` here. Typescript knows it is always defined
  return <div>{optional}</div>
}
​
// Type checking is not available.
// Can't force to `Partial<FooProps>` without another object
Foo.defaultProps = {
  optional: 'bar',
}
​
// `optional` is not required. Typescript understands `defaultProps` makes it optional
const foo = <Foo />
​
// Example 2
// * `optional` is forced to be optional through interface - default provided by deconstruction
// * Storybook Docs can understand the properties and will make a prop table
// * Storybook Docs do _not_ understand default props
// * Type-checking default values is automatic
​
interface FooProps {
  optional?: string
}
​
export function Foo({ optional = 'bar' }: FooProps) {
  // optional is `string` here. Typescript knows it is always defined
  return <div>{optional}</div>
}
​
// `optional` is not required per interface
const foo = <Foo />
```

I lean towards the first method because the default props are able to be pulled out by the storybook docs add on properly. The lack of type checking when assigning defaultProps isn't a huge deal since it's usually a pretty intentional assignment. 

Using defaults during destructuring is cleaner (and would be my preference), but unfortunately our storybook docs plugin doesn't support it yet 🙁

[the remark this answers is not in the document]
anicholls: After a bit more conversation on Slack, we would all lean towards the second option and we're going to look into updating storybook docs to support it. Let's stick with your destructured defaults instead 🙂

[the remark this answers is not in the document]
laurenraddatz: sweet! thanks

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -0,0 +1,174 @@
...
+import {IconButton} from '@workday/canvas-kit-react-button';
+import {checkIcon} from '@workday/canvas-system-icons-web';
+
+import {SwatchBook} from './parts/SwatchBook';
+import {ResetButton} from './parts/ColorReset';
+
+export interface ColorPickerProps {
+  ariaLabel?: string;
+
+  onChange: (color: string) => void;
```
NicholasBoll: This signature brings up an open question I had. I created an issue for this: #487 

It is very easy to create a signature like this, but it has ramifications downstream to consumers of form fields.

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -0,0 +1,174 @@
...
+  selectedColor?: string;
+  customColors?: string[];
+
+  showCustomInput?: boolean;
+  customInputLabel?: string;
+  onSubmitClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+
+  onReset?: Function;
+  resetColor?: string;
+  resetLabel?: string;
```
NicholasBoll: We're in a weird transition right now. We use the `README.md` file to document props, but we now have support in Storybook to document props using JSDoc inside the prop interface. We will eventually deprecate documenting props in the README (duplicated and harder to maintain). In the meantime, we've added prop documentation to _both_

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -0,0 +1,174 @@
...
+        />
+      )}
+      <SwatchBook colors={colorList} onSelect={onChange} selectedColor={selectedColor} />
+      {showCustomInput && (
+        <ColorInputWrapper>
+          <ColorInputAndLabel>
+            <Label>{inputLabel}</Label>
+            <HexColorInput
+              aria-label="Color Input"
+              id="#canvas-color-picker--custom-input"
```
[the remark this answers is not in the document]
NicholasBoll: Why is the hash symbol used here? From what I can tell, the `id` prop here is forwarded to the `ColorInput` component which is a wrapper around an `<input>` element and the `id` is forwarded to this element, so we'll see in the DOM:

```html
<input id="#canvas-color-picker--custom-input" type="text">
```

Is this the intent?

Alex is right, I think we can use a `FormField` here which will automatically take care of the `id` value for us, both making it unique and linking the `label` and `input` together for accessibility.

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -0,0 +1,174 @@
...
+  onChange,
+  onSubmitClick,
+  onReset,
+  resetLabel,
+  resetColor,
+  selectedColor,
+  showCustomInput,
+}) => {
+  const colorList = customColors ? customColors : defaultColors;
+  const [validHexValue, setValidHexValue] = React.useState<string>('');
```
NicholasBoll:
```suggestion
  const [validHexValue, setValidHexValue] = React.useState('');
```

`string` is inferred from the default value provided.

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const [customHexValue, setCustomHexValue] = React.useState<string>(
+    isCustomColor(colorList, selectedColor) ? selectedColor! : ''
+  );
```
NicholasBoll: Same

```suggestion
  const [customHexValue, setCustomHexValue] = React.useState(
    isCustomColor(colorList, selectedColor) ? selectedColor! : ''
  );
```

modules/_labs/color-picker/react/lib/parts/ColorReset.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.key === 'Enter') {
+      handleResetColor();
+    }
+  };
+
+  return (
+    <Container id={id} onClick={handleResetColor} onKeyDown={handleKeyDown}>
+      <Swatch color={resetColor} showCheck={false} />
+      <Label id="canvas-color-picker--reset-label">{label || 'Reset'}</Label>
```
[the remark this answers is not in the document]
NicholasBoll: What is the `id` for?

modules/_labs/color-picker/react/lib/ColorPicker.tsx
laurenraddatz: so, I should JSDoc these?

modules/_labs/color-picker/react/lib/ColorPicker.tsx
laurenraddatz: this current type signature also causes conflicts with https://github.com/Workday/canvas-kit/pull/462#discussion_r383600029

laurenraddatz: also, how should I handle this for now?

modules/_labs/color-picker/react/lib/parts/ColorReset.tsx
```diff
@@ -0,0 +1,73 @@
...
+import {Swatch} from './Swatch';
+
+export const ResetButton: React.FunctionComponent<Props> = ({id, onClick, resetColor, label}) => {
+  const handleResetColor = () => onClick(resetColor);
+
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.key === 'Enter') {
+      handleResetColor();
+    }
+  };
```
NicholasBoll: Did you check if this was actually necessary? It is making your `onClick` type signature more difficult than it needs to be and `<button>` elements already file the `onClick` event when they are focused and you hit the `Enter` or `Space` keys. I suggest removing this.

modules/_labs/color-picker/react/lib/parts/ColorReset.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <Label id="canvas-color-picker--reset-label">{label || 'Reset'}</Label>
+    </Container>
+  );
+};
+
+export interface Props {
+  id?: string;
+  label?: string;
+  resetColor: string;
+  onClick: Function;
```
NicholasBoll: Hmm. I thought we had a lint rule that prevents use of `Function`. It isn't very type-safe.

From the use, it looks like the signature is:
```ts
onClick: (value: string) => void
```

But that is weird for an `onClick` signature, right? If this is a reset button, why not use a `<button type="reset">` and make the popup wrapped in a `<form>` element? This way, clicking this button just invokes `onReset` of the form. This avoids the odd type signature of `onClick` passing a string rather than a value.

modules/_labs/color-picker/react/lib/parts/ColorReset.tsx
```diff
@@ -0,0 +1,73 @@
+import * as React from 'react';
+import styled from '@emotion/styled';
+
+import {borderRadius, colors, spacing, type, inputColors} from '@workday/canvas-kit-react-core';
+
+import {Swatch} from './Swatch';
+
+export const ResetButton: React.FunctionComponent<Props> = ({id, onClick, resetColor, label}) => {
```
NicholasBoll: We found several issues using `React.FunctionComponent<Props>` or `React.FC<Props>` or `React.StatelessComponent<Props>` or any of the other signatures where Typescript doesn't understand `defaultProps` properly and our Storybook Docs can't extract Prop type tables properly...

```suggestion
export const ResetButton = ({id, onClick, resetColor, label}: Props) => {
```

modules/_labs/color-picker/react/lib/parts/Swatch.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {SystemIcon} from '@workday/canvas-kit-react-icon';
+
+const whiteColor = colors.frenchVanilla100;
+
+export interface ContainerProps extends React.HTMLAttributes<HTMLDivElement> {
+  color: string;
+  style?: React.CSSProperties;
+}
+
+export const Container = styled('div')<ContainerProps>(
```
NicholasBoll: The type of element has come up before. Should it be a `button`, `li`, or `div`. Least semantic is `div`. We could easily argue it is a part of a list if the field (target element) states this is some type of combobox with a menu. This way as long as the target has focus, keyboard works (up, down, possibly left/right).

We may need a deeper talk with accessibility to find out what we should do here. I think `div` is fine for now until we get greater direction.

modules/_labs/color-picker/react/spec/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,170 @@
...
+  };
+
+  it('should include aria-label', () => {
+    const ariaLabel = 'Background Color';
+    const {container, unmount} = renderColorPicker({ariaLabel});
+
+    expect(container.querySelector(`[aria-label="${ariaLabel}"]`)).not.toBeNull();
+    unmount();
```
NicholasBoll: React Testing Library gives us nice helper functions to find things

```suggestion
    const {getByLabelText} = renderColorPicker({ariaLabel});

    expect(getByLabelText(ariaLabel)).not.toBeNull();
```

But this won't give us a very helpful error messages if it fails. It will tell us it couldn't find any element that is labelled by `ariaLabel`. Another way is to do:

```tsx
    const {container} = renderColorPicker({ariaLabel});

    expect(container).toContain(`aria-label="${ariaLabel}"`)
```

which will show the HTML and state it could not find the string you provided (at least giving some context). This can be an extremely noisy error if the HTML is large.

If you have a bit more information of _which_ element you expect to have this label, you can encode it in the test. This will help offload some of your current context about this component into the test if this test ever fails.

modules/_labs/color-picker/react/spec/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,170 @@
...
+  });
+
+  it('should call onChange handler when clicking a color swatch', () => {
+    const onChange = jest.fn();
+    const {container, unmount} = renderColorPicker({onChange});
+    const colorSwatch = container.querySelector('#canvas-color-picker--color-333333')!;
```
NicholasBoll: I think the exclamation point was a typo

```suggestion
    const colorSwatch = container.querySelector('#canvas-color-picker--color-333333');
```

NicholasBoll: Usually selectors like this means we haven't made our component accessible. There is no `aria-label` or anything else to tell a user of an assistive technology which value the picker will get. We can either label with our color names or the color value itself. This would allow us to use `getByLabelText('333333')` instead of putting `id`s in for the sole purpose of selecting in tests.

laurenraddatz: non-null assertion

modules/_labs/color-picker/react/lib/parts/Swatch.tsx
NicholasBoll: Oh. Wait. I didn't catch that these are treated like actual buttons. We had to add `tabIndex="0"` to get focus to work. Perhaps the decision isn't necessarily on this component. I see `SwatchContainer` elsewhere that also has a `div` with an `onClick` and a `tabIndex="0"` to make it act like a focusable button. We'll have to sync up with accessibility to know if all these swatches should have individual focus or the focus management and keyboard management comes from the target of this popup (and that target has a `aria-popup="true"`.

More info on the `combobox` role: https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.0pattern/combobox-autocomplete-list.html

It is up to Accessibility which role and therefore behavior the color picker field target element should get.

modules/_labs/color-picker/react/stories/stories_ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+    </div>
+  ))
+  .add('ColorPicker Popper examples', () => (
+    <div className="story">
+      <h3>Color Picker with Reset Button and Custom Input</h3>
+      <h5>as a Popup on a button target</h5>
+      <ColorPickerPopup />
+      <div style={{height: 200}}></div>
+      <h5>as a Popup on a custom color button target</h5>
+      <ColorPickerInputPopup />
```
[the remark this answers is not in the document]
NicholasBoll: Also a general note about all stories. A story's source doesn't know anything outside the `add(...)` block. So this story will show a `ColorPickerInputPopup` without any source.

Kind of an overall failing of the storybook source code plugin, really...

modules/_labs/color-picker/react/spec/ColorPicker.spec.tsx
NicholasBoll: I see why the exclamation is there. `fireEvent.click` probably expects an element and not `null`

NicholasBoll: Another reason to use `getByLabelText`. It will either return an element or throw an error. Either way, getting past that line means an element is returned

modules/_labs/color-picker/react/spec/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,170 @@
...
+    unmount();
+  });
+
+  it('should call onChange handler when selecting a color swatch with ENTER', () => {
+    const onChange = jest.fn();
+    const {container, unmount} = renderColorPicker({onChange});
+    const colorSwatch = container.querySelector('#canvas-color-picker--color-333333')!;
+
+    fireEvent.mouseEnter(colorSwatch);
+    fireEvent.keyDown(colorSwatch, {key: 'Enter'});
```
NicholasBoll: The `mouseEnter` isn't necessary here. This type of test should be covered by Cypress and not in the unit tests. If we use a `button` element, we don't have to do any special keyboard handling. `Enter` and `Space` keys already work on button elements.

modules/_labs/color-picker/react/spec/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,170 @@
...
+  });
+
+  it('should display custom hex codes in the custom input', () => {
+    const {container, unmount} = renderColorPicker({
+      selectedColor: '#abcdef',
+      showCustomInput: true,
+    });
+    const input = container.querySelector('input')!;
+
+    expect(input.getAttribute('value')).toBe('abcdef');
```
NicholasBoll: We use [jest-dom](https://github.com/testing-library/jest-dom) to give us more meaningful and contextual error messages.

```suggestion
    expect(input).toHaveAttribute('value', 'abcdef');
```

```suggestion
    expect(input).toHaveValue('abcdef');
```

These versions will fail with more context than just `Expected 'abcd' to equal 'abcdef'`

modules/_labs/color-picker/react/spec/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,170 @@
...
+      const onChange = jest.fn();
+      const {container, unmount} = renderColorPicker({onChange, showCustomInput: true});
+      const input = container.querySelector('input')!;
+
+      fireEvent.mouseEnter(input);
+      fireEvent.change(input, {target: {value: '123456'}});
+      fireEvent.keyDown(input, {key: 'Enter'});
+      expect(onChange).toHaveBeenCalledWith('#123456');
+      unmount();
+    });
```
NicholasBoll: This whole test belongs in Cypress. These types of tests are difficult to write, debug and maintain in React Testing Library

modules/_labs/color-picker/react/lib/ColorPicker.tsx
NicholasBoll: Good question. I think we can leave it for now. This is going into labs and the question is far-reaching and might not be answered quickly. I say leave it for now.

As for https://github.com/Workday/canvas-kit/pull/462#discussion_r383600029, you can extract `onClick` and pass `elemProps` (or rest props) to the container. The comment would require you to extend the `ColorPickerProps` interface from `React.DivHTMLAttributes<HTMLDivElement>`,

modules/_labs/color-picker/react/lib/ColorPicker.tsx
NicholasBoll: We are doing this to all other components, so that would help us a lot with that effort.

modules/_labs/color-picker/react/lib/parts/Swatch.tsx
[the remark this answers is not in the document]
laurenraddatz: I ended up going against this -- maybe we'd like to do it eventually, but the styling on it would have to change to meet both needs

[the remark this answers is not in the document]
anicholls: Can you elaborate on what the differences are that prevented you from doing it so it's captured in the PR at least? Thanks!

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -0,0 +1,192 @@
...
+              value={customHexValue}
+              placeholder={placeholder}
+              showCheck={!!placeholder}
+            />
+          </ColorInputAndLabel>
+          <CheckButton
+            aria-label={customHexInputLabel}
+            variant={IconButton.Variant.CircleFilled}
+            icon={checkIcon}
+            onClick={onSubmitClick}
```
anicholls: Shouldn't this basically just call `onColorChange`? I don't really see the value in this callback right now. This is the reason there are bugs in the stories (regular colors and custom colors being selected at the same time, etc.)

laurenraddatz: ah, yeah. the only reason it was separate is because the designs say that button should close the popup, so that's how I chose to accomplish that. feel free to change it!

anicholls: Ah makes sense. In that case, I think we need to call both `onColorChange` and `onSubmitClick` as long as the custom color is valid.

laurenraddatz:
> * [Considering using `button`s for the swatches](https://github.com/Workday/canvas-kit/pull/462#discussion_r384771270) to get a11y handling for free and remove the need for `tabIndex="0"`

going to hold off on this until accessibility looks at it (after this PR), it adds a lot of focusable elements and I'm not sure it's worth doing if we don't know it's what they'll want

cypress/integration/ColorPicker.spec.ts
```diff
@@ -4,8 +4,27 @@ const getColorPicker = () => {
   return cy.get('[type="text"]');
 };
 
+const expandHex = (hex: string) => {
```
laurenraddatz: not ideal, this is defined in `colorUtils.ts` but I didn't want to have to import from the main app/adjust the tsconfig

PR #457, opened 2020-02-11
feat(radio): Add theming support

anicholls:
## Summary

We're adding theming to our inputs. The consumer must wrap their components in a `CanvasProvider` and then they can pass their custom theme to any child components through the `theme` prop.

Partially addresses #358 

Blocked on https://github.com/Workday/canvas-kit/pull/456

## Checklist

- [x] branch has been rebased on the latest master commit
- [x] tests are changed or added
- [x] `yarn test` passes
- [x] code adheres to the [API & Pattern guidelines](../API_PATTERN_GUIDELINES.md)

PR #465, opened 2020-02-12
feat(switch): Add theming support

anicholls:
## Summary

We're adding theming to our inputs. The consumer must wrap their components in a `CanvasProvider` and then they can pass their custom theme to any child components through the `theme` prop.

Partially addresses #358 

Blocked on https://github.com/Workday/canvas-kit/pull/457

Notes:
- I noticed we were still using Blueberry 500 for the checked color. This is an old styling and all of our other inputs now use Blueberry 400. The checked background was updated. If you need to force Blueberry 500 for whatever reason, you can wrap the switch with a custom theme
- The checked disabled state was a light gray when it should have been a light blue. This has been updated.

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/switch/react/lib/Switch.tsx
+++ b/modules/switch/react/lib/Switch.tsx
@@ -121,13 +121,17 @@ const SwitchBackground = styled('div')<Pick<SwitchProps, 'checked' | 'disabled'>
     padding: '0px 2px',
     transition: 'background-color 200ms ease',
   },
-  ({checked, disabled, theme}) => ({
-    backgroundColor: disabled
-      ? colors.soap400
-      : checked
-      ? theme.palette.primary.main
-      : colors.licorice200,
-  })
+  ({checked, disabled, theme}) => {
+    if (checked) {
+      return {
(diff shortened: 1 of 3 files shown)


modules/switch/react/lib/Switch.tsx
```diff
@@ -130,9 +121,17 @@ const SwitchBackground = styled('div')<Pick<SwitchProps, 'checked' | 'disabled'>
     padding: '0px 2px',
     transition: 'background-color 200ms ease',
   },
-  ({checked, disabled}) => ({
-    backgroundColor: disabled ? colors.soap400 : checked ? colors.blueberry500 : colors.licorice200,
-  })
+  ({checked, disabled, theme}) => {
+    if (checked) {
+      return {
+        backgroundColor: disabled ? theme.palette.primary.light : theme.palette.primary.main,
```
mannycarrera4: so we are theming background colors? i guess for the switch it makes sense?

anicholls: Yeah - anything that's blue turns to `theme.palette.primary.main` (other than the focus outlines)

Issue #468, opened 2020-02-13
Refactor React Button


anicholls:
## Background 

As our button component has grown, it has become increasingly complex. It has reached a point where any changes or improvements are extremely hard to make. Bug fixes often turn into a game of whackamole - where changing a style in one place for one use case can ripple into other use cases. There also has been a lot of discussion about separating the inner workings of `IconButton` from the regular button.

In order to support theming within button, we will likely require this refactor.

I'm not sure exactly what this refactor should look like yet, but I will update this issue as I do more research/exploration. To start, we will need to add a static states table to ensure no regressions are introduced by the updates.

## Notes
- Dropdown button should only allow variants primary and secondary
- Dropdown button should not allow `icon` and `dataLabel`
- Delete button should not allow `icon` or `dataLabel`
- Small buttons should not allow `icon` or `dataLabel`
- We should update the button build script to babel 7
- Move `IconButtonToggleGroup` into a separate module called `SegmentedControls` and add support for text

## Questions
- Should we have a separate component for different variants? It's hard to enforce illegal prop combinations when everything is unified (i.e. `<Button>`, `<HighlightButton>`, etc.)
- Should small buttons allow for icons and data text? If not, are these just hidden at this size? Should there be a small dropdown button? Right now we show icons for small text buttons but not other small buttons.
- Should `IconButton` be a separate module? - no

anicholls:
closed by #471

PR #469, opened 2020-02-13
test(button): Add static state tables for all Button components

anicholls:
## Summary

In preparation for #468, we would like to have static state tables in place so that we can ensure no regressions are introduced in the upcoming button refactor. This PR adds them for `Button`, `DropdownButton`, and `TextButton`. It also moves the existing `IconButton` static states story to a common folder with the others (I thought this was clearer than having to dive into subfolders for each visual testing story - especially when we start to add multiple stories for theming).

Note: There are actually quite a few issues with some of the prop combinations. Some problems are stylistic, while others make it clear where we've exposed possible prop combinations that shouldn't be available. I would propose we leave them as is for now and try to address them in the refactor. Would love some input on this.

### DIFF
--- a/modules/button/react/stories/stories_visualTesting.tsx
+++ b/modules/button/react/stories/stories_visualTesting.tsx
@@ -1,26 +1,29 @@
 /// <reference path="../../../../typings.d.ts" />
 /** @jsx jsx */
 import {jsx, CSSObject} from '@emotion/core';
+import * as React from 'react';
 import {storiesOf} from '@storybook/react';
 import {StaticStates} from '@workday/canvas-kit-labs-react-core';
 import {ComponentStatesTable, permutateProps} from '../../../../utils/storybook';
 import {playCircleIcon} from '@workday/canvas-system-icons-web';
-import {Button, DropdownButton, TextButton} from '../index';
+import {Button, DropdownButton, TextButton, IconButton} from '../index';
 
-const blueBackground: CSSObject = {
+const buttonLayout: CSSObject = {
   display: 'flex',
   alignItems: 'center',
   justifyContent: 'center',
-  flexWrap: 'wrap',
-  backgroundColor: '#0875e1',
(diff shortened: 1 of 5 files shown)


modules/button/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,268 @@
...
+  .addParameters({component: DropdownButton})
+  .add('States', () => <DropdownButtonStates />);
+
+storiesOf('Components|Buttons/Button/React/Visual Testing/Text', module)
+  .addParameters({component: TextButton})
+  .add('States', () => <TextButtonStates />);
+
+storiesOf('Components|Buttons/Button/React/Visual Testing/Icon Button', module)
+  .addParameters({component: IconButton})
+  .add('States', () => <IconButtonStates />);
```
mannycarrera4: so clean

modules/button/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,268 @@
...
+            )}
+          </ComponentStatesTable>
+        </StaticStates>
+      </div>
+    ))}
+  </React.Fragment>
+);
+
+storiesOf('Components|Buttons/Button/React/Visual Testing/Button', module)
+  .addParameters({component: Button})
```
mannycarrera4: Should we put all of these under a `States` folder since we might also add a story for theming for each one of these?

anicholls: There is a `States` story for each component. I went for the following organization.

```
Visual Testing/
  Button/
    States
    Theming
  IconButton/
    States
    Theming
  ...
```

If the consensus is to swap the component and states/theming/etc categories, I can do that.

anicholls:
These tables make our button component look pretty buggy. Wondering if we should change the target to the `prerelease/v4` branch instead of putting them in master.

Issue #207, opened 2019-09-19
Segmented Control Redesign with Text Capabilities


xTena:
## 🚀 Feature Proposal

The Icon Button Toggle Group doesn't allow text. Will be great if we can pass only text, only an Icon or both. 

## Motivation
I have to build something similar recently because we need in our project a Button Toggle Group with only text. 

![Screenshot 2019-09-19 at 12 03 07](https://user-images.githubusercontent.com/55538167/65238826-7954ca80-dad5-11e9-9f4c-f07dc9aad030.png)

We'll love to use Canvas kit instead.

lychyi:
@xTena thanks for the feature request, I'll shop this around our design team. This is a use case that is slightly beyond just an extension of the icon button. Iconography is inherent to that design. Is there any context you can give for this idea? Maybe more about the surrounding feature that you'll need something like this in?

xTena:
Hi @lychyi, 

Thank you very much for your answer :) 

I understand is not possible to change this component but it would great if we can have one toggle group that allow text. 

We had to create one in the app because we didn't have it available in Canvas. Let me know which information you need and I can send it to you for a different channel (I can't post any info of the project here).

Thanks a million,
Ximena

willklein:
We restrict the allowed children in TypeScript and we can fix this here: https://github.com/Workday/canvas-kit/blob/3322ef6ba113940373412840edd14d35a7a3093c/modules/react/segmented-control/lib/SegmentedControl.tsx#L11

anishatulai:
I would like to work on this.

alanbsmith:
Making a note that Segmented Control should support theming. See #550.

RayRedGoose:
So if we closed this issue, should we create a new one for new SegmentedControl component?

Issue #378, opened 2019-12-17
Upgrade Button transpilation to babel 7


lychyi:
## 🚀 Feature Proposal

Right now we're using Babel 7 for Webpack when compiling Storybook.

We're also doing a one-off Babel transpilation of our `Button` component using Babel but that's on version 6. To clean things up we should remove Babel v6 and just use Babel v7.

## Motivation

Having Babel v6 and Babel v7 in the same repo can cause some unexpected confusion and we should probably be on the latest and greatest.

Issue #87, opened 2019-08-07
Add href prop to Button to render as an anchor element


anicholls:
## 🚀 Feature Proposal

Add a new optional `href` prop to `Button` components:
```typescript
  /**
   * The URL to link to when the button is clicked. 
   * If specified, will use an `a` element as the root of the component
   */
  href?: string;
```

It would also probably be good to add some prop that gets used as the `target` attribute on the underlying `a` tag

## Motivation

Many of our use cases of the `Button` component end up essentially being links, changing the current URL in the `onClick` handler:

```JSX
<Button onClick={() => window.location.href='url'}>Button Text</Button>

// may also use `window.open('url', '_blank')` in some cases
```

This has downsides in that it doesn't benefit from any of the native browser functionality that actual links get when using the `a` tag, such as being able to explicitly open a link in a new tab or window

One workaround to get the benefits of an `a` tag, while also getting the style of the `Button` component is to wrap the `Button` in an `a`:

```JSX
<a href='url' target='_blank'>
  <Button>Button Text</Button>
</a>
```

The downsides to this approach:
- The `a` tag may have styles that need to be suppressed (at the very least, disabling its `outline`)
- The `a` and underlying `button` elements are both focusable, causing keyboard nav to first focus the `a` and then the `button` next.

## Example
```JSX
<Button href='http://www.google.com'>Button Text</Button>

// DOM renders as: 
<a href='http://www.google.com'>
  /* ... children elements */
</a>

// while maintaining the style of the original `button` based implementation.
```

Patil2099:
@NicholasBoll  Can I work on this?

anicholls:
@Patil2099 Sure, feel free to have a go! Let us know if you have questions. Thanks 🙂

anicholls:
Closed by https://github.com/Workday/canvas-kit/pull/590

PR #471, opened 2020-02-14
refactor(button): Simplify Button components and prep for theming

### LINKED ISSUE #468: Refactor React Button

anicholls:
# Summary
This doesn't deal with any of the styling stuff yet. Just starting to get some thoughts down about how we should structure the components and types.

Closes #468 
Blocked on https://github.com/Workday/canvas-kit/pull/481

## Todo:
- [x] `IconButton`
- [x] Story file
- [x] Audit spacing. There are currently inconsistencies with the design specs
- [x] Focus Rings
- [x] Filled styling for toggled icon buttons
- [x] Change target to 4.x
- [x] Tests
- [x] Move from labs to replace existing button module
- [x] Documentation
- [x] Pull allCaps into a prop for TextButton
- Factor out `IconButtonToggleGroup` https://github.com/Workday/canvas-kit/issues/207 (separate PR)
- Cypress testing (separate PR)
- Theming (separate PR)
- Update changed story URL references in design site

## Breaking changes:
- Some of the button variants have been split into different components to prevent invalid API combinations. `DeleteButton`, and `HighlightButton` are now separate components with their own interface. Here are some of the invalid prop combinations that are no longer possible:
  - Delete button with a data label or icon
  - Dropdown button with a data label or icon
  - Highlight button with a data label
  - Highlight button without an icon
  - Dropdown with variants other than primary and secondary
  - Small buttons with an icon or data label
  - Small Highlight button
  - Small Dropdown button
  - etc.
  **Required changes:**
  - `<Button variant={Button.Variant.Delete}>` > `<DeleteButton>`
  - `<Button variant={Button.Variant.Highlight}>` > `<HighlightButton>` 
  - `<Button variant={Button.Variant.OutlineSecondary}>` > `<OutlineButton>` 
  - `<Button variant={Button.Variant.OutlinePrimary}>` > `<OutlineButton variant={OutlineButton.Variant.Primary}>` 
  - `<Button variant={Button.Variant.OutlineInverse}>` > `<OutlineButton variant={OutlineButton.Variant.Inverse}>` 
- The majority of the button variant types have changed. `BaseButtonProps` is no longer available as each button variant has their own interface.
- React >= 16.8 required for hooks
- Spacing within buttons has been corrected to match the specs. This may cause horizontal flow changes
- `TextButton` now only allows `TextButtonSize.Small` and `TextButtonSize.Medium`. 
  **Required changes:**
  - `TextButtonSize.Medium` > `"medium"` or `TextButtonSize.Small`
  - `TextButtonSize.Large` > `"small"` or `TextButtonSize.Medium`
- All caps variants for TextButton have been turned into a prop
  **Required changes:**
  - `<TextButton variant={TextButton.Variant.AllCaps}>` > `<TextButton allCaps={true}>`
  - `<TextButton variant={TextButton.Variant.InverseAllCaps}>` > `<TextButton variant={TextButton.Variant.Inverse} allCaps={true}>` 
- All caps large text buttons now correctly 16px (up from 14)

## Quality of Life changes:
With the new components for variants and the simpler types for sizes, the code for complex buttons is much more concise.

Before:
```tsx
<Button variant={Button.Variant.OutlineSecondary} size={Button.Size.Large}>Label</Button>
```
After:
```tsx
<OutlineButton size="large">Label</OutlineButton>
```

## Open questions:
- Previously we did not allow icons or data labels in small buttons. Are we sure we want to keep this restriction when we've been asked to better support tighter UIs? Note: Small text buttons have always allowed icons
- Should highlight buttons be able to accept data labels?
- Should there be a small dropdown button?

## Explanation

Alright - here goes.

First of all, I decided to split up some of the buttons into their own components to avoid invalid API combinations. For example, delete buttons shouldn't allow icons or data labels, but it was still technically possible to do with the old API. We have quite a few cases like this with the old API because we're trying to use the same base interface for everything. Each component now has its own interface specific to its API. We now have `Button`, `DeleteButton`, `DropdownButton`, `HighlightButton`, `TextButton`, and `deprecated_Button`. Note: We will need discussions with design on how to align these with their IA model.

In the old module, the spread of styling functions/data across many files made it really hard to follow what styles were coming from where. Rather than simply duplicating all of that logic for each component, I wanted a way that could keep the CSS for the basic button shape shared, while improving the clarity of what's happening. This led me to strip out ALL of the extra functions, generic styles objects, etc. and rely on React composition + the existing color objects we were already using. 

What this means in reality is we now have a `ButtonContainer` element that accepts a color object of type `ButtonColors`, which [houses all the underlying shared CSS](https://github.com/Workday/canvas-kit/pull/471/files#diff-7cbd44c0ffe0563adb4c811f426a640aR69). [Each exported button component makes use of this container to share the styling logic](https://github.com/Workday/canvas-kit/pull/471/files#diff-c6578334cd63534043f5a485b22030faR66). I updated the button color object for more clarity. It now looks like [this](https://github.com/Workday/canvas-kit/pull/471/files#diff-c6578334cd63534043f5a485b22030faR30). **Since the `ButtonContainer` just accepts a color object, this should make it really easy to theme buttons.**

In terms of the React composition that I mentioned, gone is the `ButtonBase` file with all of the label, icon, and data components. The new file structure is very clear, and has one component per file. Other than the shared CSS in `ButtonContainer`, each component/file houses all of the styling/coloring that it needs. After removing all of the functions, style object files, etc. the file structure is very easy to understand at a quick glance:
image

The `deprecated_Button` is also completely isolated from the new button styles 🎉 .

Finally, I also refactored some of the styling to remove redundant styles and pull all of the spacing, size and font styling logic out of the labels and into the button container itself.

Any questions or feedback are welcome!


modules/_labs/button/react/lib/Button.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+import styled from '@emotion/styled';
+import {ButtonVariant, ButtonSize} from './types';
+import {CanvasSystemIcon} from '@workday/design-assets-types';
+
+export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
```
anicholls: I think we're all done with the `BaseButtonProps` + generics interfaces. It was super unclear what props were where. Hoping a slight bit of repetition makes things way clearer.

modules/_labs/button/react/lib/Button.tsx
```diff
@@ -0,0 +1,57 @@
...
+  dataLabel?: String;
+  /**
+   * The icon of the Button.
+   */
+  icon?: CanvasSystemIcon;
+}
+
+const Container = styled('button')<ButtonProps>({});
+
+const Button = (props: ButtonProps) => {
```
anicholls: Will be changing everything to functional components.

modules/_labs/button/react/lib/DeleteButton.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
+import styled from '@emotion/styled';
+import {ButtonSize} from './types';
+
+export interface DeleteButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
```
anicholls: I decided to break up the different types of buttons into various components. The big reason for doing this was to  restrict prop combinations. It feels wrong to just render `null` when there's an invalid combination (or have incorrect styling). I would rather remove the possibility of invalid props from the API entirely.

modules/_labs/button/react/lib/types.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Standard button Button
+ */
+export enum ButtonVariant {
+  Primary = 'primary',
+  Secondary = 'secondary',
+  OutlinePrimary = 'outlinePrimary',
+  OutlineSecondary = 'outlineSecondary',
+  OutlineInverse = 'outlineInverse',
+}
+export enum ButtonSize {
```
anicholls: There's a bit more duplication with the types. Very little duplication when it comes to variants, but unfortunately since subsets of enums aren't really a thing in TS, I think we will need duplicated sizes.

If we were just using the sizes for interfaces, we would be fine.
```tsx
export interface TextButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
  // ...
  size: ButtonSize.Medium | ButtonSize.Small;
  // ...
}
```

The problem comes when we need to export the static member variables.

```tsx
TextButton.Size = ???
```
I don't believe there's a way to indicate a subset of the `ButtonSize` enum here.

anicholls: I guess we could do 
```tsx
export interface TextButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
  // ...
  size: ButtonSize.Medium | ButtonSize.Small;
  // ...
}

TextButton.Size = {
  Small: ButtonSize.Small,
  Medium: ButtonSize.Medium,
}
```

modules/_labs/button/react/lib/Button.tsx
```diff
@@ -0,0 +1,227 @@
...
+  );
+};
+
+Button.Variant = ButtonVariant;
+Button.Size = ButtonSize;
+
+Button.defaultProps = {
+  variant: ButtonVariant.Secondary,
+  size: ButtonSize.Medium,
+};
```
NicholasBoll: What is the difference between using `defaultProps` and using defaults in the functional component?

```tsx
export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement>, GrowthBehavior {
  /**
   * The variant of the Button.
   * @default ButtonVariant.Secondary
   */
  variant?: ButtonVariant;
  /**
   * The size of the Button.
   * @default ButtonSize.Medium
   */
  size?: ButtonSize;
  // ...
}

const {
  variant = ButtonVariant.Secondary,
  size = ButtonSize.Medium,
  buttonRef,
  dataLabel,
  icon,
  children,
  ...elemProps
} = props;
```

This is less type magic. It does have the downside of evaluating the right-hand for every render cycle (no difference in this case). I don't have a strong preference either way. I've exclusively done default assignment in the past. I'm fine with this if Typescript supports it well (makes `variant` and `size` optional externally) and documentation generators understand it.

modules/_labs/button/react/lib/DeleteButton.tsx
NicholasBoll: Why does `DeleteButton` not support icons or data labels?

NicholasBoll:
What about adding a `HyperlinkButton`? We've been asked to do this the use-case of file downloads: LRN-15985

I did this with a bit of hacking: https://codesandbox.io/s/hyperlinkbutton-jh7re

Basically the ask is an anchor tag with the styling of a button.

modules/_labs/button/react/lib/DeleteButton.tsx
anicholls: This was a decision made by design. I guess I could see it having a trash icon, but not a customizable one. The data label is really only meant for video/audio timestamps so I understand why it's not there.

modules/_labs/button/react/lib/Button.tsx
anicholls: I would be fine with this approach. Only reason I did it the other way was because that's how I've seen it done in the majority of examples I've looked at.

anicholls:
@NicholasBoll Yeah we already have an issue open for this: https://github.com/Workday/canvas-kit/issues/87. I definitely think it would be good to add, but not sure it should happen in this PR.

modules/_labs/button/react/lib/Button.tsx
anicholls: @NicholasBoll Update: Typescript doesn't seem to recognize these as default props when you do this. I get an error wherever I don't specify a variant and size after this change

modules/_labs/button/react/lib/types.ts
anicholls: Turns out this doesn't work. It gets mad when trying to equate `TextButton.Size.Small` with `ButtonSize.Small`. Any ideas to keep this static member functionality and avoid a separate size interface for each button type @NicholasBoll?

anicholls:
@lychyi Re https://github.com/Workday/canvas-kit/issues/378 - Now that we've [removed our dependency on `babel-plugin-emotion`](https://github.com/Workday/canvas-kit/pull/474), we no longer need to build buttons with Babel AFAIK. We should be able to use the TS compiler like the rest of our modules.

lychyi:
@anicholls Yup, for now I don't think we need Babel for button transpilation. It may require removing some of the files made just for that build process so the change may not be that straight forward.

If we want to add Emotion 11 into v4, then we can once again use component selectors with the babel plugin. In addition, Babel provides us an opportunity to transpile using other plugins so we may end up going that route anyway.

Personally, I don't mind leaving it. But if we want a clean slate and switch to `tsc`, I'll leave that up to you in terms of weighing the refactoring of the build vs. just leaving it and updating Babel.

modules/_labs/button/react/lib/Button.tsx
```diff
@@ -0,0 +1,206 @@
...
+  variant: ButtonVariant;
+  /**
+   * The size of the Button.
+   * @default ButtonSize.Medium
+   */
+  size: ButtonSize;
+  /**
+   * The ref to the button that the styled component renders.
+   */
+  buttonRef?: React.Ref<HTMLButtonElement>;
```
anicholls: Do we want this to remain `buttonRef`? or change to `ref`? I think this is a remnant from Emotion 9.

modules/_labs/button/react/lib/Button.tsx
```diff
@@ -0,0 +1,206 @@
+import * as React from 'react';
+import {colors} from '@workday/canvas-kit-react-core';
+import {focusRing, GrowthBehavior} from '@workday/canvas-kit-react-common';
+import {CanvasSystemIcon} from '@workday/design-assets-types';
+import {ButtonVariant, ButtonSize, ButtonColors, DropdownButtonVariant} from './types';
+import {ButtonContainer, ButtonLabel, ButtonLabelData, ButtonLabelIcon} from './parts';
+
+export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement>, GrowthBehavior {
```
anicholls: Interestingly, `React.HTMLAttributes<HTMLButtonElement>` doesn't seem to contain `disabled`. Should I switch back to `React.ButtonHTMLAttributes<HTMLButtonElement>`?

lychyi: Yes, the latter is just a superset of the former. The generic is just used to type the event handlers

modules/_labs/button/react/lib/Button.tsx
lychyi: I don't think so, this is just what we name our refs right? `<component>Ref` is the naming convention for the prop API. This demonstrates what type of element the ref is referring to by name without relying strictly on documentation to indicate what it's a ref for.

Or are you alluding to the `innerRef` to `ref` change in the styled component interface in Emotion 10?

anicholls: Ah I guess that's true. I mentioned Emotion 9 because I think we originally didn't want to expose Emotions API by using `innerRef`.

Will leave as is 👍

modules/_labs/button/react/lib/types.ts
lychyi: @anicholls @NicholasBoll would it be better in this case to just use a union of strings rather than an enum? I'm not sure the enum is buying us much here so to make it appease TS just sounds silly at this point.

modules/_labs/button/react/lib/parts/ButtonContainer.tsx
```diff
@@ -0,0 +1,214 @@
...
+import isPropValid from '@emotion/is-prop-valid';
+import {CSSObject} from '@emotion/core';
+import {styled, type} from '@workday/canvas-kit-labs-react-core';
+import canvas, {borderRadius, spacing, spacingNumbers} from '@workday/canvas-kit-react-core';
+import {GrowthBehavior, focusRing, mouseFocusBehavior} from '@workday/canvas-kit-react-common';
+import {ButtonSize, ButtonColors} from '../types';
+import {buttonLabelDataClassName} from './ButtonLabelData';
+
+export interface ButtonContainerProps
+  extends React.HTMLAttributes<HTMLButtonElement>,
```
lychyi: I believe this (and the other instances in this PR that target the `ButtonContainer` when we spread props) should be `React.ButtonHTMLAttributes` here. `HTMLAttributes` is just the base class so we would be missing some things.

modules/_labs/button/react/lib/Button.tsx
```diff
@@ -0,0 +1,206 @@
...
+   * The icon of the Button.
+   */
+  icon?: CanvasSystemIcon;
+}
+
+const Button = (props: ButtonProps) => {
+  const {variant, size, buttonRef, dataLabel, icon, children, ...elemProps} = props;
+
+  return (
+    // eslint-disable-next-line no-use-before-define
```
lychyi: We should either disable `no-use-before-define` for [functions](https://eslint.org/docs/rules/no-use-before-define#functions) in the linter or move the `getButtonColors` function prior to the class declaration.

I think this rule is generally useful to prevent "temporal dead zone" mistakes but in this case this comment seems like a lot of maintenance just to make sure our code is organized correctly. Oh JavaScript...

modules/_labs/button/react/lib/types.ts
anicholls: @lychyi So usage would turn into `<Button size="small" />`?

modules/_labs/button/react/lib/parts/ButtonContainer.tsx
anicholls: Good call. I only did it for `Button`, but I should do it everywhere. Incoming!

modules/_labs/button/react/lib/types.ts
NicholasBoll: Personally I've avoided enums in favor of union of string literals. In this case, enums only give us more typing, not more type safety. Autocomplete is actually more helpful with union of string literals.

It can be introduced as a non-breaking change by converting the type to a union of string literals and the enum to an object with an `as const` or casting manually:

```tsx
export interface TextButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
  // ...
  size: 'medium' | 'small';
  // ...
}

TextButton.Size = {
  Small: 'small',
  Medium: 'medium',
} as const

// or
TextButton.Size = {
  Small: 'small' as 'small',
  Medium: 'medium' as 'medium',
}
```

I don't remember if we can't use `as const`. Downlevel DTS should support this

modules/_labs/button/react/lib/parts/ButtonLabelIcon.tsx
```diff
@@ -0,0 +1,48 @@
...
+);
+
+export const ButtonLabelIcon = (props: ButtonLabelIconProps) => {
+  const {icon, size, dropdown, iconPosition, ...elemProps} = props;
+  /* istanbul ignore next line for coverage */
+  if (icon === undefined) {
+    return null;
+  }
+
+  const iconSize = size === ButtonSize.Small ? 20 : 24;
```
lychyi: Nit: I'm noting here that these should probably be `const`s somewhere in the file. Are the margin spacings hard coded for a reason also?

modules/_labs/button/react/lib/IconButton.tsx
```diff
@@ -0,0 +1,224 @@
...
+  // Only call onToggleChange on update - not on first mount
+  React.useEffect(() => {
+    if (isInitialMount.current) {
+      isInitialMount.current = false;
+    } else {
+      if (props.toggled && typeof props.onToggleChange === 'function') {
+        props.onToggleChange(toggled);
+      }
+    }
+  }, [props.toggled]);
```
lychyi: Would we need to add `props.onToggleChange` here too? We do use it in the effect. Dunno if it applies here but this is from the React docs:

> If you use this optimization, make sure the array includes all values from the component scope (such as props and state) that change over time and that are used by the effect.

modules/_labs/button/react/lib/parts/ButtonLabelIcon.tsx
anicholls: Good call. Changed `24` to `undefined` as well, since that's the default.

modules/_labs/button/react/lib/IconButton.tsx
anicholls: Nice catch 👍

modules/_labs/button/react/lib/IconButton.tsx
```diff
@@ -0,0 +1,224 @@
...
+          icon: toggled ? colors.blueberry400 : colors.frenchVanilla100,
+          focusRing: focusRing(2, 2, true, false, 'currentColor', colors.frenchVanilla100),
+        },
+        disabled: {
+          background: toggled ? 'rgba(255,255,255,0.75)' : 'rgba(0, 0, 0, 0.2)',
+          icon: toggled ? colors.blueberry400 : 'rgba(255, 255, 255, 0.75)',
+        },
+      };
+  }
+};
```
mannycarrera4: this is where I was trumped a bunch, mostly around toggled on styles and whatinput styles, did this end up working for all cases?

modules/_labs/button/react/lib/TextButton.tsx
```diff
@@ -0,0 +1,141 @@
...
+  border: '0',
+  padding: `0 ${spacing.xxs}`,
+  minWidth: 'auto',
+  '&:hover:not([disabled])': {textDecoration: 'underline'},
+};
+
+const TextButton = (props: TextButtonProps) => {
+  const {buttonRef, children, iconPosition, size, variant, icon, ...elemProps} = props;
+
+  // Note: We don't use ButtonLabel because they label styles differ from other button types
```
mannycarrera4: `they` -> `the`

modules/_labs/button/react/lib/deprecated_Button.tsx
```diff
@@ -0,0 +1,182 @@
...
+      }),
+    };
+  }
+);
+
+const DeprecatedButton = (props: DeprecatedButtonProps) => {
+  const {variant, size, buttonRef, children, ...elemProps} = props;
+
+  React.useEffect(() => {
+    console.warn('This component is now deprecated, consider using the new Button component');
```
mannycarrera4: since this is in labs and we have one outside of labs, should we make that distinction?

modules/_labs/button/react/lib/IconButton.tsx
anicholls: @mannycarrera4 Yeah I rewrote some of the `ButtonContainer` styles to make this work for all cases. You can take a look at the visual testing story if you want to double check 👍

modules/_labs/button/react/lib/deprecated_Button.tsx
anicholls: This is only in labs to make the comparison in this PR clearer. I'm going to replace the old button with this one right after it's merged. Most people won't see this change because it's in the `prerelease/v4` branch.

modules/_labs/button/react/lib/IconButton.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+IconButton.defaultProps = {
+  variant: IconButtonVariant.Circle,
+  size: ButtonSize.Medium,
+};
+
+export default IconButton;
+
+const getIconButtonColors = (variant: IconButtonVariant, toggled?: boolean): ButtonColors => {
```
mannycarrera4: ah this is so much cleaner, and passing the toggled in helps a bunch rather than returning an empty object if it's toggled on 🎉

modules/_labs/button/react/lib/types.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * The object used for passing in colors to the ButtonContainer
+ */
+export interface ButtonStateColors {
+  background?: string;
+  border?: string;
+  icon?: string;
+  iconFill?: boolean;
+  label?: string;
+  labelData?: string;
```
mannycarrera4: `ButtonStateColors` makes me think of what you have in `ButtonColors` is there a reason we named it states here?

modules/_labs/button/react/lib/types.ts
anicholls: `as const` came around in 3.4 so I think we should be good with that 👍

modules/_labs/button/react/lib/types.ts
anicholls: @mannycarrera4 These are the colors for a single state, `ButtonColors` are the entire set of colors for all states

PR #474, opened 2020-02-19
fix(button): Fix Component Selector Issue

lychyi:
## Summary

Due to some changes in the way we use `@emotion/styled`, we are no longer able use [component selectors](https://emotion.sh/docs/styled#targeting-another-emotion-component) in Emotion via Babel. This is because `styled` functions using Emotion's `CreateStyled`, like our own [Canvas styled component](https://github.com/Workday/canvas-kit/blob/ed41a76be50068cb1ae15b666d9305a6a97c4b72/modules/_labs/core/react/lib/theming/styled.ts#L52), does not work with the babel plugin for emotion.

This is a known issue and will be fixed in Emotion 11 by https://github.com/emotion-js/emotion/pull/1220 via a mapping configuration so the plugin knows about custom styled functions.

For now the fix is to just remove the usage of component selectors. This PR addresses that by using an identifier class name (similar to how Emotion identifies component selectors) in `IconButton` and then styling that class name in `IconButtonToggleGroup`.

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/button/react/spec/IconButton.spec.tsx
+++ b/modules/button/react/spec/IconButton.spec.tsx
@@ -24,17 +24,17 @@ describe('Icon Button', () => {
   });
 
   it('should have an identifier class', () => {
-    const {getByTestId} = render(
-      <IconButton data-testid={iconButtonIdentifier} aria-label="Activity Stream">
+    const {getByRole} = render(
+      <IconButton aria-label="Activity Stream">
         <SystemIcon icon={activityStreamIcon} />
       </IconButton>
     );
 
-    expect(getByTestId(iconButtonIdentifier).className).toContain(iconButtonIdentifier);
+    expect(getByRole('button').className).toContain(iconButtonIdentifier);
   });
   it('should compose custom classNames with the identifier class', () => {
     const testClassName = 'test classname';
-    const {getByTestId} = render(
+    const {getByRole} = render(
       <IconButton
         data-testid={iconButtonIdentifier}
         className={testClassName}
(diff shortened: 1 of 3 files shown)


modules/button/react/spec/IconButton.spec.tsx
```diff
@@ -22,6 +23,32 @@ describe('Icon Button', () => {
     component.unmount();
   });
 
+  it('should have an identifier class', () => {
+    const {getByTestId} = render(
+      <IconButton data-testid={iconButtonIdentifier} aria-label="Activity Stream">
```
anicholls: Nit: it's a little bit confusing at first glance to use the identifier for both the test ID and the classname. Any reason we can't drop the test ID and use `getByRole('button')`?

lychyi: Because I didn't think of that... 🤦‍♂ 😅

modules/button/react/spec/IconButton.spec.tsx
```diff
@@ -22,6 +23,30 @@ describe('Icon Button', () => {
...
+      </IconButton>
+    );
+
+    expect(getByRole('button').className).toContain(iconButtonIdentifier);
+  });
+  it('should compose custom classNames with the identifier class', () => {
+    const testClassName = 'test classname';
+    const {getByRole} = render(
+      <IconButton
+        data-testid={iconButtonIdentifier}
```
anicholls:
```suggestion
```

PR #477, opened 2020-02-19
chore: Release v3.4.0

anicholls:


modules/_canvas-kit-css/package.json
```diff
@@ -23,17 +23,17 @@
     "@workday/canvas-kit-css-checkbox": "^3.2.0",
     "@workday/canvas-kit-css-common": "^3.2.0",
     "@workday/canvas-kit-css-core": "^3.2.0",
-    "@workday/canvas-kit-css-form-field": "^3.3.0",
-    "@workday/canvas-kit-css-icon": "^3.2.0",
+    "@workday/canvas-kit-css-form-field": "^3.4.0",
+    "@workday/canvas-kit-css-icon": "^3.4.0",
     "@workday/canvas-kit-css-layout": "^3.2.0",
```
NicholasBoll: Why don't these get updated as well? Just strange to maintain different version numbers for these packages.

lychyi: No version change in packages that are not updated between now and the last annotated tag.

anicholls: Lerna 🤷‍♂

PR #478, opened 2020-02-21
test: Explicitly enable snapshots

lychyi:
## Summary

We are testing way too many stories that aren't useful for visual regression. This PR tells Chromatic to ignore all stories by default and only include ones that are marked.

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/form-field/react/stories/stories_Checkbox.tsx
+++ b/modules/form-field/react/stories/stories_Checkbox.tsx
@@ -171,7 +171,12 @@ const CheckboxStates = () => (
 );
 
 storiesOf('Components|Inputs/Checkbox/React/Visual Testing', module)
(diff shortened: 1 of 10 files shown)


modules/form-field/react/stories/stories_Checkbox.tsx
```diff
@@ -171,7 +171,12 @@ const CheckboxStates = () => (
 );
 
 storiesOf('Components|Inputs/Checkbox/React/Visual Testing', module)
-  .addParameters({component: Checkbox})
+  .addParameters({
+    component: Checkbox,
```
NicholasBoll: Interesting. It seems like Storybook is able to get the component just fine without this, but I'm not sure why.

PR #481, opened 2020-02-21
ci: Release canary builds for prerelease branches

anicholls:
## Summary

We recently added support for Canary builds on master, but would like to make them available for future major releases as well. We're just kicking off our `prerelease/v4` branch, so this update takes that branch name format into account.


.travis.yml
```diff
@@ -65,7 +65,7 @@ jobs:
       on:
         tags: true
 
-  - stage: master
+  - stage: trunk
```
anicholls: Was trying to think of a common term for master and prerelease branches. Open to suggestions.

.travis.yml
```diff
@@ -79,6 +79,12 @@ jobs:
         skip_cleanup: true
         on:
           branch: master
+      - provider: script
+        script: npm config set //registry.npmjs.org/:_authToken=$NPM_PUBLISH_TOKEN && node utils/publish-canary.js $SLACK_WEBHOOK $TRAVIS_BUILD_WEB_URL prerelease
+        skip_cleanup: true
+        on:
+          all_branches: true
+          condition: $TRAVIS_BRANCH =~ ^prerelease\/v\d*$
```
anicholls: Not sure if it's possible to put an inline conditional in the script line to only add the `prerelease` preid if the branch matches the regex. You also need to keep the script tag on one line for deploys. Seemed easier to just duplicate the master canary script.

Travis docs are bad 🤷‍♂️ - if anyone knows of a better way to do this, let me know.

lychyi: Would writing a script and leveraging Travis env variables accomplish this? I'm not sure if they're available node context but you could always do something like: `TRAVIS_BRANCH=$TRAVIS_BRANCH node canary-script.js`

And then somewhere in the node script:
```
if (process.env.TRAVIS_BRANCH.match(regex) !== null) // etc
```

lychyi: I feel like all Travis scripts inevitably end up devolving into a file somewhere that gets checked in...

anicholls: @lychyi Just pushed an update - let me know what you think

.travis.yml
```diff
@@ -75,10 +75,10 @@ jobs:
       - CHROMATIC_APP_CODE="m1dh5kc7oj"
     deploy:
       - provider: script
-        script: npm config set //registry.npmjs.org/:_authToken=$NPM_PUBLISH_TOKEN && node utils/publish-canary.js $SLACK_WEBHOOK $TRAVIS_BUILD_WEB_URL
+        script: npm config set //registry.npmjs.org/:_authToken=$NPM_PUBLISH_TOKEN && node utils/publish-canary.js
         skip_cleanup: true
         on:
-          branch: master
+          all_branches: true
```
anicholls: Just realized - Is this going to make the scripts above run on every branch? What keeps those scripts locked down to master currently?

.travis.yml
NicholasBoll: This stage name matches https://github.com/Workday/canvas-kit/pull/481/files#diff-354f30a63fb0907d4ad57269548329e3R26

NicholasBoll: You'll have to change L27 to something like:
```
if: branch = master OR branch =~ /^prerelease\//
```

.travis.yml
NicholasBoll: https://github.com/Workday/canvas-kit/pull/481/files#diff-354f30a63fb0907d4ad57269548329e3R26-R27

.travis.yml
NicholasBoll: Could go with my creative stage names: `branch`. I called the other stages `pull_request` and `tag`

anicholls: Ah - missed this. Thanks :)

anicholls: I think I'm going to stick with `trunk`. This stage should only run on "trunk" style branches that are being merged into (like `master` and `prerelease/v*`). If we used `branch` it makes it seem like it will run on all branches.

PR #482, opened 2020-02-24
test(form-field): Add new test for form field

mannycarrera4:
## Summary

Back when we re wrote some test to React Testing Library, we didn't get around to FormField, this PR addresses those test
Fixes FormField test in: https://github.com/Workday/canvas-kit/issues/286

## Checklist

- [x] `yarn test` passes


modules/form-field/react/spec/FormField.spec.tsx
```diff
@@ -1,152 +1,109 @@
...
-        .at(0)
-        .prop('id')
-    ).toEqual(hintId);
-    expect(component.find(InputComponent).prop('aria-describedby')).toEqual(hintId);
-
-    component.unmount();
+  describe('when rendered', () => {
+    it('should render a label with text Label', () => {
+      const label = 'Label';
+      const {container} = render(
```
anicholls: `getByLabelText` is much cleaner. `container` should be a last resort

modules/form-field/react/spec/FormField.spec.tsx
```diff
@@ -1,152 +1,109 @@
...
+        <FormField label={label}>
+          <input type="text" />
+        </FormField>
+      );
+      expect(container.querySelector('label')).toContainHTML('Label');
```
anicholls: If you've set `'Label'` to a variable above, let's use it.

modules/form-field/react/spec/FormField.spec.tsx
```diff
@@ -1,152 +1,109 @@
...
+    it('should render text below the input component', () => {
+      const label = 'Label';
+      const hintText = 'Helpful text goes here.';
+      const {container} = render(
+        <FormField hintText={hintText} label={label}>
+          <input type="text" />
+        </FormField>
+      );
+
+      expect(container.querySelector('p')).toContainHTML('Helpful text goes here.');
```
anicholls:
```suggestion
      expect(container.querySelector('p')).toContainHTML(hintText);
```

modules/form-field/react/spec/FormField.spec.tsx
```diff
@@ -1,152 +1,109 @@
...
-        <InputComponent />
-      </FormField>
-    );
-
-    expect(component.find(InputComponent).prop('id')).toEqual(inputId);
-    expect(component.find('Label').prop('htmlFor')).toEqual(inputId);
-
-    component.unmount();
+  describe('when rendered with hint text', () => {
+    it('should render text below the input component', () => {
```
anicholls: Can you also test the `hintId` here as well?

modules/form-field/react/stories/stories.tsx
```diff
@@ -2,30 +2,95 @@
...
+import {StaticStates} from '@workday/canvas-kit-labs-react-core';
+import {TextInput} from '../../../text-input/react';
 import README from '../README.md';
+import {FormFieldLabelPosition} from '../lib/types';
+
+const FormFieldStates = () => (
+  <StaticStates>
+    <ComponentStatesTable
+      rowProps={[
+        {label: 'Error States', props: {required: false}},
```
anicholls: Why is this row needed? Typo?

```suggestion
```

modules/form-field/react/stories/stories.tsx
```diff
@@ -2,30 +2,95 @@
...
+        error: [
+          {value: undefined, label: 'Default'},
+          {value: FormField.ErrorType.Alert, label: 'Alert'},
+          {value: FormField.ErrorType.Error, label: 'Error'},
+        ],
+      })}
+    >
+      {props => (
+        <FormField {...props} hintText="Helpful text goes here." label="Label">
+          <TextInput></TextInput>
```
anicholls: Nit: You may get an error for this not being `<TextInput />`

modules/form-field/react/spec/FormField.spec.tsx
mannycarrera4: In this case, I'm looking for the `Label` and the text associated with that label, `getByLabelText` will return any element associated with that label, which is why i used `container` instead. I have a talked with @NicholasBoll about this and how it was confusing to me that hat method really grabs elements associated with it and not the node `Label`

modules/form-field/react/stories/stories.tsx
mannycarrera4: i guess i wanted to show the form field errors, but it's a bit redundant, will remove!

modules/form-field/react/spec/FormField.spec.tsx
anicholls: Whoops - good call out haha. I think I was thinking of `getByText`.

PR #492, opened 2020-03-02
fix: Avoid polyfill requirements

anicholls:
We were unnecessarily using `Object.entries` in the check for an empty context in `useTheme`. This function requires a polyfill in IE 11, so I've switched it to use `Object.keys` instead, since this is equivalent for checking whether an object is empty.

Continued:
After a bit more exploration, we were using several functions that required polyfills. This PR now adds a lint check to ensure we don't do this in the future.

### DIFF
--- a/modules/icon/css/lib/utils.js
+++ b/modules/icon/css/lib/utils.js
@@ -38,11 +38,9 @@ export function getColor(color) {
   if (!/[0-9]{3}$/.test(color)) {
     findColor = `${color}500`;
   }
-  const foundColor = Object.keys(canvasColors).find(colorName => colorName === findColor);
+  const foundColor = canvasColors[findColor];
 
-  if (!foundColor) return null;
-
-  return canvasColors[foundColor];
+  return foundColor ? foundColor : null;
 }
 
 export function appendStyle(css) {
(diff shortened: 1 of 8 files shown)


NicholasBoll:
Hmm. Maybe we should update our `tsconfig.json` file to force compat mode with IE11. Polyfills would be used in Storybook and usually come with type definitions that opt into features like Promises.

Or perhaps an ESLint rule for detecting compat features? https://www.npmjs.com/package/eslint-plugin-compat

anicholls:
@NicholasBoll @lychyi After adding the linting rule, I found two more issues:
image

I fixed both, so we should be good now!

modules/icon/css/lib/utils.js
```diff
@@ -38,12 +38,11 @@ export function getColor(color) {
   if (!/[0-9]{3}$/.test(color)) {
     findColor = `${color}500`;
   }
-
-  const foundColor = Object.entries(canvasColors).find(([colorName]) => colorName === findColor);
+  const foundColor = Object.keys(canvasColors).find(colorName => colorName === findColor);
 
   if (!foundColor) return null;
```
codytowstik: Nit: since this is just an equals, can we avoid the extra call or is there an advantage to find?

```suggestion
  const foundColor = canvasColors[findColor]

  return (foundColor ? foundColor : null);
```

anicholls: Hah - yep. I just didn't take a close enough look at what I was refactoring. Thanks 👍

PR #495, opened 2020-03-04
chore: Simplify Chromatic CI runs

NicholasBoll:
We were part of the beta process for Chromatic v2 and now they have rolled those into their main offering. We can stop running both the main Chromatic offer plus the beta offering. This will speed up out CI process.

This PR also reduces the noise in the log from webpack by passing `--quiet`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,13 +40,13 @@ jobs:
   include:
   - stage: pull_request
     script:
-      - yarn lint
-      - yarn depcheck
-      - yarn test
+      # - yarn lint
+      # - yarn depcheck
+      # - yarn test
       - yarn chromatic --exit-zero-on-changes --app-code="dlpro96xybh" # v1 for cross-browser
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -43,16 +43,15 @@ jobs:
       - yarn lint
       - yarn depcheck
       - yarn test
-      - yarn chromatic --exit-zero-on-changes --app-code="dlpro96xybh" # v1 for cross-browser
-      - yarn chromatic --exit-zero-on-changes # v2 for visual review features and Storybook hosting
+      - yarn chromatic --quiet --exit-zero-on-changes
```
NicholasBoll: Added the `--quiet` option to decrease the noise in the log output from `yarn chromatic` where it runs `yarn build-storybook` which invokes `webpack`. Normally Webpack overwrites the last line in a log and we don't see a massive amount of output. In CI, a log line cannot be rewritten, so it ends up noisy.

Issue #496, opened 2020-03-05
Checkbox onChange should match the `HTMLInputElement` interface it extends from


NicholasBoll:
## 🐛 Bug Report

Checkbox `onChange` is defined as a SyntheticEvent rather than a ChangeEventHandler like the extended interface. This causes problems downstream when an handler is typed as `React.ChangeEventHandler<HTMLInputElement>` (since that's the base shape).

https://github.com/Workday/canvas-kit/blob/1d4e3d4ffc27d7743c4bfb33b67e3d7fba473758/modules/checkbox/react/lib/Checkbox.tsx#L47

Comes from this comment: https://github.com/Workday/canvas-kit/issues/487#issuecomment-594923058

We could either remove that line or change the signature to match the base signature (benefit of that is to add our own JSDoc info and have the `onChange` property show up in our docs)

anicholls:
@NicholasBoll This is a problem in more than Checkbox. We use this signature for the majority of our input elements. It should probably be updated everywhere

ziel5122:
> ## 🐛 Bug Report
> Checkbox `onChange` is defined as a SyntheticEvent rather than a ChangeEventHandler like the extended interface. This causes problems downstream when an handler is typed as `React.ChangeEventHandler<HTMLInputElement>` (since that's the base shape).
> 
> https://github.com/Workday/canvas-kit/blob/1d4e3d4ffc27d7743c4bfb33b67e3d7fba473758/modules/checkbox/react/lib/Checkbox.tsx#L47
> 
> Comes from this comment: [#487 (comment)](https://github.com/Workday/canvas-kit/issues/487#issuecomment-594923058)
> 
> We could either remove that line or change the signature to match the base signature (benefit of that is to add our own JSDoc info and have the `onChange` property show up in our docs)

It won't show up in docs if you extend?

NicholasBoll:
This line excludes a prop that originates from a `node_modules` folder: https://github.com/Workday/canvas-kit/blob/master/.storybook/docgen-plugin.js#L9

Without this line, there would be a lot of props. Every prop and attribute supported by a DOM element.

NicholasBoll:
> @NicholasBoll This is a problem in more than Checkbox. We use this signature for the majority of our input components. It should probably be updated everywhere

This might be a breaking change for those using Typescript with a workaround to this feature. We'd have to verify.

anicholls:
Closed by #499

PR #499, opened 2020-03-05
fix: Remove SyntheticEvent type usage

donovangini:
## Summary

Closes https://github.com/Workday/canvas-kit/issues/496

Removing the override type for `onChange` in `Checkbox`, and instead getting type from `InputHTMLAttributes` (`onChange?: ChangeEventHandler<T>;`)

Would love feedback on whether this is a breaking change, and how to handle! Getting from the default type made sense here to me, but happy to discuss

## Checklist

- [x] `yarn test` passes
      applicable

### DIFF
--- a/modules/_labs/combobox/react/lib/Combobox.tsx
+++ b/modules/_labs/combobox/react/lib/Combobox.tsx
@@ -215,7 +215,7 @@ const Combobox = ({
   }, [autocompleteItems, isFocused, value]);
 
   const handleAutocompleteClick = (
-    event: React.SyntheticEvent,
+    event: React.KeyboardEvent | React.MouseEvent,
     menuItemProps: MenuItemProps
   ): void => {
     if (menuItemProps.isDisabled) {
@@ -225,7 +225,7 @@ const Combobox = ({
     setIsFocused(false);
     setInputValue(getTextFromElement(menuItemProps.children));
     if (menuItemProps.onClick) {
-      menuItemProps.onClick(event);
+      menuItemProps.onClick(event as React.MouseEvent);
     }
   };
 
--- a/modules/_labs/combobox/react/lib/Combobox.tsx
+++ b/modules/_labs/combobox/react/lib/Combobox.tsx
@@ -41,7 +41,7 @@ export interface ComboboxProps extends GrowthBehavior, React.HTMLAttributes<HTML
   /**
    * The function called when the Combobox text input changes.
    */
-  onChange?: React.ChangeEventHandler<HTMLInputElement>;
+  onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
   /**
    * The function called when the Combobox text input focuses.
    */
--- a/modules/_labs/header/react/README.md
+++ b/modules/_labs/header/react/README.md
@@ -169,7 +169,7 @@ Default: `DubLogoTitle` (for "Dub" variants) or `WorkdayLogoTitle` (for "Full" v
 
 ---
 
-#### `onMenuClick: (React.SyntheticEvent) => void`
+#### `onMenuClick: (React.MouseEvent) => void`
 
 > A click handler for when the user clicks the mobile collapsed nav icon.
 
@@ -254,7 +254,7 @@ Default: `<DubLogoTitle />`
 
 Default: `justifyIcon` from `@workday/canvas-system-icons-web`
 
-#### `onMenuClick: (React.SyntheticEvent) => void`
+#### `onMenuClick: (React.MouseEvent) => void`
 
 > A click handler for when the user clicks the `menuToggle` element.
 
--- a/modules/_labs/header/react/lib/GlobalHeader.tsx
+++ b/modules/_labs/header/react/lib/GlobalHeader.tsx
@@ -16,7 +16,7 @@ export interface GlobalHeaderProps {
   /**
    * The function called when the responsive menu icon is clicked.
    */
-  onMenuClick?: (e: React.SyntheticEvent) => void;
+  onMenuClick?: (e: React.MouseEvent) => void;
   /**
    * If true, render the GlobalHeader in collapsed mode.
    * @default false
--- a/modules/_labs/menu/react/lib/MenuItem.tsx
+++ b/modules/_labs/menu/react/lib/MenuItem.tsx
@@ -15,7 +15,7 @@ export interface MenuItemProps extends React.LiHTMLAttributes<HTMLLIElement> {
   /**
    * The function called when the MenuItem is clicked. If the item is a child of the Menu component, this callback will be decorated with the onSelect and onClose Menu callbacks. This callback will not fire if the item is disabled (see below).
    */
-  onClick?: (event: React.SyntheticEvent) => void;
+  onClick?: (event: React.MouseEvent) => void;
   /**
    * The unique id for the MenuItem used for ARIA attributes. If the item is a child of the `Menu` component, this property will be generated and overridden.
    */
@@ -221,7 +221,7 @@ export default class MenuItem extends React.Component<MenuItemProps> {
     );
   }
 
-  private handleClick = (event: React.SyntheticEvent): void => {
+  private handleClick = (event: React.MouseEvent): void => {
     if (this.props.isDisabled) {
       return;
     }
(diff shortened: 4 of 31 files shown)


anicholls:
@donovangini Do you mind doing this for all components that use this `onChange` signature? It goes beyond just `Checkbox`

anicholls:
@NicholasBoll I would lean towards option 2 just because it's more obvious that it's a function at first glance

NicholasBoll:
@anicholls I could see that. I noticed these lines: https://github.com/Workday/canvas-kit/blame/master/API_PATTERN_GUIDELINES.md#L106-L108

donovangini:
@NicholasBoll and @anicholls, yea can update to:
`onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void` (don't want event type to be handler in here cuz that itself is a function)

will look at all the other places this needs to be update too! Thanks for calling out :+1:

NicholasBoll:
> @NicholasBoll and @anicholls, yea can update to:
> `onChange?: (event: React.ChangeEvent) => void`
> 
> will look at all the other places this needs to be update too! Thanks for calling out 👍

You'll want to add the element in there. Otherwise it will default to `Element`:

```ts
// input elements
onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void

// select element
onChange?: (event: React.ChangeEvent<HTMLSelectElement>) => void

// etc
```

lychyi:
@anicholls Interesting, I would guess that you would have been all over option 1 😄. It's just more efficient to me and you get the same signature in the IDE for help.

However from a docs perspective, option 2 is much clearer.

donovangini:
@anicholls, @NicholasBoll, I made updates to all the places with `onChange`. The only one I didn't convert was [this one in RadioGroup](https://github.com/Workday/canvas-kit/blob/f3f942d46edb13cc24ed5972d6dcc2bbb07a80af/modules/radio/react/lib/RadioGroup.tsx#L99), because the event is typed as `MouseEvent<HTMLButtonElement>`... I don't see why it would be a Button, and not be able to switch this over too, but wanted to be sure before I do! otherwise we can leave as synthetic event there. Let me know what you think!

anicholls:
@donovangini I think the `RadioGroup` type is a mistake. That should be updated to align with the other events.

Also, there are a few that still need to be updated that are not associated with `onChange`. If you `git grep SyntheticEvent`, you will see them.

Thanks again!

modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -215,7 +215,7 @@ const Combobox = ({
   }, [autocompleteItems, isFocused, value]);
 
   const handleAutocompleteClick = (
-    event: React.SyntheticEvent,
+    event: React.KeyboardEvent | React.MouseEvent,
```
alanbsmith, line 218: I'm open to ideas about how to better handle this, but because we use this function for keyboard and click events, I went with this typing.

modules/banner/react/stories/stories.tsx
```diff
@@ -7,7 +7,7 @@ import withReadme from 'storybook-readme/with-readme';
 import Banner from '../index';
 import README from '../README.md';
 
-const handleBannerClick = (e: React.SyntheticEvent) => {
+const handleBannerClick = () => {
```
alanbsmith, line 10: We weren't using the event here, so I removed it. I can add it back for documentation purposes if we'd like though.

modules/_labs/header/react/lib/GlobalHeader.tsx
```diff
@@ -16,7 +16,7 @@ export interface GlobalHeaderProps {
   /**
    * The function called when the responsive menu icon is clicked.
    */
-  onMenuClick?: (e: React.SyntheticEvent) => void;
+  onMenuClick?: (e: React.MouseEvent<HTMLButtonElement>) => void;
```
alanbsmith, line 19: My general rule for typing these was to provide a more specific typing if I was fairly certain about the element being passed in the click event, and more general if the element was ambiguous. We can definitely de-specify some of these types if wanted though.

modules/_labs/combobox/react/lib/Combobox.tsx
NicholasBoll: This is an internal function. Nothing is being used from the `event` object. This function does call `menuItem.onClick` which is from `MenuItem.tsx` which is expecting a mouse event. Did you run into any issues there?

The external interface is the `onChange` which is a change event on an `input` element. I think we'd be okay if we had a signature like this - either `KeyboardEvent | MouseEvent`. For the external consumer, they would have to do detection if they wanted access to something unique about the event (like which mouse button was down or what key was detected).

modules/_labs/header/react/README.md
```diff
@@ -169,7 +169,7 @@ Default: `DubLogoTitle` (for "Dub" variants) or `WorkdayLogoTitle` (for "Full" v
 
 ---
 
-#### `onMenuClick: (React.SyntheticEvent) => void`
+#### `onMenuClick: (React.MouseEvent<HTMLButtonElement>) => void`
```
NicholasBoll, line 172: Interesting side-note - a button click could be triggered by the "enter" or "space" keys. They are synthesized as a mouse click event with zeroed position information.

modules/_labs/header/react/lib/GlobalHeader.tsx
NicholasBoll: That is a good thing to point out. Being more specific does allow an app developer to get a more accurate `target` interface. It would mean that changing the element that we use would be a breaking change.

Is the interface of the target element important here?

```ts
let button = document.createElement('button')
let div = document.createElement('div')

div = button // TS error: HTMLButtonElement is missing 'align'
button = div // TS error: missing button props

type OmittedProps = keyof Omit<HTMLButtonElement, keyof HTMLDivElement>
//  "form" | "disabled" | "formAction" | "formMethod" | "formNoValidate" | "formTarget" | "name" | "type" | "value" | "formEnctype" | "labels" | "validationMessage" | "validity" | "willValidate" | "checkValidity" | "reportValidity" | "setCustomValidity"
```

Almost all of those extra properties have to do with being part of a form. I think specifying the element is extremely helpful for form fields. I'm not sure it is useful outside the context of a form field.

For example, the MenuItem - the element type might depend on the context the menu is used in. A navigation menu might have buttons as the target element, but an ARIA role of `menuitem` while a combobox might have `li` with a ARIA role of `option`. We might make a decision on the element based on some inherit properties of the element, but choose a role based on the correct ARIA context.

In some generic contexts, we may want to do either:
* Use `HTMLElement` because we're making no guarantees on the specific element interface
* Use a generic interface defaulting to `HTMLElement`: `<T = HTMLElement>`

modules/_labs/menu/react/lib/MenuItem.tsx
```diff
@@ -15,7 +15,7 @@ export interface MenuItemProps extends React.LiHTMLAttributes<HTMLLIElement> {
   /**
    * The function called when the MenuItem is clicked. If the item is a child of the Menu component, this callback will be decorated with the onSelect and onClose Menu callbacks. This callback will not fire if the item is disabled (see below).
    */
-  onClick?: (event: React.SyntheticEvent) => void;
+  onClick?: (event: React.MouseEvent<HTMLLIElement>) => void;
```
NicholasBoll, line 18: I'm not sure these interfaces should be specific to an `<li>`. I don't think there is anything inherently useful to an application developer on the `HTMLLIElement` interface. We may use different ARIA roles depending on the context the `MenuItem` is used.

I checked what special attributes/properties are available to the LI element:

```ts
type OmittedProps = keyof Omit<HTMLLIElement, keyof HTMLElement>;
// 'type' | 'value'
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li#Attributes

The `value` property is a number that represents the number of an ordered list.
The `type` property is deprecated.

modules/banner/react/stories/stories.tsx
NicholasBoll: I don't think it is necessary.

Private functions like this aren't in documentation anywhere. If the component has a specific interface, you'll have to set this correctly to assign it to the Component's prop list.

modules/_labs/combobox/react/lib/Combobox.tsx
alanbsmith: Yep, I'm fixing those issues now. Having a keyboard and mouse event call the same function is tricky to type.

NicholasBoll:
Changing the base introduced a conflict. Should be ready-to-go after that is resolved

PR #500, opened 2020-03-06
build: Improve PR build times

lychyi:
## Summary

We can speed up our PRs (and master builds) by reducing the amount of times Storybook gets built between Cypress, Chromatic, and build-storybook

### TODO
- [x] Make travis exit on first non-zero exit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,13 +41,15 @@ jobs:
   include:
   - stage: pull_request
     script:
-      - yarn lint
-      - yarn depcheck
-      - yarn test
+      - >-
+        yarn concurrently
+        "yarn lint"
+        "yarn depcheck"
+        "yarn test"
       - >- # Chromatic and Cypress rely on Storybook being built, this will exit early if build-storybook fails
         yarn build-storybook &&
-        yarn http-server docs -p 9001 & npx wait-on http://localhost:9001 &&
-        npx concurrently
+        yarn http-server docs -p 9001 & yarn wait-on http://localhost:9001 &&
+        yarn concurrently
         "chromatic --quiet --exit-zero-on-changes --storybook-build-dir docs"
         "yarn cypress run --record" &&
         kill $(jobs -p) || true
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ jobs:
   - stage: master
     script:
       - yarn build-storybook # build storybok to .storybook and then we can use it in chromatic
-      - yarn chromatic --quiet --auto-accept-changes --storybook-build-dir .storybook
+      - yarn chromatic --quiet --auto-accept-changes --storybook-build-dir docs
       - yarn build
     env:
       - CHROMATIC_APP_CODE="dlpro96xybh"
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,13 @@ addons:
     packages:
     - libgconf-2-4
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -66,8 +68,8 @@ jobs:
 
   - stage: master
     script:
-      - yarn chromatic --quiet --auto-accept-changes
-      - yarn build-storybook # chromatic runs `build-storybook`, but puts it into a random directory. We just build it again
+      - yarn build-storybook # build storybok to .storybook and then we can use it in chromatic
```
NicholasBoll: I think you could just remove this line. If you told `yarn chromatic` to build into the same directory as `build-storybook`, this will still build twice, just in the same directory (unless Storybook bails early if already built?)

.travis.yml
```diff
@@ -10,11 +10,13 @@ addons:
...
     - libgconf-2-4
 
-# cache both yarn and Cypress binary
+# cache both yarn and Cypress binar
+# also cache TypeScript builds, this only rebuilds when code changes
 cache:
   yarn: true
   directories:
     - ~/.cache
+    - .build-info
```
NicholasBoll: Will this work? The documentation wasn't clear about subdirectories. Does `.build-info` match `**/.build-info`?

NicholasBoll: Assuming this does work for Travis, I don't know much about how a TS build cache works. From what I can tell, this determines a graph for both type checking _and_ emitting. It looks like we emit code using Typescript. For this to work correctly, don't we also need to cache `dist` folders? I think timestamps and versions are fine for type-checking as type-checks have _no output_, but emitting does have an output.

.travis.yml
```diff
@@ -43,8 +44,9 @@ jobs:
       - yarn lint
       - yarn depcheck
       - yarn test
-      - yarn chromatic --quiet --exit-zero-on-changes
-      - yarn start & npx wait-on http://localhost:9001
+      - yarn build-storybook
+      - yarn chromatic --quiet --exit-zero-on-changes --storybook-build-dir docs
+      - yarn http-server docs -p 9001
```
NicholasBoll:
```suggestion
      - yarn http-server docs -p 9001 & npx wait-on http://localhost:9001
```

NicholasBoll: You'll need this. `http-server` will never return anything and will hold the Travis executor until Travis times out. `wait-on` will return execution when a ping comes back. You might not need the `wait-on`, but you will need the `&`

lychyi:
To exit early, you have to either write a bash script with `set -e` or use `&&` for multiple commands inside the .yml file.

I'm using the YAML's block style to write this in a way that's still readable in our travis.yml

package.json
```diff
@@ -122,7 +124,8 @@
     "tslint-config-prettier": "^1.14.0",
     "tslint-react": "^3.6.0",
     "typescript": "^3.5.1",
-    "typescript-eslint": "^0.0.1-alpha.0"
+    "typescript-eslint": "^0.0.1-alpha.0",
```
NicholasBoll: Hmm. This seems to be a dead npm package. I don't see anywhere where it's used in our codebase. I see use of the `@typescript-eslint` monorepo of packages. Perhaps this was a typo @alex?

anicholls: Hmm. It very well could have been. I can open an issue to remove it.

PR #504, opened 2020-03-10
ci: Travis build speed improvements

lychyi:
Things we did to speed up Travis (hopefully for real this time):
- Exit Chromatic after upload so it doesn't hold up the build
- Remove caching for yarn and npm (node_modules). Loading and building the cache is longer than the savings incurred by having a cache.
- Deduplicate instances of Webpack in our yarn.lock
- Replace `concurrently` with `run-p` (part of `npm-run-all`, a dep we already use). Concurrently was a 40mb hit to our cache and didn't add any value from what we already have.

We went from around 20 to 25 min to 15min. Not too shabby.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ addons:
 
 # cache both yarn and Cypress binary
 cache:
-  npm: false # caching node_modules is expensive and time consuming, we create longer builds with this
+  npm: true
   directories:
     - ~/.cache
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -35,23 +34,23 @@ before_install:
...
+        test
       - >- # Chromatic and Cypress rely on Storybook being built, this will exit early if build-storybook fails
         yarn build-storybook &&
-        yarn http-server docs -p 9001 & yarn wait-on http://localhost:9001 &&
-        yarn concurrently
-        "yarn chromatic --quiet --exit-zero-on-changes --storybook-build-dir docs"
-        "yarn cypress run --record" &&
+        yarn http-server docs -p 9001 & npx wait-on http://localhost:9001 &&
+        yarn run-p -l
+        "chromatic -- --quiet --exit-zero-on-changes --exit-once-uploaded --storybook-build-dir docs"
```
lychyi: Most of the work is done on Chromatic's side so this allows us to make our Chromatic build command exit after we've uploaded the Storybook source. This also ensures that we don't incur CI build time increases in Travis as we add more visual testing stories.

.travis.yml
```diff
@@ -35,23 +34,23 @@ before_install:
...
 jobs:
   include:
   - stage: pull_request
     script:
       - >-
-        yarn concurrently
-        "yarn lint"
-        "yarn depcheck"
-        "yarn test"
+        yarn run-p -l
```
lychyi: For some reason, `concurrently` is about 40mb and `npm-run-all/run-p` pretty much do the same thing at a fraction of the size.

package.json
```diff
@@ -137,6 +135,7 @@
     "create-module": "./utils/create-component/createComponent.js",
     "create-component": "./utils/create-component/createComponent.js",
     "precommit": "lint-staged",
+    "chromatic": "chromatic",
```
lychyi: Need this to run chromatic via `run-p` in Travis

.travis.yml
NicholasBoll:
```
--exit-once-uploaded
```

Genius. I didn't matter much when we only had Chrome and it took like 15s for 300 stories. IE11 is much slower. We already have a status check that is reported by Chromatic. We don't need to wait for it here. 👍

.travis.yml
```diff
@@ -74,7 +73,7 @@ jobs:
     script:
       - >- # Chromatic relies on a built Storybook, so exit early if build-storybook fails
         yarn build-storybook &&
-        yarn chromatic --quiet --auto-accept-changes --storybook-build-dir docs
+        yarn chromatic --quiet --auto-accept-changes --exit-once-uploaded --storybook-build-dir docs
```
NicholasBoll: 👍

yarn.lock
```diff
@@ -23602,65 +23475,6 @@ webpack@4.41.5:
...
-    micromatch "^3.1.10"
-    mkdirp "^0.5.1"
-    neo-async "^2.6.1"
-    node-libs-browser "^2.2.1"
-    schema-utils "^1.0.0"
-    tapable "^1.1.3"
-    terser-webpack-plugin "^1.4.1"
-    watchpack "^1.6.0"
-    webpack-sources "^1.4.1"
-
```
NicholasBoll: Nice. Less webpacks. It looks like those other webpack versions had the same dependencies, so it wasn't too bad, but nice to get it cleaned up

PR #505, opened 2020-03-10
refactor: Rename and move IconButtonToggleGroup to SegmentedControl

anicholls:
## Summary

As part of the button refactor, we've also decided to rename `IconButtonToggleGroup` to `SegmentedControl` and move it out into its own component. I placed it in labs for now, but if we're confident in it (it hasn't changed much), maybe we can skip that.

This is also in preparation for adding text support to this component (https://github.com/Workday/canvas-kit/issues/207).

Blocked on #509 

## Breaking Changes

- `IconButtonToggleGroup` is no longer included in `@workday/canvas-kit-react-button`

Before:
```tsx
import {IconButtonToggleGroup} from '@workday/canvas-kit-react-button'

<IconButtonToggleGroup>
  <IconButton icon={listViewIcon} title="List View" aria-label="List View" />
  <IconButton icon={worksheetsIcon} title="Worksheets" aria-label="Worksheets" />
</IconButtonToggleGroup>
```

After:
```tsx
import {SegmentedControl} from '@workday/canvas-kit-labs-react-segmented-control'

<SegmentedControl>
  <IconButton icon={listViewIcon} title="List View" aria-label="List View" />
  <IconButton icon={worksheetsIcon} title="Worksheets" aria-label="Worksheets" />
</SegmentedControl>
```


modules/_labs/segmented-control/react/lib/SegmentedControl.tsx
```diff
@@ -0,0 +1,108 @@
...
+    '&:first-of-type': {
+      borderRadius: `${borderRadius.m} 0 0 ${borderRadius.m}`,
+      borderLeft: `1px solid ${colors.soap500}`,
+    },
+    '&:last-of-type': {
+      borderRadius: `0 ${borderRadius.m} ${borderRadius.m} 0`,
+    },
+    '&:focus': {
+      borderRadius: borderRadius.m,
+      zIndex: 1,
```
mannycarrera4: will this by any chance affect if it's in a different z index context?

anicholls: zIndex is only relative to siblings so it should be fine 👍

PR #507, opened 2020-03-10
feat(drawer): Make the close icon in DrawerHeader optional

### LINKED ISSUE #506: Drawer Header Component has a close button by default
## 🐛 Bug Report

This is behaving as intended, however it would be nice to be able to have the option of not rendering a close icon. 

Drawer Header Component has a close button by default. 

## To Reproduce

In order to use the Drawer Header Component I have to use the close icon. 

## Expected Behavior

Make the an option to have a close icon. If we pass a name it would render the close Icon. If we don't, it doesn't render a icon. 

## Actual Results

The close icon renders.

## Browser (if applicable)

n/a

n/a

### Error Output

n/a

devonsoto77:
## Summary

We want to be able to use the side panel with the header. However we are going to keep it open and therefore we do not want to render a close button. 

Screen Shot 2020-03-10 at 2 14 46 PM

Closes #506 

## Checklist

- [x ] branch has been rebased on the latest master commit
- [ x] tests are changed or added
- [ x] `yarn test` passes
- [ x] all (dev)dependencies that the module needs is added to its `package.json`
- [ x] code has been documented and, if applicable, usage described in README.md
- [ x] module has been added to `canvas-kit-react` and/or `canvas-kit-css` universal modules, if
      applicable
- [ x] design approved final implementation
- [ x] code adheres to the [API & Pattern guidelines](../API_PATTERN_GUIDELINES.md)

### DIFF
--- a/modules/_labs/drawer/react/lib/DrawerHeader.tsx
+++ b/modules/_labs/drawer/react/lib/DrawerHeader.tsx
@@ -73,7 +73,7 @@ const CloseButton = styled(IconButton)({
 
 export default class DrawerHeader extends React.Component<DrawerHeaderProps, {}> {
   static defaultProps = {
-    iconLabel: 'Close',
+    closeIconLabel: 'Close',
     headerColor: colors.soap100,
     borderColor: colors.soap500,
     showInverseButton: false,
@@ -97,14 +97,14 @@ export default class DrawerHeader extends React.Component<DrawerHeaderProps, {}>
         <HeaderTitle id={id} inverse={inverse} title={title}>
           {title}
         </HeaderTitle>
-        {onClose && closeIconLabel ? (
+        {onClose && closeIconLabel && (
           <CloseButton
             variant={inverse ? IconButtonVariant.Inverse : IconButtonVariant.Plain}
             onClick={onClose}
             aria-label={closeIconLabel}
             icon={xIcon}
           />
-        ) : null}
+        )}
       </HeaderContainer>
     );
   }
--- a/modules/_labs/drawer/react/lib/DrawerHeader.tsx
+++ b/modules/_labs/drawer/react/lib/DrawerHeader.tsx
@@ -12,12 +12,12 @@ export interface DrawerHeaderProps extends React.HTMLAttributes<HTMLDivElement> 
   /**
    * The function called when the DrawerHeader close button is clicked.
    */
-  onClose: IconButtonProps['onClick'];
+  onClose?: IconButtonProps['onClick'];
   /**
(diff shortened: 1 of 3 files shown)


mannycarrera4:
@devonsoto77 can you also add a unit test for this

modules/_labs/drawer/react/lib/DrawerHeader.tsx
```diff
@@ -12,12 +12,12 @@ export interface DrawerHeaderProps extends React.HTMLAttributes<HTMLDivElement>
...
    * The function called when the DrawerHeader close button is clicked.
    */
-  onClose: IconButtonProps['onClick'];
+  onClose?: IconButtonProps['onClick'];
   /**
    * The `aria-label` for the DrawHeader close button. Useful for i18n.
    * @default Close
    */
-  closeIconLabel: string;
+  closeIconLabel?: string;
```
anicholls: Just noticed a bug with the default for this. It's using `iconLabel` instead of `closeIconLabel`. Do you mind updating that as well?

modules/_labs/drawer/react/lib/DrawerHeader.tsx
```diff
@@ -97,12 +97,14 @@ export default class DrawerHeader extends React.Component<DrawerHeaderProps, {}>
         <HeaderTitle id={id} inverse={inverse} title={title}>
           {title}
         </HeaderTitle>
-        <CloseButton
-          variant={inverse ? IconButtonVariant.Inverse : IconButtonVariant.Plain}
-          onClick={onClose}
-          aria-label={closeIconLabel}
-          icon={xIcon}
-        ></CloseButton>
+        {onClose && closeIconLabel ? (
```
anicholls: Nit: You can use short circuiting to simplify this: `{onClose && closeIconLabel && <CloseButton ... />}`

modules/_labs/drawer/react/spec/Drawer.spec.tsx
```diff
@@ -19,6 +19,11 @@ describe('Drawer', () => {
     expect(cb).toHaveBeenCalledTimes(1);
   });
 
+  test('should not render a close icon button', async () => {
+    const {container} = render(<DrawerHeader headerText={'Header Title'} />);
+    expect(container.firstChild.childNodes.length).toEqual(1);
+  });
+
```
mannycarrera4: Maybe try something like this?
```test('should not display a close icon', async () => {
    const {container} = render(<DrawerHeader headerText={'Header Title'}></DrawerHeader>);
    expect(container.querySelector('svg')).not.toBeInTheDocument();
  });```

PR #519, opened 2020-03-17
fix(color-picker): Design & use case improvements

anicholls:
## Summary

After #462, there are a few things we want to revisit from a design and a11y perspective.

## Checklist

Architecture:
- [x] Look at aligning with the existing color input swatch.

Design:
- [x] Reset button expands to full width to align with menus/dropdowns
- [x] Use `Popup` as the container for standard animation/transition
- [x] Bump selected icon up to 24px so it renders properly
- [x] Move to 8px rounded corners like our other popups
- [x] Add missing selected state inset border
- Is this even the right way to organize this component? It may make sense to have two different variations: custom color, and specific color.
  - **Custom Color**: Launch the picker from a ColorInput with a combobox role. You could either select a color from the popup (which would set the value of the input), or type directly into the input itself
  - **Specific Color**: A button (or ReadOnly ColorInput) launches the popup where you can only select specified colors.

Accessibility:
Currently awaiting a review, but here are a few things to consider:
- Should we use buttons for the swatches so we don't have to add `tabIndex="0"`?
- Should this be keyboard navigable (menu items) instead of having so many different focus stops? `aria-selected` attributes could be used to inform screen readers which color is currently "selected" vs. "focused". "Reset Color" is simply another option.

**Update:** As dicussed, we will leave the a11y improvements to a future PR


modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -163,7 +152,7 @@ const ColorPicker = ({
   };
 
   return (
-    <Container data-testid="canvas-color-picker" {...elemProps}>
+    <Popup width={250} padding={Popup.Padding.s} data-testid="canvas-color-picker" {...elemProps}>
```
NicholasBoll: I didn't notice the `data-testid`

NicholasBoll:
@anicholls I've had some conversations with design and accessibility.

I think we have 3 variations:
* **SelectOne**: You have to select a color from the list. No custom colors allowed. `role="listbox"`
* **ColorInput**: You can enter any valid hex. The color picker shows a list of preselected options. `role="combobox"`
* **MenuItem**: ColorPicker is launched from a menu. The target is typically an icon. This has swatches an optional "Custom Color" ColorInput inside the picker. Workday Sheets has such a menu color picker now, although the accessibility needs work.

Color Picker will have to be portalled to avoid issues with overflow containers the target element may be a part of.

modules/card/react/lib/Card.tsx
```diff
@@ -44,7 +45,9 @@ export interface CardProps extends React.HTMLAttributes<HTMLDivElement> {
   height?: number | string;
 }
 
-const Box = styled('div')<CardProps>(
+const Box = styled('div', {
+  shouldForwardProp: prop => isPropValid(prop) && prop !== 'width' && prop !== 'height',
```
alexandrzavalii: maybe without negation
```
 isPropValid(prop) || prop === 'width' || prop === 'height',
```

modules/popup/react/lib/Popup.tsx
```diff
@@ -77,7 +78,9 @@ const popupAnimation = (transformOrigin: TransformOrigin) => {
   `;
 };
 
-const Container = styled('div')<Pick<PopupProps, 'transformOrigin' | 'width'>>(
+const Container = styled('div', {
+  shouldForwardProp: prop => isPropValid(prop) && prop !== 'width',
```
alexandrzavalii: same here
```
isPropValid(prop) || prop === 'width'
```

modules/card/react/lib/Card.tsx
anicholls: Not sure I follow - `prop === 'width'` would return true for `width`, which would mean it does get forwarded to the html element (not what we want)

modules/_labs/color-picker/react/lib/ColorPicker.tsx
anicholls: Yeah - I didn't worry about it too much because it was in the story and not baked into the component. But we can rework to remove this if you'd prefer.

modules/card/react/lib/Card.tsx
alexandrzavalii: @NicholasBoll Oh Okay, my bad! I thought we want to pass it

anicholls:
@NicholasBoll I think this is good to go. All of the design updates have been made and I've updated the stories to reflect the 3 core use cases.

As discussed, I have left out the a11y improvements. We will spike on making use of a grid-based menu system in the future so we don't have so many tab stops in a future PR.

Issue #514, opened 2020-03-12
[Menu] ARIA menuitem roles are in camel case - 572


jpante:
## 🐛 Bug Report

The aria role for menu items on the menu component should not be in camel case.

## To Reproduce
https://workday.github.io/canvas-kit/?path=/story/labs-menu-react--default 
Inspect any of the menu items. 

## Expected Behavior

ARIA roles must never be capitalized or use camel case.  (e.g. menuitem)

## Actual Results

The <li> elements have ARIA menuitem roles with an incorrect capital 'I'.

PR #520, opened 2020-03-18
fix(menu): Update aria role for menu

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/514
## Checklist

### DIFF
--- a/modules/_labs/menu/react/lib/Menu.tsx
+++ b/modules/_labs/menu/react/lib/Menu.tsx
@@ -40,11 +40,6 @@ export interface MenuProps extends GrowthBehavior, React.HTMLAttributes<HTMLULis
    * The HTML `id` of the element that labels the Menu. Often used with menu buttons.
    */
   labeledBy?: string;
-  /**
-   * The role of the Menu. Use this to override the role of the item (e.g. you can use this element as an option in a Combobox).
-   * @default menu
-   */
-  role: string;
 }
 
 export interface MenuState {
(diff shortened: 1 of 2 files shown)


modules/_labs/menu/react/lib/Menu.tsx
```diff
@@ -40,6 +40,11 @@ export interface MenuProps extends GrowthBehavior, React.HTMLAttributes<HTMLULis
    * The HTML `id` of the element that labels the Menu. Often used with menu buttons.
    */
   labeledBy?: string;
+  /**
+   * The role of the Menu. Use this to override the role of the item (e.g. you can use this element as an option in a Combobox).
+   * @default menu
+   */
+  role: string;
```
anicholls: Why does this bug fix involve expanding the API? I can't imagine a menu component having any other role than `menu`

mannycarrera4: i can revert this, but i saw that the role in menu wasn't a prop

anicholls: The only reason it's configurable for `MenuItem` is because they can also be used in a combobox (who's items require a different role). I think we're good to keep `Menu`'s role static 👍

modules/_labs/menu/react/lib/Menu.tsx
```diff
@@ -125,7 +125,7 @@ export default class Menu extends React.Component<MenuProps, MenuState> {
         width={cardWidth}
       >
         <List
-          role="menu"
+          role="role"
```
anicholls: Whoops

```suggestion
          role="menu"
```

mannycarrera4: lol, that's my bad.... quarantine brain

Issue #508, opened 2020-03-11
Checkbox does not truncate long labels


ofalsap:
## 🐛 Bug Report
The label in the Checkbox component does not truncate long text when it is used inside a div with a specified width. The label is a prop of the Checkbox component. The Checkbox component may benefit to have an optional width props. I am happy to work on a fix for this if you think my suggestion is feasible.

## To Reproduce

Try to have a long text label in the Checkbox component used in a div.

## Expected Behavior
If the width prop has been set up :  the long label of the Checkbox component will truncate .

If the width prop has NOT been set up: the long label of the Checkbox component will over flow.

## Actual Results

The label in the Checkbox component does not truncate long text when in a div with specific width.

### Error Output
<img width="247" alt="Screenshot 2020-03-12 at 10 45 17" src="https://user-images.githubusercontent.com/62067641/76513703-9946c480-644e-11ea-892e-8cea9b7dc26a.png">

NicholasBoll:
This is what I see: 
<img width="296" alt="image" src="https://user-images.githubusercontent.com/338257/150229634-5c5e0c4d-5001-45df-93d1-0c310e3ef106.png">

Codesandbox link: https://codesandbox.io/s/sad-dawn-i6whr

Does your application override the wrapping of the checkbox?

jaclynjessup:
Can't reproduce & inactive issue.

PR #521, opened 2020-03-18
feat:  Add overflow behavior for label truncation for Checkbox and Radio

ofalsap:
## Summary
Fix for issue https://github.com/Workday/canvas-kit/issues/508 where the label of Checkbox component does not truncate when long text is used inside a div with a specified width. The label is a prop of the Checkbox component. 

In this PR I have added an optional width prop. If the width is set, the label may truncate and will show ellipsis. 
## Checklist

- [ ] `yarn test` passes
      applicable


modules/checkbox/react/lib/Checkbox.tsx
```diff
@@ -58,6 +58,10 @@ export interface CheckboxProps extends Themeable, React.InputHTMLAttributes<HTML
    * @default false
    */
   indeterminate?: boolean;
+  /**
+   * If set the Checkbox label may be truncated
+   */
+  width?: number;
```
mannycarrera4: @ofalsap thanks for contributing! I'm wondering if we can change this prop into a boolean `isTruncated` that would basically apply the truncated styles. The width prop isn't as clear al though it achieves what you want to.

alexandrzavalii: In order to make it responsive maybe we can have `width:100%` Container will defining the width

anicholls: I think we may also want to support wrapping as well. What if we had something like:

`modules/common/react/lib/types.tsx` (since it will need to be used by `Radio` as well)
```tsx
export interface OverflowBehavior {
  overflow?: 'default' | 'wrap' | 'truncate'
}
```

`modules/checkbox/react/lib/Checkbox.tsx` (and `Radio`)
```tsx
export interface CheckboxProps extends OverflowBehavior, Themeable, React.InputHTMLAttributes<HTMLInputElement> 

// ...

({overflow}) => {
  if (overflow === 'wrap') {
    return { 
      wordWrap: 'break-word' 
    }
  }
  else if (overflow === 'truncate') {
    return {
      width: 100%,
      whiteSpace: 'nowrap', 
      overflow: 'hidden', 
      textOverflow: 'ellipsis'
    }
  }
}

Comment (ofalsap) [reply modules/checkbox/react/lib/Checkbox.tsx]: Hey guys! Thank you for the feedback! Will give it a go and update you again :)

Comment (ofalsap) [reply modules/checkbox/react/lib/Checkbox.tsx]: > @ofalsap thanks for contributing! I'm wondering if we can change this prop into a boolean `isTruncated` that would basically apply the truncated styles. The width prop isn't as clear al though it achieves what you want to.

Hey guys I tried doing it this way, instead of having width but with the width set to 100% the text does not truncate when I check it in story book? Do you guys mind providing advice on this? Thank you :)

File: modules/form-field/react/stories/stories_Checkbox.tsx [hunk 2]
```diff
@@ -94,6 +99,11 @@ storiesOf('Components|Inputs/Checkbox/React/Left Label', module)
       {control(<Checkbox label="Checkbox option" indeterminate={true} />)}
     </FormField>
   ))
+  .add('Truncate ', () => (
+    <FormField label="Label" inputId="my-checkbox-field">
+      {control(<Checkbox label="Long label" overflow="truncate" />)}
+    </FormField>
+  ))
```
mannycarrera4: maybe this meant to test overflow?

anicholls:
Hey @ofalsap! I just pushed some changes. I moved your story to our visual testing stories and displayed all states rather than using knobs so it will get picked up by our visual testing. 

I think everything is good to go with Checkbox except for one thing.  The prop is working with Checkbox on its own, but there’s still an issue with ellipsis when it’s in a FormField with the label on the left.

If you want to have a look, the problem is that flexbox doesn’t like flexing items with `white-space: nowrap`. Haven’t figured out a way around this without setting an explicit width on the FormFieldInputContainer yet, and we would really like to avoid explicit width setting anywhere for responsiveness.

anicholls:
@ofalsap Sorry for the delay! I just pushed a fix for the issue I noted in my last comment. 

If you have time, it's ready to be applied to the Radio component 🍻

ofalsap:
> @ofalsap Sorry for the delay! I just pushed a fix for the issue I noted in my last comment.
> 
> If you have time, it's ready to be applied to the Radio component 🍻

Hey guys ! No worries and thank you so much! For the radio component, would I have to create a new PR or I can just add it here? Thanks :)

anicholls:
Thanks for adding the `Radio` updates @ofalsap! This is looking good. 

We're going to review it with design & a11y on Wednesday. There have been some concerns about not including a tooltip or title attribute when truncated, so there may be a little more work to do. We will get back to you with the outcome of that meeting :+1:

anicholls:
Hey @ofalsap! As you are aware, there are still ongoing discussions about the support of truncation for accessibility reasons. We don't want those conversations to block this PR, so I've temporarily removed the `truncate` option. We're going to merge with just the `wrap` option for now, and can add the truncation support in a follow up PR depending on the outcome of the a11y discussions.

anicholls:
Hi @ofalsap!

Deepest apologies for letting this sit so long. After my message above, we realized there are some regressions to the `FormField` component. Since the `wrap` functionality is basically the same as the default styling from before this PR, we decided to hold off on merging.

Since the fate of the truncate functionality is still up in the air with accessibility, we're going to close this PR until we have a more solid path forward. Let us know if you have any questions. Thanks for the contribution regardless!

Issue #364, opened 2019-12-10
Fix: a11y - Conflicting value announcements for color input 542


jpante:
## Bug Report

 Announces "#
Label edit
FFFFFF
blank"

## Expected Behavior

Announces "Label edit blank hex value" or announces

"Label edit
FFFFFF". If the FFFFFF value will not be inputed/accepted on validation should announce the first way with descriptive text indicating the input takes a hex value. If it will be inputed/accepted then should announce the second way with the value being displayed and not automatically removed on focus which is jarring.

anicholls:
@jpante Can you clarify the expected behavior? 

---

> will not be inputed/accepted on validation

Does this mean not editable/read only?

>  If the FFFFFF value will not be inputed/accepted on validation should announce the first way with descriptive text indicating the input takes a hex value.

Why would we describe that the input takes a hex value if you can't actually type into it? 

> not automatically removed on focus which is jarring.

What is this referring to?

---

This is my understanding, but correct me where I'm wrong:

If the hex value is not editable, it should announce "Label edit blank {HEX}" (where {HEX} is "hex value" if no value is defined, or the value of the input (e.g. FFFFFF) if it is defined)

If the hex value is editable, it should announce "Label edit {HEX}" (where {HEX} is the value of the input). It should also have descriptive text indicating the input takes a hex value.

Remaining questions: 
- What attribute should that descriptive text go in?
- Still not sure what the focus comment is referring to.

jpante:
@anicholls 
After further discussion with A11y team. the suggested approach is as follows
- Remove the placeholder attribute on input entry, to eliminate the double readout
- add aria-hidden to the # element to ignore it's read out

NicholasBoll:
Fixed by #639

PR #523, opened 2020-03-18
fix(color-picker): Hide hash sign for accessibility

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/364
It now announces `Label, edit text, FFFFFF`
## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/color-picker/react/lib/ColorInput.tsx
+++ b/modules/color-picker/react/lib/ColorInput.tsx
@@ -177,7 +177,9 @@ export default class ColorInput extends React.Component<ColorInputProps> {
             icon={checkSmallIcon}
           />
         ) : null}
-        <PoundSignPrefix disabled={disabled}>#</PoundSignPrefix>
+        <PoundSignPrefix aria-hidden={true} disabled={disabled}>
+          #
+        </PoundSignPrefix>
       </ColorInputContainer>
     );
   }


anicholls:
@mannycarrera4 I think this only partially fixes the issue. I left a comment on the issue for @jpante as the language there is quite confusing.

modules/color-picker/react/lib/ColorInput.tsx
```diff
@@ -177,7 +177,9 @@ export default class ColorInput extends React.Component<ColorInputProps> {
             icon={checkSmallIcon}
           />
         ) : null}
-        <PoundSignPrefix disabled={disabled}>#</PoundSignPrefix>
+        <PoundSignPrefix aria-hidden={true} disabled={disabled}>
```
NicholasBoll: Not sure if this is a part of this PR, but I had to look up what `PoundSignPrefix` was... (part of the problem with the Styled component API). Turns out this is a `span` element which doesn't have a `disabled` attribute. What is the `disabled={true}` supposed to do?

NicholasBoll: Oh, I see. It is for styling only: https://github.com/Workday/canvas-kit/pull/523/files#diff-f2800e7c1a68cd10843f1a74fb5e1e48R94-R97

mannycarrera4: it changes the color of the pound sign... which in no way is that clear.

anicholls:
Closing until we have more insight on https://github.com/Workday/canvas-kit/issues/364

PR #528, opened 2020-03-20
feat(tooltip): Refactor to a simpler API

### LINKED ISSUE #86: Provide built-in Popper for Tooltip
## 🚀 Feature Proposal

Create a new convenience component that wraps `Tooltip` with our [`Popper` helper component](https://github.com/Workday/canvas-kit/blob/master/modules/common/react/lib/Popper.tsx).

This change could be additive by simply overlaying the `Popper` props on top of `Tooltip`. If the `Popper` props aren't used, then the Popper.js positioning engine wouldn't be used and the consumer could implement their own positioning engine if desired.

In addition, the functionality for triggering and hiding the tooltip on hover/focus/blur should be implemented.

## Motivation

To provide the consumer with a more robust and easier to consume tooltip experience. 

## Example

```
<Tooltip placement={'right'} anchorElement={this.buttonRef.current}>
  Close
</Tooltip>
```

### LINKED ISSUE #341: Tooltip/Popper: Confined to parent with overflow: auto
## 🐛 Bug Report

Popper, by default, constrains the contents to a container if the container has an `overflow` set to `auto` or `scroll`. `hidden` works as expected where the tooltip is not constrained. 

The default example of Tooltip uses the canvas-kit Popper which uses PopperJS under the hood. By default, PopperJS constrains content, but can be overridden.

This might be more of a feature discussion. I expect any Popper content to not be bound by containers. We're essentially using Portalling to prevent constraints by containers, so I'd expect to not be restricted by said containers.

## To Reproduce

Go to https://codesandbox.io/s/tooltip-overflow-hidden-t2gjw. The left tooltip is constrained by the container with an `overflow: auto` setting. Overriding defaults allows the right tooltip to display as expected outside the bounds of the parent container.

## Expected Behavior

The Tooltip should not be constrained by any parent container unless explicitly desired.

## Actual Results

The Tooltip is constrained by the parent element if certain conditions are met (like overflow setting).

## Browser (if applicable)

All

https://codesandbox.io/s/tooltip-overflow-hidden-t2gjw

### LINKED ISSUE #349: Popper on tooltip appears sporadically
## 🐛 Bug Report

The tooltip does not always appear when hovering over the tooltip anchor object (such as the button)

if you move the mouse away toward the tooltip and then back, the tooltip no longer shows.

## To Reproduce

Go to the Canvas Kit storybook
https://workday.github.io/canvas-kit/?path=/story/tooltip--default
Follow instructions above.

## Expected Behavior

The tooltip should always show when hovering over the anchor object (button for eg).

## Actual Results

See issue description.

## Browser (if applicable)

All browsers.

See above

### Error Output

N/A

Hovering
Screen Shot 2019-12-03 at 12 02 35 PM

NOTE:
The issue stems from the Popper library (mouse event handling) and has a known fix.  It is a simple 1-line change.
https://github.com/mui-org/material-ui/issues/7680

### LINKED ISSUE #472: Upgrade Popper to v2
## 🚀 Feature Proposal

Popper is now on v2, we should upgrade this as a dep of the project.

This will affect #86 and should be taken care of at the same time perhaps?

## Motivation

Please outline the motivation for the proposal.

## Example

Please provide an example for how this feature would be used.

### LINKED ISSUE #517: Default our Popups to use React Portal
## 🚀 Feature Proposal

Based off of this PR https://github.com/Workday/canvas-kit/pull/419 there's been discussion to make all of our `Popups` to default to using [Portal](https://reactjs.org/docs/portals.html) . This issue is to convert the rest of our Popups including Toast, and Tooltips to use Portal as well.

NicholasBoll:
* Add unit tests
* Add Cypress specifications
* Add visual tests
* Upgrade Popper to v2 and use canvas-kit Popper component instead of materia-ui's
* Fixes #86
* Fixes #349 
* Fixes #341 
* Fixes #472
* Fixes #517 

Tooltip portion of #286

This change makes it easier to create accessible tooltips. It is now possible to do this:

```tsx
<Tooltip title="Close">
  <IconButton variant={IconButton.Variant.Circle} icon={xIcon} aria-label="Close" />
</Tooltip>
```

## Checklist

- [x] `yarn test` passes

## Commit Message
* Add unit tests
* Add Cypress specifications
* Add visual tests
* Upgrade Popper to v2 and use canvas-kit Popper component instead of materia-ui's

BREAKING CHANGE: Tooltip API was updated to make it easier to consume.
The original `Tooltip` did little more than add a `role="tooltip"` to a styled component. The original tooltip is now exported as `TooltipContainer` to make it easier to migrate without rewriting all tooltips. Imports will have to be updated to use the old API:
**Before:**
```ts
import { Tooltip } from '@workday/canvas-kit-react-tooltip'
```
**After:**
```ts
import { TooltipContainer as Tooltip } from '@workday/canvas-kit-react-tooltip'
```
Also with this change, the tooltip no longer gets the role `tooltip` and must be added manually.

BREAKING CHANGE: `Popper` was changed to a Functional Component with a forwarded ref. If you passed a `ref` object to `Popper` before, it will now point to the element rather than the `Popper` instance.


modules/tooltip/react/lib/Tooltip.tsx
```diff
@@ -47,12 +49,12 @@ const TooltipContainer = styled('div')<TooltipProps>(
     },
   },
   ({transformOrigin}) => ({
-    animation: tooltipAnimation(transformOrigin),
-    animationDuration: '150ms',
-    animationTimingFunction: 'ease-out',
-    transformOrigin: transformOrigin
-      ? `${transformOrigin.vertical} ${transformOrigin.horizontal}`
-      : 'top center',
+    // animation: tooltipAnimation(transformOrigin),
```
NicholasBoll: I'm going to get the position info from Popper for this

cypress/support/commands.ts
```diff
@@ -40,3 +40,70 @@ Cypress.Commands.overwrite('tab', (originalFn, subject) => {
...
+  const customMatchers = {
+    'have.ariaDescription': haveAriaDescription(args[0]),
+    'have.ariaLabel': haveAriaLabel(args[0]),
+  };
+  // See if the expectation is a string and if it is a member of Jest's expect
+  if (typeof expectation === 'string' && customMatchers[expectation]) {
+    return originalFn(subject, customMatchers[expectation]);
+  }
+
+  return originalFn(subject, expectation, ...args);
```
NicholasBoll, line 108: It is possible for these accessibility helpers to be a chai matcher library. I end up using them enough that I decided to add them here.

yarn.lock
```diff
@@ -17822,7 +17827,12 @@ polished@^3.3.1:
   dependencies:
     "@babel/runtime" "^7.4.5"
 
-popper.js@^1.14.1, popper.js@^1.14.4, popper.js@^1.14.7, popper.js@^1.15.0:
+popper.js@^1.14.1:
+  version "1.16.1"
+  resolved "https://registry.yarnpkg.com/popper.js/-/popper.js-1.16.1.tgz#2a223cb3dc7b6213d740e40372be40de43e65b1b"
+  integrity sha512-Wb4p1J4zyFTbM+u6WuO4XstYx4Ky9Cewe4DWrel7B0w6VVICvPwdOpotjzcf6eD8TsckVnIMNONQyPIUFOUbCQ==
+
+popper.js@^1.14.4, popper.js@^1.14.7:
```
NicholasBoll, line 17835: These are from Storybook

wallaby.js
```diff
@@ -33,8 +33,8 @@ module.exports = wallaby => {
 
       // Tell Jest how to resolve symlinked modules. Without this, Jest will look at source TS files and not at Wallaby's compiled & instrumented files
       (jestConfig.moduleNameMapper = {
-        '@workday/canvas-kit-react-(.*)': '<rootDir>/modules/$1/react',
-        '@workday/canvas-kit-labs-react-(.*)': '<rootDir>/modules/_labs/$1/react',
+        '@workday/canvas-kit-react-([^/]+)(/?.*)': '<rootDir>/modules/$1/react$2',
+        '@workday/canvas-kit-labs-react-([^/]+)(/?.*)': '<rootDir>/modules/_labs/$1/react$2',
```
NicholasBoll, line 37: This fixes WallabyJS module resolutions. Before we were getting mappings like this:
`@workday/canvas-kit-react-color-picker/lib/parts/ColorSwatch` -> `modules/color-picker/lib/parts/ColorSwatch/react`

cypress/integration/Tooltip.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+        beforeEach(() => {
+          cy.get('body').trigger('keydown', {
+            key: 'Escape',
+          });
+        });
+
+        // TODO: implement this functionality
+        it.skip('should close the tooltip', () => {
+          cy.findByRole('tooltip').should('not.be.visible');
+        });
```
NicholasBoll, line 55: There's some WIP on the accessibility specifications and some of them we're punting on until we have a Popup manager or even more time (like collision detection)

modules/_labs/color-picker/react/stories/stories_ColorPicker.tsx
```diff
@@ -92,8 +92,7 @@ storiesOf('Labs|Color Picker/React', module)
     ];
 
     const onBlur = (e: React.FocusEvent<HTMLInputElement>) => {
-      // @ts-ignore
-      if (!popupRef.current || !popupRef.current.popper.popper.contains(e.relatedTarget)) {
+      if (!popupRef.current || !popupRef.current.contains(e.relatedTarget as Node)) {
```
NicholasBoll, line 95: I'm unsure why `.contains` is so restrictive. It expects a `Node` which isn't directly compatible with `HTMLInputElement` without casting... Oh well, this removes the need for the `// @ts-ignore`

modules/_labs/color-picker/react/stories/stories_ColorPicker.tsx
```diff
@@ -61,8 +61,8 @@ storiesOf('Labs|Color Picker/React', module)
     const [color, setColor] = React.useState(defaultColor);
     const [colorInputValidColor, setColorInputValidColor] = React.useState(defaultColor);
     const [colorInputValue, setColorInputValue] = React.useState(defaultColor);
-    const inputRef = React.useRef(null);
-    const popupRef = React.useRef(null);
+    const inputRef = React.useRef<HTMLInputElement>(null);
+    const popupRef = React.useRef<HTMLDivElement>(null);
```
NicholasBoll, line 65: These could probably just be `Element`. I'm torn on whether we should try to keep refs 100% accurate unless we absolutely need to. It is harder to maintain

modules/common/react/lib/Popper.tsx
```diff
@@ -13,95 +13,95 @@ export interface PopperProps extends React.HTMLAttributes<HTMLDivElement> {
...
    * If true, attach the Popper to the `containerElement`. If false, render the Popper within the DOM hierarchy of its parent.
    * @default true
    */
-  portal: boolean;
+  portal?: boolean;
 }
 
-export class Popper extends React.PureComponent<PopperProps> {
-  private popper: PopperJS | null;
+export const Popper = React.forwardRef<HTMLDivElement, PopperProps>(
```
NicholasBoll, line 42: Changed to a forwardRef component to fix an `ColorPicker` story that used the `popper` property of this class's instance. It really wanted to get access to the DOM element. The story probably needs to be refactored anyway. The Color Picker can still get stuck open

modules/_labs/color-picker/react/stories/stories_ColorPicker.tsx
NicholasBoll: This is still not an ideal implementation. If you focus on the input, it will open the color picker, but hitting tab goes into the ColorPicker and it will stay open until you focus on the Color Input again and focus on something _not_ inside the color picker. Not an intuitive way to close it, but I didn't want to mess with this too much.

modules/common/react/lib/utils/getTransformFromPlacement.ts
```diff
@@ -0,0 +1,33 @@
...
+  const horizontal =
+    first === 'left'
+      ? 'right'
+      : first === 'right'
+      ? 'left'
+      : second === 'end'
+      ? 'left'
+      : second === 'start'
+      ? 'right'
+      : 'center';
```
NicholasBoll, line 27: Not the easiest to read, but well tested. I did TDD for this transformation originally manually setting up all the inputs and outputs hard-coded and then extracted the series of conditionals that lead to the transformation.

This style takes a bit to learn how to read, but it is close to pattern matching. It reads similar to a `switch` statement where if a match is true, it will return the next line, otherwise it will evaluate the next match finally defaulting to the last value.

I don't anticipate any changes to this code. If it takes a bit to understand, I think that's okay. The unit tests are very clear and each input/output is manually extracted.

modules/common/react/spec/Popper.spec.tsx
```diff
@@ -1,104 +1,79 @@
...
-  return jest.fn().mockImplementation(() => {
-    return {
-      destroy: () => {}, // eslint-disable-line no-empty-function
-    };
-  });
-});
+/// <reference types="@testing-library/jest-dom/extend-expect" />
+
+import React from 'react';
+import {render, getByTestId} from '@testing-library/react';
```
NicholasBoll, line 4: The Portal mocking that was done here and the Popper changes broke these tests in ways that I didn't understand. I rewrote the tests using React Testing Library instead and didn't need any Portal hacks. 🤷‍♂

modules/toast/react/README.md
```diff
@@ -1,7 +1,7 @@
 # Canvas Kit Toast
 
 Canvas-styled React toast component. Allows for a brief message to be be shown about a process or
-action occuring in the app.
+action occurring in the app.
```
NicholasBoll, line 4: I must have made this change a long time ago and forgot about it... I don't think it is harmful here

modules/toast/react/stories/stories.tsx
```diff
@@ -32,7 +32,7 @@ const toastData = {
 };
 
 class ToastWrapper extends React.Component<{state?: string}> {
-  public render() {
+  render() {
```
NicholasBoll, line 35: I think I changed this for Storybook MDX issues... Not really necessary, but `public` is the default if not specified. Less syntax to parse...

modules/tooltip/react/README.md
```diff
@@ -21,56 +20,17 @@ yarn add @workday/canvas-kit-react-tooltip
...
+import {Tooltip} from '@workday/canvas-kit-react-tooltip';
+import {IconButton} from '@workday/canvas-kit-react-button';
+
+const TooltipExample = () => {
+  return (
+    <Tooltip title="Close">
+      <IconButton variant={IconButton.Variant.Circle} icon={xIcon} aria-label="Close" />
+    </Tooltip>
+  );
+};
```
NicholasBoll, line 33: Wow! Isn't this simpler?!

modules/tooltip/react/stories/stories.tsx
```diff
@@ -1,76 +1,130 @@
...
   };
 
-  private open = () => {
-    this.setState({
-      open: true,
-      anchorEl: this.tooltipRef.current,
-    });
+  const createPlacement = (placement: Placement) => {
+    return (
+      <Tooltip title="Add" placement={placement}>
```
NicholasBoll, line 86: Consider using HTML here to show that HTML works inside tooltips

modules/tooltip/react/stories/stories.tsx
```diff
@@ -1,76 +1,130 @@
...
 
-interface TooltipWrapperState {
-  open: boolean;
-  anchorEl: HTMLElement | null;
-}
+export default {
+  title: 'Components|Popups/Tooltip/React',
+  component: Tooltip,
+  decorators: [withReadme(README)],
+};
```
NicholasBoll, line 20: Add chromatic parameters to show up in ChromaticQA

modules/common/react/lib/Popper.tsx
```diff
@@ -13,95 +13,95 @@ export interface PopperProps extends React.HTMLAttributes<HTMLDivElement> {
...
+  (
+    {
+      anchorElement,
+      children,
+      containerElement,
+      open = true,
```
anicholls, line 48: Nit: @mannycarrera4 has been putting all of the default props at the top in #525. We have a linting rule that requires this for function params, it just doesn't catch destructuring (maybe we should look into that?)

modules/tooltip/react/stories/stories.tsx
NicholasBoll: I reviewed this, but the visual story uses `TooltipContainer` instead of `Tooltip`. I don't think adding HTML to that story will give us much confidence unless we add configuration to `Tooltip` to show it open all the time.

modules/tooltip/react/README.md
```diff
@@ -81,25 +41,37 @@ class TooltipExample extends React.Component<{}, TooltipExampleState> {
 
 ### Required
 
-> None
+#### title: string | React.ReactNode
+
+> If present, this will switch the Tooltip into a simpler to use mode where the Tooltip component
+> can wrap a target and `title` is the contents of the tooltip. This should be a string in most
```
anicholls, line 47: Isn't this only determined by which component you use? Omitting `title` will not give you the old behavior (and the type requires `title` anyway)

```suggestion
> The text contents of the tooltip. This should be a string in most
```

modules/tooltip/react/README.md
```diff
@@ -81,25 +41,37 @@ class TooltipExample extends React.Component<{}, TooltipExampleState> {
...
 
 ---
 
 ### Optional
 
-#### `transformOrigin: TransformOrigin`
+#### type: 'label' | 'describe'
 
-> Origin from which the popup will animate from
+> If present, this will switch the Tooltip into a simpler to use mode where the Tooltip component
```
anicholls, line 66: Same here

modules/tooltip/react/README.md
```diff
@@ -81,25 +41,37 @@ class TooltipExample extends React.Component<{}, TooltipExampleState> {
...
+#### title: string | React.ReactNode
+
+> If present, this will switch the Tooltip into a simpler to use mode where the Tooltip component
+> can wrap a target and `title` is the contents of the tooltip. This should be a string in most
+> cases. HTML is supported, but only text is understood by assistive technology. This is true for
+> both `label` and `describe` modes.
+
+#### children: React.ReactNode
+
+> The contents of the target for the Tooltip.
```
anicholls, line 53: It's not really the contents. If `children` is a button and I say "the button's contents", I think of the text, not the outer element

```suggestion
> The target (anchor element) for the Tooltip.
```

modules/tooltip/react/lib/Tooltip.tsx
```diff
@@ -1,74 +1,69 @@
...
+   * * `label`: Sets the accessible name for the wrapped element. Use for icons or if tooltip `title` prop is the same as the text content of the wrapped element. E.g. IconButtons or Ellipsis tooltips.
+   * * `describe`: Sets `aria-describedby` of the wrapped element. Use if the tooltip has additional information about the target.
+   *
+   * **Note**: Assistive technology settings may ignore "aria-describedby". Consider alternative
+   * way to inform a user of additional important information.
+   * @default 'label'
+   */
+  type?: 'label' | 'describe';
+  /**
+   * If present, this will switch the Tooltip into a simpler to use mode where the
```
anicholls, line 21: Same here

modules/tooltip/react/lib/Tooltip.tsx
```diff
@@ -1,74 +1,69 @@
...
+  /**
+   * Determines the tooltip type for accessibility.
+   * * `label`: Sets the accessible name for the wrapped element. Use for icons or if tooltip `title` prop is the same as the text content of the wrapped element. E.g. IconButtons or Ellipsis tooltips.
+   * * `describe`: Sets `aria-describedby` of the wrapped element. Use if the tooltip has additional information about the target.
+   *
+   * **Note**: Assistive technology settings may ignore "aria-describedby". Consider alternative
+   * way to inform a user of additional important information.
+   * @default 'label'
+   */
+  type?: 'label' | 'describe';
```
anicholls, line 19: Nit: required props on top

modules/tooltip/react/lib/Tooltip.tsx
```diff
@@ -1,74 +1,69 @@
...
+   * if this condition isn't met
    */
-  id?: string;
+  children: React.ReactNode;
 }
 
-const tooltipAnimation = (transformOrigin: TransformOrigin) => {
-  const translate = getTranslateFromOrigin(transformOrigin, spacing.xxxs);
+export const Tooltip = ({
+  type = 'label',
```
anicholls, line 45: Nit: defaults collected on top

modules/tooltip/react/lib/TooltipContainer.tsx
```diff
@@ -0,0 +1,86 @@
...
+      right: 0,
+      bottom: 0,
+    },
+  },
+  ({transformOrigin}) => {
+    if (transformOrigin === null) {
+      return {};
+    }
+    return {
+      animation: tooltipAnimation(transformOrigin || defaultTransformOrigin),
```
anicholls, line 78: Nit: we've been moving toward destructured defaults

```suggestion
  ({transformOrigin = defaultTransformOrigin}) => {
    return {
      animation: tooltipAnimation(transformOrigin),
```

If you don't want to do this, we should remove this is a default prop in the docs (since technically the default is wrong right now)

modules/tooltip/react/stories/stories.tsx
```diff
@@ -1,76 +1,130 @@
...
+      type="describe"
+      title={
+        <div>
+          This is a <em>custom</em> tooltip with <strong>custom HTML</strong>.
+          <br />
+          Any HTML is valid here.
+          <br />
+          Keep it short and note that assistive technology will read this content as text-only and
+          maybe have verbosity turned off to the point where this content is completely ignored.
+          Consider a different way to describe important content.
```
anicholls, line 50: Can we do a less verbose custom HTML example so that we can show our true intent for this feature? I feel like this will encourage bad practice. Design tries to keep people to one line if possible. Maybe we move the messaging to below the target?

```suggestion
        <div>
          This is a <em>custom</em> tooltip with <strong>custom HTML</strong>
```

modules/tooltip/react/README.md
NicholasBoll: Ah, yes. Remnants of wip

modules/tooltip/react/README.md
NicholasBoll: That is more clear

modules/tooltip/react/lib/TooltipContainer.tsx
NicholasBoll: I can still destructure with a default though

modules/tooltip/react/lib/Tooltip.tsx
anicholls: I never really thought about it this way haha. Maybe we should discuss as a team.

I guess I look at it as "all of the things that are guaranteed to be defined come first".

PR #531, opened 2020-03-23
chore: Upgrade packages to fix vulnerabilities

### LINKED ISSUE #542: fix: Remove webpack dependency on react-scripts
Storybook 6.0 will require a CRA preset for react-scripts usage and after upgrading to 5.3, we're now seeing warnings in our logs to install this. Preferably we would remove our dependency on react-scripts entirely. 

This was evalutated in #533, but we needed to get that merged so it will have to happen in a separate PR

anicholls:
Upgrade packages to address vulnerabilities in:
- `minimist` (must be >= 0.2.1 or 1.2.3)
- `acorn` (must be >= 6.4.1)
- `kind-of` (must be >= 6.0.03)

Closes #542 

There are a few packages that are still yet to be updated:
- `commitizen>minimist`(https://github.com/commitizen/cz-cli/pull/721)
- `http-server>optimist>minimist`(https://github.com/http-party/http-server/issues/614)

I needed to use a resolution for `minimist` and `acorn` since these two don't look like they're getting much movement. We're now getting some yarn warnings, but I don't think anything is breaking and it's better than a vulnerability.

I removed `react-scripts` because we've wanted it gone for a while (#542) and it was using `acorn`. 

For the `minimist` resolution, we're waiting on a new release from `commitizen`, and I'm not sure what's going to happen for `http-server`

---

Note for future reference: For some reason upgrading `rollup-plugin-terser` was giving us problems with storybook in IE11. It was causing us to emit dynamic imports (not supported in IE) for `@storybook/addons-storysource`. I'm very confused as to why, since rollup is only used in js-build (used for one css component) and has nothing to do with webpack. It's not affected by these vulnerabilities so I've reverted the upgrade. Hopefully we don't run into this again 🤷‍♂️


.postcssrc.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  map: true,
+  plugins: {
+    'postcss-discard-duplicates': {},
+    autoprefixer: {},
+    'css-mqpacker': {},
+    'postcss-banner': {
+      banner: 'Copyright 2020 Workday, Inc.',
```
lychyi, line 8: I can't remember the exact rules for writing copyright notices but do we want to keep the year we marked it copyrighted? Or do we want to keep the year synchronized or even like a date range (e.g. `2020 - 2021` for next year).

Anyway, it's a nit, but adding a programmatic year probably wouldn't hurt

PR #532, opened 2020-03-23
feat(combobox): Convert to functional component and add translation function for status text

vibdev:
## Summary

To support extending the combobox to support more features I ported the core component to a React Functional Component. ~Nothing in behavior should be changed.~ JK updated a prop name and added a new prop. Let me know if those should go in a separate PR

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/_labs/combobox/react/lib/Combobox.tsx
+++ b/modules/_labs/combobox/react/lib/Combobox.tsx
@@ -1,13 +1,13 @@
-import * as React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import styled from '@emotion/styled';
-import {CSSObject, jsx} from '@emotion/core';
-import {accessibleHide, GrowthBehavior} from '@workday/canvas-kit-react-common';
-import {depth, spacing, commonColors, borderRadius} from '@workday/canvas-kit-react-core';
-import {MenuItemProps} from '@workday/canvas-kit-labs-react-menu';
-import {Card} from '@workday/canvas-kit-react-card';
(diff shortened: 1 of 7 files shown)


modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -121,54 +118,85 @@ const listBoxIdPart = `listbox`;
...
+  return `There ${listCount === 1 ? 'is' : 'are'} ${listCount} suggestion${
+    listCount === 1 ? '' : 's'
+  }.`;
+};
+
+const isValidSingleChild = (child: React.ReactNode) => {
+  return React.isValidElement(child) && React.Children.only(child);
+};
+
+const Combobox: React.FC<ComboboxProps> = ({
```
anicholls: Nit: We've been recommending using `const Component = (props: ComponentProps) => {}` over this format for a reason I can't seem to think of at the moment haha. I will look into it and get back to you, but do you mind updating the formatting here so it aligns with our other functional components?

vibdev: Looks like it uncovered a potential bug by typing children.props better

vibdev:
```
const Component: React.FC<ComponentProps> = (props) => {
  props.children.props.helicopter // helicopter is typed as any
}
```
```
const Component = (props: ComponentProps) => {
    props.children.props.helicopter // helicopter is typed correctly as Badass
}
```

anicholls: That was why! Thanks for testing it out 👍

PR #535, opened 2020-03-26
feat(combobox): Allow grouping of listbox items

vibdev:
## Summary

Adds groups for combobox. This is WIP so accessibility team can audit.

## Checklist

- [x] `yarn test` passes
      applicable


samsmith-workday:
Reviewed Matt's work with NVDA/JAWS. Sounds good to me.

modules/_labs/combobox/react/lib/Status.tsx
```diff
@@ -0,0 +1,50 @@
...
+      role={role}
+      aria-relevant={ariaRelevant}
+      aria-atomic={ariaAtomic}
+    >
+      {displayText}
+    </Container>
+  );
+};
+
+export default Status;
```
vibdev, line 50: Broke this out, might be nice for it to be its own component at some point.

modules/_labs/combobox/react/lib/AutocompleteList.tsx
```diff
@@ -0,0 +1,133 @@
...
+import uuid from 'uuid/v4';
+
+const Autocomplete = styled('ul')({
+  padding: 0,
+  margin: `${spacing.xxs} 0`,
+});
+
+const AccessibleHide = styled('span')({
+  ...accessibleHide,
+});
```
alexandrzavalii, line 16: will this work?
```
const AccessibleHide = styled('span')(accessibleHide);
```

modules/_labs/combobox/react/lib/AutocompleteList.tsx
```diff
@@ -0,0 +1,133 @@
...
+          onClick: (event: React.MouseEvent) => {
+            event.preventDefault();
+            handleAutocompleteClick(event, listboxItem.props);
+          },
+        })}
+      </Fragment>
+    );
+  };
+
+  if (autocompleteItems.length && 'header' in autocompleteItems[0]) {
```
alexandrzavalii, line 94: maybe `hasOwnProperty` instead? or `autocompleteItems[0]['header']` ?

modules/_labs/combobox/react/lib/AutocompleteList.tsx
```diff
@@ -0,0 +1,133 @@
...
+  } else if (autocompleteItems.length) {
+    return (
+      <Autocomplete {...listBoxProps}>
+        {(autocompleteItems as React.ReactElement<MenuItemProps>[]).map(
+          (listboxItem: React.ReactElement, index: number) => createListItem(listboxItem, index)
+        )}
+      </Autocomplete>
+    );
+  } else {
+    return null;
```
alexandrzavalii, line 129: maybe do this in the top?
```
   if (!autocompleteItems.length) return null;
```

modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -209,10 +245,9 @@ const Combobox = ({
   }, []);
 
   useEffect(() => {
-    if (autocompleteItems) {
-      setShowingAutocomplete(autocompleteItems.length > 0 && isFocused);
-    }
-  }, [autocompleteItems, isFocused, value]);
+    const items = getInteractiveAutocompleteItems();
+    setInteractiveAutocompleteItems(items);
+  }, [autocompleteItems]);
```
alexandrzavalii, line 250: maybe getInteractiveAutocompleteItems should be moved inside a hook, since it is used only here

modules/_labs/combobox/react/lib/Status.tsx
```diff
@@ -0,0 +1,50 @@
...
+  useEffect(() => {
+    if (!announcementText) {
+      return;
+    }
+    setDisplayText(announcementText);
+    const clearText = setTimeout(() => setDisplayText(''), expireMilliseconds);
+
+    return () => {
+      clearTimeout(clearText);
+    };
```
alexandrzavalii, line 36: could be
`return () => clearTimeout(clearText)`
clearTimeout returns void

modules/_labs/combobox/react/stories/stories.tsx
```diff
@@ -1,59 +1,79 @@
...
-> {
-  state = {
-    currentText: '',
-  };
+const autocompleteResult = (textModifier: number): ReactElement<MenuItemProps> => (
+  <MenuItem onClick={action(`Clicked Result ${textModifier}`)}>
+    Result{' '}
+    <span>
+      num<span>ber</span>
+    </span>{' '}
```
alexandrzavalii, line 19: some extra spaces
```suggestion
    </span>
```

modules/_labs/combobox/react/lib/Combobox.tsx
vibdev: Arn't hooks more for reuse? I am not sure the value of creating one for a one off, maybe I am missing something though.

modules/_labs/combobox/react/lib/Status.tsx
vibdev: Same as above, I kinda like making it easy to add things to the return in the future. Happy to change if folks feel strongly.

modules/_labs/combobox/react/lib/Combobox.tsx
alexandrzavalii: hmm, I guess my suggestion is coming from exhaustive-depth rule.
https://github.com/facebook/react/issues/14920

modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -267,11 +306,28 @@ const Combobox = ({
     focusInput();
   };
 
+  const getGroupIndex = (itemIndex: number | null) => {
+    if (
+      itemIndex != null &&
+      autocompleteItems &&
+      autocompleteItems.length &&
+      'header' in autocompleteItems[0]
```
alexandrzavalii, line 314: lets change this one too then? `hasOwnProperty`

modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -267,11 +306,28 @@ const Combobox = ({
     focusInput();
   };
 
+  const getGroupIndex = (itemIndex: number | null) => {
+    if (
+      itemIndex != null &&
+      autocompleteItems &&
+      autocompleteItems.length &&
```
alexandrzavalii, line 313: maybe to shorten if statement? `autocompleteItems?.length`

modules/_labs/combobox/react/lib/AutocompleteList.tsx
alexandrzavalii: just my personal opinion,I dont think its difficult to add three dots and comma when we need it. 
What if we will never need it? we will be just carrying this extra line and a unnecessary copy of the object, for indefinite time.

modules/_labs/combobox/react/lib/Status.tsx
alexandrzavalii: its fine! :) I am just afraid of the word "future" :D No one can possible predict it

modules/_labs/combobox/react/lib/Combobox.tsx
vibdev: oh shoot I thought I got them all, will do

modules/_labs/combobox/react/lib/Combobox.tsx
vibdev: Last I checked null coalescing and optional chaining wasn't available in CKR, will double check again.

anicholls: It will be in 4.0! Not in master yet

modules/_labs/combobox/react/lib/Combobox.tsx
vibdev: Oh interesting maybe we should add this to the codebase? What do you think @NicholasBoll

modules/_labs/combobox/react/lib/AutocompleteList.tsx
vibdev: I would argue that lines are free, and in this case the spread cost is a bit of a micro optimization. 
I do think there is probably some places this module can benefit from performance improvements but not sure if this would move the needle.

.eslintrc.js
```diff
@@ -81,6 +81,7 @@ module.exports = {
     'quote-props': 'off',
     'space-before-function-paren': 'off',
     'react/jsx-no-bind': 'off', // Keep perf implications in mind, but was giving too many warnings and hurting readability
+    "react-hooks/exhaustive-deps": 'warn',
```
vibdev, line 84: @anicholls @NicholasBoll adding this only adds one warning for current code
```
canvas-kit-react/modules/modal/react/lib/ModalContent.tsx
  139:6  warning  React Hook React.useLayoutEffect has a missing dependency: 'handlerRef'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```
If y'all think we should add to the codebase I can autofix the warning

modules/_labs/combobox/react/lib/Combobox.tsx
NicholasBoll: I think we should add hooks lint rules. I see we added the plugin, but didn't enable any rules

vibdev: @NicholasBoll added both hook linting rules and made the fixes.

modules/_labs/core/react/lib/theming/useTheme.ts
```diff
@@ -17,12 +17,13 @@ import {defaultCanvasTheme} from './theme';
  * Tracked on https://github.com/emotion-js/emotion/issues/1193.
  */
 export function useTheme(theme?: Object): CanvasTheme {
+  const context = React.useContext(ThemeContext);
```
NicholasBoll, line 20: This will have to be an exception, at least how `useTheme` is written because it can be used both inside and outside a Hook context. Ideally this behavior is a different function for that use-case, but this change will break functionality.

NicholasBoll: Perhaps @anicholls can add more insight

vibdev: updated

modules/_labs/core/react/lib/theming/useTheme.ts
```diff
@@ -17,13 +17,18 @@ import {defaultCanvasTheme} from './theme';
...
     return theme as CanvasTheme;
   }
 
   try {
+    // TODO: useContext should not be called conditionally,
+    // but to keep this function working we need to ignore till a refactor can happen.
+    // Ideally there is a getTheme and a useTheme separately where getTheme is used outside of functional components
+    // and useTheme is used inside functional components. Then this try won’t be necessary.
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const context = React.useContext(ThemeContext);
```
NicholasBoll: Thanks for adding this

NicholasBoll:
Paired with @vibdev to add Cypress tests, fix various bugs found while writing tests and removed Jest tests that are now covered with Cypress

cypress.json
```diff
@@ -1,5 +1,6 @@
 {
   "projectId": "odida5",
   "baseUrl": "http://localhost:9001",
-  "supportFile": "cypress/support/index.ts"
+  "supportFile": "cypress/support/index.ts",
+  "watchForFileChanges": false
```
NicholasBoll, line 5: I added this. I think it's annoying to write a bit of test code, then go back to Cypress to find out what I'm querying the DOM for and the Cypress test starts over again increasing time to write tests rather than decreasing it.

PR #545, opened 2020-04-01
refactor(popper): Update to use react portal as default

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/517

## Checklist

- [ ] `yarn test` passes
      applicable


modules/common/react/lib/Popper.tsx
```diff
@@ -15,9 +15,16 @@ export interface PopperProps extends React.HTMLAttributes<HTMLDivElement> {
...
+   * The containing element for Popper elements. The Popper uses
+   * {@link https://reactjs.org/docs/portals.html Portals} to place the DOM elements
+   * of the Popper in a different place in the DOM to prevent issues with overflowed containers.
+   * When the popper is opened, `aria-hidden` will be added to siblings to hide background
+   * content from assistive technology like it is visibly hidden from sighted users. This property
+   * should be set to the element that the application root goes - not containing element of content.
+   * This should be a sibling or higher than the header and navigation elements of the application.
+   * @default document.body
    */
-  containerElement?: Element;
```
mannycarrera4, line 20: probably keep this as element?

anicholls:
@mannycarrera4 Can you make sure you target `prerelease/v4`? Some of these updates (i.e. Popper) have already been made there

modules/common/react/README.md
```diff
@@ -74,7 +74,13 @@ includes custom `data-*` attributes such as `data-test-id` to help facilitate au
 
 #### `containerElement: Element`
 
-> The element that contains the portal children when `portal` is true.
+> The containing element for Popper elements. The Popper uses {@link
```
anicholls, line 77: Not valid link formatting for md

modules/common/react/README.md
```diff
@@ -74,7 +74,13 @@ includes custom `data-*` attributes such as `data-test-id` to help facilitate au
...
+> https://reactjs.org/docs/portals.html Portals} to place the DOM elements of the Popper in a
+> different place in the DOM to prevent issues with overflowed containers. When the popper is
```
anicholls, line 79: Nit: "in a different place in the dom" > "As the last child of the specified container element"

modules/common/react/README.md
```diff
@@ -74,7 +74,13 @@ includes custom `data-*` attributes such as `data-test-id` to help facilitate au
...
+> opened, `aria-hidden` will be added to siblings to hide background content from assistive
```
anicholls, line 80: I know we do this manually in modal - have we confirmed that it popper does this for us? Are we sure this is what we want for all popup use cases? If a tooltip pops up, I wouldn't call all of the rest of the content on the page hidden

modules/common/react/lib/Popper.tsx
```diff
@@ -15,9 +15,16 @@ export interface PopperProps extends React.HTMLAttributes<HTMLDivElement> {
    */
   children: React.ReactNode;
   /**
-   * The element that contains the portal children when `portal` is true.
+   * The containing element for Popper elements. The Popper uses
+   * {@link https://reactjs.org/docs/portals.html Portals} to place the DOM elements
+   * of the Popper in a different place in the DOM to prevent issues with overflowed containers.
```
anicholls, line 20: Note to make the same copy updates here

modules/common/react/lib/Popper.tsx
```diff
@@ -60,10 +61,6 @@ export class Popper extends React.PureComponent<PopperProps> {
       return null;
     }
 
-    if (!this.props.portal) {
-      return this.renderPopper();
-    }
-
     // do not use defaultProps for containerElement because document may not be statically available
     // at require time in some testing environments; instead we safely default at runtime
     return ReactDOM.createPortal(this.renderPopper(), this.props.containerElement || document.body);
```
anicholls, line 66: Can we combine `render` and `renderPopper` now? The only reason they were separated was due to the portal conditional

anicholls: Nit: Can we change `document.body` to a default prop?

modules/common/react/lib/Popper.tsx
anicholls: `export function createPortal(children: ReactNode, container: Element, key?: null | string): ReactPortal;`
^ Types from React.createPortal, so yeah - let's keep as `Element`

NicholasBoll:
I read the PR description and associated issue. We are already defaulting to using portals. This PR removes the option to not use a Portal. Is that what we wanted? There may be cases where not portalling content is desired.

mannycarrera4:
@NicholasBoll from what I understood, we were just using portals, but this brings up the question of if we should keep the portaling option if they don't want to portal?

NicholasBoll:
> @NicholasBoll from what I understood, we were just using portals, but this brings up the question of if we should keep the portaling option if they don't want to portal?

I'm not sure what that means. By default we use portals with the option to disable if necessary (in certain cases this can make focus management _much_ easier).

This API change still allows the original functionality of disabling the separation of child content from the parent (what portalling essentially does) by allowing a developer to set the `containerElement` to be the parent element, but this is much harder to do now:

```tsx
// before
<Popper portal={false}>

// after
<Portal containerElement={targetRef.current?.parentElement}
```

Full example here:
```tsx
// before
const BeforeComponent = () => {
  const targetRef = React.useRef<HTMLDivElement>(null)
  const [isOpen, setOpen] = React.useState(false)

  return (
    <>
      <div ref={targetRef} onClick={() => setOpen(true)}>Open</div>
      <Popper open={isOpen} anchorElement={targetRef.current} portal={false}>
        Popper content
      </Popper>
    </>
  )
}

// after
const AfterComponent = () => {
  const targetRef = React.useRef<HTMLDivElement>(null)
  const [isOpen, setOpen] = React.useState(false)

  return (
    <>
      <div ref={targetRef} onClick={() => setOpen(true)}>Open</div>
      <Popper open={isOpen} anchorElement={targetRef.current} containerElement={targetRef.current?.parentElement}>
        Popper content
      </Popper>
    </>
  )
}
```

The `after` isn't 100% accurate because we can no longer control the position in the DOM if we don't want to portal to a separate container. Portals will always be a the last child of a container.

NicholasBoll:
@anicholls Is this an okay thing? No way to disable the use of Portal on a `Popper` component? There are valid reasons to not use it, but still have the niceties of using a position library.

mannycarrera4:
@NicholasBoll I might have misinterpreted the issue, i think after your explanation on the example, i think we should just default to using portal instead of ripping it out

mannycarrera4:
@anicholls @NicholasBoll I feel like we should just close this PR, most of the changes where in popper and they're being addressed in https://github.com/Workday/canvas-kit/pull/528 thoughts? I only really just updated some readmes

anicholls:
@mannycarrera4 I'm fine with closing this in favor of #528. 

@NicholasBoll Can you make sure #517 gets closed as part of #528 as well? Squashing 10 tickets with one PR 😂

Issue #543, opened 2020-03-31
feat: Allow InputProvider to use a configurable container.


anicholls:
Our `InputProvider` does not currently work with React Portals (since the popups get placed outside of the `InputProvider` container div. Now that `InputProviders` behave as a singleton, we should be save to track the user input on the body element instead.

Let's adjust `InputProvider` to use `document.body` with a configurable container.

anicholls:
Closed by https://github.com/Workday/canvas-kit/pull/546

PR #546, opened 2020-04-01
feat(core): Allow InputProvider to use a configurable container

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/543

### DIFF
--- a/modules/core/react/README.md
+++ b/modules/core/react/README.md
@@ -418,8 +418,9 @@ will remove itself from the DOM (rendering only its children) and not attach any
 ##### `container: HTMLElement | React.RefObject<HTMLElement>`
 
 > The containing element in which the InputProvider is rendered into. This property should be set to
-> the element that the application root goes - not containing element of content. Note: If the
-> element is null, it will default to `document.body`.
+> the element that the application root goes - not containing element of content.
+
+Default: `document.body` 
 
 ### Storybook Decorator
 
--- a/modules/core/react/README.md
+++ b/modules/core/react/README.md
@@ -410,6 +410,11 @@ will remove itself from the DOM (rendering only its children) and not attach any
 > Note: detecting intent will add scroll and mouse positioning listeners which could affect
 > performance.
 
+##### `container: HTMLElement`
+
+> The containing element in which the InputProvider is rendered into. This property should be set to
+> the element that the application root goes - not containing element of content. 
+
 ### Storybook Decorator
 
 We provide a [storybook decorator](../../utils/storybook/CanvasProviderDecorator.tsx) to wrap your
--- a/modules/core/react/README.md
+++ b/modules/core/react/README.md
@@ -415,10 +415,11 @@ will remove itself from the DOM (rendering only its children) and not attach any
 > Note: detecting intent will add scroll and mouse positioning listeners which could affect
 > performance.
 
-##### `container: HTMLElement`
+##### `container: HTMLElement | React.RefObject<HTMLElement>`
 
 > The containing element in which the InputProvider is rendered into. This property should be set to
-> the element that the application root goes - not containing element of content.
+> the element that the application root goes - not containing element of content. Note: If the
+> element is null, it will default to `document.body`.
 
 ### Storybook Decorator
 
--- a/modules/core/react/README.md
+++ b/modules/core/react/README.md
@@ -344,7 +344,7 @@ import {InputProvider} from '@workday/canvas-kit-react';
 <div>;
 ```
 
-This will result in these event listeners being added to the body as default
+This will result in these data attributes being added to the body element (by default)
 
 ```html
 <body data-whatinput="mouse" data-whatinput="mouse">
--- a/modules/core/react/README.md
+++ b/modules/core/react/README.md
@@ -417,8 +417,8 @@ will remove itself from the DOM (rendering only its children) and not attach any
 
 ##### `container: HTMLElement | React.RefObject<HTMLElement>`
 
-> The containing element in which the InputProvider is rendered into. This property should be set to
-> the element that the application root goes - not containing element of content.
+> The containing element in which the InputProvider attaches its data attributes. This property should
+> be set to an element that is an ancestor of all your Canvas components.
 
 Default: `document.body` 
 
--- a/modules/core/react/README.md
+++ b/modules/core/react/README.md
@@ -338,13 +338,18 @@ directly within any Canvas Kit components.
 import * as React from 'react';
 import {InputProvider} from '@workday/canvas-kit-react';
 
-<InputProvider>{/* All your components containing any Canvas components */}</InputProvider>;
+<div>
+  <InputProvider />
+  {/* All your components containing any Canvas components */}
+<div>;
 ```
 
-This will result in a wrapping div with two data attributes:
+This will result in these event listeners being added to the body as default
 
 ```html
-<div data-whatinput="mouse" data-whatinput="mouse"><!-- All your components' HTML --></div>
+<body data-whatinput="mouse" data-whatinput="mouse">
+  <!-- All your components' HTML -->
+</body>
 ```
 
 Now in any component within this wrapper, you can use these data attributes to customize the focus
@@ -413,7 +418,7 @@ will remove itself from the DOM (rendering only its children) and not attach any
 ##### `container: HTMLElement`
 
 > The containing element in which the InputProvider is rendered into. This property should be set to
-> the element that the application root goes - not containing element of content. 
+> the element that the application root goes - not containing element of content.
 
 ### Storybook Decorator
 
--- a/modules/core/react/lib/InputProvider.tsx
+++ b/modules/core/react/lib/InputProvider.tsx
@@ -110,10 +110,6 @@ export default class InputProvider extends React.Component<InputProviderProps> {
   private mousePosY: number | null = null; // Unused if props.provideIntent is not defined
   private nested = false; // True if nested within another input provider
 
-  // static defaultProps = {
-  //   container: document.body,
-  // };
-
   constructor(props: any) {
     super(props);
 
@@ -163,13 +159,11 @@ export default class InputProvider extends React.Component<InputProviderProps> {
   provideIntent = this.props.provideIntent;
 
   getContainer(container: HTMLElement | React.RefObject<HTMLElement> | undefined): HTMLElement {
-    console.warn('container', container);
     if (container === undefined) {
-      console.warn(container);
       return document.body;
     } else if ('current' in container) {
       if (container.current === null) {
-        console.warn('You ref object can not be null, therefore, falling back to document.body');
+        console.warn('Your ref object can not be null, therefore, falling back to document.body');
         return document.body;
       } else {
         return container.current;
--- a/modules/core/react/lib/InputProvider.tsx
+++ b/modules/core/react/lib/InputProvider.tsx
@@ -108,7 +108,7 @@ export default class InputProvider extends React.Component<InputProviderProps> {
   private isScrolling = false; // Unused if props.provideIntent is not defined
   private mousePosX: number | null = null; // Unused if props.provideIntent is not defined
   private mousePosY: number | null = null; // Unused if props.provideIntent is not defined
-  private nested = false; // True if nested within another input provider
+  private isAnotherInputProvider = false; // True if there is another input provider
 
   constructor(props: any) {
     super(props);
@@ -179,7 +179,7 @@ export default class InputProvider extends React.Component<InputProviderProps> {
     elementClosestPolyfill(window);
 
     if (this.container.closest('[data-whatinput]')) {
-      this.nested = true;
+      this.isAnotherInputProvider = true;
       return;
     }
     this.container.setAttribute('data-whatinput', this.currentInput);
@@ -204,7 +204,7 @@ export default class InputProvider extends React.Component<InputProviderProps> {
   }
 
   componentWillUnmount() {
-    if (this.nested) {
+    if (this.isAnotherInputProvider) {
       return;
     }
 
--- a/modules/core/react/lib/InputProvider.tsx
+++ b/modules/core/react/lib/InputProvider.tsx
@@ -4,18 +4,7 @@ import elementClosestPolyfill from 'element-closest';
 
 export interface InputProviderProps {
   provideIntent?: boolean;
-  container?: HTMLElement;
-}
-
-export interface InputProviderState {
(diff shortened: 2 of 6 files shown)


modules/core/react/lib/InputProvider.tsx
```diff
@@ -114,6 +115,10 @@ export default class InputProvider extends React.Component<InputProviderProps, I
   private eventTimer: number | undefined;
   private ref: React.RefObject<HTMLDivElement> = React.createRef();
 
+  static defaultProps = {
+    container: document.body,
+  };
+
```
mannycarrera4, line 121: remove

utils/storybook/InputProviderDecorator.tsx
```diff
@@ -1,4 +1,6 @@
 import * as React from 'react';
 import {InputProvider} from '@workday/canvas-kit-react-core';
 
-export default (storyFn: () => React.ReactNode) => <InputProvider>{storyFn()}</InputProvider>;
+export default (storyFn: () => React.ReactNode) => (
+  <InputProvider container={document.body}>{storyFn()}</InputProvider>
```
mannycarrera4, line 5: @anicholls do we want the default now to be `document.body` ?

modules/core/react/lib/InputProvider.tsx
```diff
@@ -345,12 +370,16 @@ export default class InputProvider extends React.Component<InputProviderProps, I
...
 
     const intent = this.provideIntent ? this.state.currentIntent : null;
-    return (
+
+    return container ? (
+      container
+    ) : (
       <div
         className={containerClass}
         ref={this.ref}
```
anicholls, line 385: Now that we always have a reference to the container, we no longer need this div or ref.

modules/core/react/lib/InputProvider.tsx
```diff
@@ -4,6 +4,7 @@ import elementClosestPolyfill from 'element-closest';
 
 export interface InputProviderProps {
   provideIntent?: boolean;
+  container?: HTMLElement;
```
anicholls, line 7: This will have to accept refs as well, otherwise we ask react apps to use `findDOMNode` (which is being deprecated)

modules/core/react/lib/InputProvider.tsx
```diff
@@ -170,9 +175,17 @@ export default class InputProvider extends React.Component<InputProviderProps, I
   provideIntent = this.props.provideIntent;
 
   componentDidMount() {
+    const {container} = this.props;
+    if (container) {
+      // bail early if input provider is already handling this
+      if (container.classList.contains(containerClass)) {
```
anicholls, line 181: We can just change `this.ref.current.parentElement.closest(`.${containerClass}`)` to `container.closest('[data-whatinput]')` for the same effect (`closest` checks the element itself and all ancestors).

modules/core/react/lib/InputProvider.tsx
```diff
@@ -170,9 +175,17 @@ export default class InputProvider extends React.Component<InputProviderProps, I
...
+        return;
+      }
+      container.classList.add(containerClass);
```
anicholls, line 184: This was to provide access to the div we were rendering. Since the user controls that now, we don't need to add a classname anymore. If we need to check whether an element has an input provider attached, we can check for the `data-whatinput` attr

modules/core/react/lib/InputProvider.tsx
anicholls, line 385: Now that we always have a reference to the container, we no longer need this div or ref.

modules/core/react/lib/InputProvider.tsx
mannycarrera4: wasn't sure yet if we wanted this to be a default

modules/core/react/lib/InputProvider.tsx
```diff
@@ -345,12 +370,16 @@ export default class InputProvider extends React.Component<InputProviderProps, I
...
+    const {container} = this.props;
+    if (this.state.nested || container) {
       return this.props.children;
     }
 
     const intent = this.provideIntent ? this.state.currentIntent : null;
-    return (
+
+    return container ? (
+      container
```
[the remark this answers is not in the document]
NicholasBoll: I think the confusion is this ternary. There is an early escape a few lines up:
```ts
    if (this.state.nested || container) {
      return this.props.children;
    }
```

modules/core/react/lib/InputProvider.tsx
NicholasBoll: A ref doesn't require `findDOMNode`, it will have access to the DOM node in its `.current` property.

https://reactjs.org/docs/hooks-reference.html#useref

We could add support for a `ref` as a convenience for React apps:

```tsx
const App = () => {
  const ref = React.useRef()
  return (
    <div ref={ref}>
      <InputProvider container={ref} />
      {/* other children */}
    </div>
  )
}
```

We could update the signature to accept `HTMLElement | React.RefObject<HTMLElement>` and do detection inside `InputProvider`

modules/core/react/spec/InputProvider.spec.tsx
```diff
@@ -208,7 +208,8 @@ describe('InputProvider', () => {
         </InputProvider>
       </InputProvider>
     );
-    expect(component.find('.wdc-input-provider').length).toBe(1);
```
NicholasBoll, line 211: Was this class name used for anything other than nested detection?

mannycarrera4: i don't think so!

modules/core/react/lib/InputProvider.tsx
anicholls: I would expect it to be, yes 👍

modules/core/react/lib/InputProvider.tsx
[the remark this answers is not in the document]
anicholls: Ah - thanks, I missed that.

I think we always just want to return `children`, don't we? We can do the attribute handling in the other lifecycle events.

modules/core/react/lib/InputProvider.tsx
anicholls: You are right, sorry. I guess I was wanting to avoid them having to do `container={ref.current}`

modules/core/react/README.md
```diff
@@ -410,6 +415,12 @@ will remove itself from the DOM (rendering only its children) and not attach any
 > Note: detecting intent will add scroll and mouse positioning listeners which could affect
 > performance.
 
+##### `container: HTMLElement | React.RefObject<HTMLElement>`
+
+> The containing element in which the InputProvider is rendered into. This property should be set to
+> the element that the application root goes - not containing element of content. Note: If the
+> element is null, it will default to `document.body`.
```
anicholls, line 422:
```suggestion
> the element that the application root goes - not containing element of content.

Default: `document.body` 
```

modules/core/react/lib/InputProvider.tsx
```diff
@@ -150,16 +146,9 @@ export default class InputProvider extends React.Component<InputProviderProps, I
...
-      currentInput: storedInput || InputType.Initial,
-      currentIntent: storedIntent || InputType.Initial,
-      supportsPassive: supportsPassive || false,
-      isBuffering: false,
-      isScrolling: false,
-      mousePosX: null,
-      mousePosY: null,
-      nested: false,
-    };
+    this.currentInput = storedInput || InputType.Initial;
```
anicholls, line 149: What was the reason for changing things from state to class variables? Was it to avoid the need for `shouldComponentUpdate`?

modules/core/react/lib/InputProvider.tsx
```diff
@@ -169,48 +158,63 @@ export default class InputProvider extends React.Component<InputProviderProps, I
...
       console.warn('Failed to set input status in session storage' + e);
     }
   }
 
   componentWillUnmount() {
+    if (this.nested) {
+      return;
+    }
+
+    this.container.removeAttribute('data-whatinput');
```
anicholls, line 213: What about `data-whatintent`?

modules/core/react/lib/InputProvider.tsx
```diff
@@ -169,48 +158,63 @@ export default class InputProvider extends React.Component<InputProviderProps, I
...
-      this.ref.current &&
-      this.ref.current.parentElement &&
-      this.ref.current.parentElement.closest(`.${containerClass}`)
-    ) {
-      this.setState({nested: true});
+    if (this.container.parentElement && this.container.parentElement.closest('[data-whatinput]')) {
+      this.nested = true;
       return;
     }
+    this.container.setAttribute('data-whatinput', this.currentInput);
```
anicholls, line 188: What about `data-whatintent`?

modules/core/react/lib/InputProvider.tsx
```diff
@@ -169,48 +158,63 @@ export default class InputProvider extends React.Component<InputProviderProps, I
...
     // For IE11 and under, we'll need to polyfill element.closest
     elementClosestPolyfill(window);
-
-    if (
-      this.ref.current &&
-      this.ref.current.parentElement &&
-      this.ref.current.parentElement.closest(`.${containerClass}`)
-    ) {
-      this.setState({nested: true});
+    if (this.container.parentElement && this.container.parentElement.closest('[data-whatinput]')) {
```
anicholls, line 184: If you change this to `this.container.closest('[data-whatinput]'))`, you can remove the if statement above.

This will set `nested` to `true` if the container is already "decorated", but that's okay. The name `nested` doesn't really cover all cases anymore anyway. The intent of this variable was to track whether there was another InputProvider doing its thing. Since it's now possible for this to be true regardless of whether they are nested or not, we should probably rename this variable.

modules/core/react/lib/InputProvider.tsx
```diff
@@ -169,48 +158,63 @@ export default class InputProvider extends React.Component<InputProviderProps, I
   // Need to remember how it component was mounted so we ensure listener removal
   provideIntent = this.props.provideIntent;
 
+  getContainer(container: HTMLElement | React.RefObject<HTMLElement> | undefined): HTMLElement {
+    if (container === undefined) {
+      return document.body;
```
anicholls, line 163: Why not just use a default prop?

modules/core/react/spec/InputProvider.spec.tsx
```diff
@@ -208,7 +208,8 @@ describe('InputProvider', () => {
...
+
+    expect(document.querySelectorAll('[data-whatinput]').length).toBe(1);
```
anicholls, line 212: I feel like we need a better way of testing that these are behaving as a singleton. Now that they both point to `document.body`, even if they are both setting the attribute, there will only be one match to this query.

modules/core/react/lib/InputProvider.tsx
mannycarrera4: i think the other reason is we didn't want to force a rerender every time state changed, so having class variables made more sense.

modules/core/react/lib/InputProvider.tsx
```diff
@@ -105,14 +95,20 @@ const detectWheel = () => {
...
-  private ref: React.RefObject<HTMLDivElement> = React.createRef();
+
+  private currentInput: InputType;
+  private currentIntent: InputType;
+  private supportsPassive: boolean;
+  private isBuffering = false;
+  private isScrolling = false; // Unused if props.provideIntent is not defined
+  private mousePosX: number | null = null; // Unused if props.provideIntent is not defined
+  private mousePosY: number | null = null; // Unused if props.provideIntent is not defined
+  private isAnotherInputProvider = false; // True if there is another input provider
```
mannycarrera4, line 111: not super sure about this variable name

modules/core/react/README.md
```diff
@@ -338,13 +338,18 @@ directly within any Canvas Kit components.
...
 
-<InputProvider>{/* All your components containing any Canvas components */}</InputProvider>;
+<div>
+  <InputProvider />
+  {/* All your components containing any Canvas components */}
+<div>;
 ```
 
-This will result in a wrapping div with two data attributes:
+This will result in these event listeners being added to the body as default
```
anicholls, line 347:
```suggestion
This will result in these data attributes being added to the body element (by default)
```

modules/core/react/README.md
```diff
@@ -410,6 +415,13 @@ will remove itself from the DOM (rendering only its children) and not attach any
 > Note: detecting intent will add scroll and mouse positioning listeners which could affect
 > performance.
 
+##### `container: HTMLElement | React.RefObject<HTMLElement>`
+
+> The containing element in which the InputProvider is rendered into. This property should be set to
+> the element that the application root goes - not containing element of content.
```
anicholls, line 421:
```suggestion
> The containing element in which the InputProvider attaches its data attributes. This property should
> be set to an element that is an ancestor of all your Canvas components.
```

modules/core/react/lib/InputProvider.tsx
```diff
@@ -150,67 +146,76 @@ export default class InputProvider extends React.Component<InputProviderProps, I
...
-      this.ref.current.parentElement &&
-      this.ref.current.parentElement.closest(`.${containerClass}`)
-    ) {
-      this.setState({nested: true});
+    if (this.container.closest('[data-whatinput]')) {
+      this.isAnotherInputProvider = true;
       return;
     }
+    this.container.setAttribute('data-whatinput', this.currentInput);
+    this.container.setAttribute('data-whatintent', this.currentIntent);
```
anicholls, line 188: We should only add intent if `this.provideIntent`. Might as well just call `updateAttributes()` here
```suggestion
    this.updateAttributes()
```

modules/core/react/spec/InputProvider.spec.tsx
```diff
@@ -200,15 +200,18 @@ describe('InputProvider', () => {
...
+  test(`Multiple input provider should only attach once to the dom`, () => {
+    const ref = React.createRef<HTMLButtonElement>();
     const component = mount(
-      <InputProvider>
-        <InputProvider>
-          <h1>Test</h1>
-        </InputProvider>
-      </InputProvider>
+      <div>
+        <InputProvider></InputProvider>
```
anicholls, line 207: I'm surprised something didn't complain about not using `<InputProvider />` syntax here and below. No worries though.

modules/core/react/lib/InputProvider.tsx
anicholls: Can we call this `deferInputTracking`?

Issue #526, opened 2020-03-19
Rename label props to the appropriate aria label


mannycarrera4:
## 🐛 Bug Report

We ofter have props such as `somePropNameLabel` which is used for aria attributes. For a more clear prop name, we should rename these to mimic the aria attribute.
`submitButtonLabel` -> `submitButtonAriaLabel`

anicholls:
Closed by https://github.com/Workday/canvas-kit/pull/551

PR #551, opened 2020-04-07
Fix: Update prop names to match aria labels

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/526

### DrawerHeader

The following props where renamed for appropriate aria naming and clarity •

- ``closeIconLabel`->`closeIconAriaLabel`

---

### SearchBar

The following props where renamed for appropriate aria naming and clarity

- `submitLabel` -> `submitAriaLabel`
- `openButtonLabel` -> `openButtonAriaLabel`
- `closeButtonLabel` -> `closeButtonAriaLabel`

---

### Menu

The following props where renamed for appropriate aria naming and clarity

- `labeledBy` -> `'aria-labelledby'`

---

### Popup

The following props where renamed for appropriate aria naming and clarity

- `closeLabel` -> `closeButtonAriaLabel`

The property `title` was removed since we already had an `aria-label` attached to the close button

---

### SidePanel

The following props where renamed for appropriate aria naming and clarity

- `closeNavigationLabel` -> `closeNavigationAriaLabel`
- `openNavigationLabel` -> `openNavigationAriaLabel`

---

### Skeleton

The following props where renamed for appropriate aria naming and clarity

- `loadingLabel` -> `'aria-label`

---


modules/banner/react/lib/Banner.tsx
```diff
@@ -30,7 +30,7 @@ export interface BannerProps extends Themeable, React.ButtonHTMLAttributes<HTMLB
    */
   error?: ErrorType;
   /**
-   * The text of the Banner action.
+   * The text of the Banner action. This Text is also used as the aria-label for the action in the banner
```
mannycarrera4, line 33: rename to prop

modules/popup/react/lib/Popup.tsx
```diff
@@ -61,7 +61,7 @@ export interface PopupProps extends React.HTMLAttributes<HTMLDivElement> {
    * The `aria-label` for the Popup close button.
    * @default Close
    */
-  closeLabel?: string;
+  closeButtonAriaLabel?: string;
```
mannycarrera4, line 64: note this is also used for the title

mannycarrera4:
So some of my findings:
• Sometimes we have a prop for aria labels
• Sometimes those props are used for other properties such as titles
• Sometimes we bake in the aria labels and they're not props
• In icon button,  we have an aria label prop, but it's overridden by the aria-label attribute, look at story for example vs the component

modules/_labs/drawer/react/lib/Drawer.tsx
```diff
@@ -30,11 +30,11 @@ export interface DrawerProps extends React.HTMLAttributes<HTMLDivElement> {
   /**
    * The `aria-labelledby` of the Drawer. Set this when there is a `header` for accessibility. The `role` attribute should also be used when this attribute is present. This value should be the same as the `id` attribute of the `header` element.
    */
-  'aria-labelledby'?: string;
+  drawerArialabelledBy?: string;
```
anicholls, line 33: I'm confused by this change. Generally we avoid using the component name in its props. This was just exposing the already existing `aria-labelledby` prop extended from `React.HTMLAttributes<HTMLDivElement>` so that we could document what it's used for

modules/_labs/menu/react/lib/Menu.tsx
```diff
@@ -39,7 +39,7 @@ export interface MenuProps extends GrowthBehavior, React.HTMLAttributes<HTMLULis
   /**
    * The HTML `id` of the element that labels the Menu. Often used with menu buttons.
    */
-  labeledBy?: string;
+  menuArialabeledBy?: string;
```
anicholls, line 42: Do we need the component name included here? We could also consider falling back to the default `aria-labeledby` prop here as well. 

#526 only covers one use case, yet you've altered several other cases here. I feel like there needs to be more discussion on the direction and overall strategy here before we get to reviewing this much further. Could you please formulate and document your own opinion/reasoning on #526 for each use case before we start that discussion so we have a jumping off point?

mannycarrera4:
Prop used for aria attributes and other props
We have a few components that use a prop for aria attributes and other properties such as title. 

Question
Should we add a different prop solely for the aria attribute or just document that is used for multiple reasons.

Example:
In `Popup.tsx` we have a `closeLabe`l prop that is used for the aria label AND the `title` attribute.

The same goes for `ColorPicker.tsx`. The prop `customHexInputLabe`l is used for the label and the aria label for for the check button. 

Exposing Aria attributes for customization

In `Drawer.tsx` and `Menu.tsx` we expose existing aria attributes to allow for customization. Do we always expose these in the string literal format such as `aria-labeledBy` or camel cased?

anicholls:
@mannycarrera4 Thanks for pulling together your proposal. 

I'm a bit confused by 
> If we expose an aria attribute for flexibility, we specifically say that it's for aria purposes, closeLabel -> closeAriaLabel

vs. 

>  If the props is exposed in the API but not explicitly used for anything other than aria-label, we leave as is in string literal format

Are these not two different approaches to the same case?

mannycarrera4:
@anicholls I was looking at two different use cases and was a little confused, mostly around the example of `Drawer.tsx` vs `Popup.tsx` in popup we add it to the api, but it is passed through on the attribute `aria-label` and in Drawer we don't pass that prop directly into the corresponding aria attribute, which is a little confusing but allows the consumer the flexibility

modules/_labs/drawer/react/lib/DrawerHeader.tsx
```diff
@@ -74,7 +74,7 @@ const CloseButton = styled(IconButton)({
 export default class DrawerHeader extends React.Component<DrawerHeaderProps, {}> {
   public render() {
     const {
-      closeIconLabel = 'Close',
+      'aria-label': closeIconAriaLabel = 'Close',
```
mannycarrera4, line 77: thoughts on this approach? not my favorite, but we week the attribute the same name, and just add the type as `closeIconAriaLabel` for clarification?

NicholasBoll: Hmm. So props spread on the `HeaderContainer` element except for `title`, `id`, and now `aria-label`. I'm not sure it is obvious that `aria-label` means "`aria-label` of the close icon".

I think `closeIconAriaLabel` is more descriptive of purpose than `aria-label`.

NicholasBoll: I think if the element receiving the aria label is the spread element, I think it is better for us to simply shadow the prop so that our doc system calls it out.

Example:
```tsx
// Make sense to keep `aria-label`
const MyComponent = ({
  'aria-label': ariaLabel = 'Some Default',
  ...elemProps,
}) => {
  return (
    <div aria-label={ariaLabel} {...elemProps}>Something</div>
  )
}

// Maybe it doesn't make sense
const MyComponent = ({
  'aria-label': someAriaLabel = 'Close',
  ...elemProps,
}) => {
  return (
    <div {...elemProps}>
      <SomeOtherComponent aria-label={someAriaLabel} />
  </div>
  )
}
```

modules/_labs/drawer/react/stories/stories.tsx
```diff
@@ -64,7 +64,7 @@ storiesOf('Labs|Drawer/React', module)
         <Drawer
           header={
             <DrawerHeader
-              closeIconLabel={'Close'}
+              aria-label={'Close'}
```
anicholls, line 67: Shouldn't these be `closeIconAriaLabel` now? I would only expect to use `aria-label` if it were being applied to the header container element itself.

modules/button/react/lib/IconButton.tsx
```diff
@@ -11,7 +11,7 @@ export interface IconButtonProps extends React.ButtonHTMLAttributes<HTMLButtonEl
   /**
    * The accessibility label to indicate the action triggered by clicking the IconButton.
    */
-  'aria-label': string;
+  'aria-label'?: string;
```
anicholls, line 14: This should be required.

modules/popup/react/lib/Popup.tsx
anicholls: Honestly I don't think we need a title for this icon. We generally don't apply titles to icon buttons. We could probably remove it, and it would make this clearer.

mannycarrera4: I was still used for aria-label and I figured I'd keep it for translations

4.0-MIGRATION-GUIDE.md
```diff
@@ -170,6 +170,82 @@ PRs:
 
 ---
 
+### DrawerHeader
+
+The following props where renamed for appropriate aria naming and clarity •
+
+- ``closeIconLabel`->`closeIconAriaLabel`
```
NicholasBoll, line 177:
```suggestion
- `closeIconLabel`->`closeIconAriaLabel`
```

PR #552, opened 2020-04-07
feat: Enable backdoor font theming for internationalization

anicholls:
## Summary

Under certain location circumstances, our product needs to load extra fonts for the best experience (e.g. using Noto for CJK character support). We don't want to always load these fonts as they can take a while to load and are often unnecessary. To remedy this, we need to enable consumers to dynamically change out the font family at render time. 

This PR takes advantage of our custom `styled` function from our theming API to override any `fontFamily` styles set within our components if a custom font stack has been provided in the theme context (using an undocumented `_fontFamily` field).

**Note: This POC was making the assumption that the font needs to be changeable "on the fly". If we can assume it will never change after page load, we could just look at a `window` variable provided by the app when `@workday/canvas-kit-react-core/lib/type.js` is initially loaded. This would be simpler, cleaner and more performant.**

### Disclaimer
**Our design team has not given permission to customize the font of our components, so this is not intended for such usage. This is a "hidden" feature for our internal teams, meant only to enable better internationalization support.** 

### Usage Example
```tsx
import {
  type,
  CanvasProvider,
  createCanvasTheme,
  PartialCanvasTheme,
} from '@workday/canvas-kit-labs-react-core';

const theme: PartialCanvasTheme & {_fontFamily?: string} = {
  _fontFamily: `"Noto Sans CJK KR", ${type.body.fontFamily}`,
};

<CanvasProvider theme={createCanvasTheme(theme)}>
  {/* Components */}
</CanvasProvider>
```

### Todo:
- Since this functionality is only available to components using `styled`, we need to finish migrating our components from Emotion's `styled` function. Would like some input on whether this should happen in this PR.

### Open question
- What do we need to do for mono text?

### DIFF
--- a/modules/_labs/core/react/lib/theming/createCanvasTheme.ts
+++ b/modules/_labs/core/react/lib/theming/createCanvasTheme.ts
@@ -78,11 +78,13 @@ function fillPalette(palette?: PartialCanvasThemePalette): CanvasThemePalette | 
   };
 }
 
-export function createCanvasTheme(partialTheme: PartialCanvasTheme): CanvasTheme {
-  const {palette = {}, breakpoints = {}, direction} = partialTheme;
+type PartialCanvasThemeWithFont = PartialCanvasTheme & {_fontFamily?: string};
+
+export function createCanvasTheme(partialTheme: PartialCanvasThemeWithFont): CanvasTheme {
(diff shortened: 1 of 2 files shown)


anicholls:
After discussion - taking a different approach. Closing

PR #553, opened 2020-04-08
feat(core): Add window configuration option to inherit font family

anicholls:
## Summary

Under certain location circumstances, our product needs to load extra fonts for the best experience (e.g. using Noto for CJK character support). We don't want to always load these fonts as they can take a while to load and are often unnecessary. To remedy this, we need to enable consumers to opt out of our font stack at page load so they can rely on a font stack they've specified at a higher level.

Since this only needs to happen at page load, this PR simply updates our type hierarchy to look for a window variable set before adding the `fontFamily` property to our various levels. This method is more performant, simpler and provides less room for error than what was explored in #552. It also avoids all bespoke teams within the Workday application from having to manage the font stack themselves (since this window variable can be set at the highest level of the app and not the nearest `CanvasProvider`)

Note: we are not currently worried about mono fonts because mono text is not getting any special treatment like this in the app at the moment.

### Usage Example

Consumers simply need to set the following variable before `@workday/canvas-kit-react-core/lib/index.js` is loaded`:
```tsx
window.workday.canvas.inheritFontFamily = true;
```
Note: You may need to add this definition to your window object if using Typescript.

### DIFF
--- a/modules/core/react/lib/type.ts
+++ b/modules/core/react/lib/type.ts
@@ -1,10 +1,9 @@
+import get from 'lodash/get';
 import {default as colors, typeColors, statusColors} from '@workday/canvas-colors-web';
 import {borderRadius} from './radius';
 import {CSSProperties} from './types';
 
-const inheritFont =
-  // @ts-ignore
-  window.workday && window.workday.canvas && window.workday.canvas.inheritFontFamily;
+const inheritFont = get(window, 'window.workday.canvas.inheritFontFamily');
 export const fontFamily = inheritFont
   ? 'inherit'
   : '"Roboto", "Helvetica Neue", "Helvetica", Arial, sans-serif';
--- a/modules/core/react/lib/type.ts
+++ b/modules/core/react/lib/type.ts
@@ -2,7 +2,12 @@ import {default as colors, typeColors, statusColors} from '@workday/canvas-color
 import {borderRadius} from './radius';
 import {CSSProperties} from './types';
 
-export const fontFamily = '"Roboto", "Helvetica Neue", "Helvetica", Arial, sans-serif';
+const inheritFont =
+  // @ts-ignore
+  window.workday && window.workday.canvas && window.workday.canvas.inheritFontFamily;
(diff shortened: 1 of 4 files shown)


modules/core/react/lib/type.ts
```diff
@@ -2,7 +2,12 @@ import {default as colors, typeColors, statusColors} from '@workday/canvas-color
 import {borderRadius} from './radius';
 import {CSSProperties} from './types';
 
-export const fontFamily = '"Roboto", "Helvetica Neue", "Helvetica", Arial, sans-serif';
+const inheritFont =
+  // @ts-ignore
+  window.workday && window.workday.canvas && window.workday.canvas.inheritFontFamily;
```
NicholasBoll, line 7: Seems like there should be a better way...

First, it seems wrong to do a `// @ts-ignore`. Using `lodash.get` would be a bit better here.

Second, we are referencing paths that exist within one product. It seems like we need a different approach to allow overrides with an official API from our side. Maybe through the theme?

anicholls: > First, it seems wrong to do a `// @ts-ignore`

[We're already doing this for `useTheme`](https://github.com/Workday/canvas-kit/blob/master/modules/_labs/core/react/lib/theming/useTheme.ts#L34). The idea was that we don't want to add to the window global scope for people who aren't using this functionality. Adding these fields (i.e. `declare global { interface Window { ... }}`) would add them for all consumers. `lodash.get` would probably work if you think that's preferable.

> Second, we are referencing paths that exist within one product. It seems like we need a different approach to allow overrides with an official API from our side. Maybe through the theme?

The idea is that anyone could specify this window variable. My approach in https://github.com/Workday/canvas-kit/pull/552 was more of an official API using theming, but the performance was worse, it allowed consumers to specify _any_ font (which we want to avoid), and every bespoke team in product would need to add it to their theme object. The method in this PR is a one and done thing controlled by the UI Platform team for any apps within the Workday SPA.

PR #555, opened 2020-04-08
ci: Test github actions

NicholasBoll:


NicholasBoll:
Works, but iteration needs to be done against master

PR #556, opened 2020-04-08
ci: Add unit test workflow file

NicholasBoll:


NicholasBoll:
Need to open against master for it to register correctly

PR #557, opened 2020-04-08
ci: Enable Github Actions for pull requests

### LINKED ISSUE #351: Convert CI back to Github Actions for pull requests
When we get things approved behind-the-scenes, migrate Travis CI (except for the release stage) back to Github Actions.

They are currently in hibernation as `bak` files so they should be ready to go when we switch over.

NicholasBoll:
Fixes #351 

Some benefits of Github Actions vs TravisCI
- Github Actions run faster
- Github Actions supports [Annotations](https://github.blog/2018-12-14-introducing-check-runs-and-annotations/#annotations-in-pull-requests) based on regex matching from the CLI output (from the `setup-node` action. Works with TS and ESLint errors. We've adapted some of our output to match `eslint-stylish` output to be picked up by annotations as well)

Some benefits of this PR:
- Faster builds with special caching

| TravisCI | Github Action Cache miss | Github Action cache hit |
|---------|---------------------------|------------------------|
| 20m | 16m (includes building cache) | 10m |

There are 2 caches:
- Cypress
- node_modules

`yarn install` takes about 6.5m on Github Actions. The only way to speed that up would be to minimize duplicate dependencies or dependencies altogether. Even with a cached `node_modules`, `yarn install` takes a significant amount of time to verify all dependencies in the `yarn.lock` file with what is on the filesystem. The cache works by caching both the Cypress binary and `node_modules` and _skips_ `yarn install` on cache-hit. This is only safe as long as we don't rely on special things like local-specific versions of npm `.bin` directories. So far we do not and probably should not.

I did not optimize command steps to run in parallel as this would break the annotation pattern matching. So far, the steps we parallelized in the TravisCI build run quickly enough it just isn't worth losing annotations.

### Commit Message
Use GitHub Actions for better speed, ease of use, and annotation pattern matching. Uses a bold caching technique of skipping `yarn install` if `yarn.lock` file or Cypress version caches are primed.

Fixes #351


.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10.x
+
+      - name: Cache node modules
+        id: yarn-cache
+        uses: actions/cache@v1
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-node-modules-hash-${{ hashFiles('**/yarn.lock') }}
```
NicholasBoll, line 23: Restoring cache takes about 1 minute

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Install Packages
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
```
NicholasBoll, line 26: Skipping `yarn install` is a bold move.

Unfortunately, to ensure all dependencies are valid, `yarn install` doesn't benefit from cache. Almost all the time it takes is verifying the integrity of all the files in `node_modules` and ensuring everything matches what's in the `yarn.lock` file.

Skipping makes the following assumptions:
- `yarn.lock` + OS are the only things that effect the contents of `node_modules`
- None of our CI scripts change the contents of `node_modules` _after_ `yarn install`
- Github Actions does file integrity check both when creating and when restoring cache files (failure may result in incomprehensible errors)
- None of our scripts use the contents of `node_modules` that is local to the module itself (`uuid` typically adds a `node_modules/.bin/uuid` command local to the module)

As long as these assumptions hold true, we save about 6 minutes on a build. In this script, that's 2/3 of the total CI time!

Absolute worst case: corrupt cache. Wasted time to find out it was a cache error. Destroy the cache.
Worst case: cache miss. Build takes an extra 1 minutes to create a cache.
Best case: cache hit. Build takes 6 minutes less.

utils/check-dependencies-exist.js
```diff
@@ -42,52 +43,94 @@ const labelMap = {
...
+            const char = findCharacterInFile(contents, packageName);
+
+            return `  ${colors.dim(`${line}:${char}`)}  ${colors.red('error')}  ${colors.yellow(
+              packageName
+            )} is not used in code  ${colors.dim('check-dependencies-exist')}`;
+          })
+          .join('\n');
+      }
+    })
+    .join('\n\n');
```
NicholasBoll, line 133: These changes update the output of the script to make an `eslint-stylish`-compatible output so that Github Actions creates an Annotation for any failures. Not strictly necessary, but it is nice when we don't have to comb over CI logs to see what and where failures occur.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - master
```
anicholls, line 6: Can we add prerelease branches here too?

anicholls:
image

Do you know why this is labeled as `unit_test`? Do we have any way of configuring the title? Or is it just the file/line and code line?

.github/workflows/ci.yml
NicholasBoll: Oh, yes. I forgot about the base branch prerequisite. Do we want to limit what PRs this runs on? I think I'll just remove it.

NicholasBoll:
> image
> 
> Do you know why this is labeled as `unit_test`? Do we have any way of configuring the title? Or is it just the file/line and code line?

It was when we had 2 jobs running - `unit_test` and `integration_test`. I updated it to be `verify` now. It can be any name we want. You can see it in the list of checks now:

Screen Shot 2020-04-13 at 2 23 19 PM

.travis.yml
```diff
@@ -33,25 +33,25 @@ before_install:
...
+      #   lint
+      #   depcheck
+      #   test
+      # - >- # Chromatic and Cypress rely on Storybook being built, this will exit early if build-storybook fails
+      #   yarn build-storybook &&
+      #   yarn http-server docs -p 9001 & npx wait-on http://localhost:9001 &&
+      #   yarn run-p -l
+      #   "chromatic -- --quiet --exit-zero-on-changes --exit-once-uploaded --storybook-build-dir docs"
+      #   "cypress:run -- --record" &&
+      #   kill $(jobs -p) || true
```
NicholasBoll, line 54: I think we're ready to remove these lines

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Start Server
+        run: yarn http-server docs -p 9001 & npx wait-on http://localhost:9001
+
+      - uses: chromaui/action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          appCode: dlpro96xybh
+          storybookBuildDir: docs
+          exitOnceUploaded: true
+          exitZeroOnChanges: true
```
NicholasBoll, line 63: The chromatic CLI command doesn't detect Github Actions as a CI provider. The Github Action here was built for this purpose.

anicholls: We used to have a [separate action for chromatic](https://github.com/Workday/canvas-kit/actions?query=workflow%3AChromaticQA). I don't exactly remember why though

NicholasBoll: I think we had separate workflows and jobs so that everything would run in parallel to be faster.

I removed the integration job and merged with the `unit_test` job since unit tests don't actually take that long. We also now do `exitOnceUploaded`, so we don't wait for Chromatic anymore. We instead wait on Chromatic to set a status on the commit - it is possible for this CI job to finish before Chromatic does, but that's okay. The PR will be blocked until Chromatic sets a passing status.

The slowest part of all our CI scripts is `yarn install`. Doing that 3 times seems wasteful. Unit tests + lint + checks takes about 1.5 mins. Uploading to Chromatic takes about 14s. Running Cypress takes a few minutes. Building Storybook takes a few minutes. We have to build storybook for both Cypress and Chromatic. It seems okay to have Cypress wait 14s for Chromatic rather than running them in parallel and deal with the verbose output of `run-p`.

.travis.yml
```diff
@@ -33,31 +21,11 @@ before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.15.2
   - export PATH="$HOME/.yarn/bin:$PATH"
 
-install:
-  - yarn install
+# install:
+  # - yarn install
```
NicholasBoll, line 25: Oops. Probably don't want to comment this out

.github/workflows/ci.yml
NicholasBoll: Hmm. I think I read the wrong line. Chromatic takes about 1.5 min to upload. Should we `run-p` that with Cypress? Builds are still around 10m. Running in parallel might save about 1m off that. We don't parse or annotate based off the Cypress error since there was no useful stack trace.

Cypress 4.4.0 just added support for proper stack traces: https://github.com/cypress-io/cypress/pull/6965

We could add an annotation matcher so that failed tests will show us which line failed.

package.json
```diff
@@ -170,5 +170,6 @@
   "workspaces": [
     "modules/**"
   ],
-  "browserslist": "> 5%, last 2 firefox versions, last 2 chrome versions, last 2 safari versions, last 2 edge versions, ie 11"
+  "browserslist": "> 5%, last 2 firefox versions, last 2 chrome versions, last 2 safari versions, last 2 edge versions, ie 11",
+  "version": "0.0.0"
```
NicholasBoll, line 174: Remove this

.github/workflows/ci.yml
anicholls: Makes sense 👍  We can probably leave parallelizing Chromatic and Cypress to a separate PR if we want to do it.

Issue #544, opened 2020-03-31
chore: 4.0 function promotions out of labs


anicholls:
- Promote `useTheme`, `styled` out of labs (into `@workday/canvas-kit-react-common/lib/theming`)
  - Continue referencing types from `@workday/canvas-kit-labs-react-core`
  - We want to move some of our API into common (since it's being used in a lot of our components now), but we want to keep the theme object stuff in labs
  - Don't worry about documenting these (since they're internal to our components)
- Replace `focusRing` with `themedFocusRing` for 4.0 (keeping the name `focusRing)

anicholls:
Closed by #558

PR #558, opened 2020-04-09
chore: Promote functions out of labs

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/544


modules/common/react/lib/styles/focusRing.ts
```diff
@@ -99,9 +65,17 @@ interface ThemedFocusRingOptions {
  *
  * @returns {CSSObject} the css object for the focus ring style
  */
-export function themedFocusRing(
-  theme: CanvasTheme,
-  options: ThemedFocusRingOptions = {}
+export default function focusRing(
+  theme: CanvasTheme = useTheme(),
```
mannycarrera4, line 69: theme was passed as first param because it is later used int he second param has a default value, I'm open to re working this

anicholls:
So sorry to do this, but I actually wonder if the theming stuff should go in common rather than core...

mannycarrera4:
@anicholls move it under a theming folder in common? thoughts? no worries

mannycarrera4:
Things that might also be promoted:
• Type, spacing, and RTL? Most of these we haven't touched in a while and we use in other components
• Drawer? Also hasn't been touched, do we use drawer as a base for side panel? (this is future work)
• Is pagination ready?
• Is segmented control ready?

modules/common/react/lib/styles/focusRing.ts
anicholls: Maybe let's change this to the second param. `focusRing({}, theme)` is a lot less common than `focusRing(undefined, {...})`, so I think this will be a lot cleaner.

modules/_labs/color-picker/react/lib/parts/SwatchBook.tsx
```diff
@@ -34,7 +34,7 @@ const SwatchContainer = styled('div')<SwatchContainerProps>(
 
     '&:focus': {
       outline: 'none',
-      ...focusRing(2, 2),
+      ...focusRing({width: 2, separation: 2}),
```
anicholls, line 37: Nit: You can remove `width: 2` everywhere. This is the default and it was only needed previously because you had to pass the width to specify the separation.

modules/button/react/lib/deprecated_Button.tsx
```diff
@@ -136,11 +136,11 @@ const Container = styled('button')<DeprecatedButtonProps>(
       '&:not([disabled])': {
         '&:focus': {
           borderColor: buttonColors.focusBorder,
-          ...focusRing(2, 0),
+          ...focusRing({width: 2, separation: 0}),
         },
         '&:active': {
           borderColor: buttonColors.activeBorder,
-          ...focusRing(2, 0),
+          ...focusRing({width: 2, separation: 0}),
```
anicholls, line 143: These are the defaults. You can just use `focusRing()`

modules/common/react/lib/styles/focusRing.ts
```diff
@@ -1,29 +1,19 @@
...
-  inset: boolean;
-  innerShadowColor: string;
-  outerShadowColor: string;
-};
-
 function calculateFocusRing({
-  ringWidth,
-  separationWidth,
+  width = 2,
+  separation = 0,
```
anicholls, line 8: Why are defaults needed here? They will always be defined coming from `focusRing` since it has its own defaults

modules/core/react/package.json
```diff
@@ -53,6 +53,7 @@
     "@workday/canvas-colors-web": "^0.17.13",
     "@workday/canvas-depth-web": "^0.16.4",
     "@workday/canvas-space-web": "^0.15.5",
+    "@workday/canvas-kit-labs-react-core": "^4.0.0-beta.0",
```
anicholls, line 56: This should be in `common` now, correct?

PR #533, opened 2020-03-24
chore: Manage dependencies

anicholls:
## Summary

Upgrade & remove some dependencies for 4.0

## Checklist

- [x] Upgrade Typescript to 3.8
- [ ] Upgrade Babel to 7.9
- [x] Remove unused Babel 6 deps
- [x] [Use downlevel-dts to address breaking changes in TS@3.7](https://github.com/storybookjs/storybook/pull/9847)
- [x] Remove tslint dependency
- [x] Upgrade to react/react-dom 16.12
- [ ] Remove react-scripts dep (resolves Storybook 6.0 warning re https://www.npmjs.com/package/@storybook/preset-create-react-app and simplifies our webpack config)

Note: For reviewing, it's probably easiest to follow commit by commit rather than looking at files changed

PR #566, opened 2020-04-14
feat(badge): Add badge component

alanbsmith:
## Checklist

- [x] `yarn test` passes
      applicable


modules/_labs/badge/react/lib/Badge.tsx
```diff
@@ -0,0 +1,68 @@
...
+    minWidth: '20px',
+    padding: '0 6.5px',
+  },
+  ({variant = 'default'}) => ({
+    ...variants[variant],
+  })
+);
+
+const Badge = (props: BadgeProps) => {
+  const {count = 0, ...elemProps} = props;
```
mannycarrera4, line 58: you can add the variant default here

alanbsmith: I thought about that, but it gets used [here](https://github.com/Workday/canvas-kit/pull/566/files#diff-53b9d8f6bc67ea5a52fc795e9c4bb0bdR52), so I opted to pass it along with `{...elemProps}`. Unless you want it to be explicit?

mannycarrera4: ah this isn't yet a common practice cause we just updated this, but @anicholls came up with a type that makes typescript happy so when you pass a default down below, it won't complain in the styled component. That just got merged into pre release v4 https://github.com/Workday/canvas-kit/blob/prerelease/v4/modules/common/react/lib/types.ts#L30

I think for now it this works! not a blocker for me

alanbsmith: Oh, that's nice! Thanks!

modules/_labs/badge/css/index.scss
```diff
@@ -0,0 +1,3 @@
+@import '@workday/canvas-kit-css-core/index.scss';
```
alanbsmith, line 1: Q: Should I be more specific with which assets I import here?

modules/_labs/badge/css/lib/badge.scss
```diff
@@ -0,0 +1,53 @@
...
+
+  &:empty {
+    background: $wdc-color-french-vanilla-100;
+    border: solid 6px $wdc-color-cinnamon-500;
+    padding: 0;
+  }
+}
+
+// inverse badge styles
+.wdc-badge--inverse {
```
alanbsmith, line 42: Q: I used `--inverse` as a BEM-style syntax, but if there's another convention we're following, I can adjust it.

modules/_labs/badge/react/lib/Badge.tsx
```diff
@@ -0,0 +1,82 @@
...
+const Badge = (props: BadgeProps) => {
+  const {count, ...elemProps} = props;
+
+  if (count === undefined) {
+    return <Container {...elemProps} role="status" aria-live="polite" />;
+  }
+
+  const formattedCount = count < 1000 ? count : '999+';
+  return (
+    <Container {...elemProps} role="status" aria-live="polite">
```
alanbsmith, line 76: I decided to use `role="status"` and `aria-live="polite"` because of what I read [here](https://inclusive-components.design/notifications/).

> "Somewhat perplexingly, there are two equivalent APIs for live regions: the aria-live attribute and live region ARIA roles. In most cases, you will want to use one of role="status" or aria-live="polite". Using both simultaneously maximizes compatibility with different browser and assistive technology pairings"

modules/_labs/badge/css/lib/badge.scss
mannycarrera4: i think we just default to `.wdc-badge-inverse`

modules/_labs/badge/react/lib/Badge.tsx
alanbsmith: I also pass these to Container _after_ the spread props, so they aren't able to be overridden. But if we'd like to provide more flexibility I can put them before the spread props.

modules/_labs/badge/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,73 @@
...
+  title: 'Labs|Badge/React/',
+  decorators: [withReadme(README)],
+  parameters: {
+    chromatic: {
+      disable: false,
+    },
+  },
+};
+
+export const VisualTesting = () => {
```
alanbsmith, line 20: If there's a better pattern for visual testing, please let me know. I went with a straightforward approach to describe all the permutations in a table format.

utils/storybook/ComponentStatesTable.tsx
```diff
@@ -40,7 +40,7 @@ export interface ComponentStatesTableProps {
   children(props: Props): React.ReactNode;
 }
 
-const Table = styled('table')({
+export const Table = styled('table')({
```
alanbsmith, line 43: I didn't need all the states, as Badge isn't interactive, but this styled table was handy, so I exported it.

modules/_labs/badge/react/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+provides a quantity-based summary with dynamic values
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-labs-react-badge
+```
+
+## Usage
```
alanbsmith, line 15: Q: Should I be more descriptive about usage here? More "why" to go with the "how"?

modules/_labs/badge/react/stories/stories_visualTesting.tsx
mannycarrera4: I think this is our bad for not documenting better how we visual  test our components, at least our patterns. Instead of creating a new file just for visual test, we add another folder but just add the test within the same file. Look at `stories_Checkbox.tsx` as an example, specifically the `CheckboxStates` method.

modules/_labs/badge/css/index.scss
anicholls: Yeah if all you're using is variables, I would narrow this scope

modules/_labs/badge/css/lib/badge.scss
```diff
@@ -0,0 +1,53 @@
...
+
+@mixin wdc-badge {
+  // base styles
+  align-items: center;
+  animation: wdc-badge-grow 0.2s ease;
+  border-radius: 999px;
+  box-sizing: border-box;
+  display: inline-flex;
+  font-family: $wdc-font-family;
+  font-size: 11px;
```
anicholls, line 19: This is more of a design question, but using 11px here causes it to be visually off center since the browser won't render the font on sub-pixels. Maybe we should consider using 10 or 12?

modules/_labs/badge/css/lib/badge.scss
```diff
@@ -0,0 +1,53 @@
...
+  font-weight: 700;
+  height: 20px;
+  justify-content: center;
+  min-width: 20px;
+  padding: 0 6.5px;
```
anicholls, line 24: Another design question: 6.5 seems weird to me

modules/_labs/badge/css/lib/badge.scss
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// default badge styles
+.wdc-badge {
+  @include wdc-badge();
+  background: $wdc-color-cinnamon-500;
+  color: $wdc-color-french-vanilla-100;
+  text-shadow: 0 0 1px rgba(0, 0, 0, 0.35);
+
+  &:empty {
```
anicholls, line 34: TIL 👍

modules/_labs/badge/css/lib/badge.scss
```diff
@@ -0,0 +1,53 @@
...
+    background: $wdc-color-french-vanilla-100;
```
anicholls, line 35: Weird visual thing, but this dot seems so much larger than the blue one 😂  (though I know it's not)

modules/_labs/badge/css/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "lib",
+    "index.scss"
+  ],
+  "style": "dist/canvas-kit-css-badge.min.css",
+  "main": "dist/canvas-kit-css-badge.min.css",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Workday/canvas-kit/tree/master/modules/_labs/badge/css"
+  },
+  "scripts": {
```
anicholls, line 19: Need to add dependency here too

modules/_labs/badge/react/README.md
anicholls: Usually this would go within the props below. You only have the tables, but you're missing the normal description + default that we have for each prop

modules/_labs/badge/react/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+// empty Badge
+<Badge variant="inverse" aria-label="new unread notifications" />
+```
+
+## Static Properties
+
+| Name      | Value    |
+| --------- | -------- |
+| aria-live | 'polite' |
```
anicholls, line 35: These aren't static properties. Static properties would be `Badge.Variant`

modules/_labs/badge/react/eslintrc.js
```diff
@@ -0,0 +1,116 @@
+const prettierConfig = require('./.prettierrc.js');
+
+module.exports = {
```
anicholls, line 3: Pretty sure we can just `require('../../../../eslintrc.js')` and avoid all this duplication

modules/_labs/badge/react/lib/Badge.tsx
```diff
@@ -0,0 +1,82 @@
...
+    minWidth: '20px',
+    padding: '0 6.5px',
+  },
+  ({variant = 'default'}) => ({
+    ...variants[variant],
+  })
+);
+
+const Badge = (props: BadgeProps) => {
+  const {count, ...elemProps} = props;
```
anicholls, line 68: Generally we explicitly pull out all props so that there's no confusion around what's in `elemProps` (see next comment for an example). It may seem a bit verbose, but we've run into bugs from not doing this in the past.

modules/_labs/badge/react/lib/Badge.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  if (count === undefined) {
+    return <Container {...elemProps} role="status" aria-live="polite" />;
+  }
+
+  const formattedCount = count < 1000 ? count : '999+';
+  return (
+    <Container {...elemProps} role="status" aria-live="polite">
+      {formattedCount}
+    </Container>
```
anicholls, line 78: Nit: you don't really need this conditional with the same `Container`. This case is simple, but it could be asking for inconsistencies. It could be simplified to: 

```suggestion
  const {count, variant, aria-label: ariaLabel, ...elemProps} = props;

  return (
    <Container 
      count={count} 
      variant={variant} 
      aria-label={ariaLabel}
      role="status" 
      aria-live="polite"
      {...elemProps}
    >
      {count && count < 1000 ? count : '999+'}
    </Container>
```

modules/_labs/badge/react/package.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "keywords": [
+    "canvas",
+    "canvas-kit",
+    "react",
+    "components",
+    "workday",
+    "badge"
+  ],
+  "peerDependencies": {
```
anicholls, line 47: Need deps added

modules/_labs/badge/react/stories/stories_visualTesting.tsx
anicholls: @mannycarrera4 with the move to the CSF format, this will have to be a separate file now that the parameters are defined in the default export (which is file-wide).

Honestly this is really going to be a pain when plugins can only be configured once for a whole file. cc @NicholasBoll

modules/_labs/badge/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+import Badge from '../index';
+import README from '../README.md';
+
+export default {
+  title: 'Labs|Badge/React/',
+  decorators: [withReadme(README)],
+  parameters: {
+    chromatic: {
+      disable: false,
```
anicholls, line 15: @NicholasBoll to chime in when he figures out how to get this to work (chromatic is not catching these in the CSF format atm)

modules/_labs/badge/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,73 @@
...
+          </td>
+          <td>
+            <Badge variant="inverse" aria-label={`${999} unread notifications`} count={999} />
+          </td>
+          <td>
+            <Badge variant="inverse" aria-label={`${1000} unread notifications`} count={1000} />
+          </td>
+        </tr>
+      </tbody>
+    </Table>
```
anicholls, line 71: Nit: This would be much simpler if you used `ComponentStatesTable`. No need to export `Table` in that case

```suggestion
      <ComponentStatesTable
        rowProps={[
          {label: 'Default', props: {}},
          {label: 'Inverse', props: {variant: Badge.Variant.Inverse}},
        ]}
        columnProps={[
          {label: 'Empty', props: {}},
          {label: 'Single Digit', props: {count: 1}},
          {label: 'Double Digit', props: {count: 23}},
          {label: 'Triple Digit', props: {count: 999}},
          {label: 'Greater than 999', props: {count: 1000}},
        ]}
      >
        {props => (
          <Badge {...props} aria-label={`${props.count} unread notifications`} />
        )}
      </ComponentStatesTable>
```

modules/_labs/badge/react/stories/stories_visualTesting.tsx
mannycarrera4: ooo TIL

modules/_labs/badge/css/lib/badge.scss
alanbsmith: Good call out. I'll talk to Rick about it. Thanks!

modules/_labs/badge/css/lib/badge.scss
alanbsmith: Yep. Rick and I talked about that. 

> "6.5px (obviously obscure) that value is just based on Roboto font sizing and default letter spacing."

modules/_labs/badge/css/lib/badge.scss
alanbsmith: Right! I saw that too! Visual trickery. I'm glad I'm not the only one. There wasn't a spec for this, so I'm going to run it by Rick. 👍

modules/_labs/badge/css/lib/badge.scss
anicholls: (or make the badge 1px taller/shorter)

modules/_labs/badge/react/stories/stories_visualTesting.tsx
alanbsmith: ohhh. I was wondering if it needs to be:
```
  parameters: {
    options: {
      chromatic: {
        disable: false,
      },
  },
}
```

Like what's shown [here](https://storybook.js.org/docs/configurations/options-parameter/#per-story-options)

modules/_labs/badge/react/stories/stories_visualTesting.tsx
alanbsmith: Yeah, I thought the `ComponentStatesTable` might be misleading as Badge is non-interactive? No hover / focus states.

modules/_labs/badge/react/lib/Badge.tsx
alanbsmith: yeah, good call. 👍

modules/_labs/badge/react/stories/stories_visualTesting.tsx
anicholls: Since this is for visual testing and not really for consumers viewing, I think it's preferable to use the more concise option.

Also just because they're not interaction states doesn't mean these aren't still states (empty state, overcount state, etc.)

modules/_labs/badge/react/stories/stories_visualTesting.tsx
anicholls: It looks right, but it wasn't working for @NicholasBoll on #528. Will have to double check

modules/_labs/badge/react/lib/Badge.tsx
alanbsmith: The logic on 79 needs to be tweaked a little though. I'll grab the rest of this. 👍

modules/_labs/badge/react/README.md
alanbsmith: ohhh ok. Thank you!

modules/_labs/badge/react/eslintrc.js
alanbsmith: I wonder if that will that work if someone only pulls in this module is installed individually. 🤔

anicholls: Linting is generally only used for dev within the context of the entire monorepo. This file won't even be included in what gets published to npm

modules/_labs/badge/react/stories/stories_visualTesting.tsx
anicholls: This looks good 👍  (see https://github.com/Workday/canvas-kit/pull/585)

modules/_labs/badge/react/eslintrc.js
alanbsmith: Yep. Good call. Will update. 👍

modules/_labs/badge/css/README.md
```diff
@@ -0,0 +1,68 @@
...
+<span class="wdc-badge aria-live="polite" aria-label="1 unread notification">
+  1
+</span>
+```
+
+#### Inverse
+
+```html
+<span class="wdc-badge-inverse" aria-live="polite" aria-label="999+ unread notifications">
+  2
```
anicholls, line 38: Nit: does the aria label need to shorten as well? That's mainly for visual cleanliness
```suggestion
<span class="wdc-badge-inverse" aria-live="polite" aria-label="1000 unread notifications">
  1000
```

modules/_labs/badge/css/README.md
```diff
@@ -0,0 +1,68 @@
...
+</span>
+```
+
+#### Empty Badge
+
+```html
+<span class="wdc-badge aria-live="polite" aria-label="1 unread notification"></span>
```
anicholls, line 45:
```suggestion
<span class="wdc-badge aria-live="polite" aria-label="Unread notifications"></span>
```

modules/_labs/badge/react/README.md
```diff
@@ -0,0 +1,82 @@
+# Canvas Kit React Badge
+
+<a href="https://github.com/Workday/canvas-kit/tree/master/modules/_labs/README.md">
```
anicholls, line 3: Reminder to remove this when you take it out of labs

modules/_labs/badge/react/lib/Badge.tsx
```diff
@@ -0,0 +1,83 @@
+import * as React from 'react';
+import {keyframes} from '@emotion/core';
+import styled from '@emotion/styled';
+import {borderRadius, colors, fontFamily} from '@workday/canvas-kit-react-core';
+
+export type BadgeVariant = 'default' | 'inverse';
```
anicholls, line 6: Nit: If we're not doing an enum you could probably just inline this in the interface.

modules/_labs/badge/react/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+#### `variant: 'default' | 'inverse'`
+
+> Styled variant for the Badge
+
+| Variant   | Description                 |
+| --------- | --------------------------- |
+| `default` | Red background, White text  |
+| `inverse` | White background, Blue icon |
```
anicholls, line 68:
```suggestion
| `inverse` | White background, Blue text |
```

modules/_labs/badge/react/lib/Badge.tsx
```diff
@@ -0,0 +1,83 @@
...
+    minWidth: '20px',
+    padding: '0 6.5px',
+  },
+  ({variant = 'default'}) => ({
+    ...variants[variant],
+  })
+);
+
+const Badge = (props: BadgeProps) => {
+  const {count, 'aria-label': ariaLabel, variant, ...elemProps} = props;
```
anicholls, line 69:
```suggestion
  const {variant = 'default', 'aria-label': ariaLabel, count, ...elemProps} = props;
```

modules/_labs/badge/react/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+Default: `default`
+
+---
+
+#### `count: number`
+
+> Count displayed on the Badge
+
+Default: `undefined`
```
anicholls, line 78: If we don't define an default, we just omit this

```suggestion
```

modules/_labs/badge/react/lib/Badge.tsx
alanbsmith: Yeah, I added this for readability (and if other variants need to be added later, but I can do it inline. 👍

modules/badge/css/package.json
```diff
@@ -0,0 +1,33 @@
...
+  ],
+  "style": "dist/canvas-kit-css-badge.min.css",
+  "main": "dist/canvas-kit-css-badge.min.css",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Workday/canvas-kit/tree/master/modules/badge/css"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "node ../../../../utils/css-build.js index.scss"
```
lychyi, line 21:
```suggestion
    "build": "node ../../../utils/css-build.js index.scss"
```
It goes one level too far back.

modules/badge/react/index.ts
```diff
@@ -0,0 +1 @@
+export {default as CountBadge, CountBadgeProps} from './lib/CountBadge';
```
lychyi, line 1:
```suggestion
import CountBadge from './lib/CountBadge';

export default CountBadge;
export {CountBadge};
export * from './lib/CountBadge';
```

This should be aligned with the other modules right?

modules/badge/react/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "type": "git",
+    "url": "https://github.com/Workday/canvas-kit/tree/master/modules/badge/react"
+  },
+  "files": [
+    "dist/",
+    "lib/",
+    "index.ts",
+    "ts3.5/**/*"
+  ],
+  "typesVersions": {
```
lychyi, line 21: This is what is causing builds to fail. I'm not picking up on anything that actually needs to be down-leveled so I'm not sure of the history here.

alanbsmith: yeah, that config plus `"build:downlevel-dts": "yarn run downlevel-dts dist ts3.5/dist",` was generated by the `create-component` script. I removed the build script, but I missed this. Good catch. 👍

modules/badge/react/index.ts
alanbsmith: I assume we want to export CountBadge and CountBadgeProps, but not as default. I wouldn't want someone to `import CountBadge from 'ckr-badge'` if we later create a generic Badge that we want to export as default.

modules/badge/react/package.json
anicholls: @lychyi I added it to all components when we upgraded to TS3.8 in #533

lychyi: @anicholls @alanbsmith Sorry, should have been clearer. What I meant to say was "...as it pertains to this PR". There must have been some history to generating a component using `downlevel-dts` on this PR. TS3.8 is for the next major version right?

Anyway, no big deal, ^ that was what I was really trying to ask. Should have put more context behind it.

modules/badge/css/README.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+## Accessibility
+
+If the count is live-updated (as in the case of notifications) and not static, please add
+`aria-live="polite"` to inform screen readers to announce updated information. When this attribute
+is set the screen reader will announce changes when the user is idle.
+
+```html
+<span class="wdc-count-badge" aria-live="polite" aria-label="3 direct reports">1</span>
```
anicholls, line 31: We shouldn't use direct reports here since that is static
```suggestion
<span class="wdc-count-badge" aria-live="polite" aria-label="3 new notifications">1</span>
```

modules/badge/react/README.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+## Accessibility
+
+If the count is live-updated (as in the case of notifications) and not static, please add
+`aria-live="polite"` to inform screen readers to announce updated information. When this attribute
+is set the screen reader will announce changes when the user is idle.
+
+```jsx
+<Badge count={3} aria-live="polite" aria-label="3 direct reports" />
```
anicholls, line 24:
```suggestion
<Badge count={3} aria-live="polite" aria-label="3 new notifications" />
```

modules/badge/react/README.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+## Usage
+
+```tsx
+import * as React from 'react';
+import Badge from '@workday/canvas-kit-react-badge';
```
lychyi, line 31: Just realized that docs have to change with our new component naming and named exports

modules/badge/react/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+#### `count: number`
+
+> Count displayed on the Badge
+
+Default: `0`
+
+📝 **Note**
+
+Values greater than 999 are formatted to "999+"
```
lychyi, line 76: Note: does this need to be configurable (i18n reasons?) 🤔

alanbsmith: Yeah, I'm not sure, but it's something we should have a discussion about that.

modules/badge/css/README.md
```diff
@@ -0,0 +1,74 @@
...
+- The live region should be visually hidden and only contain text
+
+### Example
+
+```html
+<button type="button" aria-label="Alerts 3 new notifications">
+  <svg aria-hidden="true" focusable="false">...</svg>
+  <span class="wdc-count-badge" aria-hidden="true">3</span>
+</button>
+<div class="wdc-accessible-hide" role="status" aria-live="polite" aria-atomic="true">
```
NicholasBoll, line 41: Is the class called `wdc-accessible-hide`? This is supposed to be only visually hidden, right?

modules/badge/css/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+- The button should have an aria-label that updates with the count
+- The elements inside the button should have `aria-hidden`
+- The live region should be outside the button
+- The live region should be visually hidden and only contain text
+
+### Example
+
+```html
+<button type="button" aria-label="Alerts 3 new notifications">
```
NicholasBoll, line 37: Should this be "3 new notifications" or "Show 3 new notifications"?

modules/badge/react/README.md
```diff
@@ -0,0 +1,91 @@
...
+<IconButton
+  style={{ position: "relative" }}
+  aria-label={`Alerts ${count} new notifications`}
+  size={IconButton.Size.Medium}
+  variant={IconButton.Variant.Circle}
+>
+  <SystemIcon icon={notificationsIcon} aria-hidden="true" />
+  <Badge count={count} aria-hidden="true" />
+</IconButton>
+<div className="wdc-accessible-hide" role="status" aria-live="polite" aria-atomic="true">
```
NicholasBoll, line 39: Is this CSS class normally available? I don't know if we automatically include this class to be available for all React components

modules/badge/react/index.ts
NicholasBoll: I think we should consider deprecating default imports/exports

modules/badge/css/README.md
alanbsmith: This text is from the a11y spec: https://cdpn.io/lefthandev/debug/5e5c051bdde98026a4c49dd583769ed3

modules/badge/react/index.ts
anicholls: > I think we should consider deprecating default imports/exports

I would love to! Maybe for v5?

modules/badge/react/README.md
alanbsmith: That's true. In react-land, they would likely:

```tsx
import { accessibleHide } from "@workday/canvas-kit-react";

const Hidden = styled("div")({
  ...accessibleHide
});
```

and then use that component. I'll update the docs to show that. 👍

modules/badge/css/README.md
alanbsmith: Correct. That's the class specified here:
https://github.com/Workday/canvas-kit/blob/master/modules/common/css/lib/accessibility.scss#L38-L40

I assume they would import the common utils / classes, but maybe not. I can specify that.

anicholls: Probably easier to just change to `css={accessibleHide}`

anicholls: Probably easier to just change to `css={accessibleHide}`

modules/badge/react/README.md
alanbsmith: updated 👍

modules/badge/css/README.md
alanbsmith: Updated! 👍

PR #569, opened 2020-04-14
feat(labs): Add Tabs component

darcar31:
## Summary

Add Tabs component to CK labs. The implementation is not final and will still need overflow support and to meet our yet-undefined compositional API and controllable state PoC.

Adds:

Tabs, Tab, TabList, TabPanel

`Tabs` is meant to be the main entry point and used for 90% of the use cases. The stories show different ways to use the tabs and even how to do a single TabPanel mode using some controllable state.

Basic tabs example:
```tsx
<Tabs>
  <Tabs.List>
    <Tabs.Item>First Tab</Tabs.Item>
    <Tabs.Item>Second Tab</Tabs.Item>
    <Tabs.Item>Third Tab</Tabs.Item>
  </Tabs.List>
  <Tabs.Panel>Contents of First Tab</Tabs.Panel>
  <Tabs.Panel>Contents of Second Tab</Tabs.Panel>
  <Tabs.Panel>Contents of Third Tab</Tabs.Panel>
</Tabs>
```

## Checklist

- [x] `yarn test` passes

## Additional References

Visual Tab States:
Visual Tab States


modules/_labs/tabs/react/lib/TabList.tsx
```diff
@@ -0,0 +1,68 @@
...
+import styled from '@emotion/styled';
+import {useTab} from './Tabs';
+import {TabPanelProps} from './TabPanel';
+import {TabProps} from './Tab';
+
+const TabsListContainer = styled('div')({
+  display: 'inline-block',
+  position: 'relative',
+  width: `100%`,
+  borderBottom: '1px solid',
```
darcar31, line 12: these styles need some cleanup for CKR constants

modules/_labs/tabs/react/lib/Tab.tsx
```diff
@@ -0,0 +1,64 @@
...
+      outline: `none`,
+      boxShadow: isSelected
+        ? ``
+        : `0 0 0 0px ${colors.frenchVanilla100}, 0 0 0 2px ${colors.blueberry400}`,
+      borderRadius: isSelected ? `` : `${borderRadius.s} ${borderRadius.s} 0px 0px`,
+      backgroundColor: isSelected ? `` : colors.soap200,
+      color: isSelected ? `` : colors.blackPepper400,
+    },
+    '&:hover': {
+      backgroundColor: isSelected ? `` : colors.soap200,
```
darcar31, line 29: these styles are the same as focus, should consolidate into a reusable const

modules/_labs/tabs/react/README.md
```diff
@@ -0,0 +1,49 @@
...
+      <Tab index={1}>Tab Number Two</Tab>
+    </TabList>
+    <TabPanels>
+      <TabPanel key="1" index={0}>
+        Hi this is tab content
+      </TabPanel>
+      <TabPanel key="2" index={1}>
+        Hi this is another tab's content
+      </TabPanel>
+    </TabPanels>
```
NicholasBoll, line 33: I've thought a lot about tabs including the API and performance. I know Reach UI does this same API: https://reacttraining.com/reach-ui/tabs/

My concern with a `TabPanel` having all the VDOM present of every render is the potential for more expensive React renders even if non-active tabs aren't rendered to the DOM. I noticed the implementation currently renders all tab panels to the DOM and uses `hidden`. This could be a very big initial render hit on tabs for large tab panel contents.

I think iterating over `TabPanel` children can short-circuit child render functions (React does top-down rendering), but the JSON object still has to be created. Have you run into any problems with this? Also the tab content could be huge. Did you solve this by having a component per `TabPanel` content? Was there any performance concerns with iterating over children using `React.Children.map` and `React.cloneElement`?

The API is definitely easier to understand by looking at the JSX. While the JSX makes it easy to understand TabPanel content, there can be other issues understanding how content gets projected. Is the `TabPanel` that shows up in the final React render tree still a `TabPanel`? Do extra props get mixed in using `React.cloneElement`? The interface of `TabPanel` would then have to include optional props that the application developer shouldn't really be exposed to. If a different component is rendered, it gets a bit confusing how the source JSX translates to the final JSX.

I don't think there is a perfect API. Some components have child lists that could be potentially huge and we have more of a renderProp API for those instances. The API is a bit harder to understand, but does allow us to add rendering optimizations with less confusion of the final JSX and DOM.

darcar31: > This could be a very big initial render hit on tabs for large tab panel contents.

At least for how we used the tabs in our app, we had performance issues with the tabs being very laggy when clicking between them and causing really choppy interactions. We studied a couple different tab components and it seemed like rendering the panels up front as `hidden` was the fix to this problem. We were rending large lists with images and many svg icons which made the render of each panel slow. 

Surely we could have added virtual scrolling etc to make the individual panels quicker if we want to speed up the initial render. I think it may be a matter of preference if we would like to trade the speed of the initial render for the speed of renders on selecting tabs. Perhaps we prefer to speed up the first render and to force our consumers tightly manage the performance of their panels.

> Is the `TabPanel` that shows up in the final React render tree still a `TabPanel`?

Right now, the `TabPanel` components written by the user will end up in the dom as a div with the role `tabpanel`. Is this too far away from what users might associate with the `TabPanel` component itself?

> The interface of `TabPanel` would then have to include optional props that the application developer shouldn't really be exposed to. 

What kind of props do you imagine we might need on the `TabPanel`? I will need to add the accessibility hookups but I think those can be done without exposing any props on the component's API.

> Was there any performance concerns with iterating over children using React.Children.map and React.cloneElement?, Do extra props get mixed in using React.cloneElement?

I also don't see any need to cloneElement right now on the tab panels if we want to force the indexes to be required. If the indexes are optional for both the tabs and the tab panels, we will have to clone element to get the lists built with proper indexes. I think I prefer the API of optional indexes, but like you said, the panel content could be large and could slow down performance.

NicholasBoll: > This could be a very big initial render hit on tabs for large tab panel contents.

Rendering tab panel content with a `hidden` attribute will be more efficient when the browser does Layout and Paint, but won't be any faster inside React. ReactDOM will still process VDOM and turn it into DOM and will still run reconciliation on `TabPanel` children on changes.

What I mean here is APIs like making the contents of `TabPanel` be a function (render prop) or `null`ing the children inside the `Tab` component when iterating over `TabPanels`. This would still parse VDOM objects, but prevent the ReactDOM reconciler from processing and creating DOM of `TabPanel` content that is not active.

> Is the `TabPanel` that shows up in the final React render tree still a `TabPanel`?

I confirmed:
Screen Shot 2020-04-22 at 11 10 09 AM

I meant the React VDOM tree and not the DOM tree. That screenshot is of the React Component render tree using the React dev tools.

> The interface of `TabPanel` would then have to include optional props that the application developer shouldn't really be exposed to.

The accessibility attributes are what I meant. They have to be valid and optional on the `TabPanel` prop interface even if the developer never sets them. They will be overridden I'm sure. You could avoid this by modifying the `React.cloneElement` call and saying what extra props are allowed. This way the external interface won't mention these attributes are valid to set.

> Was there any performance concerns with iterating over children using React.Children.map and React.cloneElement?, Do extra props get mixed in using React.cloneElement?

I think this question covers both performance of rendering DOM (even if using `hidden`) and having to add ARIA attributes. We may be using `cloneElement` right now only to add `key`, but we'll use it to add other attributes as well.

anicholls:
Haven't got around to writing unit and/or cypress tests, and we still don't have the overflow behavior. But otherwise, this is probably ready for design/a11y review

modules/_labs/tabs/react/lib/TabPanels.tsx
```diff
@@ -0,0 +1,24 @@
...
+   */
+  children: TabPanel | TabPanel[];
+}
+
+const TabPanels = ({children}: TabPanelsProps) => {
+  const clonedChildren = React.Children.map(children, (child, index) => {
+    if (!React.isValidElement(child)) {
+      return child;
+    }
+    return React.cloneElement(child, {index});
```
anicholls, line 18: I added this so you don't need to manually pass `index`. It lessens performance, but it's nice for simple use cases. We should probably add a check here and just return the child if `child.props.index` is defined.

In the documentation for `TabPanel`, we can add a note saying performance for the tab will be better if you manually set the index

modules/_labs/tabs/react/lib/TabPanel.tsx
```diff
@@ -0,0 +1,45 @@
...
+    };
+  }, [name, registerPanel, unregisterPanel]);
+
+  return (
+    <div
+      role="tabpanel"
+      aria-labelledby={`tab-${id}-${tabName}`}
+      hidden={!!tabName && tabName !== activeTab}
+      id={`tabpanel-${id}-${tabName}`}
+      tabIndex={0}
```
anicholls, line 37: @NicholasBoll what was the reason for making this focusable? If we intend to keep it this way, we need to add `InputProvider` handling

NicholasBoll: 👍

modules/_labs/tabs/react/stories/stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+import React from 'react';
+import {spacing} from '@workday/canvas-kit-react-core';
+
+import {Tabs} from '@workday/canvas-kit-labs-react-tabs';
+
+import withReadme from 'storybook-readme/with-readme';
+import README from '../README.md';
+
+export default {
+  title: 'Labs/Tabs/React',
```
anicholls, line 13: It looks like all labs stories are now under an `OTHERS` category in this PR. I think it's because the top level category got removed here. If you change this from `Labs/` to `Labs|` I think it will fix it.

image

anicholls (approved):

modules/_labs/tabs/react/stories/stories.tsx
NicholasBoll: Storybook 6 changes this, right?

anicholls: Yeah we'll have to change it back to `/` after SB is updated

modules/_labs/tabs/react/lib/TabPanel.tsx
NicholasBoll: Accessibility says it needs to be focusable. The alternative tab stop story shows how to override this functionality: https://github.com/Workday/canvas-kit/pull/569/files#diff-37fa226a19dc556892d7d02a377f11633ffba98f9052ca038665fd8a04720ca0R93-R111

I added `InputProvider` support to remove the focus outline when using the mouse

NicholasBoll: Story in action: https://5d854c26ba934e0020f5e98a-utszliksxg.chromatic.com/?path=/story/labs-tabs-react--alternative-tab-stop

modules/_labs/tabs/react/stories/stories.tsx
NicholasBoll: Fixed. For some reason, a single `Labs/*` does this for _everything_ in Labs.

PR #578, opened 2020-04-16
ci: Fix annotation for extra dependencies

NicholasBoll:
Fix an issue where the file wasn't being output so the annotation matching wasn't picking it up correctly.


utils/check-dependencies-exist.js
```diff
@@ -118,16 +118,19 @@ function formatErrorMessage(pkgFile, errors) {
           .join('\n');
       } else if (key === 'dependencies' || key === 'devDependencies') {
         const contents = fs.readFileSync(pkgFile).toString();
-        return errors[key]
-          .map(packageName => {
-            const line = findLineInFile(contents, packageName);
-            const char = findCharacterInFile(contents, packageName);
+        return (
+          `${pkgFile}\n` +
```
NicholasBoll, line 122: This was the missing line

PR #584, opened 2020-04-16
test(toast): Write test for Toasts

mannycarrera4:
## Summary

We hadn't written test for Toast yet.  While writing some test, I noticed that adding some accessibility to the toast helped with the test. There's still some accessibility to add to the toast.


modules/toast/react/lib/Toast.tsx
```diff
@@ -97,13 +97,20 @@ export default class Toast extends React.Component<ToastProps> {
         padding={PopupPadding.s}
         handleClose={onClose}
         closeIconSize={IconButtonSize.Small}
+        role={!onClose && !onActionClick ? 'status' : undefined}
+        aria-live={!onClose && !onActionClick ? 'polite' : undefined}
+        aria-atomic={!onClose && !onActionClick ? true : undefined}
```
mannycarrera4, line 102: if the toast has no link or or focusable element, these are the attributes it needs for accessibility

cypress/integration/Toast.spec.ts
```diff
@@ -0,0 +1,93 @@
+import * as h from '../helpers';
+
+function statusToast() {
+  return cy.findByRole('status');
+}
+
+function dialogToast() {
+  return cy.findByRole('dialog');
+}
```
mannycarrera4, line 9: if the toast has focusable elements, it's role changes to dialog

cypress/integration/Toast.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+
+function getToastCloseButton() {
+  return cy.get('[aria-label="Close"');
+}
+
+function getToastActionButton() {
+  return cy.get('[aria-label="View more details"');
```
mannycarrera4, line 16: I'm wondering if there's a better way to get this element.

alanbsmith: https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change

modules/toast/react/stories/stories.tsx
```diff
@@ -81,4 +86,17 @@ storiesOf('Components|Popups/Toast/React', module)
     <div className="story">
       <ToastWrapper state={'Error'} />
     </div>
+  ))
+  .add('With close button', () => (
+    <div className="story">
+      <Toast onClose={() => console.warn('test')}>Your workbook was successfully processed.</Toast>
```
alanbsmith, line 92: just checking if you wanted to leave these `console.warn()`s here. You could also use [Storybook's action addons](https://github.com/storybookjs/storybook/tree/master/addons/actions).

modules/toast/react/lib/Toast.tsx
alanbsmith: I'm curious about `undefined` instead of `false` here. I'm not sure this is incorrect, but I'm wondering if I'm missing some details.

modules/toast/react/lib/Toast.tsx
```diff
@@ -97,13 +97,20 @@ export default class Toast extends React.Component<ToastProps> {
         padding={PopupPadding.s}
         handleClose={onClose}
         closeIconSize={IconButtonSize.Small}
+        role={!onClose && !onActionClick ? 'status' : 'dialog'}
+        aria-live={!onClose && !onActionClick ? 'polite' : undefined}
```
alanbsmith, line 101: Similar to `aria-atomic`, I wonder if this should be `'polite' : 'off'`. It's very possible I'm missing something or I'm incorrect. 😄

mannycarrera4: I guess I didn't know what the opposite of polite was to be fair

modules/toast/react/lib/Toast.tsx
```diff
@@ -97,13 +97,24 @@ export default class Toast extends React.Component<ToastProps> {
         padding={PopupPadding.s}
         handleClose={onClose}
         closeIconSize={IconButtonSize.Small}
+        role={!onClose && !onActionClick ? 'status' : 'dialog'}
```
anicholls, line 100: Nit: What do you think about turning this check into a variable so it's not duplicated?

```tsx
const isInteractive = onClose && onActionClick;

return (
      <Popup
        width={toastWidth}
        padding={PopupPadding.s}
        handleClose={onClose}
        closeIconSize={IconButtonSize.Small}
        role={isInteractive ? 'dialog' : 'status'}
        aria-live={isInteractive ? 'off' : 'polite'}
        aria-atomic={!isInteractive}
```

It just reads a little bit more clearly to me.

modules/toast/react/lib/Toast.tsx
```diff
@@ -97,13 +97,24 @@ export default class Toast extends React.Component<ToastProps> {
...
         {...elemProps}
       >
         <ToastContentContainer onClose={onClose}>
           {icon && <ToastSystemIcon color={iconColor} colorHover={iconColor} icon={icon} />}
           <Message>
             {this.props.children}
-            {onActionClick && <ActionButton onClick={onActionClick}>{actionText}</ActionButton>}
+            {onActionClick && (
+              <ActionButton
+                data-testid={'action-button'} // used to test in cypress
```
anicholls, line 111: Why can't we just use `get/findByRole('button')`? It's one thing to add these to stories, but we should really avoid putting them in our components

modules/toast/react/lib/Toast.tsx
```diff
@@ -97,13 +97,24 @@ export default class Toast extends React.Component<ToastProps> {
...
+                aria-label={actionText}
```
anicholls, line 112: We don't need an aria-label here, the text is visible
```suggestion
```

modules/toast/react/spec/Toast.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+    it('should render an action text', () => {
+      const toastMessage = 'Your workbook was successfully processed.';
+      const actionText = 'View more details';
+      const {container} = render(
+        <Toast
+          onActionClick={cb}
+          actionText={actionText}
+          onClose={cb}
+          icon={checkIcon}
+          data-testid={'myToast'}
```
anicholls, line 60:
```suggestion
```

modules/toast/react/stories/stories.tsx
```diff
@@ -72,6 +72,11 @@ class ToastWrapper extends React.Component<{state?: string}> {
...
   .addParameters({component: Toast})
   .addDecorator(withReadme(README))
+  .add('Default', () => (
+    <div className="story">
+      <Toast>Your workbook was successfully processed.</Toast>
+    </div>
+  ))
   .add('Successful', () => (
     <div className="story">
       <ToastWrapper state={'Success'} />
```
anicholls, line 82: Now that you've added the visual testing story, these are kind of redundant. Can we rename the `Default` story to successful, remove this one, change `Error` to a single case and strip out all of the `ToastWrapper` stuff?

modules/toast/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,56 @@
...
+              label: 'With Action Link',
+              props: {
+                onActionClick: action('action button clicked'),
+                actionText: 'View More Details',
+              },
+            },
+          ]}
+          columnProps={permutateProps({
+            Default: [{value: undefined, label: 'Default'}],
+          })}
```
anicholls, line 36: Nit: there's only one prop so you don't need to permutate
```suggestion
          columnProps={[
            {label: 'Default', props: {}}
          ]}
```

modules/toast/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,56 @@
...
+import * as React from 'react';
+import {storiesOf} from '@storybook/react';
+import {colors} from '@workday/canvas-kit-react-core';
+import {StaticStates} from '@workday/canvas-kit-labs-react-core';
+import {action} from '@storybook/addon-actions';
+import {ComponentStatesTable, permutateProps} from '../../../../utils/storybook';
+import {exclamationCircleIcon} from '@workday/canvas-system-icons-web';
+import {Toast} from '../index';
+
+const IconButtonStates = () => (
```
anicholls, line 13: Function name is wrong. We're only using it once so you might as well just return this JSX from the arrow function in the story below. Makes it easier to see what story this is when you first open the file.

Also, this fragment and div are unnecessary

modules/toast/react/lib/Toast.tsx
mannycarrera4: It's possible that a consumer can have an action link and the close button, both have `button` as the underlying dom element hence why i added a test id to the button. I could rewrite the story so there's only one focusable element in the toast, but then I'm missing out on testing a scenario where there's both a close button and a action button. There's nothing unique about the action button that I could grab it, and I didn't want to grab it by the string `View more details` because that would make the test flaky if we decided to change that text in the story. I agree that adding the test id wasn't the best but I'm open to changing it if I can somehow have both a close button and action link and grab them separately

modules/toast/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,47 @@
...
+import {jsx} from '@emotion/core';
+import {storiesOf} from '@storybook/react';
+import {colors} from '@workday/canvas-kit-react-core';
+import {StaticStates} from '@workday/canvas-kit-labs-react-core';
+import {action} from '@storybook/addon-actions';
+import {ComponentStatesTable} from '../../../../utils/storybook';
+import {exclamationCircleIcon} from '@workday/canvas-system-icons-web';
+import {Toast} from '../index';
+
+storiesOf('Components|Popups/Toast/React/Visual Testing | States', module)
```
anicholls, line 12:
```suggestion
storiesOf('Components|Popups/Toast/React/Visual Testing', module)
```

Issue #587, opened 2020-04-17
Add missing accessibility attribute to the Modal


alexandrzavalii:
## 🐛 Bug Report

- aria-labelledby is required on the role="dialog" 
- aria-modal="true" is required on the role="dialog"

NicholasBoll:
Close via #588

PR #588, opened 2020-04-17
Fix: Add missing accessibility aria-modal=true to the Modal and aria-label

alexandrzavalii:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/587

## Checklist

- [x] `yarn test` passes
      applicable

### DIFF
--- a/modules/modal/react/lib/ModalContent.tsx
+++ b/modules/modal/react/lib/ModalContent.tsx
@@ -9,6 +9,10 @@ import {ModalWidth} from './Modal';
 
 export interface ModalContentProps extends React.HTMLAttributes<HTMLDivElement> {
   /**
+   * Aria label will override aria-labelledby, it is used if there is no heading or we need custom label for popup
+   */
+  ariaLabel?: string;
+  /**
    * The padding of the Modal. Accepts `zero`, `s`, or `l`.
    * @default PopupPadding.l
    */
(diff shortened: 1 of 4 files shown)


lychyi:
@alexandrzavalii Thanks for the contribution again!

NicholasBoll:
I'll add a Cypress test for the new aria label we added.

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -215,6 +220,8 @@ const ModalContent = ({
         handleClose={handleClose}
         padding={padding}
         transformOrigin={transformOrigin}
+        aria-modal={true}
```
NicholasBoll, line 223: It looks like `aria-modal=true` is supposed to replace the functionality of setting `aria-hidden=true` on children, but isn't fully supported yet.

Do you know if this technique will work with multiple modal dialogs stacked? The `aria-hidden` should work in that case, but I'm not sure if there are multiple `aria-modal=true`. Should the previous modal get a set back to "false" or be removed?

NicholasBoll: I created a test: https://codesandbox.io/s/stacked-modal-qhcp6

alexandrzavalii: @lychyi Thanks for taking a look!:)
@NicholasBoll good question! I will ask Peter. I would assume since we have aria-hidden for the first popup aria-modal would be ignored.

NicholasBoll: Maybe that will work as long as we apply `aria-hidden` to things. I assume broad support for `aria-modal` is a ways off, so we can worry about it then.

PR #590, opened 2020-04-20
feat(button): Add href support

anicholls:
Proof of concept for adding `href`/anchor support to our button components.

Some UI buttons require an href (marketing sites, download buttons, etc.). This is in support of those use cases.

Adds a special `as` prop to all button components (excluding `Hyperlink`). When `as="a"` is defined, we switch from `React.ButtonHTMLAttributes<HTMLButtonElement>` to `React.AnchorHTMLAttributes<HTMLAnchorElement>`

Closes: https://github.com/Workday/canvas-kit/issues/87

### DIFF
--- a/modules/button/react/lib/Button.tsx
+++ b/modules/button/react/lib/Button.tsx
@@ -42,9 +42,13 @@ export interface ButtonProps
   icon?: CanvasSystemIcon;
 }
 
-function Button(props: ButtonProps): React.ReactElement;
-function Button(props: {as: 'a'} & AnchorButtonProps<ButtonProps>): React.ReactElement;
-function Button({
+type ButtonOverload = {
+  (props: ButtonProps): React.ReactElement;
+  (props: {as: 'a'} & AnchorButtonProps<ButtonProps>): React.ReactElement;
+  Variant: typeof ButtonVariant;
+  Size: typeof ButtonSize;
+};
+const Button: ButtonOverload = ({
   theme = useTheme(),
   variant = ButtonVariant.Secondary,
   size = 'medium',
@@ -53,20 +57,18 @@ function Button({
   icon,
   children,
   ...elemProps
-}: ButtonProps) {
-  return (
-    <ButtonContainer
-      colors={getButtonColors(variant, theme)}
-      size={size}
-      ref={buttonRef}
-      {...elemProps}
-    >
-      {icon && size !== 'small' && <ButtonLabelIcon size={size} icon={icon} />}
-      <ButtonLabel>{children}</ButtonLabel>
-      {dataLabel && size !== 'small' && <ButtonLabelData>{dataLabel}</ButtonLabelData>}
-    </ButtonContainer>
-  );
-}
+}: ButtonProps) => (
+  <ButtonContainer
+    colors={getButtonColors(variant, theme)}
+    size={size}
+    ref={buttonRef}
+    {...elemProps}
+  >
+    {icon && size !== 'small' && <ButtonLabelIcon size={size} icon={icon} />}
+    <ButtonLabel>{children}</ButtonLabel>
+    {dataLabel && size !== 'small' && <ButtonLabelData>{dataLabel}</ButtonLabelData>}
+  </ButtonContainer>
+);
 
 Button.Variant = ButtonVariant;
 Button.Size = ButtonSize;
--- a/modules/button/react/lib/Button.tsx
+++ b/modules/button/react/lib/Button.tsx
@@ -34,6 +34,7 @@ export interface ButtonProps
    * Note: not displayed at `small` size
    */
   icon?: CanvasSystemIcon;
+  href?: string;
 }
 
 const Button = ({
@@ -44,12 +45,15 @@ const Button = ({
   dataLabel,
   icon,
   children,
+  href,
   ...elemProps
 }: ButtonProps) => (
   <ButtonContainer
     colors={getButtonColors(variant, theme)}
     size={size}
     ref={buttonRef}
+    as={href ? 'a' : undefined}
+    href={href}
     {...elemProps}
   >
     {icon && size !== 'small' && <ButtonLabelIcon size={size} icon={icon} />}
--- a/modules/button/react/lib/Button.tsx
+++ b/modules/button/react/lib/Button.tsx
@@ -8,7 +8,7 @@ import {
   ButtonColors,
   DropdownButtonVariant,
   ButtonSize,
-  AnchorButtonProps,
+  ButtonOrAnchorComponent,
 } from './types';
 import {ButtonContainer, ButtonLabel, ButtonLabelData, ButtonLabelIcon} from './parts';
 
@@ -42,13 +42,7 @@ export interface ButtonProps
   icon?: CanvasSystemIcon;
 }
 
-type ButtonOverload = {
-  (props: ButtonProps): React.ReactElement;
-  (props: {as: 'a'} & AnchorButtonProps<ButtonProps>): React.ReactElement;
-  Variant: typeof ButtonVariant;
-  Size: typeof ButtonSize;
-};
-const Button: ButtonOverload = ({
+const Button: ButtonOrAnchorComponent<ButtonProps, typeof ButtonVariant> = ({
   theme = useTheme(),
   variant = ButtonVariant.Secondary,
   size = 'medium',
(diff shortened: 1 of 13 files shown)


modules/button/react/lib/parts/ButtonContainer.tsx
```diff
@@ -35,6 +41,7 @@ export interface ButtonContainerProps
    * This avoids using the inline `style` attribute when the shape needs to be customized (e.g. for IconButton)
    */
   extraStyles?: CSSObject;
+  as?: React.ElementType | keyof JSX.IntrinsicElements;
```
anicholls, line 44: Emotion's `as` + typescript don't like each other: https://github.com/emotion-js/emotion/issues/1137

anicholls:
I was playing with some typescript magic to try and avoid using multiple components like we did with `Avatar` and `AvatarButton` (#206), but it doesn't seem like it's possible. We have two options:

1. Use multiple components with different interfaces for `button` buttons vs. `a` buttons. We could maybe use some sort of factory pattern to make this easier to manage, but it's going to get really repetitive, or the types are going to get very nasty. My preference is to not go this route
2. Use a type intersection for the few attributes that are not shared between `button` and `a`. Here are the differences between the types which typescript would fail to catch in the future:
```tsx
interface AnchorHTMLAttributes {
  download?: any;
  href?: string;
  hrefLang?: string;
  media?: string;
  ping?: string;
  rel?: string;
  target?: string;
  type?: string;
  referrerPolicy?: string;
}

interface ButtonHTMLAttributes {
  autoFocus?: boolean;
  disabled?: boolean;
  form?: string;
  formAction?: string;
  formEncType?: string;
  formMethod?: string;
  formNoValidate?: boolean;
  formTarget?: string;
  name?: string;
  type?: 'submit' | 'reset' | 'button';
  value?: string | string[] | number;
}
```
`type` is the only thing that conflicts, and I think we would always defer to the `button` version in that case.

modules/button/react/lib/Button.tsx
```diff
@@ -34,6 +34,7 @@ export interface ButtonProps
    * Note: not displayed at `small` size
    */
   icon?: CanvasSystemIcon;
+  href?: string;
```
anicholls, line 37: Currently, this only exposes `href`, so you can't use things like `target="_blank"`. We could consider `Pick`ing some additional props from 
```tsx
interface AnchorHTMLAttributes {
  download?: any;
  href?: string;
  hrefLang?: string;
  media?: string;
  ping?: string;
  rel?: string;
  target?: string;
  type?: string;
  referrerPolicy?: string;
}
```

anicholls:
Based on a great recommendation from @NicholasBoll, I was able to use function overloading on our button functional component to define different interfaces depending on whether `as="a"` is passed or not. Typescript will now get mad about using any anchor attributes until this prop is added.

modules/button/react/lib/Button.tsx
```diff
@@ -36,7 +42,9 @@ export interface ButtonProps
   icon?: CanvasSystemIcon;
 }
 
-const Button = ({
+function Button(props: ButtonProps): React.ReactElement;
```
anicholls, line 45: This is the old/standard way to do function overloading. I explored how to do it with arrow functions as well:
```tsx
type ButtonOverload = {
  (props: ButtonProps): React.ReactElement;
  (props: {as: 'a'} & AnchorButtonProps<ButtonProps>): React.ReactElement;
  Variant: typeof ButtonVariant;
  Size: typeof ButtonSize;
};
const Button: ButtonOverload = ({
  theme = useTheme(),
  variant = ButtonVariant.Secondary,
  size = 'medium',
  buttonRef,
  dataLabel,
  icon,
  children,
  ...elemProps
}: ButtonProps) => (
  <ButtonContainer
    colors={getButtonColors(variant, theme)}
    size={size}
    ref={buttonRef}
    {...elemProps}
  >
    {icon && size !== 'small' && <ButtonLabelIcon size={size} icon={icon} />}
    <ButtonLabel>{children}</ButtonLabel>
    {dataLabel && size !== 'small' && <ButtonLabelData>{dataLabel}</ButtonLabelData>}
  </ButtonContainer>
);
```

I think I lean towards the arrow function for consistencies sake (even though you need to define the static variables again. Anyone have a preference??

NicholasBoll: I'm good with an arrow function. It seems like our docgen understands the props?

modules/button/react/lib/types.ts
```diff
@@ -92,3 +92,21 @@ export interface ButtonColors {
...
+  React.AnchorHTMLAttributes<HTMLAnchorElement>;
+
+/**
+ * Returns an overloaded functional component that uses props P by default,
+ * but uses AnchorButtonProps<P> when `as="a"`.
+ * Note: V must be `typeof ButtonVariant` since TS cannot assign enums as a type in a generic properly
+ */
+export type ButtonOrAnchorComponent<P, V = undefined> = {
+  (props: P): React.ReactElement;
+  (props: {as: 'a'} & AnchorButtonProps<P>): React.ReactElement;
```
anicholls, line 109: Not sure if this is a restriction with the way TS overloads work, but anchor attributes aren't allowed when `as` is not defined, but when it is defined, button attributes are still allowed.

modules/button/react/lib/Button.tsx
```diff
@@ -34,9 +40,14 @@ export interface ButtonProps
    * Note: not displayed at `small` size
    */
   icon?: CanvasSystemIcon;
+  /**
+   * The alternative container type for the button. Uses Emotion's special `as` prop.
+   * Will render an `a` tag instead of a `button` when defined.
+   */
+  as?: 'a' | undefined;
```
NicholasBoll, line 47: Do we need the `| undefined`? The `?` optional already adds `| undefined`

modules/button/react/lib/Button.tsx
```diff
@@ -34,9 +40,14 @@ export interface ButtonProps
...
 }
 
-const Button = ({
+const Button: ButtonOrAnchorComponent<ButtonProps, typeof ButtonVariant> = ({
```
NicholasBoll, line 50: A downside to this approach is react-docgen gets very confused about this type signature and doesn't do any prop table. We can re-evaluate when we look at documentation solutions.

modules/button/react/lib/types.ts
NicholasBoll: That's interesting. I don't get that behavior: [Example](https://www.typescriptlang.org/play/?jsx=2#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wChSkA7AVzwCFqYYJKA1FKYFS+Ab1LhwACpxAcAnnAC8BMKIlkAvuSq04AERxgAJhADulBkxbtO3PgLgAZDgHMk0ggBs7SJeVLAeSKJnQOjZkoRCDAAZzgkAA8YKm0I5HQYADpAlgAJABUAWSsAQSZOACNGJDCAHizctMoAUSckECoYAD44fkEANw4uHgAuOBrTXphSZU9vX38NLV0DGpDwyJi4iIB5EGAYcoWtMIAaAm6zHnw2jrhjkYHNULnDRiDh8zHyGHEwBzzKNAALaF2oQqdDaMg2Wx2hwA1khxBBMIhiCkalV8oVgCVYhVUTV6o1mi02gAyRFJZLfP7QVEFGDFUrYnL5H7-KB4po8FoqKKQWBwd6fQaPFhrTo+JwQFDaSFwVig9qWAAUciBAzoAEoBokMMktTA2c1FcrwgNeHAUGEBvgUPg4Io4CSKSzAeEdi0NaTtbr9TxLM9+jLXuQ0CwwvBbjp9A9jJRVULKCKxRKpeH7s6DnyPkh4TM7pGhj1zHKFWrpG1ytpgJ04AB6TkpvNx5J++AyevzOPN8gNeCxcCkXtgRzlNtRoJwTDQEBSABEE4g08uBZ4Ul4I5qTaXKRsUHsihaADEIBA6BxytW13HOQOhxfo+PJzO5wvzTOUAuruYV7eghuTlvXHuh7Hqe56zA20acv2jSDjIw5ge2d6-EQmCPke76bl+8Gjiwv4jMk267geR4nlAZ7fiwV7QTeWE1HASFIChs5oWaYSvuhf6YbmCE-s2+EAURwGkaBXHYZQLRAA). Perhaps the TS version is causing this difference?

JakeGinnivan:
I just saw this linked from an emotion issue. Not sure if you have seen the `withComponent('a')` function.

For example `const OutlineButtonLink = OutlineButton.withComponent('a')` should do what you are after, including dropping button only props and switching to a props.

This is just an FYI if you hadn't seen that.

Issue #591, opened 2020-04-21
Improve theming API stability


anicholls:
## 🐛 Bug Report

Currently our themed components only work if consumers provide a full Canvas theme or no theme at all (reverts to default). We require consumers to wrap a custom theme in `createCanvasTheme()` to fill out the default fields. If this is omitted, or CK components are used inside an arbitrary Emotion `ThemeContext`, they will generate a runtime error (see codesandbox below)

Currently we're assuming that [if the `ThemeContext` is not an empty object, that the theme fields from `CanvasTheme` will always be defined](https://github.com/Workday/canvas-kit/blob/master/modules/_labs/core/react/lib/theming/useTheme.ts#L26). We should be falling back to the default theme value when any of these fields doesn't exist.  `useTheme` will only return the default fields if the theme object is empty. If it has fields, but is not formatted like `CanvasTheme` (e.g. `{ nothing: ["red", "blue"] }`), we do nothing, and when we try to access expected fields it fails. 

Emotion's `ThemeProvider` and `ThemeContext` are basically named React context components. Based on this, [previously the Emotion team had indicated that they plan to do away with their theming support in future Emotion versions](https://github.com/emotion-js/emotion/pull/973), encouraging users to roll their own. It seems like that narrative has changed and based on the [current state of theming in Emotion 11](https://github.com/emotion-js/emotion/blob/next/packages/react/src/theming.js), it doesn't look like it's going anywhere. We had previously discussed rolling our own theme provider, but it looks like we should continue supporting the emotion one for now. That being said, we can run into conflicts if there are multiple Emotion contexts running at the same time (e.g. two different versions of emotion). I'm not sure what the impact of this is, but they [added a warning for it in Emotion v11](https://github.com/emotion-js/emotion/issues/1470). I would love some input on whether we should stick with Emotion's `ThemeProvider` or roll our own.

There have been some concerns about merging (whether realistically or effectively) the Canvas theme object with any custom theme object. We would obviously avoid overwriting any existing fields, but this is unavoidable unless we create our own theme context/provider.

## Examples

- https://codesandbox.io/s/agitated-cookies-y6kmy?file=/src/index.tsx 
- https://codesandbox.io/s/stacked-modal-1pnit?file=/src/index.tsx

## Required updates

To avoid runtime errors like these, I see two different options:

[...]

PR #593, opened 2020-04-22
fix(common): Improve theming API stability

anicholls:
To ensure all theme inputs are of the format we expect (to avoid runtime errors when accessing theme fields), we now call an improved (and memoized) version of createCanvasTheme() inside of useTheme() to fill out any expected fields that have not been defined. This results in every theme object having all expected fields defined. This internalizes the usage of `createCanvasTheme()` and it's no longer a requirement for consumers to call this. We are still "merging" the consumers custom theme with the default canvas theme, but it only happens within the context of our components.

There have been some concerns about merging the Canvas theme object with any custom theme object. We obviously avoid overwriting any existing fields, but this is unavoidable unless we create our own theme context/provider. Due to some restraints to how Emotion’s styled function works (and our wrapper around it), I decided not to do this for now (more below). We’ve internalized the `createCanvasTheme` part of the API so that consumers reference to their custom theme is never changed. However, we do need to merge the two themes within our Canvas Components. This is done behind the scenes, in a way where the custom theme fields always take precedence.

I looked into using a custom context, but the concern we have is that there’s no way for us to hi-jack Emotion’s `styled` function, which means we would need to wrap every single styled component in Canvas Kit in another context consumer (or we’d have to manually manage one around every exported component). It also means we can’t use the `theme` prop as this would be overridden by Emotion and we would have to use something else. This PR addresses the runtime errors noted in #591 and adds some other improvements, but I think we’ll have to do some broader “canvasing” to figure out if a separate theme context (no theme merging) is a hard requirement.

For future reference (a case against merging): https://github.com/Workday/canvas-kit/issues/591#issuecomment-618733358

Finally, after continued concern around merging the Canvas theme object, we decided to namespace it. So now everything has been moved within `theme.canvas`. This is a breaking change.

Closes: https://github.com/Workday/canvas-kit/issues/591

### Breaking Changes

- The `theme` object now needs to be namespaced under `canvas`:
```tsx
{
  theme: {
    canvas: {
      palette: { 
        // ...
      },
      breakpoints: { 
        // ...
      },
      direction: ContentDirection.LTR
    }
  }
}
```
- This means several type references have changed:
  - `CanvasTheme` > `EmotionCanvasTheme`
  - `PartialCanvasTheme` > `PartialEmotionCanvasTheme`


modules/_labs/core/react/lib/theming/createCanvasTheme.ts
```diff
@@ -78,8 +79,8 @@ function fillPalette(palette?: PartialCanvasThemePalette): CanvasThemePalette |
   };
 }
 
-export function createCanvasTheme(partialTheme: PartialCanvasTheme): CanvasTheme {
-  const {palette = {}, breakpoints = {}, direction} = partialTheme;
+function calculateCanvasTheme(partialTheme: PartialCanvasTheme): CanvasTheme {
```
alexandrzavalii, line 82: does the type PartialCanvasTheme mean that the fields that we are passing should be one of CanvasTheme?

anicholls: This type checks any fields that are in common with `CanvasTheme` (i.e. it will complain about `direction: 'foo'`), but it will still allow extra fields.

modules/_labs/core/react/lib/theming/createCanvasTheme.ts
```diff
@@ -50,50 +51,60 @@ function shiftColor(hexColor: string, direction: ColorDirection) {
...
+      error: fillPalette(defaultCanvasTheme.palette.error, error),
+      success: fillPalette(defaultCanvasTheme.palette.success, success),
+      neutral: fillPalette(defaultCanvasTheme.palette.neutral, neutral),
     },
     breakpoints,
     direction: direction === ContentDirection.RTL ? direction : ContentDirection.LTR,
   };
 
-  return merge({}, defaultCanvasTheme, mergeable) as CanvasTheme;
+  return merge({}, defaultCanvasTheme, mergeable, extraFields) as CanvasTheme;
```
alexandrzavalii, line 107: as a quick fix what if we would keep canvas theme in a nested field of provider?
```
const thene = {
  canvasTheme: { paletter, breakpoints, direction},
  ...ourProjectLevelTheme
}
```

and then we could customize useTheme function to always return canvasTheme
```
 useCanvasTheme = () => {
    const theme = useTheme():
   return theme.canvasTheme
}
```

anicholls: hah! Simple but I like it. I will run it by the team and see what they think.

package.json
```diff
@@ -99,7 +99,7 @@
     "postcss-discard-duplicates": "^4.0.2",
     "postcss-inline-svg": "^3.1.1",
     "postcss-loader": "^2.1.4",
-    "prettier": "^1.12.1",
+    "prettier": "^2.0.5",
```
anicholls, line 102: Required to support optional chaining. It was technically added in 1.19.0 but I figured why not 🤷‍♂️ . Someone tell me if you disagree.

anicholls: Ended up just using 1.19.1

modules/common/react/lib/theming/styled.ts
```diff
@@ -28,9 +28,9 @@ export const convertToStaticStates = (obj?: CSSObject): CSSObject | undefined =>
 function styled<Props>(node: any) {
   return (...args: Interpolation<Props>[]) => {
     const newArgs: Interpolations = args.map(
-      interpolation => (props: Props & {theme: CanvasTheme & {_staticStates?: boolean}}) => {
+      interpolation => (props: Props & {theme: EmotionCanvasTheme & {_staticStates?: boolean}}) => {
```
anicholls, line 31: Now that we've updated the theme to be namespaced under `canvas`, we obviously have to add the extra key every time we access a theme value (i.e. `theme.canvas.palette.primary.main`).

I tried to do some magic so that our components were only passed the `canvas` object for their theme. This function started to get quite messy (since it could be called with `theme.canvas.palette` or `theme.palette`), and it became confusing to know which version of the theme you were using in certain places. This is a bit more verbose, but I think it's clearer when there's less magic going on.

Would like some input on this.

anicholls:
Now that we've updated the theme to be namespaced under `canvas`, we obviously have to add the extra key every time we access a theme value (i.e. `theme.canvas.palette.primary.main`).

I tried to do some magic so that our components were only passed the `canvas` object for their theme. However, the `styled` function started to get quite messy (since it could be called with `theme.canvas.palette` or `theme.palette`), and it became confusing to know which version of the theme you were using in certain places. This is a bit more verbose, but I think it's clearer when there's less magic going on.

Would like some input on this.

modules/switch/react/lib/Switch.tsx
```diff
@@ -126,7 +126,9 @@ const SwitchBackground = styled('div')<Pick<SwitchProps, 'checked' | 'disabled'>
   ({checked, disabled, theme}) => {
```
alexandrzavalii, line 126: is there a reason of not using destructuring from theme? 
```
({checked, disabled, theme: {canvas:{palette}}}) => {
 backgroundColor: paletter.primary.light
 }
```

anicholls: Just pushed an update with destructuring. I couldn't do it everywhere because some of our helper functions rely on the `theme` object. Was able to update it in most places though! 👍

PR #596, opened 2020-04-23
chore: Update Cypress to 4.4.1 and remove custom TS handling

NicholasBoll:
## Summary

Cypress recently added some nice features including the ability to handle Typescript files without any extra configuration. They also now include nicer stack traces in error messages (that map to our code vs Cypress internals).

This ends up removing extra dependencies, helping maintain my position as removing more code than I add to the codebase.

https://docs.cypress.io/guides/references/changelog.html


cypress/plugins/index.js
```diff
@@ -1,24 +1,3 @@
...
-                loader: 'ts-loader',
-                options: {transpileOnly: true},
-              },
-            ],
-          },
-        ],
-      },
-    },
-  };
-  on('file:preprocessor', wp(options));
```
mannycarrera4, line 23: do we not need this anymore?

mannycarrera4: nvm, just saw the PR description, nice!

NicholasBoll: Yeah. The file needs to stick around and export something, otherwise Cypress complains, Leaving the function block empty causes ESLint to complain. So there is just a comment now

PR #599, opened 2020-04-27
fix: Fix ESLint VSCode plugin

NicholasBoll:
Fixes this issue for us: https://github.com/typescript-eslint/typescript-eslint/issues/967

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -23,7 +23,7 @@ module.exports = {
     ecmaFeatures: {
       jsx: true,
     },
-    project: ['tsconfig.json', 'cypress/tsconfig.json'],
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
     ecmaFeatures: {
       jsx: true,
     },
-    project: ['tsconfig.json', 'cypress/tsconfig.json'],
+    project: ['./tsconfig.json', './cypress/tsconfig.json'],
```
NicholasBoll, line 26: This is a subtle change, but it seems that the CLI understands, but the VSCode plugin has an issue

PR #600, opened 2020-04-27
test(popup): Add tests for Popup

mannycarrera4:
## Summary

Adds visual and cypress test for Popup and change enzyme test to react testing library

### DIFF
--- a/cypress/integration/Popup.spec.ts
+++ b/cypress/integration/Popup.spec.ts
@@ -38,7 +38,7 @@ describe('Popup', () => {
         getPopupTargetButton().click();
       });
 
-      it('should open the modal', () => {
+      it('should open the popup', () => {
         getPopup().should('be.visible');
       });
 
@@ -52,7 +52,7 @@ describe('Popup', () => {
         });
 
         it('should have an aria-labelledby attribute', () => {
-          h.modal.get().should('have.attr', 'aria-labelledby');
+          getPopup().should('have.attr', 'aria-labelledby');
         });
       });
 
@@ -67,24 +67,4 @@ describe('Popup', () => {
       });
     });
   });
-
-  //   ['Error', 'Success'].forEach(story => {
-  //     context(`given the '${story}' story is rendered`, () => {
-  //       beforeEach(() => {
-  //         h.stories.load('Components|Popups/Toast/React', story);
-  //       });
-  //       it('should not have any axe errors', () => {
-  //         cy.checkA11y();
-  //       });
-  //     });
-  //   });
-
-  //   context(`given the toast with a close button  and action button`, () => {
-  //     beforeEach(() => {
-  //       h.stories.load('Components|Popups/Toast/React', 'With action link and close icon');
-  //     });
-  //     it('should have a role of dialog', () => {
-  //       getDialogToast().should('have.attr', 'role', 'dialog');
-  //     });
-  //   });
 });
--- a/cypress/integration/Popup.spec.ts
+++ b/cypress/integration/Popup.spec.ts
@@ -51,7 +51,7 @@ describe('Popup', () => {
           getPopup().should('have.attr', 'role', 'dialog');
         });
 
-        it('should have an aria-labelledby attribute', () => {
+        it('should have an aria-labelledby attribute when a heading is provided', () => {
           getPopup().should('have.attr', 'aria-labelledby');
         });
       });
--- a/cypress/integration/Popup.spec.ts
+++ b/cypress/integration/Popup.spec.ts
@@ -4,10 +4,70 @@ function getPopup() {
   return cy.findByRole('dialog');
 }
 
+function getPopupTargetButton() {
+  return cy.contains('Delete Item');
+}
+
+function getCloseButton() {
+  return getPopup().find('[data-close]');
+}
+
 describe('Popup', () => {
-  //   before(() => {
-  //     h.stories.visit();
-  //   });
+  before(() => {
+    h.stories.visit();
+  });
+  context(`given the open popup`, () => {
+    beforeEach(() => {
+      h.stories.load('Components|Popups/Popup/React', 'Open');
+    });
+    it('should not have any axe errors', () => {
+      cy.checkA11y();
+    });
+
+    it('should have a role of status', () => {
+      getPopup().should('have.attr', 'role', 'dialog');
+    });
+  });
+
+  context(`given the default popup`, () => {
+    beforeEach(() => {
+      h.stories.load('Components|Popups/Popup/React', 'Default');
+    });
+    context('when the target buttton is clicked', () => {
+      beforeEach(() => {
+        getPopupTargetButton().click();
+      });
+
+      it('should open the modal', () => {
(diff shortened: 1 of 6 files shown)


modules/popup/react/lib/Popup.tsx
```diff
@@ -152,7 +152,6 @@ export default class Popup extends React.Component<PopupProps> {
               size={closeIconSize}
               onClick={handleClose}
               icon={xIcon}
-              title={closeLabel}
```
mannycarrera4, line 155: We already have an aria label which is why removed the title, thoughts?

modules/toast/react/spec/Toast.spec.tsx
```diff
@@ -41,8 +41,8 @@ describe('Toast', () => {
           {toastMessage}
         </Toast>
       );
-      const closeIcon = container.querySelector('[data-close="close"]'); /*? */
-      fireEvent.click(closeIcon); /*? */
```
mannycarrera4, line 45: this was some  comments that where left, seemed minor enough to remove

cypress/integration/Popup.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+  });
+  context(`given the open popup`, () => {
+    beforeEach(() => {
+      h.stories.load('Components|Popups/Popup/React', 'Open');
+    });
+    it('should not have any axe errors', () => {
+      cy.checkA11y();
+    });
+
+    it('should have a role of status', () => {
```
mannycarrera4, line 27: change to dialog

cypress/integration/Popup.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+  context(`given the default popup`, () => {
+    beforeEach(() => {
+      h.stories.load('Components|Popups/Popup/React', 'Default');
+    });
+    context('when the target buttton is clicked', () => {
+      beforeEach(() => {
+        getPopupTargetButton().click();
+      });
+
+      it('should open the modal', () => {
```
mannycarrera4, line 41: change to popup

cypress/integration/Popup.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  return cy.findByRole('dialog');
+}
+
+function getPopupTargetButton() {
+  return cy.contains('Delete Item');
+}
+
+function getCloseButton() {
+  return getPopup().find('[data-close]');
+}
```
mannycarrera4, line 13: this function and the one above are similar to the modal helper functions, should I use those instead?

modules/popup/react/package.json
```diff
@@ -35,6 +35,9 @@
     "workday",
     "popup"
   ],
+  "devDependencies": {
+    "@workday/canvas-kit-labs-react-core": "^3.6.0"
+  },
```
mannycarrera4, line 40: used for stories using static states component

mannycarrera4:
Some things to note about our Popup. Our popup is more of a styled container, our Modal component has more accessibility and functionality than our popup. After talking to Nicholas for a bit, this component seems more like a dialog container. I wrote some minimal cypress tests and wanted to get some clarification if they make sense @NicholasBoll

modules/popup/react/lib/Popup.tsx
```diff
@@ -26,7 +26,7 @@ export interface PopupProps extends React.HTMLAttributes<HTMLDivElement> {
    * The origin from which the Popup will animate.
    * @default {horizontal: 'center', vertical: 'top'}
    */
-  transformOrigin: TransformOrigin;
+  transformOrigin: TransformOrigin | null;
```
mannycarrera4, line 29: this allows to pass null in the visual stories and doesn't add the animation allowing for the snap shots to be captured

cypress/integration/Popup.spec.ts
NicholasBoll: Hmm. If I were to share some helpers I might do the opposite since `Popup` is what's common between `Popup` and `Modal`. But that's an implementation detail of the Modal. Modal uses shared helpers - those helper functions are meant to be usable eventually outside of this repository. Non-sharable helper functions should remain inside the spec file IMO.

These helper functions are different than the modal ones. They don't map as closely to the Cypress command log - these are more like "macro helpers" than just call to other commands and don't have their own logging. This isn't a huge problem on simpler tests, but can really aid in debugging larger tests.

modules/popup/react/lib/Popup.tsx
NicholasBoll: I think it was added so that a mouse user could see what the Icon would do? I agree we don't want both `title` and `aria-label`. Perhaps this should be a label Tooltip? The new Tooltip API only exists on the v4 branch, but would be what we want.

It would look something like:
```tsx
<Tooltip title={closeLabel}>
  <IconButton
    /* ... */
  />

The label is actually required on the tooltip as well as the `IconButton` because of Typescript. The Tooltip uses `React.cloneElement` and adds an `aria-label` to the child it wraps.
</Tooltip>

File: modules/popup/react/spec/Popup.spec.tsx [hunk 8]
```diff
@@ -1,25 +1,84 @@
...
+      const closeButtonAriaLabel = 'close';
+      const {getAllByRole} = render(
+        <Popup closeLabel={closeButtonAriaLabel} handleClose={cb}>
+          <div>Are you sure you'd like to delete the item titled 'My Item'?</div>
+
+          <button onClick={cb}>Delete</button>
+        </Popup>
+      );
+
+      expect(getAllByRole('button')[0]).toHaveAttribute('aria-label', closeButtonAriaLabel);
```
NicholasBoll, line 36: It looks like we added `data-close` as an attribute to help with selecting. Would that be a safer and more helpful selector here?

modules/popup/react/spec/Popup.spec.tsx
```diff
@@ -1,25 +1,84 @@
...
+    it('should call the handleClose callback when clicked', () => {
+      const closeButtonAriaLabel = 'close';
+      const {getAllByRole} = render(
+        <Popup closeLabel={closeButtonAriaLabel} handleClose={cb}>
+          <div>Are you sure you'd like to delete the item titled 'My Item'?</div>
+
+          <button onClick={cb}>Delete</button>
+        </Popup>
+      );
+      fireEvent.click(getAllByRole('button')[0]);
```
NicholasBoll, line 48: Same here

modules/toast/react/spec/Toast.spec.tsx
NicholasBoll: I think these are also removed in v4

modules/popup/react/spec/Popup.spec.tsx
mannycarrera4: i think that makes sense, I can make that change!

cypress/integration/Popup.spec.ts
mannycarrera4: ah gotcha, thanks for explaining that!

PR #606, opened 2020-05-01
test(select): Fix visual testing states for options

jamesfan:
Adds missing visual testing states for selected and assistive-focused (keyboard-focused) options.


modules/_labs/select/react/stories/stories.tsx
```diff
@@ -316,6 +323,7 @@ storiesOf('Labs|Select/React/Visual Testing', module)
           <SelectOption
             {...props}
             onChange={() => {}} // eslint-disable-line no-empty-function
+            style={{width: 100}}
```
lychyi, line 326: Curious what others think. 🤔 

I don't think it's wise to shorten the width here. That's against the grain of visually testing our components the way they're intended to render. This is like testing a specific style override.

I get that it's really wide but I would probably favor adding a second static state table below this one that just shows assistive focus if that's really what this line is mitigating.

anicholls: [We already do it on `TextInput`](https://github.com/Workday/canvas-kit/blob/master/modules/form-field/react/stories/stories_TextInput.tsx#L205) 🤷‍♂️ 

Not a big deal IMO.

NicholasBoll: Hmm. That is interesting. I'm okay with it on our interactive state tables. We should have some visual baselines of the component though. Like on Buttons - small, medium, large, small-grow, medium-grow, large-grow, etc. Those would not modify things like `width`.

jamesfan: Originally, I added that width to force all of the options to render at the same width since I noticed the more lengthy column headers were making some columns wider than others, which in turn made the options different widths (since they expanded to fill their parent). I thought it was strange to have some options wider than others. However, I just tried removing the `width: 100` and everything looks fine now, which makes me wonder if I had longer column headers at some point while I was finalizing the wording... so I may be able to remove that width anyway (though if we remove it here, we should also remove the `width` in the `States` story as well).

Typically, the option will render at the same width as the select button (`280px` or `100%` if `grow` is `true`). Here, it's just expanding to fill the width of the table cell. Incidentally, I'm just now wondering if it's proper to be rendering an `<li>` (the option is a styled `<li>`) without the parent `<ul>`... 🤔

lychyi: @anicholls I didn't know this was done elsewhere, I still don't think it's good practice to modify the style of something we're visually testing as our baseline. So I don't think while it may not be a big deal, I think it makes testing less rigorous. Like imagine if we wrote some styles that assumed that `width` wasn't modified to 100 but we force it to do that here?

I would rather find a way to modify the size of the container element rather than the component inside. But taking @NicholasBoll suggestion, having a visual baseline that we test is a good alternative also. It at least visually covers the default SelectOption.

jamesfan: I've removed the inline styles from `Select` and `SelectOption`. I also wrapped the `SelectOption` in a `<ul>` with a width of `120px` to keep the table at a reasonable width.

NicholasBoll:
@jamesfan I noticed the Select Options visual tests included Error and Alert of the parent Select state. Do we ever anticipate the options reflecting these changes? If not, we should remove the noise.

https://5d854c26ba934e0020f5e98a-onhoojqnbb.chromatic.com/?path=/story/labs-select-react-visual-testing--states-option

NicholasBoll:
There is also a lot of words to each state in the Select options visual table. It is very hard to understand that table at all. Maybe we want to stick to interaction states for the columns (hover, active, etc) and selected states or something else for the rows. Something to make that table easier to understand

jamesfan:
> @jamesfan I noticed the Select Options visual tests included Error and Alert of the parent Select state. Do we ever anticipate the options reflecting these changes? If not, we should remove the noise.
> 
> https://5d854c26ba934e0020f5e98a-onhoojqnbb.chromatic.com/?path=/story/labs-select-react-visual-testing--states-option

@NicholasBoll The Alert state of the option is _slightly_ different from the Default and Error states -- if you look really closely, you'll the text is shifted to the left by one pixel. This is due to the extra pixel of width in the two-color border of a Select in the Alert state: the text in the option has to move one pixel to the left to compensate in order to maintain alignment with the text in the button.

That said, I agree the table is pretty hard to understand in its current state. I'll see what I can do, maybe revisit Lynn's suggestion to break the table apart.

jamesfan:
Still working on this, btw. Don't merge it in quite yet ;).

modules/_labs/select/react/stories/stories.tsx
```diff
@@ -296,29 +295,38 @@ storiesOf('Labs|Select/React/Visual Testing', module)
...
-            if (props.disabled && !['', 'hover'].includes(props.className)) {
+            if (
+              props.disabled &&
+              (props['aria-selected'] === 'true' ||
+                !['', 'hover'].includes(props.className) ||
+                props.focused)
+            ) {
               return false;
             }
             return true;
```
NicholasBoll, line 317: The downside of using `permuateProps` is a very difficult to understand prop filtering. I think this means to not show column (unclear why column and not row) if the the prop merged-matrix has both `disabled` AND either `aria-selected` or default/hover classNames or focused.

NicholasBoll:
@jamesfan Are you still working on this? If so, remove the "ready for review" label.

jamesfan:
Does anyone mind if I change the base of this PR to `prerelease/v4`? We're making modifications to the stories in #648, and I feel like we could avoid some merge issues if I base off of `prerelease` instead of `master`.

modules/_labs/select/react/stories/stories.tsx
jamesfan: That's correct, although FWIW I'm no longer using `permutateProps` in the latest version of the states table. Given the smaller number of combinations for the columns and rows in the new tables, I didn't think it was necessary to use `permutateProps`.

jamesfan:
I wrestled for a while with how to best break this table apart. Before I get into the particulars, here's the [old options table](https://5d854c26ba934e0020f5e98a-onhoojqnbb.chromatic.com/?path=/story/labs-select-react-visual-testing--states-option) for reference.

First off, I've created Groups for each of the three error states: `Default` (no error), `Alert`, and `Error`. Each Group is outlined with a gray border. Currently, all three Groups are identical with the exception of a _very_ slight `1px` difference in the padding for the `Alert` Group ([more info](#issuecomment-623769247)). To Nicholas' point, we could remove the `Alert` and `Error` Groups to simplify this story, especially since any visual regressions related to the error states are likely be caught in the `States (Menu On)` story.

That said, each Group contains two tables: one table for `Disabled` states and another table for `Interaction` states (everything else).

* For the `Interaction` table, I chose to represent `Hover/Active` along the columns and `Assistive-Focus/Selected` along the rows.
* I created a separate `Disabled` table because the `Disabled` state was incompatible with values in the `Interaction` columns (`Active`) **AND** the `Interaction` rows (`Assistive-Focus` and `Selected`). Leaving `Disabled` in the `Interaction` columns would have resulted in invalid combinations in some cells (e.g., `Assistive-Focus` x `Disabled`). A separate `Disabled` table seemed like the cleanest solution.

**Questions:**
1. Thoughts on this new layout? I think it's a solid improvement over the previous mega-table, but I'm curious to get an outside perspective.
2. Any opinions on removing the `Alert` and `Error` Groups? Leave them in for technical completeness, or remove them to simplify things and rely on `States (Menu On)` to catch visual regressions?
3. And, finally, if everything else looks good, happy to listen to feedback on the styling of the Groups as well.

NicholasBoll:
I personally like the improvements to the visual table. I think the organization better explains the relationship between the snapshot and what is expected.

As for the Alert and Error groups, I think it makes sense to remove them and rely on the Menu On visual snapshot. My worry is without the context of knowing that Alert adds a `1px` border, they look the same and I'm not sure how to compare any differences. By contrast, the Menu On story shows the options in full context of the Menu so a regression of the border is more obvious and easier to compare.

I don't think we should just have all states captured for completeness (similar to code coverage issues), but have visual states captured in a way that someone can review them. When we or designers evaluate the diffs, additional context is very helpful and should be preferred.

modules/_labs/select/react/stories/stories.tsx
```diff
@@ -287,39 +286,62 @@ storiesOf('Labs|Select/React/Visual Testing', module)
...
-        )}
-      </ComponentStatesTable>
-    </StaticStates>
+    <div>
+      {[
+        {label: 'Default', error: undefined},
+        {label: 'Alert', error: Select.ErrorType.Alert},
+        {label: 'Error', error: Select.ErrorType.Error},
+      ].map(errorTable => (
+        <div style={{border: `1px solid ${colors.soap600}`, marginTop: 30, padding: '0 20px'}}>
```
anicholls, line 295: We could drop this border if we're going to remove the error and alert tables

PR #609, opened 2020-05-01
fix: Add type checking to PRs and fix type errors

NicholasBoll:
We accidentally removed any TS type checks on PRs when we moved to babel in the webpack config. TS errors are now only caught on `master` and `prerelease` builds, but that is too late and not visible. This change adds type checks back by making it part of the CI process. I also fixed Typescript errors accidentally introduced.


.eslintrc.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
     ecmaFeatures: {
       jsx: true,
     },
-    project: ['./tsconfig.json', './cypress/tsconfig.json'],
+    project: ['./tsconfig.eslint.json', './cypress/tsconfig.json'],
```
NicholasBoll, line 26: Specific eslint config because we ignore `spec` and `stories` directories for Typescript, but want to cover them with linting... Perhaps we'll revisit that when we move our build steps to using Babel instead of Typescript. That way we can type check all Typescript files.

.eslintrc.js
```diff
@@ -71,13 +71,13 @@ module.exports = {
     'no-undef-init': 'error',
     'no-unused-expressions': 'error',
     'no-unused-labels': 'error',
-    'no-use-before-define': ['warn', {functions: false, classes: true}],
+    'no-use-before-define': 'off', // TS takes care of this one...
```
NicholasBoll, line 74: Typescript ensures this is safe. Do we have a reason to have a lint rule? This lint rule triggers on completely safe patterns like:

```ts
const getFoo() => {
  return getBar()
}

const getBar() => {
  return 'bar';
}

File: .eslintrc.js [hunk 3]
```diff
@@ -71,13 +71,13 @@ module.exports = {
...
     'no-var': 'error',
     'prefer-const': 'error',
     'space-before-function-paren': 'off',
     'react/jsx-no-bind': 'off', // Keep perf implications in mind, but was giving too many warnings and hurting readability
     curly: 'error',
     radix: 'error',
-    'prettier/prettier': ['error', prettierConfig],
+    'prettier/prettier': ['warn', prettierConfig],
```
NicholasBoll, line 81: I switched this to `warn`. I'd prefer to turn these rules off entirely since formatting is automatic anyways. I tried just running `yarn format:prettier`, but it matched more files than with just `yarn format` (which runs `eslint --fix`) so I get why this was enabled. The extra files were just some JSON files and MD files. I think that'd be okay to do and remove this rule entirely.

It just happens way too often then while I'm typing I get red squigglies and instantly think the issue is a type error. It happens while screen sharing as well. This wasn't a problem until I fixed this ESLint config file to be picked up by the ESLint plugin in VSCode: https://github.com/Workday/canvas-kit/pull/599

.github/workflows/ci.yml
```diff
@@ -45,6 +45,9 @@ jobs:
       - name: Dependency Check
         run: yarn depcheck
 
+      - name: Type Check
+        run: yarn typecheck
+
```
NicholasBoll, line 50: This adds type checking to the PRs

cypress/support/commands.ts
```diff
@@ -95,13 +94,17 @@ export const haveAriaLabel = (text: string) => ($target: JQuery) => {
...
+};
+
 Cypress.Commands.overwrite('should', (originalFn, subject, expectation, ...args) => {
   const customMatchers = {
     'have.ariaDescription': haveAriaDescription(args[0]),
     'have.ariaLabel': haveAriaLabel(args[0]),
   };
   // See if the expectation is a string and if it is a member of Jest's expect
-  if (typeof expectation === 'string' && customMatchers[expectation]) {
+  if (isKeyOf(customMatchers, expectation)) {
```
NicholasBoll, line 107: I enabled strict Typescript checking on the cypress directory. This helps tell typescript that `expectation` is a valid key of `customMatchers`

cypress/tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
   "compilerOptions": {
+    "lib": ["es2015", "es2017", "dom"],
+    "strict": true,
```
NicholasBoll, line 4: Heh. Ooops. We never really had type checking on Cypress files... I had to add libs to get Typescript to pass on Cypress types since they use types from the dom lib.

modules/button/react/lib/DeleteButton.tsx
```diff
@@ -56,5 +56,7 @@ const DeleteButton: ButtonOrAnchorComponent<DeleteButtonProps> = ({
 );
 
 DeleteButton.Size = ButtonSize;
+// @ts-ignore ButtonOrAnchorComponent requires a Variant, but will complain Variable 'Variant' implicitly has an 'any' type. ts(7005)
+DeleteButton.Variant = undefined;
```
NicholasBoll, line 60: I had to make `Variant` required in order for things like `Button.Variant.Small` to not throw "possibly undefined" errors. Rather than defining a brand new interface that didn't include `Variant`, I opted for the ability to set `V` to `undefined`.

Without this line, Typescript will thrown an error on the `const DeleteButton` line saying `Variant` is missing. So I have to set it. If I set it to anything other than `undefined`, Typescript will tell me `'foo' is not compatible with undefined`. If I set it to `undefined`, it will say `Variant` is implicitly `any`... This sets it to make Typescript happy, but disables the implicit any error...

modules/button/react/lib/IconButton.tsx
```diff
@@ -46,24 +46,11 @@ export interface IconButtonProps extends React.ButtonHTMLAttributes<HTMLButtonEl
...
-  Variant: typeof IconButtonVariant;
-  Size: Partial<typeof ButtonSize>;
-};
-
-const IconButton: ButtonOrAnchorComponent = ({
+const IconButton: ButtonOrAnchorComponent<
+  IconButtonProps,
+  typeof IconButtonVariant,
+  'aria-label'
+> = ({
```
NicholasBoll, line 53: I thought this was better than defining a new type

modules/button/react/lib/types.ts
```diff
@@ -96,17 +96,22 @@ export interface ButtonColors {
...
+ *
+ * Note: `V` must be `typeof ButtonVariant` since TS cannot assign enums as a type in a generic properly
  */
-export type ButtonOrAnchorComponent<P, V = undefined> = {
+export type ButtonOrAnchorComponent<P, V = undefined, K extends keyof P = never> = {
+  (props: {as: 'a'} & AnchorButtonProps<P, K>): React.ReactElement;
   (props: P): React.ReactElement;
-  (props: {as: 'a'} & AnchorButtonProps<P>): React.ReactElement;
-  Variant?: V;
+  Variant: V;
```
NicholasBoll, line 115: I had to make `Variant` required in order for things like `Button.Variant.Small` to not throw "possibly undefined" errors. Rather than defining a brand new interface that didn't include `Variant`, I opted for the ability to set `V` to `undefined`.

modules/button/react/lib/types.ts
```diff
@@ -96,17 +96,22 @@ export interface ButtonColors {
...
+ * but uses `AnchorButtonProps<P, K>` when `as="a"`.
+ * `K` is a key of `P`
+ *
+ * Note: `V` must be `typeof ButtonVariant` since TS cannot assign enums as a type in a generic properly
  */
-export type ButtonOrAnchorComponent<P, V = undefined> = {
+export type ButtonOrAnchorComponent<P, V = undefined, K extends keyof P = never> = {
+  (props: {as: 'a'} & AnchorButtonProps<P, K>): React.ReactElement;
   (props: P): React.ReactElement;
-  (props: {as: 'a'} & AnchorButtonProps<P>): React.ReactElement;
```
NicholasBoll, line 109: I swapped these lines since Typescript seems to always pick the last override when using `styled(IconButton)`. No idea why...

modules/button/react/lib/types.ts
```diff
@@ -96,17 +96,22 @@ export interface ButtonColors {
 /**
  * Used to get the props of the anchor version of a button
  */
-export type AnchorButtonProps<P> = Omit<P, keyof React.ButtonHTMLAttributes<HTMLButtonElement>> &
+export type AnchorButtonProps<P, K extends keyof P = never> = Omit<
+  P,
+  keyof Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, K>
+> &
```
NicholasBoll, line 102: I added `K` to allow for omitting specific keys from the `P` type. This prevents the need for defining this interface again

modules/button/react/spec/Button.spec.tsx
```diff
@@ -20,7 +20,7 @@ const map: any = {
   'Deprecated Button': DeprecatedButton,
 };
 Object.keys(map).forEach(buttonName => {
-  const ButtonComponent = map[buttonName];
+  const ButtonComponent: typeof Button = map[buttonName];
```
NicholasBoll, line 23: `typeof Button` is common to all, so I used that to prevent `any` everywhere.

modules/core/react/lib/InputProvider.tsx
```diff
@@ -215,7 +215,7 @@ export default class InputProvider extends React.Component<InputProviderProps> {
     this.enableListeners(false);
   }
 
-  enableListeners(enable: boolean) {
+  enableListeners = (enable: boolean) => {
```
NicholasBoll, line 218: I got Typescript errors with out this telling me that use of `this` within this function was set to global... I'm not sure how things worked... Perhaps since there's only meant to be one on the page? I have no idea how this wasn't caught before...

modules/core/react/lib/InputProvider.tsx
```diff
@@ -231,7 +231,7 @@ export default class InputProvider extends React.Component<InputProviderProps> {
     const fn = enable ? window.addEventListener : window.removeEventListener;
 
     // pointer events (mouse, pen, touch)
-    if (window.PointerEvent) {
+    if ('PointerEvent' in window) {
```
NicholasBoll, line 234: TS complained that `PointerEvent` is always defined and is a function and that maybe we meant to call it instead.

modules/page-header/react/spec/PageHeader.spec.tsx
```diff
@@ -46,7 +46,7 @@ describe('Page Header', () => {
   });
 
   test('PageHeader should spread extra props', () => {
-    const component = mount(<PageHeader data-propspread="test" />);
+    const component = mount(<PageHeader title="Page Header" data-propspread="test" />);
```
NicholasBoll, line 49: TS complained that title is required

modules/tooltip/react/lib/Tooltip.tsx
```diff
@@ -43,7 +43,7 @@ export interface TooltipProps extends Omit<React.HTMLAttributes<HTMLDivElement>,
 }
 
 export const Tooltip = ({
-  type = 'label',
+  type = 'label1',
```
NicholasBoll, line 46: Testing annotations in Typescript for type errors

package.json
```diff
@@ -135,6 +135,9 @@
     "commitmsg": "commitlint --edit $HUSKY_GIT_PARAMS",
     "build-storybook": "build-storybook -c .storybook -o docs",
     "lint": "node utils/check-lockfile-links.js && node utils/check-mismatched-dependencies.js && eslint -c ./.eslintrc.js --ext=jsx,ts,tsx .",
+    "typecheck": "yarn typecheck:src && yarn typecheck:cypress",
+    "typecheck:src": "tsc -p . --noEmit --incremental false",
```
anicholls, line 139: Won't this also do `cypress/` since that folder is in `./`?

NicholasBoll: The `-p` means project where the path determines where the `tsconfig.json` file will be. That file ignores the `cypress` directory.

We had to split Cypress and the rest of the application up because both `jest` and `cypress` add globals with colliding definitions.

.github/tsc.json
```diff
@@ -0,0 +1,18 @@
+{
+  "problemMatcher": [
+      {
+          "owner": "tsc",
+          "pattern": [
+              {
+                  "regexp": "^(?:\\s+\\d+\\>)?([^\\s].*)\\((\\d+),(\\d+)\\)\\s*:\\s+(error|warning|info)\\s+(\\w{1,2}\\d+)\\s*:\\s*(.*)$",
```
NicholasBoll, line 7: This fixes an issue with the matcher built into `@actions/setuo-node` where the line info doesn't work

package.json
anicholls: Doh - haha sounds good.

modules/core/react/lib/InputProvider.tsx
anicholls: It seems like these changes broke the `InputProvider` tests. Not quite sure why

modules/core/react/lib/InputProvider.tsx
anicholls: Odd... If it's a class function, shouldn't `this` be correct? We should do this to all the functions if it's necessary

modules/core/react/lib/InputProvider.tsx
anicholls: I just changed this back and `yarn typecheck` works and the tests pass?

PR #611, opened 2020-05-01
fix(toast): Remove transform origin for visual stories

mannycarrera4:
When Chromatic takes its snapshot, the `Toast` hasn't animated into view so the snapshots look empty, by allowing a transform origin of null, we get to have our snapshots back.


NicholasBoll:
This contains some of the same code as #600 which is against master. Should we just merge `master` forward and close this PR?

mannycarrera4:
Works for me @NicholasBoll

mannycarrera4:
Solved in #600 once we merged master forward into pre release v4

anicholls:
@mannycarrera4 Don't you still need to pass `transformOrigin={null}` for this story? 

FYI, master changes are going in w/ https://github.com/Workday/canvas-kit/pull/623

mannycarrera4:
ah yeah, in that case, I can just delete all the transform origin stuff since that went into master, and pass null to the story

Issue #548, opened 2020-04-06
Use forwardRef for all components


NicholasBoll:
## 🚀 Feature Proposal

We use an API that allows the spread of extra attributes and properties to a "container" element of a component via `elemProps`. In some cases we even explicitly forward a reference with a naming convention like `inputRef`. `elemProps` does not forward refs for us.

Assuming `ColorInput` is a class component:

```tsx
<ColorInput ref={ref} inputRef={inputRef} />

ref; // reference to the instance of ColorInput
inputRef; // reference to the <input> element
```

React has a `forwardRef` API that allows us to formalize the use of `ref` as a property on all components. We would then apply that forwarded ref to the same containing element that extra props are applied to if it makes sense.

`forwardRef` allows us to keep the implementation of `ColorInput` private and allow us to give access to the public interface of the element.

## Motivation

There are many instances where it is desired to have a reference to an element of a component. We already do this in some cases (e.g. `inputRef` for inputs). `ref` can be used on components now, but that just encourages the use of methods or properties of the component. We mostly treat that interface as private and could introduce problems later as we refactor code.

Without using `forwardRef`, the implementation details of a component become public. For example, changing `ColorInput` from a [Class Component](https://reactjs.org/docs/react-component.html) to a [Functional Component](https://reactjs.org/docs/components-and-props.html) becomes a breaking API change.

You can always get access to any element on the page using `document.querySelector`. This change would make it easier to get access to the underlying element of each component without resorting to tricks or hacks to do so.

## Example

```tsx
// start date
<DateInput ref={startDateRef} />
<DateInput ref={endDateRef} />

// Perhaps when we choose a value for the start date, we want to focus on the end date
endDateRef.current.focus()
```

## Downsides

Another decision for Components - what element should be forwarded? Is it the same element as the element we spread extra `elemProps` over?

jpante:
Will need to look at this as a bigger milestone across all components.

alexandrzavalii:
just faced similar issue with TextButton.
TextButton does not forwardRef.

NicholasBoll:
All inputs and buttons forward as do new components. The `Svg` component still does not forward the ref.

PR #614, opened 2020-05-03
fix(avatar): Combine Avatar & AvatarButton and provide fallback image

### LINKED ISSUE #613: Provide fallback to icon when avatar image fails to load
## 🚀 Feature Proposal

If the link for the avatar fails, eg 404 etc, we are currently showing a broken link. This could be made a bit nicer by rendering the default icon instead.

## Motivation

Broken links look bad.

vibdev:
## Summary

If the link for the avatar fails, eg 404 etc, we are currently showing a broken link. This could be made a bit nicer by rendering the default icon instead.

Not sure the best way to test this, should I just add an item to the storybook for avatar image?

Also open question if the icon should have a label with the altText passed in on failure?

Closes #613 

## BREAKING CHANGES
- `<AvatarButton>` has been removed. By default `<Avatar>` will now be a button. If you need the old plain div version you can pass the prop `as="div"`. 

- The component is now a functional component instead of a class. If you are using ref on the class version it will not be pointing to the same thing. `buttonRef` has changed to `ref` since it could now reference a button or a div

- Visual change: Avatar images appear once they are load. While loading or if they fail to load the default icon will be shown. So you may want to check which variant you are using even in the image case.

## Checklist

- [x] `yarn test` passes

## Example
Avatar loading


modules/avatar/react/lib/AvatarButton.tsx
```diff
@@ -47,6 +47,21 @@ export default class AvatarButton extends React.Component<AvatarButtonProps> {
     altText: 'Avatar',
   };
 
+  constructor(props: AvatarButtonProps) {
```
vibdev, line 50: There is a lot of duplication in Avatar and AvatarButton. I wanted to make this a functional component but not 100% sure how y'all want to deal with the static methods.

anicholls: We just figured out a better way to do this, consolidating it into a single component! Check out https://github.com/Workday/canvas-kit/blob/prerelease/v4/modules/button/react/lib/DropdownButton.tsx#L33. 

It uses overload types to allow for different types depending on whether `as="a"` is passed. We weren't in a rush to update this one, but feel free if you'd like.

vibdev: oh sweet, I can give it a go

modules/avatar/react/lib/AvatarButton.tsx
```diff
@@ -47,6 +47,21 @@ export default class AvatarButton extends React.Component<AvatarButtonProps> {
...
+    super(props);
+    this.state = {
+      hasImageError: false,
+    };
+  }
```
NicholasBoll, line 55: Are you okay with the following instead?

```suggestion
  state = {
    hasImageError: false,
  }
```

It removes the `readonly` from `state`, but it is much cleaner.

modules/avatar/react/lib/AvatarButton.tsx
NicholasBoll: We have a proposal to convert to functional components and forward refs: #548

We don't really know if people are using `ref` on class components - it would be a breaking change if they are.

We do static properties like this:
```tsx
function MyComponent(props: Props) {
  return <div />
}

MyComponent.Variant = MyComponentVariant

File: modules/avatar/react/lib/Avatar.tsx [hunk 3]
```diff
@@ -72,14 +76,29 @@ export default class Avatar extends React.Component<AvatarProps> {
...
     const background = variant === AvatarVariant.Dark ? colors.blueberry400 : colors.soap300;
     return (
       <StyledContainer variant={variant} size={size} aria-label={altText} {...elemProps}>
-        {url ? (
-          <img src={url} alt={altText} />
+        {url && !this.state.hasImageError ? (
+          <img src={url} alt={altText} onError={this.setFallbackIcon} />
         ) : (
           <SystemIconCircle icon={userIcon} background={background} size={size} />
         )}
```
NicholasBoll, line 104: Is it possible to fade in a loaded avatar when loaded and not show on error instead of DOM swapping? If there is a URL, is the area simply blank until the URL is loaded?

NicholasBoll:
Overall, I like the idea. A few questions on the implementation to make it a better experience.

modules/avatar/react/lib/Avatar.tsx
vibdev: Oh that is a good idea. Let me play around with it

modules/avatar/react/lib/AvatarButton.tsx
vibdev: Would just moving to functional component be breaking? Or would the forward ref part be?

vibdev: I guess it doesn't matter since changing to `as=` would be breaking anyway.

NicholasBoll: Moving to a Functional Component changes what `ref` points to. In a class it is the instance of that class. In a FC, it doesn't point to anything unless it is forwarded to an element. It would be breaking only if someone used the `ref` for something. It is unlikely as there isn't anything useful on the instance.

NicholasBoll: I missed @anicholls comment. That is a nice way to support both. If added to `master`, it could be a non-breaking change by one component being an `as=` configuration and we add a deprecation notice to it.

anicholls: If it's ready in time, I'd also be cool with getting this in v4 and just doing removing `AvatarButton` entirely

vibdev:
@anicholls @NicholasBoll Updated and unified as a single functional component, let me know what you think.

modules/avatar/react/lib/AvatarButton.tsx
vibdev: I went ahead and just removed AvatarButton and defaulted to it being a button with optional div'ing. It seemed like button was a bit more common use case and the typescript was a lot easier.

modules/avatar/react/lib/Avatar.tsx
vibdev: Added a fade in over the top of the default icon, lmkwyt

modules/_labs/header/react/README.md
```diff
@@ -27,7 +27,6 @@ This component renders a responsive, Canvas-style header.
 ```tsx
 import {Header} from '@workday/canvas-kit-labs-react-header';
-import {AvatarButton} from '@workday/canvas-kit-react-avatar';
```
anicholls, line 30: This line shouldn't have been deleted - should have just been changed to `Avatar`

modules/avatar/react/lib/Avatar.tsx
```diff
@@ -10,80 +11,136 @@ export enum AvatarVariant {
...
    */
-  altText: string;
+  altText?: string;
   /**
    * The url of the Avatar image.
    */
   url?: string;
+  /**
+   * The alternative container type for the button. Uses Emotion's special `as` prop.
+   * Will render an `button` tag instead of a `div` when defined.
```
anicholls, line 36: As you have it, this is backwards

modules/avatar/react/lib/Avatar.tsx
```diff
@@ -10,80 +11,136 @@ export enum AvatarVariant {
...
+   */
+  as?: 'div';
+  /**
+   * The function called when the Avatar is clicked.
+   */
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+  /**
+   * The ref to the button that the styled component renders.
+   */
+  buttonRef?: React.Ref<HTMLButtonElement>;
```
anicholls, line 46: I wonder if this should be called `buttonRef` if it could be a div 🤔

modules/avatar/react/lib/Avatar.tsx
```diff
@@ -10,80 +11,136 @@ export enum AvatarVariant {
...
+    <StyledContainer
+      variant={variant}
+      size={size}
+      aria-label={altText}
+      onClick={onClick}
+      disabled={onClick ? false : true}
+      ref={buttonRef}
+      {...elemProps}
+    >
+      <StyledStack size={size}>
```
anicholls, line 131: Where does the term `Stack` come from?

modules/avatar/react/lib/Avatar.tsx
```diff
@@ -10,80 +11,136 @@ export enum AvatarVariant {
...
+    height: '100%',
+    borderRadius: borderRadius.circle,
+    transition: 'opacity 200ms linear',
+  },
+  ({isLoaded}) => ({
+    opacity: isLoaded ? 1 : 0,
+  })
+);
+
+const Avatar = ({
```
anicholls, line 102: Am I missing the function overloading we discussed? If it's not included, when I pass `as="div"`, all of the button attributes will still be valid which is not what we want

modules/avatar/react/lib/Avatar.tsx
```diff
@@ -10,80 +11,136 @@ export enum AvatarVariant {
...
 
-export default class Avatar extends React.Component<AvatarProps> {
-  static Variant = AvatarVariant;
-  static Size = SystemIconCircleSize;
+const StyledImage = styled('img')<{isLoaded: boolean}>(
+  {
+    width: '100%',
+    height: '100%',
+    borderRadius: borderRadius.circle,
+    transition: 'opacity 200ms linear',
```
anicholls, line 95: Nit: Can we try 150ms? it feels a little slow atm

anicholls: Also, this is going to break our snapshot tests 🤔

modules/avatar/react/lib/Avatar.tsx
vibdev: Sorry I don't remember that conversation? Are you saying we should null out onClick if as='div' is passed in?

modules/avatar/react/lib/Avatar.tsx
vibdev: I will remove for now, we can think of a better way to handle this later

modules/avatar/react/lib/Avatar.tsx
vibdev: 🤷‍♂️ it is used to stack elements on top of each other using absolute position. Have I ever mentioned how much I hate naming styled components.

modules/avatar/react/lib/Avatar.tsx
vibdev: will rename to elementRef, maybe? In the DropdownButton example it was named buttonRef.

modules/avatar/react/lib/Avatar.tsx
vibdev: 👍 how do you deal with animations and transitions usually? Is there a class that overrides things like this for testing?

modules/avatar/react/lib/Avatar.tsx
anicholls: Ah - makes sense. Haha I feel you 😂

modules/avatar/react/lib/Avatar.tsx
anicholls: I wonder if we should just call it `ref` since we'll be doing that a lot with #548. Any thoughts @NicholasBoll?

modules/avatar/react/lib/Avatar.tsx
anicholls: Maybe my initial explanation wasn't clear. Check out https://github.com/Workday/canvas-kit/blob/prerelease/v4/modules/button/react/lib/DropdownButton.tsx#L33 and https://github.com/Workday/canvas-kit/blob/prerelease/v4/modules/button/react/lib/types.ts#L112.

`ButtonOrAnchorComponent` provides different interfaces depending on whether `as` is defined by using function overrides. We'll probably want to do something similar here. Since it's only in one place and not shared a bunch of components (like buttons), you could probably just put the type inline. Something like

```tsx
type AvatarDivProps<P> = Omit<
  P,
  keyof React.ButtonHTMLAttributes<HTMLButtonElement>
> &
  React.DivHTMLAttributes<DivAnchorElement>;

type AvatarOverload = {
(props: AvatarProps): React.ReactElement;
(props: {as: 'div'} & AvatarDivProps<ButtonProps>): React.ReactElement;
  Variant: typeof AvatarVariant;
  Size: typeof AvatarSize;
};
const Avatar: AvatarOverload = ({
//...
```

We'll likely want a more generic type for this in the future, but I wouldn't worry about it in this PR.

modules/avatar/react/lib/Avatar.tsx
anicholls: Everything in the past has been tied to popups and there is a `transformOrigin` field. We support `transformOrigin={null}`, which removes the animation. Not sure how we'd want to do it here 🤔

modules/avatar/react/lib/Avatar.tsx
vibdev: done, one thing I noticed is I had to  `Omit<AvatarProps, 'as'>` from the button version otherwise ts wouldn't complain about adding ButtonHTMLAttributes to as='div' Avatars.

modules/avatar/react/lib/Avatar.tsx
vibdev: We could override the css, but snaps are done in storybook too right? Is there a way to conditional pass something in just for tests and not for a normal storybook run?

modules/avatar/react/README.md
```diff
@@ -101,106 +104,20 @@ Default: `Avatar.Size.m`
...
-#### `altText: string`
-
-> Text describing what the avatar is showing.
-
----
-
-#### `url: string`
-
-> The URL of the user's photo. Expects a square (1:1) photo.
+> A ref to the underlying `<button>` element.
```
anicholls, line 111:
```suggestion
> A ref to the underlying `<button>` or `<div>` element.
```

modules/avatar/react/lib/Avatar.tsx
anicholls: I just realized that since we switched to opt-in snapshots (to keep our chromatic costs down), this component isn't captured anymore. We're going to be re-evaluating our storybook approach in the near-ish future as we continue work on a documentation site, so I wouldn't worry about this for now.

modules/avatar/react/README.md
vibdev: The inline suggestion commit suggestion is pretty neat!

modules/avatar/react/lib/Avatar.tsx
```diff
@@ -30,60 +31,132 @@ export interface AvatarLocalProps {
...
+   */
+  as?: 'div';
+  /**
+   * The function called when the Avatar is clicked.
+   */
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+  /**
+   * The ref to the button or div that the styled component renders.
+   */
+  ref?: React.Ref<HTMLButtonElement>;
```
NicholasBoll, line 46: I don't think this will work. Let me test it out.

vibdev: I considered adding the ref in the Avatar overload but wasn't sure if it was needed since in Alex's example it wasn't split up.

NicholasBoll: I have confirmed this will not work and React will throw the following warning:
```
react.development.js:167 Warning: Avatar: `ref` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop.
```

I think `ref` is the correct prop, but we'll have to use `forwardRef` for the component to make that work. Some more details here: #548

anicholls: Good catch 👍  We should make sure we have a test for this as well.

Issue #592, opened 2020-04-22
Pagination: Announce change to screen readers


NicholasBoll:
## 💬 Questions and Help

The Pagination component does not announce any changes to the DOM when changing pages. The question is should the Pagination component announce something, or should this be up to the application?

## Pagination component announces changes
This could be generic like "Showing Page 2 of 3". It could be implemented using `aria-live=polite`. The downside here could be the application also announces changes and the announcement could be out-of-order and possibly confusing:

"The table was updated to Page 2. Showing Page 2 of 3" or "Showing Page 2 of 3. The table was updated to Page 2" depending on the order of the Pagination component compared to the Table component.

## Application announces changes
This is more burden on the application developers, but would allow more semantic announcements. It could avoid announcing pages again if the application's announcements were more semantic.

NicholasBoll:
Closed via #620

PR #620, opened 2020-05-05
fix(pagination): Provide aria live attribute for accessbility

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/592

Provide an aria live announcement when the label changes announcing the current page you're on out of how many potential pages. This is an `opt-out` by default to provide as much information as we can. If you'd like to opt out, the consumer can set `announceLabelToScreenReaders ` to false.


modules/_labs/pagination/react/lib/Pagination.tsx
```diff
@@ -123,7 +123,9 @@ const Pagination = (props: PaginationProps) => {
           />
         </ButtonsContainer>
         {showGoTo && <GoTo onSubmit={onPageChange} max={numPages} label={goToLabel} />}
-        {showLabel && <StyledLabel>{customLabel(labelFrom, labelTo, total)}</StyledLabel>}
+        {showLabel && (
+          <StyledLabel aria-live="polite">{customLabel(labelFrom, labelTo, total)}</StyledLabel>
+        )}
```
NicholasBoll, line 128: Is this an okay opt-in? If the label is shown, it will get announced to screenreaders on update. That does allow an application developer to disable only by providing their own label...

If this is an acceptable solution, I think we should add something to the README.md about this accessibility feature so at least we let people know it is happening and it could possibly double announce depending on what the rest of the application does when the page value changes.

mannycarrera4:
So after talking to accessibility, there's a few things to consider:

A lot of this depends on context and it always seems like it's up to the consumer to opt in.

• Some think that we shouldn't be using aria-live status updates for literally everything that happens on the page

• Some of it should just come down to the user exploring page content.

• Some felt that more information is always better than not enough information, and providing status updates on page content offers a more informative, better user experience.

Contextual questions might include:

Is this component being used in a single page app? Yes/No
If yes, does the single page app already use another live region to announce page updates? Yes/No
Is this component being used inside of an iframe such that the rest of the page isn't updated? Yes/No

I propose one of two things:

1. We either add a prop to turn on the `aria-live` and document it
2. Leave code as is and try to document as much as possible since it's very dependent on the use case and context of the application.

I personally prefer two since it's impossible to account for all use cases.

NicholasBoll:
That is a tough one. We want things to be more accessible by default.

Reading documentation and opting in is a lot to ask of all developers (opt-in). It is also hard to find where the magic is happening and turn it off (opt-out). Would accessibility prefer we are cautious and give more info allowing opt-out if the application provides better contextual info? Or would they rather opt-in.

Ultimately applications go under accessibility review and audits. I think it is easier to determine which information is superfluous than what information is lacking.

My personal vote would be to default on. If Accessibility is familiar with how Canvas Kit behaves (because they've helped to shape it), they should have an idea what would need to be disabled or if the information provided needs to be more contextual (that would then be up to app developers to provide).

mannycarrera4:
Also, there seems to be an issue with how React batches updates to the DOM because it announces it twice?

lychyi:
This is tough. But what @NicholasBoll said may be important to think about:
> Ultimately applications go under accessibility review and audits. I think it is easier to determine which information is superfluous than what information is lacking.

Even if applications don't go through good accessibility reviews, which is the greater evil? Not announcing the pagination changes or announcing them unnecessarily? Also, this is something that only gets announced after user interaction right? It's not like we're building a component that could possibly announce unnecessarily without a user interacting with the page (like when a login session is about to expire).

mannycarrera4:
@lychyi Correct, it announces on change

modules/_labs/pagination/react/lib/Pagination.tsx
```diff
@@ -36,6 +36,15 @@ export interface PaginationProps extends React.HTMLAttributes<HTMLElement> {
...
   /** Optional width to pass to component. This is the width the container deems is available. You can use a measure component to get this. */
   width?: number;
+  /**
+   * Announces the change to the label below the pagination bar describing on what page
+   * you are on out how many possible pages.
+   * Note: We default this to `true` to provide more information for accessibility.
+   * However, the context of your application might mean that this information is superfluous.
+   * If you choose to now want this, you can opt out by setting this to `false`
+   * @default true
+   */
```
mannycarrera4, line 46: Should I reference this PR in this comment for future reference?

mannycarrera4:
The other thing I thought about doing is adding an `aria-label` to the input that announces something like this `aria-label={!value ? 'Label edit blank hex value' : `edit hex value ${formattedValue}`}`

modules/_labs/pagination/react/lib/Pagination.tsx
NicholasBoll, line 46: What do you think of the following?

```suggestion
  /**
   * Announces page changes to screen readers using aria-live
   * Note: Your application may already announce page changes to screen readers through
   * other means like focus changes or other aria-live regions. Set this to `false` to remove
   * redundant announcement to screen reader users.
   * @default true
   */
```

modules/_labs/pagination/react/lib/Pagination.tsx
```diff
@@ -36,6 +36,15 @@ export interface PaginationProps extends React.HTMLAttributes<HTMLElement> {
...
+  announceAriaLiveOfLabel?: boolean;
```
NicholasBoll, line 47: The name doesn't flow in my brain very well. What about `shouldAnnounceLabel`? Or `shouldAnnounceLabelToScreenReaders`? The second is much more verbose. We could drop the `should` (I personally like prefixing `should` or `is` for booleans)

PR #631, opened 2020-05-06
chore: Upgrade Babel and presets to support optional chaining

### LINKED ISSUE #605: Add support for optional chaining
## 🚀 Feature Proposal

With the update to Typescript 3.8 in v4, we can now use optional chaining (i.e. `foo?.bar?.baz`).

However, for this to work fully, we need to upgrade prettier and babel. Let's do that so we can use this 🙂

NicholasBoll:
* Upgrade Storybook
* Upgrade Prettier
* Clean yarn.lock file

Fixes #605


babel.config.js
```diff
@@ -2,12 +2,7 @@ module.exports = {
   env: {
     test: {
       presets: ['@babel/preset-env', '@babel/preset-typescript', '@babel/preset-react'],
-      plugins: [
-        'emotion',
-        '@babel/proposal-class-properties',
-        '@babel/proposal-object-rest-spread',
```
NicholasBoll, line 8: Object rest spread is already a part of `'@babel/preset-env'`

package.json
```diff
@@ -4,25 +4,24 @@
...
-    "@storybook/react": "^5.3.17",
-    "@storybook/source-loader": "^5.3.14",
+    "@storybook/addon-actions": "^5.3.18",
+    "@storybook/addon-docs": "^5.3.18",
+    "@storybook/addon-knobs": "^5.3.18",
+    "@storybook/addon-storysource": "^5.3.18",
+    "@storybook/addon-viewport": "^5.3.18",
+    "@storybook/components": "^5.3.18",
+    "@storybook/react": "^5.3.18",
+    "@storybook/source-loader": "^5.3.18",
```
NicholasBoll, line 24: Some of these monorepo dependencies were on `5.3.14` and others on `5.3.17`. I just updated to all be latest

package.json
```diff
@@ -69,7 +68,7 @@
     "enzyme-adapter-react-16": "^1.5.0",
     "eslint": "^6.8.0",
     "eslint-config-prettier": "^6.3.0",
-    "eslint-plugin-compat": "^3.5.1",
+    "eslint-plugin-compat": "next",
```
NicholasBoll, line 71: https://github.com/amilajack/eslint-plugin-compat/issues/307

.eslintrc.js
```diff
@@ -78,6 +76,5 @@ module.exports = {
     'react/jsx-no-bind': 'off', // Keep perf implications in mind, but was giving too many warnings and hurting readability
     curly: 'error',
     radix: 'error',
-    'prettier/prettier': ['warn', prettierConfig],
```
NicholasBoll, line 81: I found the linting squigglies to be distracting for issues that would be fixed on save or during the commit stage. I opted to remove from the default config and move to a separate config specifically for formatting. The `.eslintrc-prettier.js` config file is called for formatting purposes only.

modules/avatar/react/spec/Avatar.spec.tsx
```diff
@@ -29,7 +29,13 @@ describe('Avatar Accessibility', () => {
...
   test('Avatar should pass axe DOM accessibility guidelines', async () => {
     const html = ReactDOMServer.renderToString(<Avatar />);
-    expect(await axe(html)).toHaveNoViolations();
+    expect(
+      await axe(html, {
+        rules: {
+          region: {enabled: false},
+        },
+      })
+    ).toHaveNoViolations();
```
NicholasBoll, line 38: I upgraded `jest-axe` which introduced the `region` rule. I could either roll back that update, or update usages to ignore that rule. Our small examples tend to not have landmark regions, so this rule is not helpful in this context. `jest-axe` doesn't allow global configuration without a central export. We're deprecating the usage of `jest-axe` in favor of `cypress-axe`, so I didn't think updating the 5 usages that are still left would be too bad.

Issue #182, opened 2019-09-11
Add a popup stacking manager to canvas-kit


NicholasBoll:
## 🚀 Feature Proposal

Add a Popup manager to help with focus management of popup content and handle proper stacking context of Popups

## Motivation

A Popup is a component that uses Portals to avoid constraints of parent overflow containers. This is done by adding a DOM element to the end of the `<body>` tag and often set a `z-index` to ensure the Popup is rendered on top of other content.

### Rendering
A popup can receive a `zIndex` value to be set on the container based on a stacking algorithm set by the Popup Manager to ensure proper rendering of Popups. Ex: this helps avoid the issue where a Tooltip renders behind the Modal.

### Accessibility
Accessibility of Popup content often includes the use of the Escape key to close a Popup, but simply attaching a handler for the Escape key would cause all Popups to close rather than only the top-most Popup. A Popup manager is used to handle a stack of Popups to know which one should receive the close event.

Additionally when a Popup is closed, focus should be moved back to the focusable target. In the case of a Modal, this would be the button that was clicked to to trigger the button. A Popup Manager helps track previously focused targets to ensure focus returns when a Popup is closed.

## Example

A Modal contains a Select component or a DatePicker component. After the Modal is opened, the user clicks on a DatePicker and a calendar opens. That calendar should be rendered on top of the Modal and not constrained by the bounding box of the Modal. If the user hits the Escape key, the calendar Popup should close and focus returns to the DatePicker input and the Modal should still be open. If the user hits the Escape key again, the modal will now close and focus should return back to the button that triggered the Modal.

### Requirements
- **Maintain a stack of active popups globally**. All popups on a page (whether from Canvas Kit or outside) will use this stack. It should be basic and simple enough that additional features can built on top of it.
- **The stack needs to maintain references to elements**. A common requirement of popups is to close the top-most when clicking outside the popup. In order for this to work properly, the clicked element is checked against currently opened popups within the stack to see if the click originated inside or outside opened popups.

### Questions
1. Is a FILO stack powerful enough for all use-cases? Do relationships need to be maintained in trees?
1. Do we need a concept of closable via ESC key or click outside to help aid functions built on top of the stack to know which elements are closed? Or should each component keep track of its element and close only if it's element is currently at the top of the stack
1. Do we need global event/click listeners as part of the stack or can each individual component keep track of those things?

NicholasBoll:
Closed by #670

PR #641, opened 2020-05-09
feat(select): Render menu using a portal

### LINKED ISSUE #571: Add Portal Support to Select component
## 🚀 Feature Proposal

Add portal support to select component, as per other components in canvas kit.

## Motivation

Currently select component menu expands outside the bounds of the visible area.

jamesfan:
Closes #571.

Modifies the Select (with Canvas-styled Menu) to render its menu using a portal. Positioning is handled by the Canvas Kit Popper wrapper.

Note: this was branched off of `prerelease/v4` in order to leverage Popper v2.

_Last updated: June 8, 2020_

## Checklist

- [x] Fix IE11 issue where the selected option isn't scrolled into view when you activate the menu
- [x] Integrate with Popup Stack Manager
- [x] Design and Accessibility review
- [x] Update documentation
- [x] Clean up stories


modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -1,17 +1,45 @@
 import * as React from 'react';
-import {ErrorType, styled} from '@workday/canvas-kit-react-common';
+import {Rect} from '@popperjs/core';
```
jamesfan, line 2: Should we export `Rect` from `@popperjs/core` in the Canvas Kit Popper (in addition to `Placement` and `PopperOptions`)? Currently, I'm using `Rect` to type the `reference` and `popper` arguments in the `offset` modifier.

modules/_labs/select/react/package.json
```diff
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "@emotion/core": "^10.0.28",
+    "@popperjs/core": "^2.1.1",
```
jamesfan, line 42: `@popperjs/core` can be removed as a dependency if we export `Rect` from the CK Popper.

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -134,21 +196,131 @@ const MenuList = styled('ul')<Pick<SelectProps, 'error'>>(
...
+    return buttonRef.current ? (
+      <Popper
+        placement={isFlipped ? 'top-start' : 'bottom-start'}
+        open={!isHidden}
+        anchorElement={buttonRef.current}
+        popperOptions={this.generatePopperOptions()}
+        // zIndex is necessary in order for the menu to be displayed
+        // properly if Select is placed within a CK Modal (necessary
+        // to override zIndex of 1 for the Modal)
+        style={{zIndex: 2}}
```
jamesfan, line 316: Wanted to call out the `zIndex` here. See code comment for an explanation.

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -352,6 +368,11 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
         value: option.value,
         ...(onOptionSelection
           ? {
+              // Prevent click from propagating (and, say, dismissing
+              // a modal containing the Select)
+              onClick: (event: React.MouseEvent) => {
+                event.stopPropagation();
+              },
```
NicholasBoll, line 375: We will need a better solution here. We have the same issue with other Popup systems and we shouldn't have to solve it one-off in all of them. We should add some requirements and fix it here: #182

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -352,6 +368,11 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
...
               // mouseDown provides a slightly better UX than click
               // since visual feedback of selected option is more
               // immediate
```
NicholasBoll, line 378: Not really part of this PR, but I did notice this while testing. Right-clicking selects a menu item. Or more accurately, right-mouse down selects an item. Why did we decide to do mouse down instead of a click? A native select waits for a click. If you press your mouse down, but don't release, the item is selected and the menu closes. It is surprising.

At minimum, we should not select on any other mouse button than the left one. At maximum, we should consider selection on click.

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -273,12 +264,22 @@ export default class Select extends React.Component<SelectProps, SelectState> {
...
+    this.toggleMenu(isMenuHidden);
   };
 
   handleMouseDown = (event: React.MouseEvent<HTMLButtonElement>): void => {
-    // Cancel default handling of the event to ensure we maintain
-    // control of focus
+    // Cancel default handling of the event to ensure we maintain control
+    // of focus (i.e., so focus is immediately transferred to the menu when
+    // opening the menu, rather than briefly being applied to the button
+    // and creating a momentary flash of a focus ring around the button)
```
NicholasBoll, line 282: Great explanation! I'd like to see more of these types of comments in the codebase.

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -20,7 +48,18 @@ interface SelectMenuProps
...
+
+interface SelectMenuState {
+  // We need the width state to support the use case where the menu is
+  // displayed without user interaction (e.g., if we want to show the menu
+  // from the start). buttonRef.current is set to null during the first
+  // render, which prevents Popper from rendering due to the lack of a
+  // reference element. By the time componentDidMount runs, buttonRef.current
+  // will have been set -- we can then update the width state and trigger
+  // a render with the Popper menu.
+  width: number;
```
NicholasBoll, line 62: This is an example of where hooks are easier and don't require state. `useEffect` and `useLayoutEffect` are run after render where an element's dimensions can be measured. `useLayoutEffect` will trigger before DOM changes are rendered to the page and `useEffect` will trigger after DOM changes are rendered to the page.

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -57,11 +96,21 @@ const menuBorderCSS = (error?: ErrorType): CSSObject => {
...
-      borderLeft: dividerBorder,
-      borderRight: dividerBorder,
+      position: 'absolute',
+      width: '100%',
+
+      '[data-popper-placement^="bottom-start"] &': {
+        top: 0,
+      },
+      '[data-popper-placement^="top-start"] &': {
+        bottom: 0,
```
NicholasBoll, line 112: I'm curious about the `^` in this selector. Aren't the values only either `top-start` or `bottom-start`?

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -134,21 +196,131 @@ const MenuList = styled('ul')<Pick<SelectProps, 'error'>>(
...
+            // Skid menu along the edge of the button to account
+            // for fractional x-positioning of the button (e.g.,
+            // if the button is in a horizontally-centered modal)
+            // and to ensure proper alignment between the button
+            // and the menu
+            const skidding = reference.x % 1 === 0 ? 0 : 1;
+            // Displace menu towards the button to obscure the bottom
+            // edge of the button and to create a smooth visual
+            // connection between the button and the menu
+            const distance = -parseInt(borderRadius.m, 10);
```
NicholasBoll, line 250: Clever hack here. `parseInt` will convert a string to a number from left-to-right until it encounters a non-number and will stop parsing and return the result so far. This effectively strips out the `px` part of the value.

modules/_labs/select/react/lib/SelectMenu.tsx
NicholasBoll: We cannot hard-code without an option to override. `zIndex` 2 will work in a Canvas Kit-only application with no other `zIndex` overrides, but will not work in an application with other z index values. Modal allows the z-index to be overridden and we should also do that here. Until we have a Popup manager #182 that works globally across all apps, we need to allow z-index to be overridden.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Extraneous code left over from a copy-and-paste. I'll remove the carets.

modules/_labs/select/react/lib/SelectBase.tsx
jamesfan: Option selection will be modified to trigger on click instead of mousedown in #672.

Previously, I had set it to mousedown to avoid the brief moment when two options share the same blue background color (the just-clicked-on option and the assistive-focused option), but this is an extremely minor and subjective visual nit. I agree consistency with the native select should win out here.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: SelectMenu has been converted to a function component. State is now maintained using the `useState` hook.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Removed the zIndex after merging in the stack manager.

modules/_labs/select/react/lib/SelectBase.tsx
jamesfan: Removed stopPropagation call after merging in the stack manager.

modules/_labs/select/react/stories/stories_visualTesting.tsx
```diff
@@ -87,20 +87,36 @@ export const SelectStatesMenuOn = () => (
...
+        return (
+          <div
+            style={{
+              height: 500,
+              display: 'flex',
+              alignItems: 'center',
+              justifyContent: 'flex-start',
+            }}
+          >
+            <SelectBase
```
jamesfan, line 106: I noticed the menu occasionally becomes slightly misaligned with its corresponding button if the button has a fractional position (issue filed at #728). I originally noticed this in the IE visual snapshot testing for the `Select States Menu On` story and assumed it was specific to IE, but it also affects Firefox (Chrome is unaffected).

I don't _think_ this should block the introduction of this PR into v4 (not sure how common it will be for the Select button to have a fractional position, though Selects in tables may be a reasonably common use case), but curious what others think.

jamesfan: I've tweaked the positioning logic of the menu and it appears to be better now in IE/Firefox (though not perfect). [Old](https://5d854c26ba934e0020f5e98a-kvhfcczlnk.chromatic.com/?path=/story/testing-react-labs-select--select-states-menu-on) vs [New](https://5d854c26ba934e0020f5e98a-yorrfvmfvx.chromatic.com/?path=/story/testing-react-labs-select--select-states-menu-on).

However, I've also since noticed the issue seems to be dependent on high-DPI vs low-DPI screens (see #728).

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -110,6 +111,21 @@ export interface SelectBaseProps extends CoreSelectBaseProps {
    * @default true
    */
   isMenuAnimated: boolean;
+  /**
+   * If true, automatically flip the SelectBase menu to keep it visible if necessary (e.g., if the the menu would otherwise display below the visible area of the viewport).
+   * @default true
+   */
+  isMenuAutoFlipped: boolean;
```
NicholasBoll, line 118: These config options read like state vs props. `isMenuAutoFlipped` vs `shouldMenuAutoFlip`.

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -110,6 +111,21 @@ export interface SelectBaseProps extends CoreSelectBaseProps {
...
+  /**
+   * If true, focus the SelectBase menu when it's shown. Set to false if you don't want to focus the menu automatically (for visual testing purposes, for example).
+   * @default true
+   */
+  isMenuAutoFocused: boolean;
+  /**
+   * If true, flip the SelectBase menu so it extends upwards from the button.
+   * @default false
+   */
+  isMenuFlipped: boolean;
```
NicholasBoll, line 128: Are these all config for visual testing? If so, we should consider a way to do static visual things without requiring public props for this purpose. Context is a possibility.

modules/_labs/select/react/lib/SelectMenu.tsx
NicholasBoll: Feel free to do that

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -1,18 +1,47 @@
...
+  /**
+   * If true, automatically flip the SelectMenu to keep it visible if necessary (e.g., if the the SelectMenu would otherwise display below the visible area of the viewport).
+   * @default true
+   */
+  isAutoFlipped: boolean;
+  /**
+   * If true, focus the SelectMenu when it's shown. Set to false if you don't want to focus the SelectMenu automatically (for visual testing purposes, for example).
+   * @default true
+   */
+  isAutoFocused: boolean;
```
NicholasBoll, line 34: Same with these. The intent is more like a `shouldAutoFocus` - a configuration rather than a changing state.

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -133,20 +189,152 @@ const MenuList = styled('ul')<Pick<SelectProps, 'error' | 'theme'>>(
...
+    {
+      name: 'offset',
+      options: {
+        offset: ({
+          placement,
+          reference,
+          popper,
+        }: {
+          placement: Placement;
+          reference: Rect;
```
NicholasBoll, line 218: Is `Rect` necessary if you inline these modifiers? I don't see any benefit extracting out into a variable if inlining can drop extra TS burden. The config is already called `modifiers`, so we're not loosing the abstraction of a name

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -438,13 +416,19 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
...
+            isAutoFocused={isMenuAutoFocused}
+            isFlipped={isMenuFlipped}
+            isHidden={isMenuHidden}
             isHiding={isMenuHiding}
-            menuRef={menuRef}
+            menuRef={this.menuRef}
             onBlur={onMenuBlur}
             onKeyDown={onKeyDown}
+            onCloseOnEscape={onMenuCloseOnEscape}
           >
```
NicholasBoll, line 432: This is starting to be a lot of config options

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -133,20 +189,152 @@ const MenuList = styled('ul')<Pick<SelectProps, 'error' | 'theme'>>(
...
+  const handleWidthChange = () => {
+    if (buttonRef.current) {
+      const newMenuWidth = buttonRef.current.clientWidth + 2 * buttonBorderWidth;
+      setWidth(newMenuWidth);
+    }
+  };
+
+  useEffect(() => {
+    handleWidthChange();
+  }, [width]);
```
NicholasBoll, line 278: This seems like it could be a reusable hook without the `buttonBorderWidth` and the magic `2`.

NicholasBoll, line 278: This dependency array is incorrect. It calculates `width` and depends on `buttonRef`
```suggestion
  const [width, setWidth] = useState(0);

  const handleWidthChange = () => {
    if (buttonRef.current) {
      const newMenuWidth = buttonRef.current.clientWidth + 2 * buttonBorderWidth;
      setWidth(newMenuWidth);
    }
  };

  useEffect(() => {
    handleWidthChange();
  }, [buttonRef]);
```

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -133,20 +189,152 @@ const MenuList = styled('ul')<Pick<SelectProps, 'error' | 'theme'>>(
...
+  const [width, setWidth] = useState(0);
+
+  const handleWidthChange = () => {
+    if (buttonRef.current) {
+      const newMenuWidth = buttonRef.current.clientWidth + 2 * buttonBorderWidth;
+      setWidth(newMenuWidth);
+    }
+  };
+
+  useEffect(() => {
```
NicholasBoll, line 276: Should this be `useLayoutEffect` instead? `useEffect` will always have a 1 frame delay which will cause a "flickering" effect.

```suggestion
  useLayoutEffect(() => {
```

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -133,20 +189,152 @@ const MenuList = styled('ul')<Pick<SelectProps, 'error' | 'theme'>>(
...
+    return () => {
+      window.removeEventListener('resize', handleWidthChange);
+    };
+  }, []);
+
+  useCloseOnEscape(popupRef, () => {
+    if (onCloseOnEscape) {
+      onCloseOnEscape();
+    }
+  });
```
NicholasBoll, line 299: A bit cleaner is:
```suggestion
  useCloseOnEscape(popupRef, () => onCloseOnEscape?.());
```

I think that is a worthwhile terseness. It uses the optional chaining feature to call a callback if the function is defined. Perhaps we should allow `useCloseOnEscape` and `useCloseOnOutsideClick` to have an optional callback? This would drop the requirement to define an empty callback each render.

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -1,18 +1,47 @@
...
+  /**
+   * If true, focus the SelectMenu when it's shown. Set to false if you don't want to focus the SelectMenu automatically (for visual testing purposes, for example).
+   * @default true
+   */
+  isAutoFocused: boolean;
+  /**
+   * If true, flip the SelectMenu so it extends upwards from the button.
+   * @default false
+   */
+  isFlipped: boolean;
```
NicholasBoll, line 39: I think I'd prefer to have a disjoint union of the options allowed:

```ts
placement: 'top-start' | 'bottom-start'
```

Or since `top` and `bottom` are the only ones with meaning since we make the width the same as the parent:

```ts
placement: 'top' | 'bottom'
```

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -1,18 +1,47 @@
...
+  /**
+   * If true, hide the SelectMenu.
+   * @default false
+   */
+  isHidden: boolean;
```
NicholasBoll, line 44: What is the difference between `isHidden` and `isHiding`?

modules/_labs/select/react/lib/scrolling.ts
```diff
@@ -0,0 +1,59 @@
+const clientHeight = (elem: HTMLElement): number => {
+  const computedStyle = window.getComputedStyle(elem, null);
+  const borderTopWidth = parseInt(computedStyle.getPropertyValue('border-top-width'), 10);
+  const borderBottomWidth = parseInt(computedStyle.getPropertyValue('border-bottom-width'), 10);
+  return elem.offsetHeight - borderTopWidth - borderBottomWidth;
+};
+
+// Modified from https://gist.github.com/hsablonniere/2581101
+export const scrollIntoViewIfNeeded = (elem: HTMLElement, centerIfNeeded = true): void => {
+  const parent: HTMLElement | null = elem.parentElement;
```
NicholasBoll, line 10: You generally shouldn't provide a type to a `const` and let it infer from the assignment. You could inadvertently widen the type:

```suggestion
  const parent = elem.parentElement;
```

modules/_labs/select/react/stories/stories_testing.tsx
```diff
@@ -0,0 +1,136 @@
...
+      {...elemProps}
+    >
+      {children}
+    </div>
+  );
+};
+
+const SelectModal = () => {
+  const [open, setOpen] = React.useState(false);
+  const buttonRef = React.useRef<HTMLButtonElement>() as React.RefObject<HTMLButtonElement>; // cast to keep buttonRef happy
```
NicholasBoll, line 37: Weird little trick:
```suggestion
  const buttonRef = React.useRef<HTMLButtonElement>(null)
```

There are 3 overloads to `useRef`:

```ts
    function useRef<T>(initialValue: T): MutableRefObject<T>;
    function useRef<T>(initialValue: T|null): RefObject<T>;
    function useRef<T = undefined>(): MutableRefObject<T | undefined>;
```

If you pass `null`, it will pick the `RefObject` version.

modules/_labs/select/react/stories/stories_testing.tsx
```diff
@@ -0,0 +1,136 @@
...
+        </Button>
+        <Button onClick={closeModal} variant={Button.Variant.Secondary}>
+          Cancel
+        </Button>
+      </Modal>
+    </div>
+  );
+};
+
+export const PortalTest = () => (
```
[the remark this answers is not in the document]
NicholasBoll: Only `visualTesting` stories have state tables. Non-visual testing stories are tested manually or by Cypress.

modules/_labs/select/react/stories/stories_testing.tsx
```diff
@@ -0,0 +1,136 @@
...
+      }}
+      {...elemProps}
+    >
+      {children}
+    </div>
+  );
+};
+
+const SelectModal = () => {
+  const [open, setOpen] = React.useState(false);
```
NicholasBoll, line 36: There's a `useModal` hook for just such an occasion: https://github.com/Workday/canvas-kit/blob/53b7f49e96540850c48c90465fe7ae9498e6953c/modules/modal/react/stories/stories.tsx#L20-L37

modules/_labs/select/react/stories/stories_testing.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const buttonRef = React.useRef<HTMLButtonElement>() as React.RefObject<HTMLButtonElement>; // cast to keep buttonRef happy
+  const openModal = () => {
+    setOpen(true);
+  };
+  const closeModal = () => {
+    setOpen(false);
+    if (buttonRef.current) {
+      buttonRef.current.focus();
```
NicholasBoll, line 44: Not sure why this line is here? The Modal should take care of focus changes for you.

modules/_labs/select/react/stories/stories_visualTesting.tsx
NicholasBoll: It seems like it might be related to how browsers choose to handle sub-pixel positioning like Flexbox.

Chrome seems to snap to whole numbers while Firefox allows fractional values.

Here's `getBoundingClientRect()` in Chrome:
Screen Shot 2020-06-11 at 2 53 40 PM

Here's `getBoundingClientRect()` in Firefox:
Screen Shot 2020-06-11 at 2 53 51 PM

Subpixel values are happening in Firefox on both normal and high-DPI displays.

NicholasBoll: From a test of one not showing correctly:

CSS
```css
transform: translate(196px, 345px);
```

`getBoundingClientRect()`
```ts
{
  x: 195.39999389648438
  y: 309
}
```

NicholasBoll: It appears low-DPI displays will snap to pixel values making this a rounding error.

High-DPI displays can do sub-pixel rendering in Firefox (but not Chrome). Overriding the `transform` value to the exact same value as the `x` makes it render properly in both High and low DPI in Firefox.  Perhaps there's a bug in Popper?

NicholasBoll:
Overall, I'm concerned with the number of props being added to the menu as well as some of the prop names.

modules/_labs/select/react/lib/SelectBase.tsx
jamesfan: Renamed `isMenuAutoFlipped` to `shouldMenuAutoFlip`.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Renamed `isAutoFocused` to `shouldAutoFocus` (as well as `isAnimated` to `shouldAnimate`, and `isAutoFlipped` to `shouldAutoFlip`) to better convey intent.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Fixed.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Fixed.

modules/_labs/select/react/lib/scrolling.ts
jamesfan: Fixed.

modules/_labs/select/react/stories/stories_testing.tsx
jamesfan: Used the `useModal` hook.

modules/_labs/select/react/stories/stories_testing.tsx
jamesfan: Removed `buttonRef` now that we're using `useModal`.

modules/_labs/select/react/stories/stories_testing.tsx
jamesfan: Not sure, I copied code from an older (pre-stack-manager) Modal story without much thought. This code no longer exists now that we're using `useModal`.

modules/_labs/select/react/lib/SelectBase.tsx
jamesfan: That's correct. I'll note this down to tackle later.

modules/_labs/select/react/lib/SelectBase.tsx
jamesfan: Agreed, the visual testing props are a contributor. Will look into this in the future.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: `isHiding` indicates the menu is animating (fading) out. `isHidden` indicates the menu is completely hidden.

jamesfan:
> Overall, I'm concerned with the number of props being added to the menu as well as some of the prop names.

I've renamed the prop names in question to better convey intent. Agreed on there being too many props for the menu -- I'll look into context as a possible solution for the future.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: I'm not seeing a flickering effect here. Are you seeing something on your end?

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Filing this away for later.

modules/_labs/select/react/lib/SelectMenu.tsx
lychyi: Yup. _Technically_, it can flicker because the width state of the menu initializes at 0 and then gets calculated and set asynchronously.

Yet, I'm not sure if that flicker ever shows up through regular usage of a `Select`. We animate the menu in via a fade so if the menu width is 0 for one frame... it won't be perceptible.

Clicking through tests was the only way I could get it to show up. If you click on the visual testing story "Select States Menu On" you can see the 1 frame difference between `useEffect` and `useLayoutEffect`.

Here's me scrubbing through a timeline in Chrome (very slowly). You'll see that the menu is supposed to be open when the story loads but it takes one frame before it shows.

Screen-Recording-2020-06-11-at-11 10 44-PM

Here it is with `useLayoutEffect`, the menu paints synchronously to having its width set.

Screen-Recording-2020-06-11-at-11 14 53-PM

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: What would be the advantage of a union over a boolean in this case?

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Do you mean having `generatePopperOptions` return like so:
```
return {
  modifiers: [
    {
      name: 'offset',
      options: {
        offset: ({reference}) => {
...
```
Or even inlining `modifiers` within the `popperOptions` prop:
```
<Popper
  popperOptions={{
    modifiers: [
      {
        name: 'offset',
        options: {
          offset: ({reference}) => {
...
```
In both cases, TS complains about `reference` implicitly having an `any` type.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Done.

modules/_labs/select/react/package.json
jamesfan: Removed `@popperjs/core` as a dependency now that `Rect` is exported from the CK Popper.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Ah, I didn't realize that was the flicker you were referring to. I had noticed the menu delay earlier when pulling up the Menu On story, but I didn't realize it was due to `useEffect`. @lychyi Thanks for the detailed explanation! I've replaced `useEffect` with `useLayoutEffect`.

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -304,8 +299,8 @@ const SelectMenu = (props: SelectMenuProps) => {
   // Render nothing if buttonRef.current hasn't been set
   return buttonRef.current ? (
     <Popper
-      placement={isFlipped ? 'top-start' : 'bottom-start'}
-      open={!isHidden}
+      placement={isFlipped ? 'top' : 'bottom'}
+      open={!isHidden && width !== 0}
```
NicholasBoll, line 303: This makes sure the Popper is not activated until the width of the button has been calculated since we're using `top` and `bottom` which are "centered" Popper types.

`top-start` and `bottom-start` are incorrectly rounded which causes problems for HighDPI displays in Firefox and Edge. The problem comes from the following lines in Popper.js: https://github.com/popperjs/popper-core/blob/master/src/utils/computeOffsets.js#L68-L81

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -124,6 +135,25 @@ function getFirstElementToFocus(overlayEl: HTMLElement): HTMLElement {
...
+    window.addEventListener('resize', onResize);
+
+    return () => {
+      window.removeEventListener('resize', onResize);
+    };
+  }, []);
+
+  return {width, height};
+};
+
```
NicholasBoll, line 156: This should eventually be broken out into a sharable hook.

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -178,17 +209,21 @@ const ModalContent = ({
...
-      <Popup
-        width={width}
-        heading={heading}
-        handleClose={handleClose}
-        padding={padding}
-        transformOrigin={transformOrigin}
-        aria-modal={true}
-        ariaLabel={ariaLabel}
+      <CenteringContainer
+        style={{width: windowSize.width % 2 === 1 ? 'calc(100vw - 1px)' : '100vw'}}
```
NicholasBoll, line 213: One of the worse things about the `styled` API is what it does for dynamic styles. Emotion will inject or remove stylesheets from the page when a dynamic style changes which will trigger a full-page style recalculation. Sometimes that doesn't matter because the styles don't change often, but the habit can have huge negative performance implications. An inline style for critical styles prevents full page style recalculations.

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -208,13 +208,8 @@ const generatePopperOptions = (
...
       options: {
-        offset: ({reference}: {reference: Rect}) => {
-          // Skid menu along the edge of the button to account
-          // for fractional x-positioning of the button (e.g.,
-          // if the button is in a horizontally-centered modal)
-          // and to ensure proper alignment between the button
-          // and the menu
-          const skidding = reference.x % 1 >= 0.5 ? 1 : 0;
+        offset: () => {
+          const skidding = 0;
```
NicholasBoll, line 212: We don't use `skidding` anymore. Skidding originally fixed a Chrome issue where some elements would render subpixel in some circumstances by not others. For instance, the centered dialog element in the Modal component would receive a `0.5px` subpixel offset if the window had an odd number for its width, but placing a Popper with the same subpixel offset would round to a whole number. Other browsers do not suffer this rounding issue.

The Modal now compensates by forcing an even number when centering the modal dialog element so we don't need this offset anymore. It didn't work in all circumstances anyways.

modules/_labs/select/react/lib/SelectMenu.tsx
NicholasBoll: Booleans convey less meaning and context that string literals. What does "isFlipped" do? Is positioning on the top of the target mean `isFlipped=true` or `isFlipped=false`? Sure, the JSDoc may explain what `true` and `false` mean, but now you have to go to the docs every time. It just is less obvious and thus more mental energy to understand.

modules/_labs/select/react/lib/SelectMenu.tsx
NicholasBoll: I see. Now we're getting into state machine territory. You can usually tell if you have a bunch of booleans to determine state. But we have a weird issue here. Not all permutations of `isHidden` and `isHiding` are valid. We only have 3 states, but there are 4 permutations.

For example,
|  | isHidden=true  | isHidden=false  |
|---|---|---|
| isHiding=true  | N/A | Currently hiding |
| isHiding=false  | Hidden | Not Hidden |

Also I noticed a lot of negative variable names. Example `isHidden` vs `isOpen`. It is generally considered a bad practice as negating a negative variable name results in a double-negative that is harder to understand: https://softwareengineering.stackexchange.com/questions/196830/boolean-method-naming-affirmative-vs-negative

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -133,21 +188,152 @@ const MenuList = styled('ul')<Pick<SelectProps, 'error' | 'theme'>>(
...
+  const handleWidthChange = () => {
+    if (buttonRef.current) {
+      const newMenuWidth = buttonRef.current.clientWidth + 2 * buttonBorderWidth;
+      setWidth(newMenuWidth);
+    }
+  };
+
+  useLayoutEffect(() => {
+    handleWidthChange();
+  }, [buttonRef]);
```
NicholasBoll, line 272: An optimization here would be to only measure the width when the menu is open rather than on initialization (the most performance sensitive phase). The optimization would be as follows:

```suggestion
  const handleWidthChange = () => {
    if (buttonRef.current && !isHidden) { // Oops, a double-negative - "not is hidden" means "is open", right?
      const newMenuWidth = buttonRef.current.clientWidth + 2 * buttonBorderWidth;
      setWidth(newMenuWidth);
    }
  };

  useLayoutEffect(() => {
    handleWidthChange();
  }, [buttonRef, isHidden]);
```

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -133,21 +188,152 @@ const MenuList = styled('ul')<Pick<SelectProps, 'error' | 'theme'>>(
...
+
+  // TODO: Figure out a better way to handle width changes in the reference button.
+  // Seems like we should resize the menu when the reference button width changes, not
+  // necessarily when the window resizes. Resizing the menu on window resize addresses
+  // the case when `grow = true` and the user resizes the browser window while the
+  // menu is open, but doesn't address cases where the reference button size changes
+  // through other means.
```
NicholasBoll, line 279: I'm guessing this is where we want a way to force Popper to update? Popper will update on window resize, but not when the target element width changes.

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -133,21 +188,152 @@ const MenuList = styled('ul')<Pick<SelectProps, 'error' | 'theme'>>(
...
+  // correct behavior for now, we're dismissing the Menu on blur instead of
+  // using the hook.
+
+  // Make sure we only open the Popper when we know the width of the button element. We need to use
+  // `top` and `bottom` instead of `top-start` and `bottom-start` placements because PopperJS
+  // incorrectly rounds `start` and `end` modifiers:
+  // https://github.com/popperjs/popper-core/blob/38914aae7a2e91715c6eb2b563517082a40cfa64/src/utils/computeOffsets.js#L68-L81
+  // This rounding causes problems with browsers that allow subpixel values for elements like
+  // Firefox and Edge.
+  const open = !isHidden && width !== 0;
```
NicholasBoll, line 305: Double-negatives...

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Correct.

modules/_labs/select/react/lib/SelectMenu.tsx
anicholls: +1. I was also confused by the terminology "flipped"

modules/modal/react/lib/ModalContent.tsx
anicholls: @NicholasBoll I feel like this comment should be inline, otherwise it could easily be mistakenly turned into `css` or a styled component prop

NicholasBoll: I think that should be documented at a higher level. I helped another team with a major performance issue when using `styled`/`css` for x/y coordinates while dragging. The default is to put dynamic styles in Emotion, but we should probably recommends putting dynamic styles in a `style` attribute instead. I think certain properties are dynamic for the sake of the API, but won't change in practice. Like theme or our design system constants.

Maybe a best practice around using Emotion or CSS-in-JS. I could add that comment here, but it would likely be missed and we accidentally introduce performance problems in applications where we may not notice as much because our examples have less styles and DOM elements than an application.

modules/_labs/select/react/lib/SelectMenu.tsx
NicholasBoll: I updated the Menu to use `bottom` instead of `bottom-start` which means Popper could open before the width is set causing issues where the wrong position was chosen. I could see this in IE11 where the menu would open with a `0` width causing the Popper's left side to be in the middle of the button.

I fixed this issue by keeping the Popup closed until width was known. This would cause the issue @lychyi showed. This wouldn't be a problem normally as chances are a user could not open a menu faster than `useEffect` calculating width.

https://github.com/Workday/canvas-kit/pull/641/commits/0fa2671a701ba4e4a284afe4c07f72988fff7ab8 introduced an optimization that delays the calculation of width to when we want to open the menu, making it possible for IE to choose the wrong location. But the Popper doesn't actually "open" until width is non-zero, so this shouldn't happen in practice, but `useLayoutEffect` does prevent the extra frame.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Removed the `width !== 0` check in #743 since it's no longer necessary now that we're using `useLayoutEffect` instead of `useEffect` when setting the width of the menu (the check was preventing a newly activated menu from scrolling to the selected item if that item was initially out of view).

Issue #570, opened 2020-04-14
Add theming support to Select


jpante:
## 🚀 Feature Proposal

Add theming support to the Select component

### Motivation

Add theming to the final core input component for theming support on product teams.

jamesfan:
Merged into `prerelease/v4`.

PR #648, opened 2020-05-12
feat(select): Add theming to select in labs

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/570

Add theming support for select

## Checklist

- [ ] `yarn test` passes
      applicable


modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -192,25 +185,46 @@ const SelectButton = styled('button')<Pick<SelectBaseProps, 'error' | 'grow' | '
...
+  ({theme}) => {
+    const themedFocusOutlines = theme.canvas.palette.common.focusOutline;
+    return {
+      '&:focus:not([disabled])': {
+        borderColor: themedFocusOutlines,
+        boxShadow: `inset 0 0 0 1px ${themedFocusOutlines}`,
+        outline: 'none',
+      },
+    };
+  },
```
mannycarrera4, line 197: I tried to move this bit of code down below rather than having two places where theme is being called, however, it wasn't adding the focus state to the select, any idea why?

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -35,18 +36,28 @@ const fadeOutAnimation = keyframes`
...
-    dividerBorderWidth = 2;
+    if (theme) {
+      borderColor = theme.canvas.palette.alert.main;
+      dividerBorderColor = theme.canvas.palette.alert.main;
+      dividerBorderWidth = 2;
+    } else {
+      borderColor = colors.cantaloupe600;
+      dividerBorderColor = inputColors.warning.border;
+      dividerBorderWidth = 2;
+    }
```
mannycarrera4, line 60: Same as the code below, theme had to be passed to be able to style different things. Both the methods are different enough however, I would like to simplify these into a cleaner way so I'm open to suggestions

modules/_labs/select/react/stories/stories_VisualTesting.tsx
```diff
@@ -0,0 +1,190 @@
...
+  return {
+    data: {},
+    disabled: option.disabled || false,
+    id: option.value,
+    label: option.label || option.value,
+    value: option.value,
+  };
+});
+
+storiesOf('Labs|Select/React/Visual Testing', module)
```
mannycarrera4, line 38: Should i change this to storybook csf?

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -35,18 +36,28 @@ const fadeOutAnimation = keyframes`
...
   if (error === ErrorType.Error) {
-    borderColor = inputColors.error.border;
-    dividerBorderColor = inputColors.error.border;
+    if (theme) {
+      borderColor = theme.canvas.palette.error.main;
+      dividerBorderColor = theme.canvas.palette.error.main;
+    } else {
+      borderColor = inputColors.error.border;
+      dividerBorderColor = inputColors.error.border;
+    }
```
mannycarrera4, line 50: i can probably just to a ternary here?

modules/_labs/select/react/lib/SelectMenu.tsx
mannycarrera4, line 60: same here, do a ternary that way I don't have to specify the border twice

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -36,28 +36,18 @@ const fadeOutAnimation = keyframes`
...
   } else if (error === ErrorType.Alert) {
-    if (theme) {
-      borderColor = theme.canvas.palette.alert.main;
-      dividerBorderColor = theme.canvas.palette.alert.main;
-      dividerBorderWidth = 2;
-    } else {
-      borderColor = colors.cantaloupe600;
-      dividerBorderColor = inputColors.warning.border;
-      dividerBorderWidth = 2;
-    }
```
mannycarrera4: if no theme is provided, there's no default color which is why i initially added the if!

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: I made `theme` a required argument, so no need for a ternary any more.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Same as above: `theme` is now a required argument.

modules/_labs/select/react/lib/SelectBase.tsx
jamesfan: Hmm, my guess would be the focused button styles weren't being applied somewhere. Should be working now though with the recent changes.

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: Cleaned up with recent changes.

modules/_labs/select/react/stories/stories_VisualTesting.tsx
jamesfan: Saving CSF for later after visual testing stories have stabilized a bit from changes in other PRs (#606 and #641 ).

modules/_labs/select/react/lib/SelectMenu.tsx
jamesfan: `theme` should always be provided now that it's a required argument to the helper function.

modules/_labs/select/react/lib/SelectOption.tsx
```diff
@@ -62,12 +63,17 @@ const Option = styled('li')<SelectOptionProps>(
...
     } else {
       // ...else return hover and selected (via aria-selected) styles
       const selectedStyles = {
         '&[aria-selected="true"]': {
+          // Currently, there doesn't seem to be an appropriate color for
+          // the background of a selected option in the canvas theme object.
+          // `primary.lightest` happens to match from the default theme, but
+          // I'm not sure the selected color should be based off of the primary
+          // color.
           backgroundColor: colors.blueberry100,
```
jamesfan, line 77: Any thoughts on which (if any) theme color we should use for the background color of a selected option? The current color of the selected option (`blueberry100`) happens to match `primary.lightest` of the default canvas theme, but it doesn't seem like the selected color should be based off of the primary color (if anything, it should be based off of `common.focusOutline`, or maybe there needs to be a new color in `common` which can be used as a background color for selected things).

modules/_labs/select/react/stories/stories.tsx
```diff
@@ -322,4 +326,10 @@ storiesOf('Labs|Select/React/Visual Testing', module)
...
       </ComponentStatesTable>
     </StaticStates>
-  ));
+  ))
+  .addParameters({
+    canvasProviderDecorator: {
+      theme: customColorTheme,
+    },
+  })
+  .add('Theming', () => <SelectStates />);
```
jamesfan, line 335: I feel like there should be Theming states tables for `States (Menu On)` and `States (Options)` as well, but I'm leaning towards handling them later given both tables are being modified considerably in other PRs (`States (Menu On)` in #641, and `States (Options)` in #606).

modules/_labs/select/react/lib/SelectOption.tsx
anicholls: I would use `primary.lightest` for now, but I agree with you. We'll probably do a larger audit for how theming colors are applied in the future.

jamesfan: Ok, get ready for some unattractive Theming states tables. 🤣 Like purple and turquoise (I think that's what's being used in `customColorTheme`).

modules/_labs/select/react/stories/stories.tsx
jamesfan: I went ahead and added the states tables. We'll resolve conflicts later.

Issue #640, opened 2020-05-08
ActionBar component throw an error message to dev console


xTena:
## 🐛 Bug Report

Recently update to lastest version of @workday/canvas-kit-react-action-bar ActionBar component and when the component is render the following error is thrown in the dev console:

The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type".

## To Reproduce

Render ActionBar component

## Expected Behavior

No errors in dev console

## Actual Results

![Screenshot 2020-05-08 at 18 10 08](https://user-images.githubusercontent.com/55538167/81430372-32484200-9157-11ea-877e-7fa349fe65c7.png)

I would like to contribute!!! Possible solution: 
- Replace lines 35 and 45 of ActionBar.tsx with ':first-of-type':

PR #649, opened 2020-05-13
feat: Support SSR

anicholls:
## Summary

Currently our components don't technically support server-side rendering (eg. Gatsby). This PR updates several components/modules so that we don't try to access `document` or `window` during the initial render (before things are mounted). It also swaps `:first-child` for `:first-of-type` since that is an issue with Emotion SSR.

I've added [`eslint-plugin-react-ssr`](https://github.com/ytanruengsri/eslint-plugin-react-ssr) to catch some of these errors in the future. However, it doesn't cover every single case so I [opened an issue](https://github.com/ytanruengsri/eslint-plugin-react-ssr/issues/2) to have this addressed. I tried adding our own custom linting rule, which errors on any references to `window` and `document` outside `componentDidMount`, `useEffect`, and `useLayoutEffect` if they are not checked for `undefined`. However, I ran into issues with an "allow list" approach instead of `react-ssr`'s "deny list" approach.

To help protect us against SSR regressions, I have added simple SSR tests that use jest's node environment + react-doms `renderToString` function for all of our components that reference `window` or `document`. We probably want to add this coverage for all components in the future, but for now I figured this was good enough. I also added this as a template in the create-component script.

Closes: #640 

---

Also a gotcha I discovered for anyone working on custom linting rules in the future - you need to run `yarn add -WD file:./utils/custom-lint-rules` after each change for your updates to be reflected in `node_modules/`


lychyi:
I'd definitely call this a feat :)

Anything else we need to be aware of for SSR? I would probably include documentation and a linter too as part of this PR if it's not too much trouble. It would be a shame to make it SSR-ready and then regress after another PR.

alexandrzavalii:
is there anything blocking this PR?

alexandrzavalii:
we are not doing SSR, we are just trying to get rid of any console errors/warnings

anicholls:
Tests are failing due to some `PopupStack` SSR issues. Waiting for the fixes for these to be contributed by @NicholasBoll

NicholasBoll:
Finishing up #535 and then I'll update this PR with the SSR fix

modules/modal/react/spec/SSR.spec.tsx
```diff
@@ -0,0 +1,13 @@
+/**
+ * @jest-environment node
+ */
+import React from 'react';
+import {renderToString} from 'react-dom/server';
+import Modal from '../';
+
+describe('Modal', () => {
+  it('should render on a server without crashing', () => {
+    const ssrRender = () => renderToString(<Modal open={true} heading="test" />);
```
NicholasBoll, line 10: Will we ever render an open `Modal` from the server? Does that even make sense? Typically modals are triggered by some type of client event.

anicholls: Probably not, I was just trying to properly test `ModalContent` since it references `document`. Could drop this prop if you think that makes more sense

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -113,16 +113,78 @@ interface Stack {
   };
 }
 
+/**
+ * Get a deeply nested dot-notation path from an arbitrary object safely. Will return `undefined` if
+ * path does not exist. This function is not meant to be type-safe. Use with caution.
+ * @param obj Any object
+ * @param path dot-notation path of a deep property
+ */
+function get(obj: any, path: string): any {
```
anicholls, line 122: @NicholasBoll I assume you didn't use lodash's `get` because you didn't want to add the dependency for this module?

NicholasBoll: Yes. The following code would actually compress better though (since spreads can explode in Babel for IE11 support):

```ts
function get(obj: any, path: string): any {
  const parts = path.split('.')
  const first = parts.splice(0, 1)[0];
  if (parts.length && obj[first]) {
    return get(obj[first], parts.join('.'));
  } else {
    return obj[first];
  }
}

function set(obj: any, path: string, value: any): any {
  const parts = path.split('.')
  const first = parts.splice(0, 1)[0];
  if (parts.length) {
    if (obj[first] === undefined) {
      obj[first] = {};
    }
    set(obj[first], parts.join('.'), value);
  } else {
    obj[first] = value;
  }
  return value;
}
```

What do you think?

Here's the difference in Babel's REPL:

With Spreads:
```js
function _toArray(arr) { return _arrayWithHoles(arr) || _iterableToArray(arr) || _unsupportedIterableToArray(arr) || _nonIterableRest(); }

function _nonIterableRest() { throw new TypeError("Invalid attempt to destructure non-iterable instance.\nIn order to be iterable, non-array objects must have a [Symbol.iterator]() method."); }

function _unsupportedIterableToArray(o, minLen) { if (!o) return; if (typeof o === "string") return _arrayLikeToArray(o, minLen); var n = Object.prototype.toString.call(o).slice(8, -1); if (n === "Object" && o.constructor) n = o.constructor.name; if (n === "Map" || n === "Set") return Array.from(o); if (n === "Arguments" || /^(?:Ui|I)nt(?:8|16|32)(?:Clamped)?Array$/.test(n)) return _arrayLikeToArray(o, minLen); }

function _arrayLikeToArray(arr, len) { if (len == null || len > arr.length) len = arr.length; for (var i = 0, arr2 = new Array(len); i < len; i++) { arr2[i] = arr[i]; } return arr2; }

function _iterableToArray(iter) { if (typeof Symbol !== "undefined" && Symbol.iterator in Object(iter)) return Array.from(iter); }

function _arrayWithHoles(arr) { if (Array.isArray(arr)) return arr; }

function set(obj, path, value) {
  var _path$split = path.split('.'),
      _path$split2 = _toArray(_path$split),
      first = _path$split2[0],
      rest = _path$split2.slice(1);

  if (rest.length) {
    if (obj[first] === undefined) {
      obj[first] = {};
    }

    set(obj[first], rest.join('.'), value);
  } else {
    obj[first] = value;
  }

  return value;
}
```

With `splice`:
```js
function get(obj, path) {
  var parts = path.split('.');
  var first = parts.splice(0, 1)[0];

  if (parts.length && obj[first]) {
    return get(obj[first], parts.join('.'));
  } else {
    return obj[first];
  }
}
```

NicholasBoll: That compresses to like 400 bytes with no depencencies

anicholls: If the behavior works, I don't think anyone will ever look too closely at these functions so I'm not too worried about the perceived complexity between the two. In this case I'd probably learn towards `splice` if we're trying to keep this package super minimal 👍

PR #651, opened 2020-05-13
fix: Align hover ripple for checkbox and radio components in IE11

### LINKED ISSUE #650: Circular shape on Hover next to All radio buttons and checkboxes in IE
## To Reproduce

See Chromatic Report - https://www.chromatic.com/component?appId=5d854c26ba934e0020f5e98a&name=Components%7CInputs%2FCheckbox%2FReact%2FVisual%20Testing&mode=snapshot&buildNumber=1573&browser=ie

## Expected Behavior

No Hover effect should be centered on the checkbox and radio button in IE
correct

## Actual Results

Hover effect is not centered and is offset in IE

error

## Browser (if applicable)

IE11

lychyi:
## Summary

Fixes #650 

## Explanation
I'll do my best to explain how this fix came to be for both `Radio` and `Checkbox` because they are implemented similarly.

### Align browsers by explicit positioning (i.e. change top: auto to top: 0px)
The first thing was to fix ripple positioning in IE11. This means that we have to make sure that the ripple is placed intentionally by setting `top: 0px` instead of the default `top:auto` which is browser controlled.

### Unfortunately, that introduces offset errors due to border-width changes
Hang on, not that simple though. The width of the border for the parent element (`RadioBackground`, `CheckboxBackground`) will render with offset errors on all browsers when this happens.

Why? This is because absolutely positioned element coordinates are based on the parent's content not the content + border width (border-box doesn't help this). So to align correctly you have to account for the border's width, i.e. `top: -1px`. If the border changes to `2px` (which it does), this becomes problematic since you also have to adjust the ripple's positioning. They're tightly coupled.

I tried, and as I adjusted for the changing `border-width` the visual state tables would look correct, I got offset errors while transitioning between states with different border sizes! This is not good and really hard to reason about and impossible to catch with our tests today.

Offset errors look something like this:

Screen Shot 2020-05-20 at 7 23 16 AM

### Solution: Move it so it wouldn't be influenced by changes in parent's border-width
The solution was to move the ripple to a place where it wouldn't be influenced by changes in `border-width` from the background components. Moving it to be a sibling of the background component makes positioning a bit easier to reason about since changes in one doesn't affect the other.

### Still needs to be connected to the input though
But one requirement is that the ripple animates via the `input` element's state changes (hover, active, disabled, etc.) so instead of a keeping it as a pseudo element (originally `::after`), it was changed to a concrete element (`span`) so it could be targeted in a CSS sibling selector (`~ span:first-of-type`).

### An alternative solution
An alternative was to transition everything to `box-shadow`s because that doesn't influence positioning. You can't change individual components of it like you can a border (`border-color`, `border-width`, etc.) so that change worked but was also more code and nasty-looking ternaries when we wanted to have new `box-shadow`s based on props.

## Checklist

- [x] `yarn test` passes (No tests were added because this regression already exists as a baseline 😓 )
- [x] Fix for `Radio`
- [x] Fix for `Checkbox`
- [ ] Inconsistencies with the Figma spec (could be a different issue/PR)

### DIFF
--- a/modules/radio/react/lib/Radio.tsx
+++ b/modules/radio/react/lib/Radio.tsx
@@ -84,7 +84,7 @@ const RadioRipple = styled('span')<Pick<RadioProps, 'disabled'>>({
   width: radioWidth,
   position: 'absolute',
   pointerEvents: 'none',
-  zIndex: 1,
+  zIndex: -1,
 });
 
 const RadioInput = styled('input')<RadioProps>(
--- a/modules/radio/react/lib/Radio.tsx
+++ b/modules/radio/react/lib/Radio.tsx
@@ -105,6 +105,8 @@ const RadioInput = styled('input')<RadioProps>(
       transition: 'box-shadow 150ms ease-out',
       width: radioWidth,
       position: 'absolute',
+      top: -1,
+      left: -1,
       zIndex: -1,
     },
   },
--- a/modules/radio/react/lib/Radio.tsx
+++ b/modules/radio/react/lib/Radio.tsx
@@ -82,7 +82,7 @@ const RadioRipple = styled('span')<Pick<RadioProps, 'disabled'>>({
   transition: 'box-shadow 150ms ease-out',
   width: radioWidth,
   position: 'absolute',
-  pointerEvents: 'none',
+  pointerEvents: 'none', // This is a decorative element we don't want it to block clicks to input
   zIndex: -1,
 });
 
--- a/modules/radio/react/lib/Radio.tsx
+++ b/modules/radio/react/lib/Radio.tsx
@@ -78,7 +78,6 @@ const RadioInputWrapper = styled('div')<Pick<RadioProps, 'disabled'>>({
 const RadioRipple = styled('span')<Pick<RadioProps, 'disabled'>>({
   borderRadius: borderRadius.circle,
   boxShadow: `0 0 0 0 ${colors.soap200}`,
-  content: '""',
   height: radioHeight,
   transition: 'box-shadow 150ms ease-out',
   width: radioWidth,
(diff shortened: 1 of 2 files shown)


lychyi:
Do not merge, bugs still be out there 🐛

modules/radio/react/lib/Radio.tsx
```diff
@@ -75,10 +75,18 @@ const RadioInputWrapper = styled('div')<Pick<RadioProps, 'disabled'>>({
...
- */
+const RadioRipple = styled('span')<Pick<RadioProps, 'disabled'>>({
+  borderRadius: borderRadius.circle,
+  boxShadow: `0 0 0 0 ${colors.soap200}`,
+  content: '""',
+  height: radioHeight,
+  transition: 'box-shadow 150ms ease-out',
+  width: radioWidth,
+  position: 'absolute',
+  pointerEvents: 'none',
```
lychyi, line 86: Don't want the ripple to be clickable. It alternates from being behind the `RadioBackground` to in front depending on the state of the input.

NicholasBoll: I think that's fine. The radio input is in the same position with a opacity of 0, right? If you think that's a questionable point, you could add a comment. Cypress will also catch errors here if we have a test that makes sure clicking on the input works. It would fail if something is preventing the click from reaching the input.

lychyi: Inline comment added. That's a good call-out.

lychyi:
BTW, Chromatic is not catching changes with Checkbox even though I'm clearly fixing it (changing it) here. It's not catching all the changes with Radio either. The diffs don't catch the hover ripple moving places which is a little troubling. I've opened up an issue with the Chromatic team, for now we'll just have to use our eyes to make sure the changes are good to go. 👀

PR #652, opened 2020-05-13
chore: Add invalid import rule

### LINKED ISSUE #581: Add lint rule to avoid invalid imports
## 🚀 Feature Proposal

Add a custom lint rule to avoid importing across modules. A common mistake is to include a file from another module that is not exported from the `index` file. 

## Motivation

This issue is a result of #575. It is easy to accidentally import from another package's `lib` directory which will not exist in a deployed package (it is a `dist` folder after compilation`). A lint rule can be a low-effort reminder that this is will cause issues.

## Example

```tsx
import { ColorSwatch } from '@workday/canvas-kit-react-color-picker/lib/parts/ColorSwatch.tsx'
// this line will cause a lint error that will suggest exporting `ColorSwatch` from the main export of `@workday/canvas-kit-react-color-picker`
```

alanbsmith:
## Issue

Resolves #581

## Summary

Adds a custom lint rule that flags workday imports from `/lib`.

`utils/custom-lint-rules/restricted-imports.js`

## Testing Manually

You'll need to `yarn install` to add this lint rule. It's not an external dependency, but it's in the package.json

Run the test:
```sh
yarn test utils/custom-lint-rules/ --coverage=false
```

Or break it yourself: 😄 

Add something like this to a component:
```js
import radius from '@workday/canvas-kit-react-core/lib/radius';
```
And run `yarn lint`

## New Dependencies

n/a

## Checklist

- [x] `yarn test` passes
      applicable


package.json
```diff
@@ -78,6 +78,7 @@
     "eslint-plugin-prettier": "^3.1.1",
     "eslint-plugin-react": "^7.16.0",
     "eslint-plugin-react-hooks": "^2.1.2",
+    "eslint-plugin-workday-custom-rules": "./utils/custom-lint-rules/",
```
NicholasBoll, line 81: TIL

alanbsmith: Same, honestly.

Issue #610, opened 2020-05-01
Add toolbar square icon


mannycarrera4:
## 🚀 Feature Proposal

There are various cases through out Workday where there's a format bar. A lot of teams have heavily customized our icons and icon buttons to fit their use case. I propose we add another variant to our icon button that allows for specific styling and proper size. By default our IconButtons are 40px by 40px and when used with the focus ring, becomes too large for a small area like a format bar. 

Attached below are some examples and proposed solutions.
For reference: 
![image (2)](https://user-images.githubusercontent.com/7966550/82371449-73eaaf80-99d7-11ea-82d0-5939374974fc.png)

mannycarrera4:
Closing since this is now merged into `v4`

PR #658, opened 2020-05-15
feat(button): Add Toolbar Square Icon

mannycarrera4:
## Summary

Fixes:https://github.com/Workday/canvas-kit/issues/610
This new alternate button provides some uses cases for Toolbars and format bars. Al though it's similar, the toggled on state is quite different.
## Checklist

- [x] `yarn test` passes


modules/button/react/lib/ButtonStyles.ts
```diff
@@ -335,6 +335,11 @@ export const iconButtonStyles: ButtonGenericStyle = {
         height: canvas.spacing.l,
         ...getIconButtonStateStyle(IconButtonVariant.Square),
       },
+      [IconButtonVariant.Alternate]: {
+        borderRadius: borderRadius.m,
+        width: '32px !important', // override width auto for small buttons
```
mannycarrera4, line 340: any reason why we have width auto on icon button size small? All of our icons are always 24 by 24

modules/button/react/lib/IconButton.tsx
```diff
@@ -76,6 +76,11 @@ export const IconButtonCon = styled('button', {
   iconButtonStyles.styles,
   ({variant}) => getButtonStyle(iconButtonStyles, variant),
   ({size, variant}) => {
+    if (variant === IconButtonVariant.Alternate) {
+      return {
+        ...iconButtonStyles.variants!.sizes.small,
+      };
+    }
```
mannycarrera4, line 83: Basically always default to small is there a better way to do this?

modules/button/react/lib/types.ts
```diff
@@ -44,6 +44,7 @@ export enum IconButtonVariant {
   CircleFilled = 'circleFilled',
   Inverse = 'inverse',
   InverseFilled = 'inverseFilled',
+  Alternate = 'alternate',
```
mannycarrera4, line 47: Maybe `SquareAlternate` ?

modules/button/react/lib/utils.ts
```diff
@@ -245,7 +249,7 @@ export function getButtonStateStyle(variant: AllButtonVariants): CSSObject {
 
 export function getIconButtonStateStyle(variant: AllButtonVariants): CSSObject {
   const buttonColors = ButtonColors[variant];
-
+  console.warn(buttonColors);
```
mannycarrera4, line 252: remove

modules/button/react/README.md
```diff
@@ -184,15 +184,16 @@ import {activityStreamIcon} from '@workday/canvas-system-icons-web';
...
-| `Square`        | Square, white background, dark gray icon         |
-| `SquareFilled`  | Square, gray background, dark gray icon          |
-| `Plain`         | Dark gray icon                                   |
-| `Circle`        | Circular, dark gray icon                         |
-| `CircleFilled`  | Circular, gray background, dark gray icon        |
-| `Inverse`       | Circular, transparent background, white icon     |
-| `InverseFilled` | Circular, semitransparent background, white icon |
+| Theme           | Description                                                |
+| --------------- | ---------------------------------------------------------- |
+| `Alternate`     | Square, white background, dark gray icon used for toolbars |
```
mannycarrera4, line 189: I made this specific to toolbars because it has specific styles for that reason, thoughts on the wording? @jpante

modules/button/react/lib/types.ts
lychyi: Naming things is part art, part best practices, and really hard _sometimes_. With button variants, we're at capacity for intuitive names. So disclaimer: I'm not sure I have a good suggestion right now.

My opinion on things to think about when naming things like props, enums, etc., would be curious to hear what others have to say or add:
- **The name should attempt to stand alone. Or be clear on what it relates to.** When using `Alternate` or `SquareAlternate`, it just sounds like you're saying `Other`.
- **It should try to accommodate future changes if/when they happen.** The word `Alternate` is sort of a catch-all. It doesn't allow us to have any expansion in our variants in the future.
- **It should be descriptive without locking you into a particular implementation.** `Alternate` doesn't lock us into a particular implementation (like if `Primary` was named `Blue` instead) but it isn't particularly descriptive. It's almost too vague.

jamesfan: Is `Toolbar` too specific? It satisfies #*1* in Lynn's comment above. As far as #*2* and #*3*, can anyone think of another context where a button like this would be used outside of a Toolbar (or anything which could reasonably be called a Toolbar)?

Another suggestion would be `SquareCompact`, but I wonder if it should even use the shared `Square` prefix given all the visual differences in states between this variant and all the current `Square*` variants -- in addition to the differences in the Toggled states, it looks like Focus and Click are also different on their own.

mannycarrera4: i liked the suggestion of `compact` as far as I'm aware a lot of products in workday would be using this as a sub set to a toolbar, it's distinct enough and it avoids confusion with the other variants, another option was `groupedSquare`

mannycarrera4:
Closing this PR, after discussion with design and development, we'll be changing this against pre release v4, updating some styles, creating a new variant for icon button that is specific for tool bar and adding that to the documentation. Will open up a new PR against pre release with the updated issue based on the design discussion.

PR #660, opened 2020-05-19
chore: Fix create-module and promote-module

NicholasBoll:
The API of mkdirp changed with version 1.0 from callbacks to promises, but we didn't update our use-cases. The change of version was in #531. I tested this with a new module

### DIFF
--- a/utils/create-component/createComponent.js
+++ b/utils/create-component/createComponent.js
@@ -59,8 +59,10 @@ inquirer
     const react = targets.includes('React');
     const componentPath = path.join(cwd, unstable ? `modules/_labs/${name}` : `modules/${name}`);
 
+    console.log(name, unstable, targets, componentPath);
+
     if (!fs.existsSync(componentPath)) {
-      mkdirp(componentPath);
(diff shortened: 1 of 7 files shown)


utils/create-component/createComponent.js
```diff
@@ -59,8 +59,10 @@ inquirer
     const react = targets.includes('React');
     const componentPath = path.join(cwd, unstable ? `modules/_labs/${name}` : `modules/${name}`);
 
+    console.log(name, unstable, targets, componentPath);
```
alanbsmith, line 62: Do you want to keep this console.log?

alanbsmith, line 62:
```suggestion

```

utils/create-component/createComponent.js
```diff
@@ -59,8 +59,9 @@ inquirer
     const react = targets.includes('React');
     const componentPath = path.join(cwd, unstable ? `modules/_labs/${name}` : `modules/${name}`);
 
+
```
NicholasBoll, line 62:
```suggestion

```

PR #662, opened 2020-05-19
feat(button): Add toolbar square icon

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/610

Add a toolbar square icon as a new variant to icon button

• This new Icon Button has slightly different styling for toggled on.
• It's default size is set to `small` to fit it's intended space of a toolbar

## Checklist

- [x] `yarn test` passes
      applicable


modules/button/react/README.md
```diff
@@ -712,6 +713,9 @@ Default: `IconButtonVariant.Circle`
 
 Default: `'medium'`
 
+Note: If the Icon Button is of variant `ToolbarSquare` it will default to `'small'` to fit it's
```
mannycarrera4, line 716: edit grammar....

modules/button/react/lib/IconButton.tsx
```diff
@@ -52,7 +52,7 @@ const IconButton: ButtonOrAnchorComponent<
 > = ({
   theme = useTheme(),
   variant = IconButtonVariant.Circle,
-  size = 'medium',
+  size = IconButtonVariant.ToolbarSquare ? 'small' : 'medium',
```
mannycarrera4, line 55: Thoughts on this approach? I wanted to default to small always, but should we leave it open and have the consumer always set the size?

modules/button/react/lib/IconButton.tsx
```diff
@@ -95,7 +97,7 @@ const IconButton: ButtonOrAnchorComponent<
   return (
     <ButtonContainer
       colors={getIconButtonColors(variant, theme, toggled)}
-      size={size}
+      size={variant === IconButton.Variant.ToolbarSquare ? IconButton.Size.Small : size}
```
mannycarrera4, line 100: remove

modules/button/react/lib/IconButton.tsx
```diff
@@ -81,7 +81,9 @@ const IconButton: ButtonOrAnchorComponent<
     width: size === 'small' ? spacing.l : spacing.xl,
     height: size === 'small' ? spacing.l : spacing.xl,
     borderRadius:
-      variant === IconButtonVariant.Square || variant === IconButtonVariant.SquareFilled
+      variant === IconButtonVariant.Square ||
+      variant === IconButtonVariant.SquareFilled ||
+      variant === IconButtonVariant.ToolbarSquare
```
alanbsmith, line 86: This is starting to get big enough to warrant a separate function for readability (IMO).

Something like:
```js
function getBorderRadius(variant) {
  switch (variant) {
    case IconButtonVariant.Square:
    case IconButtonVariant.SquareFilled:
    case IconButtonVariant.ToolbarSquare:
      return borderRadius.m;
    default:
      return borderRadius.circle;
  }
}
```

modules/button/react/stories/stories_IconButton.tsx
```diff
@@ -224,6 +224,39 @@ storiesOf('Components|Buttons/Button/React/Icon Button', module)
...
+      setToggled(!toggled);
+    };
+
+    return (
+      <div className="story">
+        <h3>Small Toolbar Square</h3>
+        <IconButton
+          aria-label="Activity Stream"
+          icon={activityStreamIcon}
+          // size={IconButton.Size.Small}
```
alanbsmith, line 239: Should this be removed?

mannycarrera4: yes

modules/button/react/lib/IconButton.tsx
mannycarrera4: ok i really like this approach because now that function can be edited in the future if need be, thanks for the suggestion!

modules/button/react/lib/IconButton.tsx
```diff
@@ -116,6 +113,25 @@ IconButton.Size = {
...
 
+const getIconButtonBorderRadius = (variant: IconButtonVariant) => {
+  switch (variant) {
+    case IconButtonVariant.Square:
+    case IconButtonVariant.SquareFilled:
+    case IconButtonVariant.ToolbarSquare:
+      return {borderRadius: borderRadius.m};
+    default:
+      return {borderRadius: borderRadius.circle};
+  }
```
alanbsmith, line 124: 👍

modules/button/react/lib/IconButton.tsx
```diff
@@ -116,6 +113,25 @@ IconButton.Size = {
...
+};
+
+const getIconButtonSize = (variant: IconButtonVariant, size: 'small' | 'medium') => {
+  if (variant === IconButtonVariant.ToolbarSquare) {
+    return 'small';
+  } else {
+    return size;
+  }
```
alanbsmith, line 132: 👍

modules/button/react/stories/stories_IconButton.tsx
```diff
@@ -224,6 +224,51 @@ storiesOf('Components|Buttons/Button/React/Icon Button', module)
...
+          aria-label="Activity Stream"
+          icon={activityStreamIcon}
+          disabled={true}
+          variant={IconButton.Variant.ToolbarSquare}
+        />
+        <h3>Medium Toolbar Square</h3>
+        <IconButton
+          aria-label="Activity Stream"
+          icon={activityStreamIcon}
+          size={IconButton.Size.Medium}
```
anicholls, line 251: Why have a medium version if it's not supposed to be medium?

anicholls: Also your `getIconButtonSize` function doesn't seem to be working because this is still rendering a larger icon.

modules/button/react/lib/IconButton.tsx
```diff
@@ -116,6 +113,25 @@ IconButton.Size = {
...
+    case IconButtonVariant.SquareFilled:
+    case IconButtonVariant.ToolbarSquare:
+      return {borderRadius: borderRadius.m};
+    default:
+      return {borderRadius: borderRadius.circle};
+  }
+};
+
+const getIconButtonSize = (variant: IconButtonVariant, size: 'small' | 'medium') => {
+  if (variant === IconButtonVariant.ToolbarSquare) {
```
anicholls, line 128: I'm hesitant to do this. A big part of the v4 button refactor was to avoid API restrictions like this where some combinations aren't valid (i.e. a medium toolbar icon button).

I think we should either allow medium toolbar icon buttons, or this should be broken into it's own `ToolbarIconButton` component

modules/button/react/lib/IconButton.tsx
```diff
@@ -52,7 +52,7 @@ const IconButton: ButtonOrAnchorComponent<
 > = ({
   theme = useTheme(),
   variant = IconButtonVariant.Circle,
-  size = 'medium',
+  size = variant === IconButtonVariant.ToolbarSquare ? 'small' : 'medium',
```
anicholls, line 55: Same here. If this has different defaults, it should be a separate component. Thoughts on pulling this out of `IconButton` and adding a `ToolbarIconButton`?

modules/button/react/lib/types.ts
```diff
@@ -45,6 +45,7 @@ export enum IconButtonVariant {
   CircleFilled = 'circleFilled',
   Inverse = 'inverse',
   InverseFilled = 'inverseFilled',
+  ToolbarSquare = 'toolbarSquare',
```
anicholls, line 48: Unless we're planning on adding a `ToolbarCircle`, we should just call this `Toolbar`

modules/button/react/lib/IconButton.tsx
```diff
@@ -258,5 +274,30 @@ const getIconButtonColors = (
           icon: toggled ? themePrimary.main : 'rgba(255, 255, 255, 0.75)',
         },
       };
+
+    case IconButtonVariant.ToolbarSquare:
+      return {
+        default: {
+          icon: toggled ? themePrimary.dark : colors.licorice200,
```
anicholls, line 281: Why not `themePrimary.main`? All of the other icons use blueberry400, not blueberry500.

modules/button/react/lib/IconButton.tsx
alanbsmith: Yeah, agreed. In the example I suggested:
```js
  if (size) {
    // always allow an override
    return size;
  }
```

That allows us some flexibility to potentially add a medium sized toolbar icon without worrying about breaking changes down the road.

The other option is to create a separate component here, which is fine too.

modules/button/react/lib/IconButton.tsx
```diff
@@ -116,6 +113,16 @@ IconButton.Size = {
...
+const getIconButtonBorderRadius = (variant: IconButtonVariant) => {
+  switch (variant) {
+    case IconButtonVariant.Square:
+    case IconButtonVariant.SquareFilled:
+      return {borderRadius: borderRadius.m};
+    default:
+      return {borderRadius: borderRadius.circle};
+  }
+};
+
```
mannycarrera4, line 125: This was some minor clean up when I was thinking of it as a variant, I can remove this if it's not relevant.

modules/button/react/lib/ToolbarIconButton.tsx
```diff
@@ -0,0 +1,120 @@
...
+   */
+  buttonRef?: React.Ref<HTMLButtonElement>;
+  /**
+   * The icon of the IconButton.
+   */
+  icon?: CanvasSystemIcon;
+  /**
+   * The function called when the IconButton toggled state changes.
+   */
+  onToggleChange?: (toggled: boolean | undefined) => void;
```
mannycarrera4, line 33: I removed the `as` prop. Otherwise I had to use `ButtonOrAnchor` and that requires  a variant and size, but that's not relevant in this specific component, thoughts?

modules/button/react/lib/ToolbarIconButton.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+const getToolbarIconButtonColors = (theme: EmotionCanvasTheme, toggled?: boolean): ButtonColors => {
+  const {
+    canvas: {
+      palette: {primary: themePrimary},
+    },
+  } = theme;
+  return {
+    default: {
+      icon: toggled ? themePrimary.dark : colors.licorice200,
```
mannycarrera4, line 99: Blueberry500 is used in this instance because when it's toggled on the background is also blue, so it just helps it stand out a little bit more

modules/button/react/lib/IconButton.tsx
mannycarrera4: Design updated this to 500 because when active, it also have a blue background, this allows the icon to stand out a little bit more

PR #665, opened 2020-05-20
ci: Improve canary builds & publish behavior

### LINKED ISSUE #661: ci: Improve prerelease versioning logic
Our existing structure for canary versioning has been hard/confusing to debug/reason about lately. We’ve been hacking around Lerna’s `canary` solution, and I think it’s time to drop that entirely and calculate our own versions because it’s not working. I would like to propose a few changes/improvements:

- Consider dropping Lerna `--canary` usage and calculate the version for canary builds ourselves (we've had lots of issues around the lack of configurability here - see https://github.com/lerna/lerna/issues/1893 for an example)
- Change the prerelease dist-tags to follow the same style as master
- Don't do canary builds for release merge commits
- Delete `canary` preids accidentally created in #619 
- Unrelated, but while we're at it can we add `--force-publish="*"` for release/tag builds?

## Canary versions

To customize the version at all when using `lerna publish`, the `--canary` flag is required. Alternatively, we could run `lerna version` manually first, and then publish that version (the same flow we use for releases). This may not be necessary depending on what we decide for other solutions below. 

With the improvement to dist-tags (see below), I don't think the sha is completely necessary, so maybe we can continue using `--canary`. Previously, we were using it so that we didn't have conflicts after a "prerelease release" (since that resets the count). We should fix this properly by appending the previous release to the preid (e.g. `4.0.0-beta.4-next.0`). Also, with the improvements below, I think it's much more obvious what the affixed number means.

## Prerelease dist-tags

Our current dist-tag release strategy for prerelease branches does not line up with how we treat master. We have a tag for the canary build (`prerelease`), but no tag for alpha/beta releases. This is a bit confusing, and I think we should update to match. Proposal: 

`latest` = The latest released version (e.g. `3.7.0` )
`next` = The latest master merge (e.g. `3.8.1-next.0`)
`prerelease` = The latest prerelease released version (e.g. `4.0.0-beta.3`)
`prerelease-next` = The latest prerelease merge (e.g. `4.0.0-beta.4-next.0`)

Note: the number after the `next` tags indicates the number of canary builds since the last release (either `latest` or `prerelease`)

## Merge commit builds
It's confusing when we generate two builds for a release merge commit (one canary and one for the release itself). We should prevent canary builds in this instance.

We need to determine a way to detect a release merge commit (since the tag is applied after merging). We could either look for "chore: Release" in the commit message, or check if the commit contains changes to `CHANGELOG.md` (can assume this is a release).

anicholls:
## Summary

Addresses everything in https://github.com/Workday/canvas-kit/issues/661:

- Change prerelease canary dist tag to `prerelease-next` so it doesn't clobber `prerelease` releases
- Use previous release for canary preids instead of the sha (which can result in unpredictable resolutions). Example: `4.0.0-beta.4-next.0`
- Skip canary builds if last merge commit was a release (if `CHANGELOG.md` was changed)
- **Disable prerelease canaries for now**. The above changes will not be reflected as we've decided to disable prerelease canary builds temporarily since we've had some issues and they're not being used.
- **Change default publish behavior of Lerna to publish all modules regardless of whether they've been updated (`--force-publish="*"`)**

Closes #661

### DIFF
--- a/utils/publish-canary.js
+++ b/utils/publish-canary.js
@@ -18,7 +18,9 @@ let distTag;
 if (TRAVIS_BRANCH === 'master') {
   distTag = 'next';
 } else if (isPrerelease) {
-  distTag = 'prerelease-next';
+  // distTag = 'prerelease-next';
+  console.error('Prerelease canary builds disabled.');
+  process.exit(0);
 } else {
   console.error('No travis branch provided');
   process.exit(1);
--- a/utils/publish-canary.js
+++ b/utils/publish-canary.js
@@ -11,13 +11,14 @@ const {
   TRAVIS_BUILD_URL = 'https://travis-ci.org/Workday/canvas-kit/branches',
 } = process.env;
 
+const isPrerelease = TRAVIS_BRANCH.match(/^prerelease\/v\d*$/g);
 const data = {};
-let preid;
+let distTag;
 
 if (TRAVIS_BRANCH === 'master') {
-  preid = 'next';
-} else if (TRAVIS_BRANCH.match(/^prerelease\/v\d*$/g)) {
-  preid = 'prerelease-next';
+  distTag = 'next';
+} else if (isPrerelease) {
+  distTag = 'prerelease-next';
 } else {
   console.error('No travis branch provided');
   process.exit(1);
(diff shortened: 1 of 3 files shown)


utils/publish-canary.js
```diff
@@ -11,28 +11,72 @@ const {
...
+    return cmd(`git describe --abbrev=0`);
+  })
+  .then(releaseTag => {
+    const lastReleasePreId = releaseTag.trim().split('-')[1];
+
+    const lastReleaseNum = parseInt(lastReleasePreId.match(/\d+$/), 10);
+    const pos = lastReleasePreId.indexOf(lastReleaseNum);
+    const nextReleasePreId = lastReleasePreId.slice(0, pos) + (lastReleaseNum + 1);
+
+    const preid = isPrerelease ? `${nextReleasePreId}-next` : 'next';
```
anicholls, line 72: Grabs last release tag (e.g. `4.0.0-beta.3`), and grabs the prerelease suffix (`beta.3`). It then grabs the number at the end, and increments it. Finally it appends `-next` (e.g. `beta.4-next`).

This preid will result in a version in the form of `4.0.0-beta.4-next.X` (where X is the integer number of commits since the `4.0.0-beta.3` release

lychyi: I think this can be done using the [semver package](https://www.npmjs.com/package/semver#prerelease-identifiers) too. You can probably get rid of a lot of the custom logic in case it has some bugs that we're not catching now.

utils/publish-canary.js
```diff
@@ -11,28 +11,72 @@ const {
...
+      if (error) {
+        throw error;
+      }
+    }
+  );
+};
+
+cmd('git diff --name-only HEAD HEAD^')
+  .then(filesChangedInMerge => {
+    if (filesChangedInMerge.includes('CHANGELOG.md')) {
```
lychyi, line 53: What if we change/fix the `CHANGELOG.md` file as part of previous commit? Is there a better way to determine if the last merge was a commit?

anicholls: Not quite sure what you mean here. Are you talking about the case where we have a merge commit that updates the changelog that is not a release?

If so, I discussed that case with @NicholasBoll and we agreed it was probably fine if we cause a canary build to be skipped every once in a while. It seemed like a better option than the alternative of looking for `chore: Release` in the commit message (less room for user error)

PR #667, opened 2020-05-21
chore: Add public config to create-component script

### LINKED ISSUE #643: Automating new component release
## Summary
We need to test a way to automatically make a new scoped component public.  As part of this we can define a checklist of tasks for a new component release and use this check list.

Investigate if there's potential to use this as a key in our create components package generator

## Reason
Automating the process will remove some of the missed steps we currently face when releasing new components.

alanbsmith:
## Summary

Resolves #643

Adds a step in the create-component cli to add a public config to new modules

`utils/create-component/createComponent.js`

## Testing Manually

Run the script!

```sh
yarn create-component
```

If the module should be public (the default), this config will be added to the package.json

```json
  "publishConfig": {
    "access": "public"
  },
```

Otherwise, it will be omitted, and the module will be private when published.

## New Dependencies

n/a

## Thank You Gif
_Share a fun [gif](https://giphy.com) to say thanks to your reviewer:_

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/utils/create-component/createComponent.js
+++ b/utils/create-component/createComponent.js
@@ -6,6 +6,7 @@ const path = require('path');
 const mkdirp = require('mkdirp');
 const inquirer = require('inquirer');
 const cmd = require('node-cmd');
+const colors = require('colors');
 
 const createReactModule = require('./createReactModule');
 const createCssModule = require('./createCssModule');
--- a/utils/create-component/createComponent.js
+++ b/utils/create-component/createComponent.js
@@ -46,7 +46,7 @@ const questions = [
   {
     type: 'confirm',
     name: 'publicModule',
-    message: 'Should this module be public?:',
+    message: 'Make access public when publishing?:',
     default: true,
   },
   /**
--- a/utils/create-component/createComponent.js
+++ b/utils/create-component/createComponent.js
@@ -6,7 +6,6 @@ const path = require('path');
 const mkdirp = require('mkdirp');
 const inquirer = require('inquirer');
 const cmd = require('node-cmd');
-const colors = require('colors');
 
 const createReactModule = require('./createReactModule');
 const createCssModule = require('./createCssModule');
@@ -44,6 +43,12 @@ const questions = [
     message: 'Is this an unstable component (i.e. should it go in Canvas Kit Labs)?:',
     default: false,
   },
+  {
+    type: 'confirm',
+    name: 'publicModule',
+    message: 'Should this module be public?:',
+    default: true,
+  },
   /**
    * Add question to add deps
    * React: CK core, emotion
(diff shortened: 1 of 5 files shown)


utils/create-component/createComponent.js
```diff
@@ -6,7 +6,6 @@ const path = require('path');
 const mkdirp = require('mkdirp');
 const inquirer = require('inquirer');
 const cmd = require('node-cmd');
-const colors = require('colors');
```
alanbsmith, line 9: This wasn't being used, so I removed it.

utils/create-component/createComponent.js
```diff
@@ -44,6 +43,12 @@ const questions = [
     message: 'Is this an unstable component (i.e. should it go in Canvas Kit Labs)?:',
     default: false,
   },
+  {
+    type: 'confirm',
+    name: 'publicModule',
+    message: 'Should this module be public?:',
```
lychyi, line 49: I think this question is a tad vague since this really relates only to publishing and not anything else. What we're really asking is do you want this module to be public upon publishing. Maybe add a brief explanation, something like:

`Publish this as a public module? (scoped modules are restricted access by default when published)`

I know it's a tad verbose but you really only skim through this prompt once and then you kind of get it later.

lychyi: Or maybe `Make access public when publishing?` since `Publish this as a public module?` may indicate that we're doing this right now.

alanbsmith: Sounds good. Will update. 👍

utils/create-component/createComponent.js
anicholls: @alanbsmith This is a special import. It extends the string prototype, but doesn't actually get called specifically. This allows us to do things like `'This is a warning'.yellow`, and it's used several times in this file. I'm pretty sure removing this will break that functionality.

utils/create-component/templates/css/packageJson.js
```diff
@@ -21,7 +21,14 @@ module.exports = (name, moduleName, description, unstable) => `
...
     "test": "echo \\"Error: no test specified\\" && exit 1",
     "build": "node ../../../${unstable ? '../' : ''}utils/css-build.js index.scss"
-  },
+  }, ${
+    public
+      ? `
+  "publishConfig": {
+    "access": "public"
+  },`
+      : ``
```
anicholls, line 30: Nit:
```suggestion
    public && `
  "publishConfig": {
    "access": "public"
  },`
```

would be cleaner

utils/create-component/createComponent.js
alanbsmith: ohhh. Thanks for catching that! I'll put it back!

utils/create-component/templates/css/packageJson.js
alanbsmith: I thought about that, but I was worried that the template literal will but a "false" in the json. I'll test that though.

alanbsmith: confirmed this.

Issue #657, opened 2020-05-15
Switch Component requires two clicks to change checked state.


aaroncalderon:
## 🐛 Bug Report

When implementing the Switch component, it requires two clicks to change the checked state/value.

## To Reproduce

See CodeSandbox.io [Use Case: Workday Canvas-kit Switch](https://codesandbox.io/s/use-case-workday-canvas-kit-switch-0np5l), click on the switch component until it changes state several times from false to true and from true to false.

## Expected Behavior

As per your documentation only one click is required to change state.

## Actual Results

The component must be clicked twice to change checked state.

## Browser (if applicable)

Chrome

## Link to repl or repo (highly encouraged)

CodeSandbox.io [Use Case: Workday Canvas-kit Switch](https://codesandbox.io/s/use-case-workday-canvas-kit-switch-0np5l)

### Error Output

n/a

mannycarrera4:
@aaroncalderon thanks for submitting an issue!

In the sandbox you've provided a noticed a couple things:
1) The click target  seems to be off which was preventing clicking on the whole switch:
<img width="917" alt="Screen Shot 2020-05-21 at 2 49 15 PM" src="https://user-images.githubusercontent.com/7966550/82609930-bac4da80-9b7a-11ea-8c97-db44195caad4.png">

We have a fix in the works for this here: https://github.com/Workday/canvas-kit/pull/669

2) By adding `event.preventDefault()` to your handleChange function, this was breaking the event listener because it's not a `cancelable` event and therefore stopping the natural behavior of an on change event. 

The PR should fix the click target and removing `event.preventDefault()` should do the trick!

aaroncalderon:
@mannycarrera4 thanks so much for pointing this out.

I put together these CodeSandbox samples based off other react sample codes for `this.handleChange` functions.

Looks like on the canvas-kit documentation the `this.handleCheck` is the function name of choice. But, I have not been able to see any code sample for this.

Now the question is...

So, when a form contains additional code-kit components, like radio or select, and all components use the same onChange function, only the switch component (so far) displays this odd behavior when we have an `event.preventDefault();` in there. 

I believe that the `event.preventDefault();` is there to handle buttons, and prevent form submission.

- Does this mean that we must handle switch components differently?
- Is there a best practice regarding this?

[...]

PR #669, opened 2020-05-21
fix: Update click target for switch

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/657

The click target is off when the label and and switch are stacked, making it hard to click sometimes. By adding display block, the click target is now the full width of the input.


lychyi:
@mannycarrera4 Does it fix that issue or just a part of that issue? Using the 'Fixes' keyword will close it. Just making sure this is the only fix for that issue.

mannycarrera4:
@lychyi I believe so, i think we can also set `left: 0` which will shift the input to the left of it's container which has position relative.

lychyi:
@mannycarrera4 Can we base this off of master and target this PR there?

mannycarrera4:
Closing to base off of master instead

PR #670, opened 2020-05-21
feat: Add a Popup Stack manager to Canvas Kit

### LINKED ISSUE #182: Add a popup stacking manager to canvas-kit

NicholasBoll:
Fixes #182 

## Summary

Adds a popup stack manager to Canvas Kit. This popup stack is framework agnostic and meant to be used globally throughout an entire application stack. It takes care to normalize z-index values, escape key handling, and click outside handling.

Components have been modified to use this stack so that escape key and click outside listeners work with other popups in the stack. This is needed for accessibility requirements and so that Canvas Kit can work with other popup systems (so long as those popup systems consume this stack).

## Checklist

- [x] `yarn test` passes
      applicable

## PR Squash commit message:
Fixes #182

Adds a popup stack manager to Canvas Kit. This popup stack is framework agnostic and meant to be used globally throughout an entire application stack. It takes care to normalize z-index values, escape key handling, and click outside handling.

Components have been modified to use this stack so that escape key and click outside listeners work with other popups in the stack. This is needed for accessibility requirements and so that Canvas Kit can work with other popup systems (so long as those popup systems consume this stack).

BREAKING CHANGES:
* Popper has been moved to the Popup module and out of common
* `closeOnEscape` has been removed from the Modal component as it is no longer needed


.eslintrc.js
```diff
@@ -67,7 +67,8 @@ module.exports = {
     'no-param-reassign': 'error',
     'no-redeclare': 'error',
     'no-undef-init': 'error',
-    'no-unused-expressions': 'error',
+    'no-unused-expressions': 'off',
+    '@typescript-eslint/no-unused-expressions': 'error',
```
NicholasBoll, line 71: This was added for optional chaining

4.0-MIGRATION-GUIDE.md
```diff
@@ -329,9 +332,14 @@ PRs:
 ### Popper
 
 `Popper` was changed to a Functional Component with a forwarded ref. If you passed a `ref` object to
-`Popper` before, it will now point to the element rather than the `Popper` instance.
+`Popper` before, it will now point to the element rather than the `Popper` instance. Popper was
+moved to the `@workday/canvas-kit-react-popup module. This change aligns with the concept that Popup
```
NicholasBoll, line 336:
```suggestion
moved to the `@workday/canvas-kit-react-popup` module. This change aligns with the concept that Popup
```

modules/_labs/popup-stack/react/README.md
```diff
@@ -0,0 +1,36 @@
...
+<a href="https://github.com/Workday/canvas-kit/tree/master/modules/_labs/README.md">
+  <img src="https://img.shields.io/badge/LABS-alpha-orange" alt="LABS: Alpha" />
+</a>  This component is work in progress and currently in pre-release.
+
+Stack for managing popup UIs to coordinate global concerns like escape key handling and rendering order
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-labs-react-popup-stack
```
NicholasBoll, line 12: Change all modules to be `@workday/canvas-kit-popup-stack` instead

modules/_labs/popup-stack/react/lib/PopupStack.ts
```diff
@@ -0,0 +1,175 @@
...
+ * This type is purposely an interface so that it can be extended for a
+ * specific use-case
+ */
+export interface PopupStackItem {
+  /**
+   * Ephemeral type popups usually have global listeners like click outside
+   * or escape key handling. Ephemeral type popups include Tooltips, Modals,
+   * DropDowns, etc. Persistent type popups include Windows. Persistent can
+   * only be dismissed explicitly by a close button.
+   */
```
NicholasBoll, line 11: Update documentation to make more sense

modules/_labs/popup-stack/react/lib/PopupStack.ts
```diff
@@ -0,0 +1,175 @@
...
+  /**
+   * **Do not change unless you know what you're doing**. This sets the
+   * minimum z-index value of all popup elements on the whole page,
+   * not just your application. The stack is global. This should only be
+   * changed by the owners of the entire application stack.
+   * @param min minimum z-index value of any element in the stack
+   */
+  setZIndexMin(min: number) {
+    stack.zIndex.min = min;
+  },
```
NicholasBoll, line 143: Consider removing and add them in if necessary

modules/_labs/popup-stack/react/lib/PopupStack.ts
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * Set the z-index of each item in the stack by changing setting
+ * `element.style.zIndex`. This should work for a web target across all
+ * frameworks. The strategy
+ */
+export function defaultGetValue(index: number, length: number): number {
+  const {min, max} = stack.zIndex;
+
+  return Math.max(min, max + index - length + 1);
+}
```
NicholasBoll, line 32: Consider changing z-index assignment strategy to not change z-index of all other popups

modules/_labs/popup-stack/react/lib/PopupStack.ts
```diff
@@ -0,0 +1,175 @@
...
+  if (items.length) {
+    return items[items.length - 1];
+  }
+  return null;
+}
+
+/**
+ * Set the z-index of each item in the stack by changing setting
+ * `element.style.zIndex`. This should work for a web target across all
+ * frameworks. The strategy
```
NicholasBoll, line 26: Complete this...

modules/_labs/popup-stack/react/lib/PopupStack.ts
```diff
@@ -0,0 +1,175 @@
...
+ */
+export function defaultGetValue(index: number, length: number): number {
```
NicholasBoll, line 28: Consider changing the name to `defaultGetZIndexValue`

modules/_labs/popup-stack/react/lib/PopupStack.ts
```diff
@@ -0,0 +1,175 @@
...
+    min: number;
+    max: number;
+    getValue: typeof defaultGetValue;
+  };
+}
+
+// We need to make sure only one stack is ever in use on the page - ever.
+// If a stack is already defined on the page, we need to use that one.
+// Never, ever, ever change this variable name on window
+const stack: Stack = (window as any).__popupStack || {
```
NicholasBoll, line 53: See if this should go on the `window.workday` namespace.

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -35,8 +39,9 @@ export interface ModalContentProps extends React.HTMLAttributes<HTMLDivElement>
    * However, we cannot guarantee that it is safe to simply bind an event listener and close in all
    * cases. Some applications may use a Popup manager to make sure the correct popup is receiving
    * the close command. If your application uses custom popup stacking, do not set this to true.
+   * @deprecated
    */
-  closeOnEscape: boolean;
+  closeOnEscape?: boolean;
```
NicholasBoll, line 44: Remove this property

NicholasBoll: Document as a breaking change

modules/popup/react/lib/Popper.tsx
```diff
@@ -33,29 +41,63 @@ export interface PopperProps extends React.HTMLAttributes<HTMLDivElement> {
    */
   popperOptions?: Partial<PopperOptions>;
   /**
-   * If true, attach the Popper to the `containerElement`. If false, render the Popper within the DOM hierarchy of its parent.
+   * If true, attach the Popper to the `containerElement`. If false, render the Popper within the
+   * DOM hierarchy of its parent.
    * @default true
    */
   portal?: boolean;
+  type?: 'ephemeral' | 'persistent';
```
NicholasBoll, line 49: Consider removing these

modules/popup/react/lib/usePopupStack.ts
```diff
@@ -0,0 +1,36 @@
...
+export const usePopupStack = <E extends HTMLElement>(
+  ref: React.RefObject<E>,
+  type: 'ephemeral' | 'persistent' = 'ephemeral'
+) => {
+  // We useLayoutEffect to ensure proper timing of registration
+  // of the element to the popup stack. Without this, the timing is unpredictable
+  // when mixed with other frameworks. Other frameworks should also register as soon
+  // as the element is available
+  React.useLayoutEffect(() => {
+    PopupStack.add({element: ref.current!, type});
```
NicholasBoll, line 19: Consider adding a raf for writing to the DOM to avoid layout thrashing

package.json
```diff
@@ -138,7 +138,7 @@
     "typecheck:src": "tsc -p . --noEmit --incremental false",
     "typecheck:cypress": "tsc -p cypress --noEmit",
     "format:prettier": "prettier --config=./.prettierrc.js --ignore-path=./.prettierignore \"**/*.{js,jsx,json,ts,tsx,md}\" --write",
-    "format": "yarn lint --fix -c .eslintrc-prettier.js",
+    "format": "yarn eslint --ext=jsx,ts,tsx --fix -c .eslintrc-prettier.js",
```
NicholasBoll, line 141: Look into this

modules/_labs/menu/react/package.json
```diff
@@ -46,7 +46,8 @@
   ],
   "devDependencies": {
     "@material-ui/core": "^4.9.7",
-    "@workday/canvas-system-icons-web": "^1.0.20"
+    "@workday/canvas-system-icons-web": "^1.0.20",
+    "@workday/canvas-kit-react-popup": "^4.0.0-beta.3"
```
NicholasBoll, line 50: Fix this version

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -85,17 +81,17 @@ const Container = styled('div')({
...
   horizontal: 'center',
   vertical: 'bottom',
 } as const;
 
-function getFirstElementToFocus(modalEl: HTMLElement): HTMLElement {
-  const firstFocusable = modalEl.querySelector<HTMLElement>(
-    `[data-close=close],[id="${modalEl.getAttribute('aria-labelledby')}"]`
+function getFirstElementToFocus(overlayEl: HTMLElement): HTMLElement {
+  const modalEl = overlayEl.querySelector('[role=dialog]');
+  const firstFocusable = modalEl?.querySelector<HTMLElement>(
```
NicholasBoll, line 93: This change was necessary because the `modalRef` is now pointing to the overlay instead of the `[role=dialog]` element

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -169,30 +154,23 @@ const ModalContent = ({
...
+  const onOverlayClick = (event: React.MouseEvent<HTMLElement>) => {
+    if (event.target === modalRef.current) {
+      onClose();
     }
   };
 
-  useKeyDownListener(handleKeydown);
+  usePopupStack(modalRef);
+  useEscapeKey(modalRef, onClose);
+  useFocusTrap(modalRef);
```
NicholasBoll, line 168: Moved many sharable behaviors for Modal into Popup hooks

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -212,9 +190,8 @@ const ModalContent = ({
   }, []);
 
   const content = (
-    <Container onClick={handleOutsideClick} {...elemProps}>
+    <Container {...elemProps} ref={modalRef} onClick={onOverlayClick}>
```
NicholasBoll, line 193: Change the element of `modalRef`. This is because the PopupStack sets the z-index of the ref that is handed to it. If that was the Modal and not the overlay, the overlay could appear behind content it is supposed to cover.

modules/modal/react/stories/stories.tsx
```diff
@@ -214,4 +214,32 @@ storiesOf('Components|Popups/Modal/React', module)
...
+              Cancel
+            </Button>
+          </Modal>
+          <Button onClick={modal1.closeModal} variant={Button.Variant.Secondary}>
+            Cancel
+          </Button>
+        </Modal>
+      </>
+    );
+  });
```
NicholasBoll, line 245: Add a test for this case to make sure we handle it correctly

modules/popup-stack/stories/stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+import README from '../README.md';
+
+export default {
+  title: 'Labs|Popups/Stack Manager',
+  decorators: [withReadme(README)],
+};
+
+export const MixedPopupTypes = () => {
+  return <div />;
+};
```
NicholasBoll, line 15: Need to add more here

modules/popup-stack/README.md
```diff
@@ -0,0 +1,38 @@
+# Canvas Kit Popup Stack
+
+Stack for managing popup UIs to coordinate global concerns like escape key handling and rendering
+order. The popup stack is framework agnostic. The popup stack uses a shared array as the stack. This
+sharing is enforced. If the stack variable has already been declared, the first-instantiated will be
```
anicholls, line 5:
```suggestion
sharing is enforced. If the window stack variable (`window.workday.__popupStack`) has already been declared, the first-instantiated instance will be
```

modules/popup-stack/README.md
```diff
@@ -0,0 +1,38 @@
...
+used. There is no need to worry about bootstrapping at a specific time.
+
+All Popup-style UIs within a page should register with this stack in order to properly function
+together.
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-popup-stack
```
anicholls, line 14: Any thoughts on whether this should be bundled in the bundle modules?

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -0,0 +1,127 @@
...
+
+function getLast<T>(items: T[]): T | null {
+  if (items.length) {
+    return items[items.length - 1];
+  }
+  return null;
+}
+
+/**
+ * Set the z-index of each item in the stack by changing setting
```
anicholls, line 17: This doesn't set the z-index

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -0,0 +1,127 @@
...
+ * `element.style.zIndex`. This is a framework-agnostic approach to changing
+ * zIndex. This strategy will start the popup stack at 30 and each popup will
```
anicholls, line 19: sentence cut off

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -0,0 +1,127 @@
...
+ */
+export function defaultGetValue(index: number, length: number): number {
```
anicholls, line 21: Couldn't overriding this function make the behavior inconsistent? Can you remind me of the use case for this?

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -0,0 +1,127 @@
...
+   */
+  remove(element: HTMLElement): void {
+    stack.items = stack.items.filter(item => item.element !== element);
+
+    setZIndexOfElements(PopupStack.getElements(), stack.zIndex.getValue);
+  },
+
+  /**
+   * Is the current element the top in the stack? Use `isTopmostEphemeral`
+   * to handle detection of escape key or click outside.
```
anicholls, line 77:
```suggestion
   * Is the current element the top in the stack?
```

modules/popup/react/lib/Popper.tsx
anicholls: Shouldn't we still be removing these?

modules/popup/react/lib/Popper.tsx
```diff
@@ -33,29 +41,63 @@ export interface PopperProps extends React.HTMLAttributes<HTMLDivElement> {
...
+
+    if (!portal) {
+      return contents;
+    }
+
+    return ReactDOM.createPortal(contents, containerElement || document.body);
+  }
+);
+
+const OpenPopper = React.forwardRef<HTMLDivElement, PopperProps>(
```
anicholls, line 85: Can you add some documentation on why you had to break these out? It's a bit confusing at first glance

modules/popup/react/lib/useEscapeKey.ts
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import {PopupStack} from '@workday/canvas-kit-popup-stack';
+
+/**
+ * Use this on ephemeral popups. It will add escape key handling and will call the `onClose`
```
anicholls:
```suggestion
 * Adds escape key handling and will call the `onClose`
```

package.json
```diff
@@ -156,6 +156,10 @@
         "yarn format",
         "git add"
       ],
+      "*.json": [
+        "node utils/check-mismatched-dependencies.js",
```
anicholls, line 160: We should be running check-lockfile at this step too. It's important because otherwise we end up with artifactory URLs in github

NicholasBoll: That looks like it is handled in the `yarn.lock` step. Should we also add that here?

modules/popup-stack/lib/PopupStack.ts
anicholls: If we've removed the ability to change this, can we rename it to `getValue` or bake its behavior into `setZIndexOfElements`?

modules/popup-stack/stories/stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+import {PopupStack} from '@workday/canvas-kit-popup-stack';
+import README from '../README.md';
+
+export default {
+  title: 'Components|Popups/Stack Manager',
+  decorators: [withReadme(README)],
+};
+
+export const MixedPopupTypes = () => {
+  return <div />;
```
anicholls, line 14: What are the plans for this story?

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -0,0 +1,131 @@
...
+    max: number;
+    getValue: typeof defaultGetValue;
+  };
+}
+
+// We need to make sure only one stack is ever in use on the page - ever.
+// If a stack is already defined on the page, we need to use that one.
+// Never, ever, ever change this variable name on window
+(window as any).workday = (window as any).workday || {};
+const stack: Stack = (window as any).workday.__popupStack || {
```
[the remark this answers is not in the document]
NicholasBoll: Yes. @nmoresco told me underscores mean private variables. It won't have to be officially documented.

modules/popup-stack/stories/stories.tsx
NicholasBoll: My plan is to test functionality of a few popup types - modals, selects, tooltips and a special Window component to show that it works correctly. Similar to this:

modules/popup-stack/README.md
NicholasBoll: Do you mean re-exported from `@workday/canvas-kit-react`?

modules/popup-stack/lib/PopupStack.ts
NicholasBoll: I have tests for this function, so I'm not sure it makes sense to bake it into `setZIndexOfElements`. I could rename it to `getValue`.

.storybook/config.js
```diff
@@ -17,8 +17,8 @@ function loadStories() {
   const allExports = [];
 
   allReqs.forEach(req => {
-    req.keys().forEach(fname => {
-      const story = req(fname);
+    req.keys().forEach(filename => {
+      const story = req(filename);
```
NicholasBoll, line 21: At first I thought this was "function name"

.storybook/manager-head.html
```diff
@@ -1,3 +1,9 @@
 <script>
   document.title = 'Canvas Kit Storybook';
 </script>
+
+<style>
+  [title^='_hidden_'] {
+    display: none !important;
+  }
+</style>
```
NicholasBoll, line 9: This style rule will hide any Storybook item with a name that starts with "_hidden_". The `testingOnly` utility function will apply this.

cypress/helpers/modal.ts
```diff
@@ -1,16 +1,3 @@
...
- * @example
- * h.modal.get()
- *   .should('have.attr', 'role', 'dialog')
- */
-export function get(testId?: string): Cypress.Chainable<JQuery> {
-  const selector = testId ? `[data-testid='${testId}'] [role=dialog]` : `[role=dialog]`;
-
-  return cy.get(selector);
-}
-
```
NicholasBoll, line 13: This helper is not needed. We can actually use this instead!:

```ts
cy.findByLabelText(modalTitle)
```

cypress/integration/Modal.spec.ts
```diff
@@ -387,20 +367,68 @@ describe('Modal', () => {
...
+            .parent() // overlay
+            .click('top');
+        });
+
+        it('should close the second modal', () => {
+          cy.findByLabelText('Really Delete Item').should('not.exist');
+        });
+
+        it('should not close the first modal', () => {
+          cy.findByLabelText('Delete Item').should('exist');
```
NicholasBoll, line 431: These tests are now testing for the stacking functionality of escape key and clickOutside handling. If 2 modals are stacked, only the topmost modal should close.

cypress/integration/Popup.spec.ts
```diff
@@ -65,6 +65,28 @@ describe('Popup', () => {
...
+
+      context('when the user clicks outside the popup', () => {
+        beforeEach(() => {
+          cy.get('body').click('topLeft');
+        });
+
+        it('should close the popup', () => {
+          getPopup().should('not.visible');
+        });
+      });
```
NicholasBoll, line 89: I added escape and click outside behaviors to our popup example. These tests verify the functionality

cypress/integration/PopupStack.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+          key: 'Escape',
+        });
+      });
+
+      it('should close Delete Item popup', () => {
+        cy.findByLabelText('Delete Item').should('not.exist');
+      });
+    });
+  });
+});
```
NicholasBoll, line 153: Not all of these behaviors have been fully verified by accessibility and UX. I figure that will be a much longer conversation involving planning. Canvas Kit doesn't have the concept of a `Window` anyways, so I thought I'd put up an initial specification to get things started. Changes to some of these specs will not break anything since `Window` doesn't exist yet.

cypress/integration/Tooltip.spec.ts
```diff
@@ -49,8 +49,7 @@ describe('Tooltip', () => {
           });
         });
 
-        // TODO: implement this functionality
-        it.skip('should close the tooltip', () => {
+        it('should close the tooltip', () => {
```
NicholasBoll, line 52: Woot! Being able to now run skipped tests FTW!!

modules/_labs/color-picker/react/stories/stories_ColorPicker.tsx
```diff
@@ -5,7 +5,7 @@ import {action} from '@storybook/addon-actions';
 import withReadme from 'storybook-readme/with-readme';
 import {ColorInput} from '@workday/canvas-kit-react-color-picker';
 import {colors} from '@workday/canvas-kit-react-core';
-import {Popper} from '@workday/canvas-kit-react-common';
+import {Popper} from '@workday/canvas-kit-react-popup';
```
NicholasBoll, line 8: `Popper` is moving from common to popup. This is documented in the migration guide. Docs have been updated

modules/common/react/lib/utils/mergeCallback.ts
```diff
@@ -0,0 +1,23 @@
...
+export function mergeCallback<T extends Function>(callback1: T, callback2?: Function): T {
+  if (callback2) {
+    return (((...args: any[]) => {
+      callback2?.apply(null, args);
+      callback1.apply(null, args);
+    }) as any) as T;
+  } else {
+    return callback1;
+  }
+}
```
NicholasBoll, line 23: First merge strategy for a callback being defined in the component and also in `elemProps`. Calls both if both exist. Bonus the `@example` shows in intellisense:

Screen Shot 2020-06-04 at 5 32 12 PM

modules/modal/react/stories/stories.tsx
```diff
@@ -1,13 +1,43 @@
...
-import {storiesOf} from '@storybook/react';
+
 import withReadme from 'storybook-readme/with-readme';
 
+import {Modal, useModal} from '@workday/canvas-kit-react-modal';
+
 import {Button, DeleteButton} from '../../../button/react';
-import Modal, {useModal} from '..';
+import {FormField} from '../../../form-field/react';
+import {TextInput} from '../../../text-input/react';
```
NicholasBoll, line 10: There is examples of both relative imports and package name imports in stories. Personally I think we should do only package name imports, but it is annoying to maintain `devDependencies` that are only added for stories. I think we should have a lint rule for only package name imports in stories and remove the requirement for `devDependencies` to reflect what's in story files.

modules/modal/react/stories/stories.tsx
```diff
@@ -39,7 +69,7 @@ const DefaultModalExample = () => {
   );
 };
 
-const AccessibleModalExample = () => {
+export const AccessibilityTest = () => {
```
NicholasBoll, line 72: I'm not sure if this should be a testing only story. Sure, we run tests against this, but the contents of the story were requested by the accessibility team for testing (to make sure our content hiding methods work for mouse users, keyboard users, and users of all kinds of assistive technology).

Stories that are marked as "test only" are still viewable with the URL, but they just won't show up in the navigation.

modules/popup-stack/README.md
NicholasBoll: I don't think we'll re-export.

modules/popup/react/lib/Popper.tsx
```diff
@@ -0,0 +1,169 @@
...
+   * The reference element used to position the Popper. Popper content will try to follow the
+   * `anchorElement` if it moves and will reposition itself if there is no longer room in the
+   * window.
+   */
+  anchorElement: React.RefObject<Element> | Element | null;
+  /**
+   * The content of the Popper. If a function is provided, it will be treated as a Render Prop and
+   * pass the `placement` chosen by PopperJS. This `placement` value is useful if your popup needs
+   * to animate and that animation depends on the direction of the content in relation to the
+   * `anchorElement`.∑
```
NicholasBoll, line 21: Whoops

modules/popup/react/lib/useBringToTopOnClick.ts
```diff
@@ -0,0 +1,27 @@
...
+ * opened from another Popup. Usually this is a Tooltip or Select component inside something like a
+ * Modal.
+ */
+export const useBringToTopOnClick = <E extends HTMLElement>(ref: React.RefObject<E>) => {
+  let timer: number;
+  const onClick = (event: MouseEvent) => {
+    timer = requestAnimationFrame(() => {
+      if (ref.current?.contains(event.target as Node)) {
+        PopupStack.bringToTop(ref.current!);
+      }
```
NicholasBoll, line 17: This delay allows for other actions to be processed first. For example, if a `ClickOutside` is registered, it will be handled first - allowing another popup to close before being brought to top, which would cancel the closing (because it was no longer the topmost).

Here's an example without this delay.
bringToFront is processed before closeOnOutsideClick

Priority and cancellability of behaviors might be necessary. `requestAnimationFrame` isn't a very obvious solution to this problem. But we have yet to define how behaviors may interact. I think this is fine for now?

modules/popup/react/lib/useCloseOnOutsideClick.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @param onClose
+ */
+export const useCloseOnOutsideClick = <E extends HTMLElement>(
+  ref: React.RefObject<E>,
+  onClose: () => void
+) => {
+  const onClick = (event: MouseEvent) => {
+    if (
+      PopupStack.isTopmost(ref.current!) &&
+      !PopupStack.contains(ref.current!, event.target as HTMLElement)
```
NicholasBoll, line 20: Workday's Standard Tooling calls this an "autohidePartner". It included the `owner` element and children as an "inside" click. This allows the application decide if clicking on a popup's target should toggle the existence of the popup or if it should stay open. Without this, clicking the target would toggle the popup and there would be no way to opt-out of the behavior. With this logic, it is an opt-in to toggle.

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -0,0 +1,220 @@
...
+      stack.items = [...stack.items.filter(i => i !== item), item];
+
+      // Also bring children to top. There are a few cases where stacking might break otherwise:
+      // - Clicking a Popup calls `bringToTop`, but mouse is over a Tooltip so that Tooltip is now
+      //   under the Popup
+      // - Clicking a button opens a new Popup, but that click bubbles up to a `bringToTop` call
+      //   putting the new popup under an existing one
+      getChildPopups(item, stack.items).forEach(popup => {
+        PopupStack.bringToTop(popup.element);
+      });
```
NicholasBoll, line 183: If we didn't do this, there could be a race condition that would cause the new popup to appear behind a window with `bringToFront`:

popup_dom_event_bubbling

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -0,0 +1,220 @@
...
+   * to this item. If the previous are true, all "children" stack items will be brought to top as
+   * well and will be on top of the element passed to `bringToTop`. This maintains stack item
+   * "hierarchy" so that stack items like Popups and Tooltips don't get pushed behind elements they
+   * are supposed to be on top of.
+   *
+   * This does not need to be called when a popup is added since added popups are already place on
+   * the top of the stack.
+   */
+  bringToTop(element: HTMLElement): void {
+    const item = stack.items.find(i => i.element === element);
```
NicholasBoll, line 171: `.find` is not available in IE. Fall back to `indexOf`

modules/popup/react/README.md
```diff
@@ -124,3 +257,90 @@ Default:
...
+This should be used stacked UI elements that are dismissible like Tooltips, Modals, non-modal
+dialogs, dropdown menus, etc.
+
+## useCloseOnOutsideClick
+
+```ts
+useCloseOnOutsideClick(ref: React.RefObject<HTMLElement>, onClose: () => void): void
+```
+
+Registers global lister for all clicks. It will only call the `onClose` callback if the click
```
jamesfan, line 327: Should read `listener` instead of `lister`.

modules/popup/react/lib/useCloseOnOutsideClick.ts
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import {PopupStack} from '@workday/canvas-kit-popup-stack';
+
+/**
+ * Registers global lister for all clicks. It will only call the `onClose` callback if the click
```
jamesfan, line 6: Should read `listener` instead of `lister`.

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -169,52 +155,24 @@ const ModalContent = ({
...
-          el.removeAttribute('aria-hidden');
-        }
-      });
-    };
-  }, []);
+  useAssistiveHideSiblings(modalRef);
 
   const content = (
-    <Container onClick={handleOutsideClick} {...elemProps}>
+    <Container {...elemProps} ref={modalRef} onClick={onOverlayClick}>
```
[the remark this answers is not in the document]
anicholls: I would say it's unlikely that people would apply an onClick listener here, but it wouldn't hurt to merge.

modules/modal/react/stories/stories.tsx
anicholls: Maybe it doesn't need to be a testing only story, but I think it could live in the `Testing` section (which would require another file

modules/popup/react/lib/useBringToTopOnClick.ts
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import {PopupStack} from '@workday/canvas-kit-popup-stack';
+
+/**
+ * This hook will bring an element to the top of the stack when any element inside the provided
+ * `ref` element is clicked. If `Popper` was used or `PopupStack.add` provided an `owner`, all
+ * "child" popups will also be brought to the top. A "child" popup is a Popup is a Popup that was
```
anicholls, line 7:
```suggestion
 * "child" popups will also be brought to the top. A "child" popup is a Popup that was
```

modules/popup/react/README.md
```diff
@@ -1,9 +1,21 @@
...
+different requirements of behavior for UX and accessibility - we can call them capabilities,
+behaviors, or traits. Canvas Kit comes with a number of [behaviors](#hooks) in the form of React
+Hooks.
 
-Note: This popup does not include a positioning engine. In our example we use our popper utility,
-which is a wrapper to Popper.js, to wrap our Popup component and position it.
+If you are building your own custom stacked UI components, use the [Popper](#popper) component along
+with our [hooks](#hooks). The `Popper` component and hooks work with the stack management system for
+correct rendering and accessibility behavior. If you cannot use `Popper`, use the
+[usePopupStack](#usepoupstack) hook. to properly register and deregister the popup at the correct
```
anicholls, line 13:
```suggestion
[usePopupStack](#usepoupstack) hook to properly register and deregister the popup at the correct
```

modules/popup/react/README.md
```diff
@@ -124,3 +257,90 @@ Default:
...
+
+## useBringToTopOnClick
+
+```ts
+useBringToTopOnClick(ref: React.RefObject<HTMLElement>): void
+```
+
+This hook will bring an element to the top of the stack when any element inside the provided `ref`
+element is clicked. If `Popper` was used or `PopupStack.add` provided an `owner`, all "child" popups
+will also be brought to the top. A "child" popup is a Popup is a Popup that was opened from another
```
anicholls, line 303:
```suggestion
will also be brought to the top. A "child" popup is a Popup that was opened from another
```

modules/popup/react/README.md
```diff
@@ -124,3 +257,90 @@ Default:
...
+Popup. Usually this is a Tooltip or Select component inside something like a Modal.
+
+This should be used on stacked UI elements that are meant to persist, like Windows.
+
+## useCloseOnEscape
+
+```ts
+useCloseOnEscape(ref: React.RefObject<HTMLElement>, onClose: () => void): void
```
anicholls, line 311: Many of these don't return `void` - they return a function to remove the listeners

modules/popup-stack/README.md
```diff
@@ -0,0 +1,181 @@
...
+clicked on. This is how operating system "windows" work. The behavior requires global coordination
+to determine the order that these popups should be rendered.
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-popup-stack
+```
+
+## Usage
```
anicholls, line 57: I think we should probably call out to the hooks here for React users

modules/modal/react/stories/stories.tsx
```diff
@@ -156,62 +201,81 @@ const NoCloseModalExample = () => {
...
+          <Button onClick={modal2.closeModal} variant={Button.Variant.Secondary}>
+            Cancel
+          </Button>
+        </Modal>
+      </Modal>
+    </>
+  );
+};
+
+testingOnly(StackedModals);
```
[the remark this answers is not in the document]
NicholasBoll: I think if we move to a testing category, we don't need `testingOnly`

modules/popup-stack/stories/stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+            Delete
+          </DeleteButton>
+          <Button onClick={onClose}>Cancel</Button>
+        </Popup>
+      </Popper>
+    </div>
+  );
+};
+
+export const MixedPopupTypes = () => {
```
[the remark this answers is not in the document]
NicholasBoll: Hmm, well there isn't any story other than this one, so there's no place for the Readme to go...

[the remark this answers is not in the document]
anicholls: Could we just link tot he stack readme from popup, modal, toast, etc.?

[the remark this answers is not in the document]
NicholasBoll: I created a new `Default` story using the raw `PopupStack` with no React (except to bootstrap and cleanup a story). I can attach a readme to this.

modules/popup/react/README.md
NicholasBoll: I think you're talking about callback functions given to `use[Layout]Effect`. The callback here is usually an event listener or a callback function that takes no arguments. For example, `handleClose` in `Popup.tsx` is `() => void`.

This is the definition of a `MouseEvent` in React:
```ts
type React.MouseEventHandler<T = Element> = (event: MouseEvent<T, globalThis.MouseEvent>) => void
```

There is nothing to capture the return value of the callback.

modules/popup-stack/README.md
NicholasBoll: Hooks are in the Popup module. There isn't really a good way to link to other docs currently.

modules/popup/react/lib/useBringToTopOnClick.ts
lychyi: Does this need to be part of code comments? I think I would have a hard time understanding this without this PR comment. I don't want this rationale to get lost.

modules/popup/react/lib/useCloseOnOutsideClick.ts
lychyi: This sounds like it's an important part of the codebase for future knowledge?

> Without this, clicking the target would toggle the popup and there would be no way to opt-out of the behavior. With this logic, it is an opt-in to toggle.

NicholasBoll: I'll also add that to the JSDoc of `PopupStack.contains`

NicholasBoll: Screen Shot 2020-06-09 at 1 10 37 PM

modules/popup/react/lib/useBringToTopOnClick.ts
NicholasBoll:
```tsx
    // This delay allows for other actions to be processed first. For example, if a `ClickOutside`
    // is registered, it will be handled first - allowing another popup to close before being
    // brought to top, which would cancel the closing (because it was no longer the topmost).

    // Here's an example without this delay. ![bringToFront is processed before
    // closeOnOutsideClick](https://user-images.githubusercontent.com/338257/83921416-87b64580-a73b-11ea-913b-6bd0592def74.gif)

    // Priority and cancellability of behaviors might be necessary. `requestAnimationFrame` isn't a
    // very obvious solution to this problem. But we have yet to define how behaviors may interact.
    // I think this is fine for now?
    timer = requestAnimationFrame(() => {
      if (ref.current?.contains(event.target as Node)) {
        PopupStack.bringToTop(ref.current!);
      }
    });
```

modules/popup/react/README.md
anicholls: Whoops - you're right. For some reason I missed the `use[Layout]Effect` when I saw the return statement and commented here without really thinking things through. You're good 👍

modules/popup-stack/README.md
anicholls: Yeah I didn't mean document them here. Can't we just do something like:

```
If you're using the Popup Stack within a React app, 
we provide [convenient hooks](../popup/react/README.md#hooks) 
for these behaviors in `@workday/canvas-kit-react-popup`
```

modules/modal/react/stories/stories_testing.tsx
```diff
@@ -0,0 +1,166 @@
...
+import {
+  Popup,
+  Popper,
+  usePopup,
+  useCloseOnOutsideClick,
+  useCloseOnEscape,
+} from '@workday/canvas-kit-react-popup';
+
+export default {
+  title: 'Testing|Popups/Modal/React',
```
jamesfan, line 15: Should this be `Testing|React/Popups/Modal` instead?

modules/popup/react/stories/stories_testing.tsx
```diff
@@ -0,0 +1,35 @@
+/// <reference path="../../../../typings.d.ts" />
+import * as React from 'react';
+
+import {Button} from '@workday/canvas-kit-react-button';
+import {Popper, Popup, usePopup, useCloseOnOutsideClick} from '@workday/canvas-kit-react-popup';
+
+export default {
+  title: 'Testing|Popups/Popup/React',
```
jamesfan, line 8: Should this be `Testing|React/Popups/Popup` instead?

modules/modal/react/stories/stories_testing.tsx
NicholasBoll: Ah. Good catch. I didn't see the change to navigation hierarchy here

NicholasBoll: There is still going to be a `Popups` top-level since `Popup Stack` is not React specific... That seems odd... Maybe just `Popup Stack`?

anicholls: Or we could have `Testing|JavaScript/...`

PR #671, opened 2020-05-21
fix(switch): Fix click target for switch

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/657

Make the input display block pinning the input to the left

## Before:
Screen Shot 2020-05-21 at 2 49 15 PM

## After: 
Screen Shot 2020-05-21 at 2 53 07 PM

### DIFF
--- a/modules/switch/react/lib/Switch.tsx
+++ b/modules/switch/react/lib/Switch.tsx
@@ -63,6 +63,7 @@ const SwitchInput = styled('input')<SwitchProps>(
     marginLeft: spacing.xxxs,
     borderRadius: borderRadius.circle,
     opacity: 0,
+    display: 'block', // make click target width of input
   },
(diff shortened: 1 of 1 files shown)


modules/switch/react/lib/Switch.tsx
```diff
@@ -63,6 +63,7 @@ const SwitchInput = styled('input')<SwitchProps>(
     marginLeft: spacing.xxxs,
     borderRadius: borderRadius.circle,
     opacity: 0,
+    display: 'block', // make click target width of input
```
lychyi, line 66: I don't think this comment is quite accurate. 1) It doesn't change the width and 2) this _is_ the input 😄 

This is makes the input a block element which fixes this by pinning the input to the left (in left-to-right that is)

PR #672, opened 2020-05-22
fix(select): Fix miscellaneous UI issues

jamesfan:
## Summary

Packages a few UI fixes to the Select component:

1. Improves the button border animation when blurring off of the menu. Previously, the button animation (from a thicker blue border to a thinner gray border) triggered _after_ the menu faded out on blur, creating a disjointed visual. Now, the button animation is synced with the menu fade out animation.

2. Previously, the button was not receiving focus after it had been clicked to dismiss the menu. This has been fixed.

3. Adjusts option selection to occur on click rather than mousedown to be more consistent with how a native select option behaves.

4. Prevents options from inheriting `text-align`.

## Checklist

- [x] `yarn test` passes


lychyi:
Do you think we can fix another UI issue here? It can be a separate PR too.

When the container specifies the `text-align` property the SelectOptions inherit that alignment too. We found this as part of an issue identified in #657.

Screen Shot 2020-05-20 at 3 12 08 PM

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -187,12 +189,12 @@ const SelectButton = styled('button')<Pick<SelectBaseProps, 'error' | 'grow' | '
       },
     },
   },
-  ({error, isMenuHidden}) => {
+  ({error, isMenuHidden, isMenuHiding}) => {
```
alanbsmith, line 192: This isn't relevant to this PR, but I really don't like that `isMenuHidden` and `isMenuHiding` are decoupled bits of state. This feels like something we should fix later on. I would opt for a `menuVisibility` state that can either be `visible`, `hiding`, or `hidden`.

jamesfan: Good call, might be a quick fix I can address in this PR -- otherwise, I'll make a note of it for a later PR (or maybe package it into #641 since I'm already making adjustments to the menu there).

alanbsmith: Yeah, no worries. I don’t want to introduce scope creep on what would be an isolated change. Totally up to you. 👍

jamesfan:
@anicholls I'll handle the suggestions from @alanbsmith in a separate PR. Merged.

Issue #675, opened 2020-05-25
Avatar misaligned on IE


alexandrzavalii:
## 🐛 Bug Report

Avatar is missaligned on IE with version 4.

<img width="433" alt="Screen Shot 2020-05-25 at 3 58 15 PM" src="https://user-images.githubusercontent.com/10633451/82824231-de509380-9ea0-11ea-9637-163c695c43eb.png">

## Browser (if applicable)

IE11

anicholls:
Closed by #676

PR #676, opened 2020-05-25
fix(avatar): Fix misalignment on ie11

alexandrzavalii:
## Summary

fixes: #675 

## Checklist

- [x] `yarn test` passes


lychyi:
We will have to add some visual tests against this before merging. Someone on the core team can do that. But I just did a very similar fix so this PR looks like the right direction. We just got to be careful if borders can change width in different states.

alexandrzavalii:
@lychyi I can try adding visual testing myself.
is it by just adding additional "Visual Testing" story?

modules/avatar/react/lib/Avatar.tsx
```diff
@@ -68,26 +67,28 @@ const StyledContainer = styled('button', {
...
-    height: size,
-    width: size,
-  }),
-  ({variant, onClick}) => ({
-    cursor: onClick ? 'pointer' : 'default',
     '&:not([disabled])': {
       '&:focus': {
         outline: 'none',
-        ...(variant === AvatarVariant.Dark ? focusRing({separation: 2}) : focusRing()),
+        ...focusRing({separation: 2}),
```
anicholls, line 73: Note for posterity: I realized that there are no guarantees of the avatar color regardless of the variant so it doesn't make sense for this to be conditional

modules/common/react/lib/styles/hideMouseFocus.ts
```diff
@@ -5,9 +5,9 @@ import {CSSObject} from '@emotion/core';
  * Requires wrapping your components in an InputProvider component.
  */
 export const hideMouseFocus: CSSObject = {
-  [`[data-whatinput='mouse'] &:focus,
-    [data-whatinput='touch'] &:focus,
-    [data-whatinput='pointer'] &:focus`]: {
+  [`[data-whatinput="mouse"] &:focus,
```
lychyi, line 8: Can we back out of the formatting changes here and in `styled.ts` below?

anicholls: These aren't formatting - this was intentional. I ran into issues with the `StaticStates` wrapper because it the regex was looking for `"` and not `'`. 

I updated this to be more technically correct (attributes generally use `"`), and updated the regex in `styled` to catch either

Issue #678, opened 2020-05-26
ToastPopup actionLink on same line as Toast message


Andy93:
## 🐛 Bug Report

Action Link is on the same line as the toast message now and it should be below

## To Reproduce

Steps to reproduce the behavior: Create a toast component with Actionlink and toast message : https://codesandbox.io/s/toast-regression-on-v4-l6cv8?file=/src/index.tsx

## Expected Behavior

Action link to be below the toast message

## Actual Results

Action link is on the same line as the toast message

## Browser (if applicable)

Chrome

## Link to repl or repo (highly encouraged)

Toast component

![Screenshot 2020-05-26 at 15 38 10](https://user-images.githubusercontent.com/1674471/82918289-68b1f980-9f6c-11ea-8687-2bb5dc0e509f.png)

mannycarrera4:
hey @Andy93 thanks for submitting the issue!

Looks like in `v4` we migrated some `type` styles and since the action button extends our `link` type style, it's adding a `display: inline-block` : https://github.com/Workday/canvas-kit/blob/prerelease/v4/modules/core/react/lib/type.ts#L142 resulting in the behavior that you're seeing.

This should be a simple fix to override that display property to `display: block` or `display:flex`

alexandrzavalii:
Thanks for getting back! this is the style for the button.
```
const ActionButton = styled('button')({
  display: 'block',
  backgroundColor: 'transparent', // To prevent Safari from rendering grey 'buttonface' as bgcolor
  border: 'none',
  padding: 0,
  marginTop: spacing.xxxs,
  ...type.body2,
  ...type.variant.link,
});
```
Should we move all the destructuring at the top of the object?

mannycarrera4:
@alexandrzavalii I was just discussing this with ! That might be the approach we take. Still figuring out if  that's what we want to do or use something like `HyperLink` component, however, that changes the design a bit.

anicholls:
Closed by #682

PR #682, opened 2020-05-27
fix(toast): Action link align on new line

alexandrzavalii:
## Summary

fixes: #678 

## Checklist

- [x] `yarn test` passes
      applicable


modules/toast/react/stories/stories_visualTesting.tsx
```diff
@@ -39,7 +39,7 @@ storiesOf('Components|Popups/Toast/React/Visual Testing', module)
       >
         {props => (
           <Toast transformOrigin={null} aria-label="Play" {...props}>
-            Your workbook was successfully processed.
+            Successfully processed.
```
mannycarrera4, line 42: This is to prove that when the text is a single line, the link won't wrap up. 👍 

You can maybe add a story that has multiple lines to cover that use case as well

alexandrzavalii: will do!
thanks

modules/core/react/lib/type.ts
```diff
@@ -149,6 +149,7 @@ const variants: CanvasTypeVariant = {
     },
     '&:focus': {
       boxShadow: `0 0 0 2px ${colors.blueberry400}`,
+      outline: 'none',
```
alexandrzavalii, line 152: no need for outline when there is boxShadow 
Screen Shot 2020-05-27 at 3 36 29 PM

PR #684, opened 2020-05-28
feat: Add toolbar dropdown button

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/685
Add toolbar dropdown button

Blocked on #688


modules/button/react/README.md
```diff
@@ -760,7 +760,9 @@ Default: `undefined`
 
 ---
 
-# ToolbarIconButton
+# Toolbar Buttons
+
+## ToolbarIconButton
```
mannycarrera4, line 766: need to add documentation on the dropdown button

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  };
+
+  const chevronStyles = {
+    margin: '0 2px 0 0',
+  };
+
+  const customIconStyles = {
+    marginLeft: '2px',
+    marginRight: 0,
+  };
```
mannycarrera4, line 47: I needed these styles for the right styling per design, I can sync up with design again to see if this is necessary

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  return (
+    <ButtonContainer
+      colors={getToolbarIconButtonColors(theme)}
+      ref={buttonRef}
+      extraStyles={containerStyles}
+      aria-label={iconArialabel}
+      {...elemProps}
+    >
+      {icon ? <SystemIcon css={customIconStyles} icon={icon} /> : children}
+      <SystemIcon css={chevronStyles} icon={chevronDownSmallIcon} />
```
mannycarrera4, line 58: The chevron is always on the right, should RTL come in to play here?

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+    </ButtonContainer>
+  );
+};
+
+export default ToolbarDropdownButton;
+
+const getToolbarIconButtonColors = (theme: EmotionCanvasTheme): ButtonColors => {
```
mannycarrera4, line 65: rename to dropdown, bad copy pasta

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const containerStyles = {
+    padding: 0,
+    minWidth: spacing.l,
+    width: 'auto',
+    height: spacing.l,
+    borderRadius: borderRadius.m,
+    'span > *': {
+      margin: 0,
+    },
+    ['& .wd-icon']: {
```
alexandrzavalii, line 32:
```suggestion
    '& .wd-icon': {
```

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+      height: '20px',
+    },
+  };
+
+  const chevronStyles = {
+    margin: '0 2px 0 0',
+  };
+
+  const customIconStyles = {
+    marginLeft: '2px',
```
alexandrzavalii, line 45:
```suggestion
    marginLeft: 2
```

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  theme = useTheme(),
+  buttonRef,
+  'aria-label': iconArialabel,
+  icon,
+  children,
+  ...elemProps
+}: ToolbarDropdownButtonProps) => {
+  const containerStyles = {
+    padding: 0,
+    minWidth: spacing.l,
```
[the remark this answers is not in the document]
mannycarrera4: the spacing?

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+      height: 20,
+    },
+  };
+
+  const chevronStyles = {
+    margin: '0 2px 0 0',
+  };
+
+  const customIconStyles = {
+    marginLeft: `${spacing.xxxs} !important`,
```
mannycarrera4, line 42: ok so I added this cause button constainer adds
 ```
'& > *:first-of-type': {
      marginLeft: 0,
    },
```
I needed to override this per the design, thoughts on thios?

File: modules/button/react/lib/ToolbarDropdownButton.tsx (hunk 7 above)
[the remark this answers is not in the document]
Comment (alexandrzavalii) [reply modules/button/react/lib/ToolbarDropdownButton.tsx:25]: yes

File: modules/button/react/lib/ToolbarDropdownButton.tsx [hunk 9]
```diff
@@ -0,0 +1,87 @@
...
+const ToolbarDropdownButton = ({
+  theme = useTheme(),
+  buttonRef,
+  'aria-label': iconArialabel,
+  icon,
+  children,
+  ...elemProps
+}: ToolbarDropdownButtonProps) => {
+  const containerStyles = {
+    padding: spacing.zero,
```
alexandrzavalii, line 24: should we consistently use `spacing.zero` when we want to set 0? 
If yes then we should change this.
```
const chevronStyles = {
    margin: '0 2px 0 0',
}
```

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+    borderRadius: borderRadius.m,
+    '& .wd-icon': {
+      display: 'inline-block',
+      verticalAlign: 'middle',
+      width: 20,
+      height: 20,
+    },
+  };
+
+  const chevronStyles = {
```
alexandrzavalii, line 37: is there a reason why we have the styles inside the function? they are getting recreated every render

modules/button/react/lib/ToolbarDropdownButton.tsx
mannycarrera4: There's other instances in our code base where we use `0` instead of `spacing.zero`

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+import {colors, spacing, borderRadius} from '@workday/canvas-kit-react-core';
+import {focusRing, useTheme, Themeable, EmotionCanvasTheme} from '@workday/canvas-kit-react-common';
+import {SystemIcon} from '@workday/canvas-kit-react-icon';
+import {ButtonColors} from './types';
+import {ButtonContainer} from './parts';
+import {IconButtonProps} from './IconButton';
+import {chevronDownSmallIcon} from '@workday/canvas-system-icons-web';
+
+export interface ToolbarDropdownButtonProps
+  extends Omit<IconButtonProps, 'size' | 'variant' | 'toggled' | 'onToggleChange'>,
```
anicholls, line 13: Nit: This will be cleaner if you extend `ToolbarIconButtonProps`

anicholls: Also, are we certain there is no use case for toggling these buttons? If there were, I would rather this just be a `dropdown?: boolean` prop in `ToolbarIconButton`

modules/button/react/README.md
```diff
@@ -760,6 +761,8 @@ Default: `undefined`
 
 ---
 
+# Toolbar Buttons
+
```
anicholls, line 765:
```suggestion
```
Nit: I don't think this is necessary, especially when it's at the same header level as the components themselves

modules/button/react/README.md
```diff
@@ -825,3 +828,51 @@ Default: `undefined`
 > The icon of the button. Optional because ToolbarIconButton can also wrap a SystemIcon component.
 
 ---
+
+# ToolbarDropdownButton
+
+> Button containing an icon or custom element. Icon may be a component from
```
anicholls, line 834: Can you add a note here and for `ToolbarIconButton` about their specific usage for toolbars?

modules/button/react/README.md
```diff
@@ -825,3 +828,51 @@ Default: `undefined`
...
+
+## Usage
+
+```tsx
+import * as React from 'react';
+import {ToolbarDropdownButton} from '@workday/canvas-kit-react-button';
+import {activityStreamIcon} from '@workday/canvas-system-icons-web';
+<ToolbarDropdownButton
+  icon={activityStreamIcon}
+  title="Activity Stream"
```
anicholls, line 846: Generally we suggest just aria-label (and maybe a Tooltip), don't we? The `title` tag is a little ugly :P

modules/button/react/README.md
```diff
@@ -825,3 +828,51 @@ Default: `undefined`
...
+
+### Optional
+
+#### `buttonRef: React.Ref<HTMLButtonElement>`
+
+> Returns the ref to the rendered HTMLButtonElement.
+
+---
+
+### `icon: CanvasSystemIcon`
```
anicholls, line 873:
```suggestion
#### `icon: CanvasSystemIcon`
```

modules/button/react/lib/ToolbarDropdownButton.tsx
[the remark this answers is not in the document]
anicholls: @alexandrzavalii Spacing doesn't live in the theme object ATM

modules/button/react/lib/ToolbarDropdownButton.tsx
anicholls: I would just give this element a class below (e.g. `wdc-toolbar-dropdown-btn-arrow`) and move this to `containerStyles`. We should avoid `!important` wherever possible.

modules/button/react/stories/stories_ToolbarButttons.tsx
```diff
@@ -1,17 +1,22 @@
 /// <reference path="../../../../typings.d.ts" />
```
anicholls, line 1: There are 3 t's in the name of this file 😂

modules/button/react/stories/stories_visualTesting.tsx
```diff
@@ -364,6 +365,27 @@ const buttonStories = [
       </React.Fragment>
     ),
   },
+  {
+    name: 'Toolbar Dropdown Button',
```
anicholls, line 369: Note: This will have to change with #688

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+export default ToolbarDropdownButton;
+
+const getToolbarDropdownButtonColors = (theme: EmotionCanvasTheme): ButtonColors => {
+  return {
+    default: {
+      icon: colors.licorice200,
+      background: 'transparent',
+    },
+    hover: {
+      icon: colors.licorice500,
```
anicholls, line 71: Active and hover are using the same color (same goes for `ToolbarIconButton` actually). Is this intended?

modules/button/react/stories/stories_ToolbarButttons.tsx
```diff
@@ -1,17 +1,22 @@
...
+import {activityStreamIcon, paragraphIcon} from '@workday/canvas-system-icons-web';
 
-import {ToolbarIconButton} from '../index';
+import {ToolbarIconButton, ToolbarDropdownButton} from '../index';
 
 import README from '../README.md';
 
-storiesOf('Components|Buttons/Button/React/Toolbar Button', module)
+const customElementStyles = {
+  margin: '0 16px 0 2px',
```
anicholls, line 15: This left margin feels really tight - was this specced?

modules/button/react/README.md
mannycarrera4: ah i think this was copy pasta from icon button, I'll check those examples as well to remove title!

modules/button/react/lib/ToolbarDropdownButton.tsx
mannycarrera4: it's per design! i Will double confirm to make sure this is the intended styles

modules/button/react/lib/ToolbarDropdownButton.tsx
mannycarrera4: I've talked to design and this button is not toggleable, so I just extended the props from toolbarIconButton which did clean up the omits.

modules/button/react/lib/ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const ToolbarDropdownButton = ({
+  theme = useTheme(),
+  buttonRef,
+  'aria-label': iconArialabel,
+  icon,
+  children,
+  ...elemProps
+}: ToolbarDropdownButtonProps) => {
+  const containerStyles = {
```
alexandrzavalii, line 23: is there a reason of keeping `containerStyles` inside a function?

mannycarrera4: nope! I moved them up

modules/button/react/stories/stories_visualTesting.tsx
mannycarrera4: @anicholls i changed this but I have one question, for the themed visual stories, why is the pattern `Testing|React/Buttons/Button` instead of something like `Testing|React/Buttons/Hyper link/theming` ? 

Also i noticed that for `IconButton` theming this is the path: https://github.com/anicholls/canvas-kit/blob/update-prerelease/modules/button/react/stories/visual-testing/theming/stories_IconButton.tsx#L7

And it doesn't match the other theming paths : https://github.com/anicholls/canvas-kit/blob/update-prerelease/modules/button/react/stories/visual-testing/theming/stories_DeleteButton.tsx#L7 not sure if this is intentional or not

anicholls: @mannycarrera4 the hierarchy is as follows:

`Testing|<TARGET>/<CATEGORY>/<MODULE>/<COMPONENT>`

It's a bit weird to see `Buttons/Button`, but it makes more sense when you see something like `Inputs/TextInput`. The full name of the stories comes from the exported story names, so we end up with `Testing|React/Buttons/Button/Hyperink/Hyperlink States` and `Testing|React/Buttons/Button/Hyperink/Hyperlink Theming States` (note - we repeat the component name in the final story name so that it's easier to read in Chromatic).

Let me know if you still have questions after this explanation.

---

Good catch on the IconButton path. We can update that in #688.

mannycarrera4: awesome thanks for explaining that!  That makes sense

modules/button/react/stories/visual-testing/stories_ToolbarIconButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+import {Container, stateTableColumnProps} from './utils';
+
+export default withSnapshotsEnabled({
+  title: 'Testing|React/Buttons/Button/Toolbar Icon Button',
+  component: ToolbarIconButton,
+});
+
+export const ToolbarIconButtonStates = () => (
+  <React.Fragment>
+    {[false, true].map(toggled => (
```
anicholls, line 22: Does toggled really need to be another table? We did it for icon buttons so it was a little easier to read, but when there is only one row per table, why not just combine them and make `toggled` a row prop?

modules/button/react/stories/visual-testing/theming/stories_ToolbarIconButton.tsx
```diff
@@ -0,0 +1,15 @@
+/// <reference path="../../../../../../typings.d.ts" />
+import {customColorTheme, withSnapshotsEnabled} from '../../../../../../utils/storybook';
+import {ToolbarIconButton} from '../../../';
+import {ToolbarIconButtonStates} from '../stories_ToolbarIconButton';
+
+export default withSnapshotsEnabled({
+  title: 'Testing|React/Buttons/Button',
```
anicholls, line 7:
```suggestion
  title: 'Testing|React/Buttons/Button/Toolbar Icon Button',
```

utils/storybook/withSnapshotsEnabled.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+/**
+ * A function to enable chromatic snapshots for a single story or a set of stories.
+ * To apply to all stories in a file, wrap the default export of a CSF story file.
+ * To apply to a single story, wrap the story fn.
+ *
+ * TODO: Clean up after optional chaining is enabled
+ */
+export const withSnapshotsEnabled: {
```
anicholls, line 26: I figured you would just rebase on #688 rather than pulling its files in. Do you mind if we hold off on merging this until that goes in so that these files are associated with that PR instead of this one?

modules/button/react/stories/visual-testing/theming/stories_ToolbarDropdownButton.tsx
```diff
@@ -0,0 +1,15 @@
+/// <reference path="../../../../../../typings.d.ts" />
+import {customColorTheme, withSnapshotsEnabled} from '../../../../../../utils/storybook';
+import {ToolbarDropdownButton} from '../../../';
+import {ToolbarDropdownButtonStates} from '../stories_ToolbarDropdownButton';
+
+export default withSnapshotsEnabled({
+  title: 'Testing|React/Buttons/Button',
```
anicholls, line 7:
```suggestion
  title: 'Testing|React/Buttons/Button/Toolbar Dropdown Button',
```

utils/storybook/withSnapshotsEnabled.ts
mannycarrera4: that works for me

Issue #683, opened 2020-05-27
Improve visual testing coverage


anicholls:
In an effort to reduce the number of snapshots we were generating, we disabled chromatic snapshots by default and made them opt-in (#478). To improve coverage, while limiting the number of snapshots, we've added the `StaticStates` and `ComponentStatesTable` components. However, these haven't been used on a large number of components, so since #478, we haven't had the greatest coverage in our visual testing.

In preparation for v4 (`prerelease/v4`) being merged, we want to improve our coverage so we're more likely to catch regressions when releasing the update.

---

### Visual Testing Infrastructure
- [x] Add helper to enable chromatic snapshots as it's easy to typo right now
- [x] Use component + variant name in story name so it's more obvious in Chromatic
- [x] Pull visual testing stories into their own `Testing` top level section within storybook.
- [x] Switch to CSF format for these stories
- [x] Update create-component to use CSF and create a visual testing story for you

### Visual Testing Coverage
Here are the modules in the `master` branch that are missing snapshot tests at the moment:
- [ ] labs/combobox
- [x] labs/core (#709)
- [x] labs/drawer (#712)
- [ ] labs/header
- [ ] labs/menu
- [ ] action-bar
- [x] avatar (#690)
- [x] button
- [x] card
- [ ] common
- [ ] cookie-banner
- [x] core (#696)
- [ ] icon
- [x] SegmentedControl (#710)
- [ ] layout
- [ ] loading-animation
- [x] modal (#693)
- [ ] Menu (#714)
- [ ] SearchBar (#713)
- [ ] page-header
- [x] Popup (#691)
- [x] side-panel
- [x] skeleton (#717)
- [x] status-indicator (#692)
- [ ] table
- [x] tooltip (#691)

Of these modules, these are the modules who've had snapshot tests added in the `prerelease/v4` branch. These will need to be moved to master.
- avatar (#690)
- tooltip (#691)

---

We will need to add a new visual testing story for each of these modules. Ensure we make use of `StaticStates` and `ComponentStatesTables` to limit the number of visual testing stories (ideally one story per module). Each of these stories will have to manually enable chromatic snapshots:
```tsx
  .addParameters({
    component: Foo,
    chromatic: {
      disable: false,
    },
  })
```

Also, make sure we consider how each component could be broken with changes to subcomponents when structuring the visual testing story (for example, see #682)

lychyi:
This was erroneously closed in #727

PR #686, opened 2020-05-29
test: Improve snapshot infrastructure

anicholls:
Addresses infrastructure improvements in https://github.com/Workday/canvas-kit/issues/683

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -265,12 +265,12 @@ Component helpers come in 3 flavors:
 
 ## Visual Tests
 
-Canvas Kit uses [ChromaticQA](https://www.chromaticqa.com/) for visual tests which are run on stories
-that are opted-in through a special parameter. All the visual states should be represented at least
-one story. This way all of the visual states of a component can be visually regression tested without
-requiring the test to interact with the UI. These states should include loading states, error states,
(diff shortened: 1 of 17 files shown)


TESTING.md
```diff
@@ -292,10 +292,8 @@ export const MyVisualStory = () => {
 
 MyVisualStory.story = {
   parameters: {
-    chromatic: {
-      disabled: false,
-    },
-  }
+    ...enableSnapshots(),
+  },
```
NicholasBoll, line 296: An alternate would be a CSF-only helper:

```ts
enableSnapshots(MyVisualStory)
```

That seems a little more magical, but I'm looking at a helper function to hide a story and currently that cannot be done with parameters:

```ts
MyVisualStory.story = {
  ...testingOnly(),
  parameters: {
    ...enableSnapshots(),
  }
}
```

This is less "magic", but doesn't seem easier to understand. Alternatively, the "decorator"-type pattern seems like the helper functions are at the same level of abstraction:

```ts
testingOnly(MyVisualStory)
enableSnapshots(MyVisualStory)

// or
testingOnly(enableSnapshots(MyVisualStory))
```

PR #687, opened 2020-05-29
test: Improve snapshot infrastructure and organization

anicholls:
Addresses infrastructure improvements in https://github.com/Workday/canvas-kit/issues/683

- [x] Use component + variant name in story name so it's more obvious in Chromatic
- [x] Pull visual testing stories into their own `Testing` top level section within storybook.
- [x] Switch to CSF format for these stories
- [x] Update create-component to use CSF and create a visual testing story for you
- [x] Add [helper to enable chromatic snapshots](https://github.com/Workday/canvas-kit/pull/687/files#diff-a0609b5321f153d7f887ac35e7646044R26) as it's easy to typo right now:
- > The CSF decorator can be used on an entire file of stories, or a single story. It only works for CSF formats.
- All stories in a file:
  ```tsx
  export default withSnapshotsEnabled({
    // CSF details (title, component, etc.)
  });
  ```
- Specific story:
  ```tsx
  export const MyVisualStory = () => {
    // story contents
  };
  
  withSnapshotsEnabled(MyVisualStory);
  ```

New "Testing" section: 
image


anicholls:
One thought: I wonder if we could update our storybook config to automatically enable chromatic if the story is in the `Testing|` top level category.

utils/storybook/withSnapshotsEnabled.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+/**
+ * A function to enable chromatic snapshots for a single story or a set of stories.
+ * To apply to all stories in a file, wrap the default export of a CSF story file.
+ * To apply to a single story, wrap the story fn.
+ *
+ * TODO: Clean up after optional chaining is enabled
+ */
+export const withSnapshotsEnabled: {
```
anicholls, line 26: Note: I looked at making this a regular storybook decorator so you could do something like
```tsx
export default {
  title: 'My Story',
  decorators: [withSnapshotsEnabled]
}
```
However, editing the `storyFn.story` object doesn't seem to work from within the decorator function. This must be a limitation of the Storybook decorator API.

mannycarrera4:
NB: Within our testing is it worth noting the difference between `States` as in hover, active, ect.. and then other stories where States mean visual states where the component changes based on props?

anicholls:
@mannycarrera4 Yeah it's worth considering. I don't think it's been a point of confusion so far though. Not sure what I would want to see instead.

PR #690, opened 2020-06-02
test(avatar): Add snapshot tests for Avatar and AvatarButton

anicholls:
Partially addresses #683 

These tests are based on the ones added to `prerelease/v4` in #676


modules/avatar/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,93 @@
...
+  ComponentStatesTable,
+  permutateProps,
+  withSnapshotsEnabled,
+} from '../../../../utils/storybook';
+import {Avatar, AvatarButton} from '../index';
+
+// @ts-ignore: Cannot find module error
+import testAvatar from './test-avatar.png';
+// eslint-disable-next-line no-empty-function
+const noop = () => {};
```
alexandrzavalii, line 15: this could be somewhere in test utils folder?

modules/avatar/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+export default withSnapshotsEnabled({
+  title: 'Testing|React/Indicators/Avatar',
+  component: Avatar,
+  parameters: {
+    chromatic: {
+      delay: 300, // Ensure we capture the image after loading and transition
```
mannycarrera4, line 22: NB: I could see this being a utils as well since we have a few components that have transitions

anicholls: Most of the time we turn off transitions w/ `transformOrigin={null}` rather than using a delay though. This is the only place we use a chromatic delay atm so going to leave as is for now

PR #693, opened 2020-06-02
test(modal): Add modal snapshots

anicholls:
Partially addresses #683


modules/modal/react/stories/stories_VisualTesting.tsx
```diff
@@ -0,0 +1,61 @@
...
+      })}
+      columnProps={[{label: '', props: {className: ''}}]}
+    >
+      {({heading, ...props}) => (
+        <Modal
+          {...props}
+          heading={heading}
+          open={true}
+          handleClose={noop}
+          style={{position: 'relative', width: 'auto', height: 'auto', padding: 32}}
```
anicholls, line 50: Note: This isn't ideal - I have to apply these styles to get the overlays to show up nicely all on one page. Open to other ideas, but this seemed like the only way without creating a story for each.

alexandrzavalii: Looks like default for `width` and `height` is `auto`

modules/modal/react/stories/stories_VisualTesting.tsx
```diff
@@ -0,0 +1,61 @@
...
+        >
+          <p>Are you sure you'd like to delete the item titled 'My Item'?</p>
+          <Button style={{marginRight: '16px'}} variant={Button.Variant.Delete}>
```
alexandrzavalii, line 53: Maybe for the sake of consistency just always pass a number for default units?  ```marginRight: 16 ```

modules/modal/react/stories/stories_VisualTesting.tsx
anicholls: This is getting applied to the wrapper overlay, not the modal itself. The defaults for these are `100vw/h`

NicholasBoll:
For whatever reason, the Chrome and Firefox snapshots are appearing empty: https://www.chromatic.com/snapshot?appId=5d854c26ba934e0020f5e98a&id=5ed7d0ea8e6105002284bdd4

I think that's because of the animation. On IE it is showing up.

Issue #694, opened 2020-06-03
Modal is not sending focus back to the target button, in responsive


alexandrzavalii:
## 🐛 Bug Report

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/10633451/83637885-3291fc80-a5a0-11ea-8075-1687f180c820.gif)

## To Reproduce

 Make sure you use ios device with voiceover.
 Use this codesandbox link to reproduce: https://hqlce.csb.app/

## Expected Behavior

Focus should be sent back to the target, assumingly because missing the 300ms timeout on focus

NicholasBoll:
#618 might have some related implementation

NicholasBoll:
This is fixed in v5

PR #695, opened 2020-06-03
fix(modal): Fix focus by adding timeout

alexandrzavalii:
## Summary

fix: #694

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/modal/react/lib/ModalContent.tsx
+++ b/modules/modal/react/lib/ModalContent.tsx
@@ -139,7 +139,7 @@ const useInitialFocus = (
   modalRef: React.RefObject<HTMLElement>,
   firstFocusRef: React.RefObject<HTMLElement> | undefined
 ) => {
-  const handlerRef = document.activeElement instanceof HTMLElement ? document.activeElement : null;
+  const handlerElem = document.activeElement instanceof HTMLElement ? document.activeElement : null;
 
   React.useLayoutEffect(() => {
     if (modalRef.current) {
@@ -148,8 +148,10 @@ const useInitialFocus = (
       elem.focus();
     }
     return () => {
-      if (handlerRef) {
-        handlerRef.focus();
+      if (handlerElem) {
+        setTimeout(() => {
+          handlerElem.focus();
+        }, 300);
       }
     };
   }, [modalRef, firstFocusRef]);
@@ -192,8 +194,9 @@ const ModalContent = ({
 
   React.useEffect(() => {
     const siblings = [...((container.children as any) as HTMLElement[])].filter(
-      el => el !== modalRef.current!.parentElement
+      el => el.tagName.toLowerCase() !== 'script' && el !== modalRef.current!.parentElement
     );
+
     const prevAriaHidden = siblings.map(el => el.getAttribute('aria-hidden'));
     siblings.forEach(el => {
       el.setAttribute('aria-hidden', 'true');


cypress/integration/Modal.spec.ts
```diff
@@ -103,6 +105,10 @@ describe('Modal', () => {
           });
 
           it('should trap focus inside the modal element', () => {
+            h.modal
+              .get()
+              .pipe(h.modal.getCloseButton)
+              .should('have.focus');
```
NicholasBoll, line 111: This is tested in the previous test, why is it tested again?

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -135,23 +135,27 @@ const useKeyDownListener = (handleKeydown: EventListenerOrEventListenerObject) =
   }, [handleKeydown]);
 };
 
+const focusWithTimeout = (el: HTMLElement | null) => {
+  setTimeout(() => {
+    if (el) {
+      el.focus();
+    }
+  }, 300);
+};
```
NicholasBoll, line 144: Why is a timeout needed? Why 300ms instead of any other number? Do we even need `useLayoutEffect` if we're going to delay execution of the `focus` call?

There is nothing to cancel this timeout when the component is unmounted. It is possible for the component to unmount before this timeout callback is called. In reality, this shouldn't be an issue because the browser will silently ignore a call to focus on a detached element (this will maintain a reference to the element even after unmounting). But all timers should be cleaned up.

NicholasBoll: After looking into the issue more, I think the 300ms magic number is from the delay incurred by touch devices that have double-tap-to-zoom enabled to determine if a tap is intended as a click or intended as part of a double-tap.

There are a number of ways to deal with this. Are we sure a 300ms delay on focus switching is the best one? Seems like a subpar experience for mobile users - we're fixing the focus issue by introducing a delay to compensate for the delay rather than removing the delay.

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -192,8 +196,9 @@ const ModalContent = ({
 
   React.useEffect(() => {
     const siblings = [...((container.children as any) as HTMLElement[])].filter(
-      el => el !== modalRef.current!.parentElement
+      el => el.tagName.toLowerCase() !== 'script' && el !== modalRef.current!.parentElement
```
NicholasBoll, line 199: I've noticed script tags are hidden by `aria-hidden`, but what accessibility problems does this cause?

cypress/integration/Modal.spec.ts
alexandrzavalii: because of the timeout, we have to first wait for the focus to be moved to closing button.

modules/modal/react/lib/ModalContent.tsx
alexandrzavalii: I dont think it causes any problems. Its something that I noticed. But having it on the script does not bring any value either?

modules/modal/react/lib/ModalContent.tsx
alexandrzavalii: Thanks for taking a look at it!  This 300ms is fixing the focus bug that Peter found.
Peter Heery:
it's primarily for VoiceOver on both MacOS and iOS - it won't hinder any other screen reader and browser combos  - I think it should be kept for all, it's standard practice

alexandrzavalii: You are right we might not need `useLayoutEffect` in this case.

alexandrzavalii: I did not clear setTimeout, because it will be ignored

mathalete: Browsers on touchscreen devices typically have a 300ms delay between the moment the pointing device leaves the screen, and the mouse compatibility and click events are fired. (Note that, to be precise, on iOS the delay is historically 350ms).

As noted, the delay is in place because browsers need to wait if this was a single tap, or if this was simply the first instance of a double-tap gesture. 

If after 300ms - 350ms delay there was no further tap, it is then assumed that it was meant as an activation, and the mouse and click events are fired.

Just to be clear, the issue here is a different issue - VoiceOver on iOS and MacOS (I don't have an instance of iPadOS to check) historically has quirks when taking stock of changes to the accessibility tree, which in turns causes subpar experience for screen reader users.

For example, to ensure that setting programmatic focus() works consistently it needs to be set after a 300 - 350 millisecond delay. 

Or, if injecting content into a live region generally we'd do so after a delay from the event that triggers that live region.

I wouldn't class these quirks necessarily as bugs, as far as I know there's nothing in the spec about when something should be become available to the accessibility api -  it's just there's a variety of nuances depending what screen reader / browser / OS being used.

In the case of the modal here, you'd set focus() to the modal close <button> after a 300 - 350 millisecond delay, then when closing the modal, when the click event is fired on the close <button>, set focus back to the open <button> after a 300 - 350 millisecond delay.

modules/modal/react/lib/ModalContent.tsx
mathalete: Sometimes in older versions of iOS when using explore by touch with VoiceOver running, scripts included within the <body> tag would incorrectly be announced, so as a belt and braces approach aria-hidden="true" would be added to the <script> element.

This is since been resolved in the later iterations - there's no impact here for a11y.

cypress/integration/Modal.spec.ts
NicholasBoll: Ah. This is a wait.

modules/modal/react/lib/ModalContent.tsx
NicholasBoll: I think we should move this function to the common module and document why it is necessary. It seems we don't always know _when_ it will be necessary, but unloading some of this context into the documentation of the function should help when we run into this in the future.

cypress/integration/Modal.spec.ts
NicholasBoll: I debated adding this explicit wait in the `beforeEach`, but decided against that. We should either add a comment that this group is to wait for focus transfer, or extract to a function called `waitForFocusTransfer`

PR #696, opened 2020-06-03
test(core): Add new stories and enable snapshots

anicholls:
Partially addresses #683.

Adds missing stories for color, border radius and spacing tokens. Enables snapshots for all core react stories

### DIFF
--- a/modules/core/react/stories/stories.tsx
+++ b/modules/core/react/stories/stories.tsx
@@ -92,6 +92,72 @@ export const Type = () => (
   </React.Fragment>
 );
 
+const Shape = styled('div')<{radius?: string | number; size?: string | number}>(
+  {
+    ...type.h4,
+    margin: spacing.m,
+    background: colors.blueberry400,
+    color: colors.frenchVanilla100,
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    flexDirection: 'column',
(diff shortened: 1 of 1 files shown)


modules/core/react/stories/stories.tsx
```diff
@@ -1,41 +1,71 @@
 /// <reference path="../../../../typings.d.ts" />
 /** @jsx jsx */
 import {jsx} from '@emotion/core';
```
mannycarrera4, line 3: Dowe want to put these under `visual-testing/core` ?

modules/core/react/stories/stories.tsx
```diff
@@ -44,62 +74,166 @@ export const type = (hierarchy: any) => (
...
+
+export const Spacing = () => (
+  <React.Fragment>
+    {Object.keys(spacing).map(size => (
+      <div css={{display: 'flex'}}>
+        <SizeLabel>
+          {size}
+          <span>({(spacing as any)[size]})</span>
+        </SizeLabel>
+        <Shape size={(spacing as any)[size]} radius={borderRadius.m} />
```
lychyi, line 155: This seems to make `xxxs` and `xxs` look round and less clear on the sizing. Is there a specific reason that there's a non-zero border radius for this shape?

image

PR #700, opened 2020-06-04
fix(common): Auto-generate contrast color for partial theme

### LINKED ISSUE #697: Theming contrast color is not based off palette main color
## 🐛 Bug Report

When using `createCanvasTheme` to fill out a theme object, we automatically extrapolate all colors from a single `main` color. However, unless `contrast` is specified, we always default to white. We should use the `getForegroundColor` function from `@workday/canvas-kit-react-common` to determine this. 

Line to fix: https://github.com/Workday/canvas-kit/blob/prerelease/v4/modules/common/react/lib/theming/createCanvasTheme.ts#L86
Note this change should happen in the `prerelease/v4` branch.

donovangini:
fix #697

Default a contrast color in `createCanvasTheme` when none is passed in. Updated logic in `pickForegroundColor` to choose the highest contrast between dark and light.

## Checklist

- [x] branch has been rebased on the latest `prerelease/v4` commit
- [x] `yarn test` passes
      applicable

## Additional References

[Candidate Experience Figma](https://www.figma.com/file/2whhdLvk03y91IH9DTi49Z/%F0%9F%8C%8EApply-Flow-(PM%2FDev)?node-id=5734%3A3806)


modules/common/react/lib/utils/colorUtils.ts
```diff
@@ -8,34 +9,20 @@ export const expandHex = (hex: string) => {
...
 ) => {
-  const rgbColor = hexToRgb(background);
-  if (rgbColor) {
-    const r: number = rgbColor.r;
-    const g: number = rgbColor.g;
-    const b: number = rgbColor.b;
-    /** Based on : https://www.w3.org/TR/WCAG20-TECHS/G18.html */
-    return r * 0.299 + g * 0.587 + b * 0.114 > 186 ? darkColor : lightColor;
+  if (chroma.valid(background)) {
+    return chroma.contrast(background, darkColor) >= chroma.contrast(background, lightColor)
```
anicholls, line 23: Hmm - this is interesting. When they are really close sometimes it is choosing the opposite color than what we expect. For example:

image

(see https://github.com/Workday/canvas-kit/blob/prerelease/v4/modules/icon/react/lib/SystemIconCircle.tsx#L72)

White is ~4.5 and black is ~4.6. Oddly enough, the contrast for white seems stronger. I think our old calculation favored white on colors vs. just choosing the highest ratio. Will have to think about how to handle this.

anicholls: Hey @donovangini - I caught up with the design team and we have a path forward. Can we update this to always select `lightColor` if its contrast is >= 4.5, otherwise take the highest of the two?

donovangini: cool, yep! updated and added a couple tests

PR #702, opened 2020-06-05
fix: Fix default prop bug

### LINKED ISSUE #705: Default props refactor bugs
## 🐛 Bug Report

A regression was introduced when we refactored how we set default props in #525. The three affected components are Layout, RadioGroup, and SidePanel. More detailed information can be found in #702

There is also a related issue with the Label component not being passed a `labelPosition` in RadioGroup.

alanbsmith:
## ISSUE

Resolves #705 

## Overview

We discovered an issue with RadioGroup's default props as a result of the refactor in #525. After a review of the components in the PR, we found other components that were also affected:

* Layout
* SidePanel

If you'd like to see a simplified version of this issue, check out [this sandbox](This is a simplified version of the default props issue: https://codesandbox.io/s/default-props-issue-2br8q?file=/src/App.js)

`modules/layout/react/lib/Layout.tsx`

## Testing Manually

* `yarn start`
* view the [Layout stories](http://localhost:9001/?path=/story/components-containers-layout-react--page-layout)
* view the [SidePanel stories](http://localhost:9001/?path=/story/components-containers-side-panel-react--configurable)
* view the [RadioGroup Stories]()

## New Dependencies

n/a

## Screenshots (if applicable)

n/a


modules/layout/react/lib/Layout.tsx
```diff
@@ -73,8 +73,10 @@ export default class Layout extends React.Component<LayoutProps> {
         return child;
       }
 
+      const {spacing = spacingNumbers.xs} = this.props;
```
anicholls, line 76: I'm hesitant to define the default in two places like this. What if we passed the props into this function with something like 
```tsx
{React.Children.map(children, this.renderChild.bind(undefined, this.props))}
```

modules/radio/react/lib/RadioGroup.tsx
```diff
@@ -76,13 +76,11 @@ export default class RadioGroup extends React.Component<RadioGroupProps> {
   }
 
   private renderChild = (child: React.ReactElement<RadioProps>, index: number): React.ReactNode => {
+    const {value = 0, name: parentName} = this.props;
```
anicholls, line 79: Same thing goes here

modules/layout/react/lib/Layout.tsx
alanbsmith: yeah, I don't love the duplication either. Will update 👍

alanbsmith: Turns out I can't use bind here, but I'll just make renderChild a pure function instead. 👍

modules/radio/react/lib/RadioGroup.tsx
alanbsmith: Resolved. 👍

modules/form-field/react/stories/stories_Radio.tsx
```diff
@@ -14,142 +13,233 @@ const hintId = 'error-desc-id';
...
+            <Radio id="1" value="email" label="E-mail" />
+            <Radio id="2" value="phone" label="Phone" />
+            <Radio id="3" value="fax" label="Fax (disabled)" disabled={true} />
+            <Radio id="4" value="mail" label="Mail" />
+          </RadioGroup>
+        </FormField>
+      );
+    },
+    {
+      chromatic: {
```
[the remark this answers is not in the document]
anicholls: Also, since we're trying to limit the number of testing stories, I think we need to be more intentional about how we test this (rather than just turning chromatic on for each existing story). We will likely need to consider other input components inside of FormField as well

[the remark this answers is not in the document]
alanbsmith: Yep. Good call. Will update.

alanbsmith:
@NicholasBoll The tests existed, but were flawed. So I updated them here: modules/form-field/react/stories/visual-testing/stories_Radio.tsx

NicholasBoll:
The visual diffs were a combination font-rendering issues and @anicholls recent changes to theming... I wish we could change how Chromatic chooses baselines.

PR #715, opened 2020-06-08
test(side-panel): Add stories and enable snapshots

mannycarrera4:
## Summary

Partially fixes: #683
Add snapshots for side panel component


mannycarrera4:
This doesn't test the breakpoints, any ideas how to best test that in chromatic, should we?

NicholasBoll:
Did you want to target `master` with this?

Issue #698, opened 2020-06-04
Components using white against main theme color should be using contrast color


anicholls:
## 🐛 Bug Report

Several of our themed components use a hardcoded white color against configurable theme colors. If these theme colors are configured to be lighter than expected, this will fail accessibility. We should either be using the theme `contrast` color, or our `pickForegroundColor` utility.

Examples:
- Checkbox
- Radio
- Switch
- ???

anicholls:
Closed by #719

PR #719, opened 2020-06-09
fix: Use theme contrast color for input "checks"

anicholls:
Several of our themed components (`Checkbox`, `Radio`, and `Select`) use a hardcoded white color against configurable theme colors for their "check". If the background theme color is configured to be lighter than expected, this will fail accessibility. This PR updates these components to pull the contrast color from `theme.canvas.palette.primary.contrast`.

Note: I updated the components to be functional to make the theming API stuff a little cleaner.

Closes https://github.com/Workday/canvas-kit/issues/698


modules/common/react/lib/utils/uniqueId.ts
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
+import uuid from 'uuid/v4';
+
+export const uniqueId = () => {
+  // https://codesandbox.io/s/react-functional-component-ids-p2ndq
+  const [id] = React.useState(() => uuid().replace(/^[0-9]*/gi, ''));
```
anicholls, line 6: Note: this strips leading digits from the id so we always produce valid IDs

Issue #134, opened 2019-08-19
Fix delayed initial skeleton animation


lychyi:
## 🐛 Bug Report

Perception-wise, the diagonal sheen of the Skeleton loader does not fire immediately when the component renders. Given the 5 second duration of the animation, it has a considerable delay before you see it. If the placeholder is meant to substitute content for sub-second load times, most users will never see the sheen.

## To Reproduce

https://workday.github.io/canvas-kit/?path=/story/skeleton--complete

## Expected Behavior

The sheen should animate as soon as the component renders/appears.

## Actual Results

The sheen takes a bit before it animates. Long enough so that content may already replace what's behind it before you see it.

lychyi:
From @anicholls:

The animation is also size dependent. We should be calculating a pixel per second speed/duration so the animation is the same at all screen sizes

willklein:
This will be addressed with a different animation in #1001.

PR #720, opened 2020-06-09
test(skeleton): Add stories and enabled snapshots

### LINKED ISSUE #717: Improve visual snapshot coverage for skeleton loader
Improve coverage of snapshot tests for skeleton loader module..

See #683

mannycarrera4:
## Summary

Fixes #717 
Partially fixes: #683
Add snapshots for skeleton component


mannycarrera4:
Thoughts on the way it's being broken up?

modules/skeleton/react/stories/visual-testing/stories_Skeleton.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export default withSnapshotsEnabled({
+  title: 'Testing|React/Indicators/Skeleton',
+  component: Skeleton,
+});
+
+export const SkeletonStates = () => (
+  <div>
+    <h3>With Custom Line Count</h3>
+    <StaticStates>
```
NicholasBoll, line 15: `StaticState` isn't used anywhere. This only does something if `className` is used as a prop and only for pseudo states like `.focus` or `.hover`.

I agree with @lychyi where the table isn't necessary to show all component states - especially in this example where there isn't any matrix. But it does show nice text descriptions of what we're looking at. I think visual tests, like our functional tests, should have context easily available. The way the text is laid out helps describe what's in the visual snapshot.

The confusing part here is which skeleton types are being used.

First table is `SkeletonText`, but the label says "With Custom Line Count" even though the "Default" has no custom line count. Your variant would be your line count of 4.

Second table is `SkeletonShape`. with a few different widths, heights, and border radii.

Third table has `SkeletonShape`, `SkeletonHeader`, and `SkeletonText`. I can see by the story that `SkeletonHeader` is not obvious and makes me wonder if there are different header types (heights). I'd like to see side-by-side comparisons of skeletons to actual content. A common thing I see people get wrong with skeleton loading is that the "skeleton" isn't actually a skeleton of the layout - so layout shifts. Sometimes dramatically which defeats the entire purpose of skeleton loaders. We should have stories that show what they look like side-by-side so we can see that they line up correctly. We should have some stories that show loading at different loading times (best practices of _when_ to show them) with the ability to reset.

Skeleton loaders are abused so many times and we don't help the situation since we don't provide any good examples.

modules/skeleton/react/lib/skeleton.tsx
```diff
@@ -13,7 +13,7 @@ export interface SkeletonProps {
 
 const TRANSPARENCY_POSITION = 45;
 const WHITE_SHEEN_WIDTH = 10;
-const DURATION = 5;
+const DURATION = 2;
```
NicholasBoll, line 16: Did you meant to change this value?

mannycarrera4: Yes, i ran it by the design team in PT, but Alex brought up a point that it might behave weird on larger screens?

NicholasBoll: I think changing that value should be from a GH issue explaining why it needs to change. Otherwise that context gets lost. There are so many things that exist without any context.

mannycarrera4: that's understandable, I'll revert this change until we have some more context

NicholasBoll: I found this: https://github.com/Workday/canvas-kit/issues/134

I still think changing this value should be a different PR addressing this issue as well as a rationale for what the final number becomes.

PR #726, opened 2020-06-10
fix(button): Misc. styling fixes and update to focusRing API

anicholls:
## Summary

- Updates background colors of `ToolbarIconButton` and `ToolbarDropdownButton` to use the same backgrounds as icon buttons per the spec.
- Updates `focusRing` API to add more control over how the shadows are inset
- Shifts inverse `TextButton` variant focus outline 2px inwards (using new `focusRing` API) per the spec

### Breaking Changes
- `focusRing`'s `inset` option has changed slightly. It used to be a boolean type. Now it is typed as `undefined | 'inner' | 'outer'`. This allows consumers to specify whether both shadows are on the outside (`undefined`), just the inner shadow is inset (`'inner'`), or both shadows are inset (`'outer'`). This was option was unused throughout most of our codebase so it shouldn't have much impact.

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/common/react/spec/focusRing.spec.ts
+++ b/modules/common/react/spec/focusRing.spec.ts
@@ -6,10 +6,10 @@ describe('memoizedFocusRing', () => {
       ringWidth: 3,
       separationWidth: 1,
       animate: true,
-      inset: 'outer' as const,
+      inset: 'outer',
       innerShadowColor: '#333333',
       outerShadowColor: '#666666',
-    };
+    } as const;
     const received0 = memoizedFocusRing(args);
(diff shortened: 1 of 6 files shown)


modules/common/react/spec/focusRing.spec.ts
```diff
@@ -6,7 +6,7 @@ describe('memoizedFocusRing', () => {
       ringWidth: 3,
       separationWidth: 1,
       animate: true,
-      inset: true,
+      inset: 'outer' as const,
```
NicholasBoll, line 9: Could put `as const` on the whole object.

Issue #710, opened 2020-06-08
Improve visual testing coverage for IconButtonToggleGroup


jpante:
Improve coverage of snapshot tests for IconButtonToggleGroup module..

See #683

anicholls:
Since this component/story was renamed, and there are differences between the story layouts between `master` and `prerelease/v4`, automated regression will not work. Also, there were several UI bug fixes/improvements made during the move. For these reasons, testing will have to be manual. 

I did a manual review and things look good. However, I found a small bug in the story logic. Going to close this issue as there are no visual regressions, but will open a PR to fix the bug and enable snapshots on the new SegmentedControl component.

PR #730, opened 2020-06-11
fix(segmented-control): Misc. fixes and story improvements

anicholls:
## Summary

In #710 I found a few issues that I wanted to resolve here:
- I thought the selection of the first button wasn't working, but it appears the storybook `action` call is introducing a delay. It still works, but it just takes 1-2 seconds to become toggled. Not sure this is worth dealing with so I just added a comment
- There was a bug with theming
- There were a few labs references left over (Readme and story name)

Also:
- Switched story to CSF
- Added a visual testing story

### DIFF
--- a/modules/segmented-control/react/stories/stories.tsx
+++ b/modules/segmented-control/react/stories/stories.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import {storiesOf} from '@storybook/react';
 import {action} from '@storybook/addon-actions';
 import withReadme from 'storybook-readme/with-readme';
+import {withSnapshotsEnabled} from '../../../../utils/storybook';
 
 import {
   listViewIcon,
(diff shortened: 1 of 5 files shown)


modules/segmented-control/react/stories/stories.tsx
```diff
@@ -16,32 +16,36 @@ import {SegmentedControl} from '../index';
...
+    action('Segmented Control selection change')(selectedValue);
+  };
+
+  return (
+    <SegmentedControl value={value} onChange={handleToggle}>
+      <IconButton
+        icon={listViewIcon}
+        value="list-view"
+        aria-label="List View"
+        // Note: For some reason this action call causes a delay in storybook
```
NicholasBoll, line 38: I'll run some analysis on this

NicholasBoll: Initial investigation shows terrible performance around strigifying a large JSON object. I'll have to pull down locally to get a proper stack trace though

NicholasBoll: OMG. What is Storybook doing?! Clicking is trigging an action which is running through `JSON.stringify` a bunch of times with rather huge objects and then posing a message across the iframe.

Storybook_slowness

NicholasBoll: Screen Shot 2020-06-11 at 12 51 33 PM

Storybook seems wildly inefficient

NicholasBoll: Removing the 2 `action` calls drastically reduces the time spent in Storybook code, but there's still a lot that Storybook is doing, decreasing any performance of our stories. This makes it very difficult to measure meaningful performance of our components.

anicholls: Well that's a bummer. We use `action` in quite a few places. Are you proposing we remove it here? It's bad, but it's not terribly bad 🤷‍♂️

NicholasBoll: I think we'll save that for when we redo documentation.

PR #731, opened 2020-06-11
ci(avatar): Fixed threshold setting

NicholasBoll:
Unfortunately Storybook configuration isn't strongly typed.

### DIFF
--- a/modules/avatar/react/stories/stories_visualTesting.tsx
+++ b/modules/avatar/react/stories/stories_visualTesting.tsx
@@ -19,7 +19,7 @@ export default withSnapshotsEnabled({
   component: Avatar,
   parameters: {
     chromatic: {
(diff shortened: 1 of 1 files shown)


modules/avatar/react/stories/stories_visualTesting.tsx
```diff
@@ -19,7 +19,7 @@ export default withSnapshotsEnabled({
   component: Avatar,
   parameters: {
     chromatic: {
```
anicholls, line 21: We could manually type this [here](https://github.com/Workday/canvas-kit/blob/master/utils/storybook/withSnapshotsEnabled.ts#L6) if we want to avoid issues like this in the future.

NicholasBoll: Yeah. Chances are high that `withSnapshotsEnabled` will have `chromatic` and other stories won't.

PR #692, opened 2020-06-02
test(status-indicator): Add snapshots

anicholls:
Partially addresses #683

PR #732, opened 2020-06-11
test(status-indicator): Fix visual testing story icons

anicholls:
## Summary

I accidentally merged https://github.com/Workday/canvas-kit/pull/692 without addressing my comment about the icons. This should fix the problem

### DIFF
--- a/modules/status-indicator/react/stories/stories_VisualTesting.tsx
+++ b/modules/status-indicator/react/stories/stories_VisualTesting.tsx
@@ -36,7 +36,7 @@ export const StatusIndicatorStates = () => (
         ],
       })}
     >
-      {({type, emphasis, props}) => (
+      {({type, emphasis, ...props}) => (
         <StatusIndicator type={type} emphasis={emphasis} {...props} label="Status" />
       )}
     </ComponentStatesTable>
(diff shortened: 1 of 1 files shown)


modules/status-indicator/react/stories/stories_VisualTesting.tsx
```diff
@@ -36,7 +36,7 @@ export const StatusIndicatorStates = () => (
         ],
       })}
     >
-      {({type, emphasis, props}) => (
+      {({type, emphasis, ...props}) => (
         <StatusIndicator type={type} emphasis={emphasis} {...props} label="Status" />
```
lychyi, line 40:
```suggestion
      {props => (
        <StatusIndicator {...props} label="Status" />
```

This works too right?

anicholls: Yep - I was getting type errors last time from this but I think it's because I forgot the `...`

modules/status-indicator/react/stories/stories_VisualTesting.tsx
```diff
@@ -36,8 +36,8 @@ export const StatusIndicatorStates = () => (
         ],
       })}
     >
-      {({type, emphasis, ...props}) => (
-        <StatusIndicator type={type} emphasis={emphasis} {...props} label="Status" />
+      {props => (
+        <StatusIndicator {...props} label="Status" />
```
lychyi, line 40: I can take care of this and merge

PR #706, opened 2020-06-08
chore: Update canvas-colors-web dependencies

anicholls:
Updates `@workday/canvas-colors-web` dependencies to 2.0.0

Closes #539 

### Breaking Changes:
- `canvas.colors.primary` & `colors.primary` were previously deprecated and are no longer available under this namespace. All of these semantic colors are still accessible via the semantic colors exports (`buttonColors`, `inputColors`, etc.)
- `canvas.colors.gradients` & `colors.gradients` exports have been moved to `canvas.gradients` or `gradients`.
- `canvas.inputColors.warning` & `inputColors.warning` exports have been changed to `*inputColors.alert` to match other conventions
- Narrow incorrect `CanvasColor` type from `string | undefined` to a list of all canvas colors

PR #733, opened 2020-06-11
fix(icon): Fix icon color references

anicholls:
## Summary

I missed a few things in #706. This addresses those issues.

### DIFF
--- a/modules/icon/css/README.md
+++ b/modules/icon/css/README.md
@@ -193,14 +193,14 @@ All system icon colors will accept colors from the
 > Icon color. This will define `data-accent-color` and `data-fill-color`. `data-accent-color` and
 > `data-fill-color` will override this property if defined.
 
-Default: `colors.primary.iconStandard`
+Default: `iconColors.standard`
 
 #### `data-hover-color`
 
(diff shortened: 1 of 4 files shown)


modules/icon/css/README.md
```diff
@@ -253,7 +253,7 @@ Default: `24`
 ## Default Options
 
 ```js
-initializeIcons((iconPath = null), (selector = '.wdc-icon'), iconRoot = document);
+initializeIcons((iconPath = null), (selector = '.wdc-icon'), (iconRoot = document));
```
lychyi, line 256: 😬

anicholls: The linter did this, not me!

lychyi:
@anicholls If this addresses #706, I'm not clear on why this PR is against `master`? Are some things broken today that we need to address before we merge `master` into `prerelease/v4`?

The resulting colors seem to be the same regardless of whether we use `primary.` or `iconColors.` right?

anicholls:
@lychyi Oh wow. Major brain fart. Sorry, I will change the base to prerelease.

> The resulting colors seem to be the same regardless of whether we use `primary.` or `iconColors.` right?
Yes, however `primary` no longer exists in 2.0

PR #741, opened 2020-06-12
fix: Fix issues after canvas-colors-web-upgrade

anicholls:
## Summary

Found some more issues with this during testing. Also removed our `.find()` usage within theming so we don't have to worry about polyfills.


modules/common/react/lib/theming/createCanvasTheme.ts
```diff
@@ -13,17 +13,14 @@ import {
...
   Darken,
   Brighten,
 }
 
 function shiftColor(hexColor: string, direction: ColorDirection) {
-  const canvasColor = Object.keys(allColors).find(
-    key => allColors[key as CanvasColor] === hexColor
-  );
+  const canvasColor =
+    colors[Object.keys(colors).filter(color => colors[color] === hexColor) as CanvasColor];
```
lychyi, line 23: I think this can just use lodash's `findKey`. We use `lodash` in this package anyway

modules/common/react/lib/theming/createCanvasTheme.ts
```diff
@@ -19,8 +20,7 @@ enum ColorDirection {
 }
 
 function shiftColor(hexColor: string, direction: ColorDirection) {
-  const canvasColor =
-    colors[Object.keys(colors).filter(color => colors[color] === hexColor) as CanvasColor];
+  const canvasColor = findKey(colors, hexColor);
```
lychyi, line 23: I don't think the second argument is right here, it is either a function or an iteratee shorthand. In this case you're asking to find the key in which a `hexColor` is `true`.

Can you add a quick test for this? This code is uncovered so that's why this isn't failing on this change. I think testing to see what something like `colors.soap300` would output would be good enough for now.

Ideally I would like to cover the three functions as individual units in this file, but I can open up an issue to take care of that for later.

Screen Shot 2020-06-15 at 9 48 10 AM


```suggestion
  const canvasColor = findKey(colors, color => color === hexColor);
```

anicholls: Hah - you are correct. I thought the tests would have caught it so I assumed this was right. Good call 👍  Will update

modules/core/react/stories/stories.tsx
```diff
@@ -207,7 +207,6 @@ const Swatch = styled('li')<{bg: string; primary?: boolean}>(
 
 const colorNames = Object.keys(colors)
   .map(color => color.replace(/\d+$/, '')) // Remove shade numbers
-  .filter(key => key !== 'primary' && key !== 'gradients')
```
NicholasBoll, line 210: I'm guessing this line is no longer needed?

PR #747, opened 2020-06-16
fix(select): Fix theming colors for proper contrast

jamesfan:
Fixes the Select component to style text colors using `pickForegroundColor` to ensure proper contrast.

### DIFF
--- a/modules/_labs/select/react/lib/SelectOption.tsx
+++ b/modules/_labs/select/react/lib/SelectOption.tsx
@@ -81,7 +81,7 @@ const Option = styled('li')<SelectOptionProps>(
       const selectedStyles = {
         '&[aria-selected="true"]': {
           backgroundColor: selectedBgColor,
-          color: pickForegroundColor(selectedBgColor),
+          color: pickForegroundColor(selectedBgColor, colors.blackPepper300),
         },
       };
       // Only display interactive (hover/active) styles if the option is interactive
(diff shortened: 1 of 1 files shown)


modules/_labs/select/react/lib/SelectOption.tsx
```diff
@@ -68,9 +77,11 @@ const Option = styled('li')<SelectOptionProps>(
       };
     } else {
       // ...else return hover and selected (via aria-selected) styles
+      const selectedBgColor = theme.canvas.palette.primary.lightest;
       const selectedStyles = {
         '&[aria-selected="true"]': {
-          backgroundColor: theme.canvas.palette.primary.lightest,
+          backgroundColor: selectedBgColor,
+          color: pickForegroundColor(selectedBgColor),
```
jamesfan, line 84: @anicholls To Donovan's point in the slack conversation you shared, it does look a little odd for the text color in selected options (`blackPepper600`) to be slightly darker than the text color in non-selected options (`blackPepper300`). Subtle, though -- probably ok for now until we figure out the potential changes to the contrast logic.

Also, should I be using `pickForegroundColor` for the hover options? I'm guessing not since the hover bg color currently isn't themeable, but wanted to double-check.

anicholls: Yes I wouldn't worry about the `blackPepper300` vs. `blackPepper600` here. I'm having a larger conversation about how this works later this week.

jamesfan: Sounds good.

anicholls: Actually, maybe for this case let's use `pickFoergroundColor(selectedBgColor, colors.blackPepper300)`.

I worry less about `blackPepper600` on a random theme color, but would prefer to use `blackPepper300` if it's dark on light with the canvas default theme.

jamesfan: I've updated `pickForegroundColor` to use `colors.blackPepper300` for the selected option.

PR #749, opened 2020-06-17
fix: Address various publish-canary.js bugs

lychyi:
## Summary

`publish-canary.js` had a few errors introduced by #665 

These have been fixed in this PR but I'm really not sure why we haven't run into this within the last few weeks. Still investigating that part.

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/utils/publish-canary.js
+++ b/utils/publish-canary.js
@@ -70,7 +70,7 @@ cmd('git diff --name-only HEAD HEAD^')
     const nextReleaseVersion = semver.inc(releaseTag, 'prerelease', 'beta'); // eg. 4.0.0-beta.3 > 4.0.0-beta.4
     const nextReleasePreid = nextReleaseVersion.split('-')[1];
 
-    if (isPrerelease & !nextReleasePreid) {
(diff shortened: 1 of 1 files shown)


utils/publish-canary.js
```diff
@@ -14,7 +15,9 @@ const {
 
 const isPrerelease = TRAVIS_BRANCH.match(/^prerelease\/v\d*$/g);
 const data = {};
+
 let distTag;
+let preid;
```
lychyi, line 20: Choosing to update `preid` as a side-effect here. If there's any other way we want to pass `preid` around to each promise callback (like promise inside a promise so the nested one has the parent's scope), I'm all ears.

utils/publish-canary.js
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+// @ts-check
```
lychyi, line 2: This helps illuminate the errors in this file. It doesn't seem to block any of our linter checks (which is good), just helpful.

PR #752, opened 2020-06-18
refactor(common): Update pickForegroundColor to use fallback color logic

donovangini:
If no colors provided or neither meets contrast requirements, falls back to the first that meets
contrast requirements of frenchVanilla100, blackPepper300, blackPepper400, blackPepper500,
blackPepper600.

## Summary

Based on our conversation today, implemented the fallback logic into `pickForegroundColor`

## Checklist

- [x] `yarn test` passes
      applicable

## Additional References

Screen Shot 2020-06-18 at 2 33 45 PM
Screen Shot 2020-06-18 at 2 34 08 PM

### DIFF
--- a/modules/common/react/lib/utils/colorUtils.ts
+++ b/modules/common/react/lib/utils/colorUtils.ts
@@ -9,6 +9,15 @@ export const expandHex = (hex: string) => {
   });
 };
 
+// Used to as a fallback to determine foreground color
+const colorPriority = [
+  colors.frenchVanilla100,
+  colors.blackPepper300,
+  colors.blackPepper400,
+  colors.blackPepper500,
+  colors.blackPepper600,
+];
+
 /**
  *
  * Chooses foreground color with accesible contrast against background. If contrast ratio
@@ -28,14 +37,6 @@ export const pickForegroundColor = (
     } else if (darkColor && chroma.contrast(background, darkColor) >= 4.5) {
       return darkColor;
     } else {
-      const colorPriority = [
-        colors.frenchVanilla100,
-        colors.blackPepper300,
-        colors.blackPepper400,
-        colors.blackPepper500,
-        colors.blackPepper600,
-      ];
-
       for (let i = 0; i < colorPriority.length; i++) {
         const color = colorPriority[i];
 
--- a/modules/common/react/lib/utils/colorUtils.ts
+++ b/modules/common/react/lib/utils/colorUtils.ts
@@ -12,22 +12,38 @@ export const expandHex = (hex: string) => {
 /**
  *
  * Chooses foreground color with accesible contrast against background. If contrast ratio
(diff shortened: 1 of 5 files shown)


modules/common/react/lib/utils/colorUtils.ts
```diff
@@ -12,22 +12,38 @@ export const expandHex = (hex: string) => {
...
-      ? darkColor
-      : lightColor;
-  } else {
-    return;
+    if (lightColor && chroma.contrast(background, lightColor) >= 4.5) {
+      return lightColor;
+    } else if (darkColor && chroma.contrast(background, darkColor) >= 4.5) {
+      return darkColor;
+    } else {
+      const colorPriority = [
```
anicholls, line 31: Nit: Can you define this outside the function so it doesn't get redefined every time it's called?

modules/common/react/lib/utils/colorUtils.ts
```diff
@@ -12,22 +12,38 @@ export const expandHex = (hex: string) => {
...
+        colors.blackPepper600,
+      ];
+
+      for (let i = 0; i < colorPriority.length; i++) {
+        const color = colorPriority[i];
+
+        if (chroma.contrast(background, color) >= 4.5) {
+          return color;
+        }
+      }
```
anicholls, line 45: Nit:
```suggestion
      colorPriority.forEach(color => {
        if (chroma.contrast(background, color) >= 4.5) {
          return color;
        }
      });
```

anicholls:
Hmm it looks like it changed the default dark avatar icon back to gray. Can you look into why that's happening?

modules/common/react/lib/utils/colorUtils.ts
donovangini: So the reason I went with the `for` loop instead of a `forEach` is that the `forEach` won't break on return. so it'll just keep evaluating til the last one.

> There is no way to stop or break a forEach() loop other than by throwing an exception. If you need such behavior, the forEach() method is the wrong tool.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

donovangini:
> Hmm it looks like it changed the default dark avatar icon back to gray. Can you look into why that's happening?

Ah looks like https://github.com/Workday/canvas-kit/blob/ce866a1820dbbe4b3a7430202d0861fcb2a978f4/modules/icon/react/lib/SystemIconCircle.tsx#L72 is the culprit. Passes a dark color that meets contrast acceptance, but no light color to override it. should i add white here to this line in the lightColor arg?

modules/common/react/lib/utils/colorUtils.ts
anicholls: TIL. I always thought it just broke early behind the scenes

PR #765, opened 2020-06-29
feat(popup): Add getAnchorClientRect prop for custom positioning

### LINKED ISSUE #762: Add Virtual Elements to Popper component
## 🚀 Feature Proposal

Add the ability to provide custom positioning for the Popper component.
 
## Motivation

Currently the Popper component only allows positioning elements by means of an `anchorElement`. This is great for positioning things relative to other elements, but sometimes I would like to tell Popper exactly where to place things. For example when implementing a context menu, I want to pass in the exact `x` and `y` coordinates of the mouse for placement.

## Example

The way I would imagine this working is by means of a prop that can act in place of `getBoundingClientRect`. A discriminated union would be used to prevent the user from also specifying an `anchorElement` when using this prop.

```tsx
// Correct
const withAnchorElement = <Popper anchorElement={myElement} />

// Correct
const withOverride = (
  <Popper
    overrideGetBoundingClientRect={() => {
      top: 10,
      left: 10,
      bottom: 100,
      right: 100,
      width: 12,
      height: 12,
    }}
  />
)

// Type error
const withBoth = (
  <Popper
    anchorElement={myElement}
    overrideGetBoundingClientRect={() => {
      top: 10,
      left: 10,
      bottom: 100,
      right: 100,
      width: 12,
      height: 12,
    }}
  />
)
```

Under the hood, Canvas would then take this function and create an object to pass to the underlying Popper instance to use as an Element. The `PopupStack` would not receive an owner element.

## Inspiration

This proposal was largely inspired by `react-popper`: (https://popper.js.org/react-popper/v2/virtual-elements/). However, unlike `react-popper` I am proposing that we don't ask the consumer to pass a "Virtual Element". Attempting to do this in the current version of Canvas works, but it's not correct. What is being passed in is not an element and it creates problems for the `PopupStack`. I think this should be communicated to the consumer and I think it feels kind of hacky to pass in an object with just `getBoundingClientRect` declared. Better to hide this in the component 😉

luislikescoffee:
## Summary

This closes #762 
It adds the ability to provide a custom DOMRect to the underlying Popper instance in the Popper wrapper in order to position things wherever we'd like on the screen.

Added visual test:
Screen Shot 2020-07-01 at 10 14 19 AM

## Checklist

- [x] branch has been rebased on the latest master commit (I need help with this as I am new to a forking workflow)
- [x] `yarn test` passes


mannycarrera4:
@luislikescoffee  a couple things:

Looks like the snapshot is getting cut off:
Screen Shot 2020-06-30 at 4 37 12 PM

You can get around this by adding some margin to the element so that the snapshot can capture what we see in storybook.

Also, I know that you're positioning the popup to some coordinates, but to have a clearer snapshot, it would be a nice to have some element or red circle indicating that you're positioning the popup from this point. It's more of a visual indicator in the snapshot than an implementation details.

NicholasBoll:
The snapshots can be seen from the "UI Tests" GH check on each commit. The latest commit points to this: https://www.chromatic.com/snapshot?appId=5d854c26ba934e0020f5e98a&id=5efbc537b5f9920022398fef

From here, Firefox looks off. I see a small red rectangle. Is that supposed to indicate where the Popper is virtually positioned to?

NicholasBoll:
Story now looks like: 
Screen Shot 2020-07-01 at 10 14 19 AM

PR #767, opened 2020-06-30
fix: Add explicit arg to initEvent in ComboBox and header SearchBar

lychyi:
## Summary

We need to add an explicit third arg for `initEvent` otherwise it bombs in IE11.

Screen Shot 2020-06-29 at 5 18 58 PM

This also fixes various layout issues for `Combobox` and `SearchBar` (not officially supported as a standalone component yet).

## Checklist

- [ ] `yarn test` passes
      applicable


modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -100,9 +100,12 @@ const AutocompleteList = styled('ul')({
 const ResetButton = styled(IconButton)<{shouldShow: boolean}>(
   {
     width: spacing.l,
+    minWidth: spacing.l,
     height: spacing.l,
     position: 'absolute',
-    margin: 0,
+    margin: `auto ${spacing.xxxs}`,
+    top: 0,
+    bottom: 0,
```
lychyi, line 108: Previously, with `top` and `bottom` not set, things just magically centered vertically in non-IE browsers. This is an explicit way to make sure we're centering the button inside the height of the `Combobox` input.

modules/_labs/combobox/react/lib/Combobox.tsx
```diff
@@ -195,7 +198,7 @@ const Combobox = ({
       } else {
         // IE 11
         event = document.createEvent('Event');
-        event.initEvent('input', true);
+        event.initEvent('input', true, true);
```
lychyi, line 201: This is where we bomb out if we don't set the third arg explicitly. 💀

modules/_labs/combobox/react/stories/stories.tsx
```diff
@@ -28,7 +27,7 @@ class Autocomplete extends React.Component<
 
   render() {
     const autocompleteResult = (textModifier: string) => (
-      <MenuItem onClick={action(`Clicked Result ${textModifier}`)}>
+      <MenuItem onClick={() => console.log(`Clicked Result ${textModifier}`)}>
```
lychyi, line 30: The `addon-action` import will essentially crash IE11 with a script lockup. I'm removing it here and replacing it with a `console.log`. It looks like the only reason this exists is to log out what happens so this should be the same functionality.

modules/_labs/header/react/lib/parts/SearchBar.tsx
```diff
@@ -80,6 +80,11 @@ export interface SearchBarProps extends GrowthBehavior, React.FormHTMLAttributes
    * @default true
    */
   showClearButton?: boolean;
+  /**
+   * Height of the Search Bar in pixels
+   * @default: 40
+   */
+  height?: number;
```
lychyi, line 87: The `SearchBar` height does need to be configurable depending on different contexts (Full Headers vs. Dub Headers for example). Generally speaking, Canvas will have an opinion on form inputs' dimensions. This is to make sure they work well together. However, I don't think this should include special inputs like `SearchBar` since they exist in the context of something else (in this case, a header).

Since `SearchBar` not officially supported as a standalone component, this API change will also probably remain under the radar and should be changeable (if we decide to remove it, fix it to a set of values, etc.)

modules/_labs/header/react/lib/parts/SearchBar.tsx
```diff
@@ -145,35 +149,49 @@ const SearchForm = styled('form')<
...
+    const collapsedSize = 40;
+    const size = 32;
     const collapseStyles: CSSObject = isCollapsed
       ? {
-          width: spacing.xl,
-          height: spacing.xl,
+          minWidth: collapsedSize,
+          width: collapsedSize,
+          minHeight: collapsedSize,
+          height: collapsedSize,
```
lychyi, line 179: I'm avoiding using `spacing` values for height and width measurements. Just because `spacing.xl` = 40 pixels doesn't mean that for every CSS value that uses 40px we need to use `spacing.xl`

modules/_labs/header/react/lib/parts/SearchBar.tsx
```diff
@@ -195,61 +213,63 @@ const CloseButton = styled(IconButton)<
 
   return {
     position: `absolute`,
-    top: spacing.zero,
-    right: spacing.zero,
-    bottom: spacing.zero,
+    top: 0,
+    bottom: 0,
+    right: 0,
```
lychyi, line 218: At this point, I think `spacing.zero`, `spacingNumbers.zero`, and `0` are all interchangeable, I was trying to figure out a rationale for all this but ended up making it more inconsistent. I'm okay leaving unless someone can tell me the official reason why we have the `zero`.

Even if we were to change our spacing system, I would imagine that `zero` always refers to `0`?

modules/_labs/header/react/lib/parts/SearchBar.tsx
```diff
@@ -195,61 +213,63 @@ const CloseButton = styled(IconButton)<
...
+  Pick<SearchBarProps, 'isCollapsed' | 'grow' | 'height'> & {
+    inputColors: ReturnType<typeof getInputColors>;
+  }
+>(({isCollapsed, inputColors, grow, height}) => {
   const collapseStyles: CSSObject = isCollapsed
     ? {
-        ...type.h3,
-        fontWeight: 400,
+        fontSize: '20px',
+        lineHeight: '20px', // For ie11, line-height needs to match font-size
```
lychyi, line 248: We were using `type.h3` _just_ to get the font-size. Anything else that `type.h3` added we just overwrote below. Probably not a good idea. I'm explicitly spelling out the size we want here.

I could have done `type.h3.fontSize` I suppose but it's a sure way to get a JavaScript error if that ever changes?

modules/_labs/header/react/stories/stories.tsx
```diff
@@ -108,16 +108,15 @@ class SearchWithAutoComplete extends React.Component<
...
         autocompleteItems={Array.apply(null, Array(this.state.currentText.length))
           .map((x: any, i: string) => autocompleteResult(i))
           .splice(0, 5)}
         isCollapsed={boolean('isCollapsed', false)}
         onInputChange={this.autocompleteCallback}
         placeholder={`Search with Autocomplete`}
         grow={true}
-        searchTheme={SearchBar.Theme.Dark}
         onSubmit={this.onSubmit}
+        {...this.props}
```
lychyi, line 119: This allows the stories to have props that override the above settings.

modules/_labs/header/react/lib/parts/SearchBar.tsx
anicholls: We will eventually want to migrate _everything_ (including widths and heights) to use `spacing`. We will ultimately move towards `rems` (like we are with typography), and it would be weird if the whole page scaled (typography, spacing, sizes, etc.) except for a few elements.

Not a big deal for this to go in as is though, because we'll have to do a serious audit for this in the future anyway.

modules/_labs/header/react/lib/parts/SearchBar.tsx
anicholls: Fine in this case since it's a one-off component.

modules/_labs/combobox/react/stories/stories.tsx
NicholasBoll: I narrowed down this problem to how Storybook sends events. It uses post message to send messages. Sometimes multiple megabytes with each thing that happens inside Storybook (`action` is part of that). What is slowing things down is the `JSON.stringify` and another stringify function that is replacing strings. It can take even a 1000ms in Chrome occasionally.

jamesfan:
Did a spot check in IE, everything looks good 👍.

I noticed the position of the `searchIcon` (magnifying glass) changed in all browsers, but maybe that's ok?

Before:
image

After:
image

PR #772, opened 2020-07-07
test(text-area): Added placeholder with value unit test

amritbhullar:
## Summary

To improve code coverage of the text area component, we should have a functional unit test that ensures the text area value is persisted even when there is a placeholder. 

## Checklist

- [x] `yarn test` passes


NicholasBoll:
This doesn't add to any code coverage values, but I don't see the harm in adding to our test-area coverage.

PR #775, opened 2020-07-08
Merge/support/v3.x into master

NicholasBoll:
Do not squash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# 3.9.1 (2020-07-07)
```
anicholls, line 6: Shouldn't this be 4.0.1?

anicholls: Or maybe not... Should it go below 4.0? I almost feel like we need a different changelog for major versions if we're supporting both...

PR #776, opened 2020-07-08
feat(breadcrumbs): Add Breadcrumbs component

### LINKED ISSUE #612: Add Breadcrumbs Component
## 🚀 Feature Proposal

Add the breadcrumb component to Canvas Kit React and Canvas Kit CSS. 

## Example

Screen Shot 2020-05-01 at 2 55 33 PM

Designer - Mele H.

alanbsmith:
## Issue

Fixes #612

## Overview

This PR adds the Breadcrumb nav component to labs.

_e.g. `/src/components/SomeComponent.js`_

## Testing Manually

Run `yarn cypress:open` and select `Breadcrumbs.spec.ts` from the list of Cypress tests.

## New Dependencies

n/a

## Screenshots (if applicable)

breadcrumbs demo gif

## Thank You Gif
_Share a fun [gif](https://giphy.com) to say thanks to your reviewer:_

goose eating breadcrumbs

## Checklist

- [ ] `yarn test` passes
      applicable


lychyi (commented):
Initial comments, my review's still WIP. Nice work though, it's a really clean looking PR.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/List/Link.tsx
```diff
@@ -0,0 +1,75 @@
...
+    event.preventDefault();
+    // allow an override to the hard redirect
+    if (onAction) {
+      onAction(href);
+    } else {
+      // default to hard redirecting
+      window.location.href = href;
+    }
+    // don't block the onClick event if it's provided
+    onClick(event);
```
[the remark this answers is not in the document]
alanbsmith: No real preference. I could do `if (onClick)`. 👍

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Button.tsx
```diff
@@ -0,0 +1,42 @@
...
+      variant={variant}
+      icon={buttonIcon}
+      color={colors.licorice200}
+      buttonRef={buttonRef}
+      toggled={toggled}
+      aria-pressed={undefined}
+      aria-expanded={toggled}
+      aria-haspopup
+      aria-controls="menu"
+      {...props}
```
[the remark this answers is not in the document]
alanbsmith: Yeah, I think `otherProps` would work here as `elemProps` usually means native props.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Menu.tsx
```diff
@@ -0,0 +1,116 @@
...
+          return resetFocus();
+        }
+        return (window.location.href = item.link);
+      default:
+        break;
+    }
+  };
+
+  return (
+    <MenuWrapper>
```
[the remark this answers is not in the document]
alanbsmith: It doesn't need to be named. I think this is left over from when I thought these would be exported primitives, but they aren't now. 

Just a side-note: The `Menu` in labs uses `Card` as its wrapper.

modules/_labs/breadcrumbs/react/README.md
```diff
@@ -0,0 +1,271 @@
...
+        <Breadcrumbs.ListItem>
+          <Breadcrumbs.Link href="/invoices">Invoices</Breadcrumbs.Link>
+        </Breadcrumbs.ListItem>
+        <Breadcrumbs.CurrentItem>
+          2020_08_01_invoice.pdf
+        </Breadcrumbs.CurrentItem>
+      </BreadCrumbs.List>
+    </Breadcrumbs.Nav>
+  );
+}
```
jamesfan, line 56: What was the reasoning behind having `Breadcrumbs.Link` in addition to `Breadcrumbs.ListItem` (as opposed to just `ListItem`)? Is there a situation (aside from `Breadcrumbs.CurrentItem`) where you would have an item with no `Link`?

Speaking of `CurrentItem`, I was curious about the need for a distinct `CurrentItem` component. Wouldn't the last `ListItem` always be the current item?

Or, are these both cases of us shifting towards more composable components (more verbose markup, but more flexibility as you mention in the README)? Am I meant to be able to have a `ListItem` with no `Link`, or a `List` with no `CurrentItem`?

jamesfan (commented):
Nice work. 👏  Still making my way through this as well, but I had some preliminary questions.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/List/Link.tsx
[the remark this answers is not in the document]
NicholasBoll: We have some other areas that are a bit cleaner now with the existential operator:

```tsx
onClick?.(event)
```

It looks a bit funky at first, but has pretty clear intent with minimal syntax.

The transpiled output is a bit verbose though:

```js
onClick === null || onClick === void 0 ? void 0 : onClick(event);
```

Honestly not sure why the output doesn't use double equals against `null` to catch `undefined` implicitly... Perhaps a performance decision.

modules/_labs/breadcrumbs/react/README.md
alanbsmith: > Q: Why `ListItem` and `Link`?

Short answer: Composability
Long answer: I broke up the items semantically instead of one large component because they're doing different things. ListItem is mostly a styled `li`. It also adds a `>` separator after children are rendered and handles the truncation + tooltip functionality. The composable pattern also allows users to add anchor tag props directly to the `Link` instead of passing them through to the `ListItem`.

> Q: Why `CurrentItem` and not just another `ListItem`. You're correct that `CurrentItem` is mostly styled `li`, but it has some special behavior. It will never contain an `a`, so it renders a span, and there's no need for a `>` separator. It also needs to be smart about setting `tabindex` to `0` when the item is truncated. The other `ListItem`s don't have to think about that because their `Link` children act as the tab stops. I could have made the `Menu` smarter, mapped over the children, cloned the element + the needed props, but this felt simpler.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Button.tsx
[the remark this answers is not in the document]
NicholasBoll: Most of the extra props are going to be applied to the `button` element. In the case of extra props, a consistent name does decrease cognitive overhead. I have no problem with `otherProps`, but I think we should use the same name everywhere. We've even called them "elem props" in conversations.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Menu.tsx
```diff
@@ -0,0 +1,116 @@
...
+  toggleActiveItemDown,
+  toggleActiveItemUp,
+  ...props
+}: DropdownMenuProps) => {
+  const handleItemKeyUp = (e: React.KeyboardEvent<HTMLLIElement>) => {
+    switch (e.key) {
+      case 'ArrowUp':
+      case 'ArrowLeft':
+      case 'Up': // IE/Edge specific value
+        return toggleActiveItemUp();
```
NicholasBoll, line 73: Do you mean to include `'Left'` as well?
```suggestion
      case 'ArrowUp':
      case 'ArrowLeft':
      case 'Up': // IE/Edge specific value
      case 'Left': // IE/Edge specific value
        return toggleActiveItemUp();
```

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Menu.tsx
```diff
@@ -0,0 +1,116 @@
...
+      case 'ArrowDown':
+      case 'Down': // IE/Edge specific value
+      case 'ArrowRight':
+        return toggleActiveItemDown();
```
NicholasBoll, line 77: Do you mean to include `'Right'` as well?
```suggestion
      case 'ArrowDown':
      case 'ArrowRight':
      case 'Down': // IE/Edge specific value
      case 'Right': // IE/Edge specific value
        return toggleActiveItemDown();
```

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Menu.tsx
```diff
@@ -0,0 +1,116 @@
...
+      default:
+        break;
+    }
+  };
+
+  const handleItemKeyDown = (e: React.KeyboardEvent<HTMLLIElement>, item: Breadcrumb) => {
```
NicholasBoll, line 83: A small nit, but I prefer `event` vs `e`.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Menu.tsx
```diff
@@ -0,0 +1,116 @@
...
+      case 'Tab':
+        e.preventDefault();
+        break;
+      case 'Enter':
+        e.preventDefault();
+        if (item.onAction) {
+          item.onAction(item);
+          return resetFocus();
+        }
+        return (window.location.href = item.link);
```
NicholasBoll, line 94: I'm guessing this is a shorthand? It doesn't matter in this specific case, but the return of this function is `undefined | string`. I thought that was an odd signature, but then the function has no explicit return signature.

The function is isolated, so there's currently no side effect of this case, but I think we should have a `break` or an explicit `return;` instead. Same with the `return resetFocus()`. It makes it seem the return value is important.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Menu.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export const DropdownMenu = ({
+  activeDropdownItemRef,
+  activeDropdownItem,
+  dropdownItems = [],
+  resetFocus,
+  setActiveDropdownItem,
+  toggleActiveItemDown,
+  toggleActiveItemUp,
+  ...props
```
NicholasBoll, line 66: The convention in other components is to use `elemProps`. The consistent naming helps to understand the intent of the variable. I'd be fine with `otherProps` if all other components also used that convention.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Menu.tsx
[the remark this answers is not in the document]
NicholasBoll: I'm definitely in that camp where naming elements is confusing. If `Menu` is a component with an explicit interface, I'm okay with the added abstraction layer. If there is no added benefit to the abstraction, it should be removed.

In the future, it is possible to replace most low-level elements with `Box` for the added convenience our tokens which would be an abstraction we could understand well although it would be harder to read at first: `<Box as="ul">`. But in that case I'm not wondering what `Box` is or what element it is representing.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Button.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <IconButton
+      type="button"
+      css={customStyles}
+      variant={variant}
+      icon={buttonIcon}
+      color={colors.licorice200}
+      buttonRef={buttonRef}
+      toggled={toggled}
+      aria-pressed={undefined}
```
lychyi, line 35: Any reason this needs to be `undefined`, I assume it's to override what `IconButton` naturally does when `toggled` is defined but this seems mysterious to me without knowing the context.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/hooks.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (activeDropdownItemRef.current) {
+      return activeDropdownItemRef.current.focus();
+    }
+  });
+};
+
+export const useDropdown = (
+  activeDropdownItemRef: React.RefObject<HTMLAnchorElement>,
+  buttonRef: React.RefObject<HTMLButtonElement>,
+  buttonId: string = ''
```
lychyi, line 22: Nit: I don't think this default parameter is needed here. `undefined` should get you a generated id.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Menu.tsx
```diff
@@ -0,0 +1,116 @@
...
+  return <div css={menuWrapperStyles} {...props} />;
+};
+
+const menuStyles = css({
+  boxSizing: 'border-box',
+  listStyle: 'none',
+  margin: `${spacing.xxs} 0`,
+  padding: 0,
+  width: 280,
+  wordBreak: 'break-word',
```
jamesfan, line 51: Looks like there are some wrapping issues in IE11:
image

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break), `break-word` isn't supported by IE (and is in fact deprecated).

I think I'll need to take a look at the Select menu as well. I don't think I addressed this edge case of long items with no spaces.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/List/Link.tsx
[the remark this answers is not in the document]
alanbsmith: Yeah, the transpiled output is really verbose, not that I about that much.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Button.tsx
alanbsmith: Good question. Yes, that's exactly it. We're using `aria-expanded` here instead of `aria-pressed`, and `undefined` removes the prop. We're doing that in the [Pagination component](https://github.com/Workday/canvas-kit/blob/master/modules/_labs/pagination/react/lib/Pages.tsx#L78) as well. But I agree the intent is unclear. I'll add a comment.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Button.tsx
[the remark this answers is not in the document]
alanbsmith: Agreed here. Opting for `elemProps` for consistency. 👍

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Menu.tsx
alanbsmith: Good catch! Thank you. Looks like I can use [`overflowWrap: 'break-word'](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) for IE.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/Menu.tsx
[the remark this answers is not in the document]
alanbsmith: Agreed that it's opaque. I'll style the `div` directly with `css`.

utils/storybook/ComponentStatesTable.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import styled from '@emotion/styled';
+import {styled} from '@workday/canvas-kit-react-common';
```
alanbsmith, line 2: This doesn't appear to have affected any of our visual tests in Chromatic.

cypress/integration/Breadcrumbs.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+        });
+      // should see tooltip for the last item (current crumb)
+      cy.get('li:last')
+        .focus()
+        .then($currentCrumb => {
+          const text = $currentCrumb.text();
+          cy.findByRole('tooltip').should('contain', text);
+        });
+    });
+  });
```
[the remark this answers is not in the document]
alanbsmith: Will update the test to have one item that truncates and one that doesn't.

modules/_labs/breadcrumbs/react/stories/stories_visualTesting.tsx
```diff
@@ -0,0 +1,212 @@
...
+      </StaticStates>
+    </CanvasProvider>
+  );
+};
+
+const DropdownMenuRenderer = ({items, direction = ContentDirection.LTR}) => {
+  const activeItemRef = React.useRef();
+  const buttonRef = React.useRef();
+  const {dropdownMenuProps} = useDropdown(activeItemRef, buttonRef, items);
+  console.log(dropdownMenuProps);
```
NicholasBoll, line 173:
```suggestion
```

NicholasBoll (approved):

PR #782, opened 2020-07-16
feat(popup-stack): Add adapter API to integrate with other popup systems

### LINKED ISSUE #750: Create a wrapper to integrate with Princess popups
## 🚀 Feature Proposal

Create a wrapper or adapter API for the stack manager to use when integrating into existing popup stack systems.

## Motivation

Ideally popup stack systems would be built on top of CK's system, but that might require a lot of QA effort for existing systems. A workaround would be to have a wrapper API that allows the Popup stack to use existing systems.

## Example

```ts
interface Adapter {
  // should put an item on your existing stack
  add(item: PopupStackItem): void

  // should remove an item on your existing stack
  remove(element: HTMLElement): void

  // The the provided element is the topmost in the stack, return true
  isTopmost(element: HTMLElement): boolean

  // should return an array of elements in the order from bottom to top
  getElements(): HTMLElements[]

  // should rearrange the stack to that the provided element is on top. `owner`
  // was provided to help so that children are also brought to top
  bringToTop(element: HTMLElement): void

  // should return true if a target is found within a stack item referenced by `element`
  contains(element: HTMLElement, target: HTMLElement): boolean
}
```

```tsx
import { createAdapter, Adapter, PopupStackItem } from '@workday/canvas-kit-popup-stack'

// should be called once and before anything is registered with the popup stack
createAdapter(adapter: Adapter);
```

This API will ensure the existing functionality is not effected. I'm not entirely sure if this is the best implementation, but it is a starting point.

NicholasBoll:
Fixes #750

## Checklist

- [x] `yarn test` passes
      applicable

## Commit message
feat(popup-stack): Add adapter API to integrate with other popup systems (#782)

### BREAKING CHANGE
This is a potentially breaking  change if you use `usePopupStack` and `ReactDOM.createPortal(contents, document.body)`:
We added `createContainer` to `PopupStack`. React no longer controls the element that is given to the the `PopupStack`. The `PopupStack` will now create a containing element that your content should render into. If you recognize this pattern, you'll need to render into `stackRef.current` instead of `document.body`.

Before:
```tsx
const ref = React.createRef<HTMLDivElement>(null);

usePopupStack(ref)

React.createPortal(contents, document.body)
```

Now
```tsx
const stackRef = usePopupStack()

React.createPortal(contents, stackRef.current)
```

### DIFF
--- a/cypress/helpers/modal.ts
+++ b/cypress/helpers/modal.ts
@@ -28,5 +28,16 @@ export function getCloseButton($modal: JQuery): JQuery {
  * @param $modal Modal element with [role=dialog]
  */
 export function getOverlay($modal: JQuery): JQuery {
-  return $modal.parent().parent();
+  const bodyEl = Cypress.$('body')[0];
+  return getElement($modal, $el => $el.parent()[0] === bodyEl);
+}
+
+function getElement(element: JQuery, predicate: (element: JQuery) => boolean): JQuery {
+  if (predicate(element)) {
+    return element;
+  } else if (element.parent().length) {
+    return getElement(element.parent(), predicate);
+  } else {
+    return Cypress.$();
+  }
 }
--- a/cypress/integration/Modal.spec.ts
+++ b/cypress/integration/Modal.spec.ts
@@ -110,6 +110,17 @@ describe('Modal', () => {
           });
         });
 
+        context('when clicking inside the modal', () => {
+          beforeEach(() => {
+            // click somewhere on the modal where there shouldn't be a close target
+            cy.findByLabelText('Delete Item').click('top');
+          });
+
+          it('should not close the modal', () => {
+            cy.findByLabelText('Delete Item').should('be.visible');
+          });
+        });
+
         context('when the close button is clicked', () => {
           beforeEach(() => {
             cy.findByLabelText('Delete Item')
@@ -141,7 +152,7 @@ describe('Modal', () => {
         context('when the overlay is clicked', () => {
           beforeEach(() => {
             cy.findByLabelText('Delete Item')
-              .pipe(h.modal.getOverlay)
+              .parent()
               .click('top');
           });
 
@@ -281,7 +292,7 @@ describe('Modal', () => {
       context('when the overlay is clicked', () => {
         beforeEach(() => {
           cy.findByLabelText('Delete Item')
-            .pipe(h.modal.getOverlay)
+            .parent()
             .click('top');
         });
 
@@ -374,7 +385,7 @@ describe('Modal', () => {
       context('when the overlay is clicked', () => {
         beforeEach(() => {
           cy.findByLabelText('Delete Item')
-            .pipe(h.modal.getOverlay)
+            .parent()
             .click('top');
         });
 
@@ -419,7 +430,7 @@ describe('Modal', () => {
       context('when the overlay is clicked', () => {
         beforeEach(() => {
           cy.findByLabelText('Really Delete Item')
-            .pipe(h.modal.getOverlay)
+            .parent()
             .click('top');
         });
 
@@ -452,7 +463,7 @@ describe('Modal', () => {
       context('when the modal overlay is clicked', () => {
         beforeEach(() => {
           cy.findByLabelText('Delete Item')
-            .pipe(h.modal.getOverlay)
+            .parent()
             .click('top');
         });
 
--- a/cypress/integration/Modal.spec.ts
+++ b/cypress/integration/Modal.spec.ts
@@ -149,7 +149,7 @@ describe('Modal', () => {
           });
         });
 
-        context('when the overlay is clicked', () => {
+        context('when clicking outside the modal', () => {
           beforeEach(() => {
             cy.get('body').click('top');
           });
@@ -287,7 +287,7 @@ describe('Modal', () => {
         });
       });
 
-      context('when the overlay is clicked', () => {
+      context('when clicking outside the modal', () => {
         beforeEach(() => {
           cy.get('body').click('top');
         });
@@ -378,7 +378,7 @@ describe('Modal', () => {
         });
       });
 
-      context('when the overlay is clicked', () => {
+      context('when clicking outside the modal', () => {
         beforeEach(() => {
           cy.get('body').click('top');
         });
@@ -421,7 +421,7 @@ describe('Modal', () => {
         });
       });
 
-      context('when the overlay is clicked', () => {
+      context('when clicking outside the modal', () => {
         beforeEach(() => {
           cy.get('body').click('top');
         });
--- a/cypress/integration/Modal.spec.ts
+++ b/cypress/integration/Modal.spec.ts
@@ -151,9 +151,7 @@ describe('Modal', () => {
 
         context('when the overlay is clicked', () => {
           beforeEach(() => {
-            cy.findByLabelText('Delete Item')
-              .parent()
-              .click('top');
+            cy.get('body').click('top');
           });
 
           it('should close the modal', () => {
@@ -291,9 +289,7 @@ describe('Modal', () => {
 
       context('when the overlay is clicked', () => {
         beforeEach(() => {
-          cy.findByLabelText('Delete Item')
-            .parent()
-            .click('top');
+          cy.get('body').click('top');
         });
 
         it('should not close the modal', () => {
@@ -384,9 +380,7 @@ describe('Modal', () => {
 
       context('when the overlay is clicked', () => {
         beforeEach(() => {
-          cy.findByLabelText('Delete Item')
-            .parent()
-            .click('top');
+          cy.get('body').click('top');
         });
 
         it('should not close the modal', () => {
@@ -429,9 +423,7 @@ describe('Modal', () => {
 
       context('when the overlay is clicked', () => {
         beforeEach(() => {
-          cy.findByLabelText('Really Delete Item')
-            .parent()
-            .click('top');
+          cy.get('body').click('top');
         });
 
         it('should close the second modal', () => {
@@ -462,9 +454,7 @@ describe('Modal', () => {
 
       context('when the modal overlay is clicked', () => {
         beforeEach(() => {
-          cy.findByLabelText('Delete Item')
-            .parent()
-            .click('top');
+          cy.get('body').click('top');
         });
 
         it('should close the popup', () => {
--- a/cypress/integration/Modal.spec.ts
+++ b/cypress/integration/Modal.spec.ts
@@ -452,7 +452,7 @@ describe('Modal', () => {
         cy.findByLabelText('Really Delete Item').should('exist');
       });
 
-      context('when the modal overlay is clicked', () => {
+      context('when clicking outside the modal', () => {
         beforeEach(() => {
           cy.get('body').click('top');
         });
--- a/cypress/integration/PopupStack.spec.ts
+++ b/cypress/integration/PopupStack.spec.ts
@@ -1,14 +1,25 @@
 import * as h from '../helpers';
 import {queries} from '@testing-library/dom';
 
+const getStackedParent = (element: HTMLElement): HTMLElement | null => {
+  if (element.style.zIndex !== '') {
+    return element;
+  } else {
+    if (element.parentElement) {
+      return getStackedParent(element.parentElement);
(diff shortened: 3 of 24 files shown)


cypress/helpers/modal.ts
```diff
@@ -28,5 +28,16 @@ export function getCloseButton($modal: JQuery): JQuery {
...
+}
+
+function getElement(element: JQuery, predicate: (element: JQuery) => boolean): JQuery {
+  if (predicate(element)) {
+    return element;
+  } else if (element.parent().length) {
+    return getElement(element.parent(), predicate);
+  } else {
+    return Cypress.$();
+  }
```
NicholasBoll, line 42: We're just assuming the `overlay` is the topmost child of the modal that is under the `document.body`. We don't really have a way of detecting parent elements unless we add non-semantic attributes or classnames to these elements.

cypress/integration/Modal.spec.ts
```diff
@@ -141,7 +152,7 @@ describe('Modal', () => {
         context('when the overlay is clicked', () => {
           beforeEach(() => {
             cy.findByLabelText('Delete Item')
-              .pipe(h.modal.getOverlay)
+              .parent()
               .click('top');
```
NicholasBoll, line 156: This should be `cy.get('body').click('top')` instead. I don't think we should specify _which_ element is clicked, just that we're clicking outside the modal.

cypress/integration/PopupStack.spec.ts
```diff
@@ -1,14 +1,25 @@
...
+  if (element.style.zIndex !== '') {
+    return element;
+  } else {
+    if (element.parentElement) {
+      return getStackedParent(element.parentElement);
+    } else {
+      return null;
+    }
+  }
+};
```
NicholasBoll, line 14: This function is resilient to how many DOM elements there are between the element and the stacked parent. As long as `z-index` is set somewhere, we'll compare it.

modules/modal/react/lib/Modal.tsx
```diff
@@ -27,12 +27,7 @@ const Modal = ({
 }: ModalProps): JSX.Element | null =>
   // Only render if on the client and `open` is `true`
   open && typeof window !== 'undefined' ? (
-    <ModalContent
-      container={container || document.body}
-      padding={padding}
-      width={width}
-      {...modalContentProps}
-    />
+    <ModalContent container={container} padding={padding} width={width} {...modalContentProps} />
```
NicholasBoll, line 30: We can no longer set the default of the container to `document.body` here. It will short-circuit `ModalContent`. We'll need to look into deprecating `container` entirely. Setting `container` will probably seriously mess things up.

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -60,7 +59,7 @@ export interface ModalContentProps extends React.HTMLAttributes<HTMLDivElement>
    * This should be a sibling or higher than the header and navigation elements of the application.
    * @default document.body
    */
-  container: HTMLElement;
+  container?: HTMLElement;
```
NicholasBoll, line 62: This should be optional and probably removed

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -193,13 +192,14 @@ const ModalContent = ({
...
   const modalRef = React.useRef<HTMLDivElement>(null);
+  const centeringRef = React.useRef<HTMLDivElement>(null);
 
   const onClose = () => handleClose?.();
 
   // special handling for clicking on the overlay
   const onOverlayClick = (event: React.MouseEvent<HTMLElement>) => {
     // Detect clicks only on the centering wrapper element
-    if (event.target === modalRef.current?.children[0] && PopupStack.isTopmost(modalRef.current)) {
+    if (event.target === centeringRef.current && PopupStack.isTopmost(modalRef.current!)) {
```
NicholasBoll, line 202: Better than `children[0].children[0]`

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -234,7 +235,8 @@ const ModalContent = ({
 
   // only render something on the client
   if (typeof window !== 'undefined') {
-    return ReactDOM.createPortal(content, container || document.body);
+    console.log(container, modalRef.current);
```
NicholasBoll, line 238: Remove this

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -212,10 +212,11 @@ const ModalContent = ({
   useAssistiveHideSiblings(modalRef);
 
   const content = (
-    <Container {...elemProps} ref={modalRef}>
+    <Container {...elemProps}>
       <CenteringContainer
+        ref={centeringRef}
         style={{width: windowSize.width % 2 === 1 ? 'calc(100vw - 1px)' : '100vw'}}
-        onClick={mergeCallback(onOverlayClick, elemProps.onClick)}
+        onMouseDown={onOverlayClick}
```
NicholasBoll, line 219: Princess Popups close on `mousedown` and if we don't also use `mousedown`, they 2 systems will not work together very well. It is a slight change to how things close - `mousedown` vs `click`. Will this be a problem?

@jamesfan I know we talked about this before. I think you originally choose an item in the select on mousedown and we switched it to `click` instead...

modules/modal/react/stories/stories_VisualTesting.tsx
```diff
@@ -17,12 +18,33 @@ export default withSnapshotsEnabled({
...
 const noop = () => {}; // eslint-disable-line no-empty-function
 
+const TestContent = () => {
+  const content = (
+    <div
+      style={{
+        color: 'white',
+        zIndex: 1,
+        position: 'relative',
+      }}
```
NicholasBoll, line 28: This ensures the Modal always renders on top of other content. This proves it still works. It would show as a visual regression otherwise

modules/modal/react/lib/ModalContent.tsx
jamesfan: @NicholasBoll That's correct. It shouldn't be an issue to change it back to `mousedown`.

cypress/integration/Modal.spec.ts
```diff
@@ -140,9 +151,7 @@ describe('Modal', () => {
 
         context('when the overlay is clicked', () => {
           beforeEach(() => {
-            cy.findByLabelText('Delete Item')
-              .pipe(h.modal.getOverlay)
-              .click('top');
+            cy.get('body').click('top');
```
NicholasBoll, line 154: We really only care about clicking outside the modal. I don't think we need to encode anything about the Modal outside that

modules/popup-stack/stories/stories_testing.tsx
```diff
@@ -3,7 +3,7 @@
 
 import React from 'react';
 import ReactDOM from 'react-dom';
-import {css, jsx} from '@emotion/core';
+import {css, jsx, CSSObject} from '@emotion/core';
```
NicholasBoll, line 6: `CSSObject` is unused

modules/popup/react/lib/Popup.tsx
```diff
@@ -103,6 +103,10 @@ const Container = styled('div', {
       animationDuration: '150ms',
       animationTimingFunction: 'ease-out',
       transformOrigin: `${transformOrigin.vertical} ${transformOrigin.horizontal}`,
+      // Allow overriding of animation in special cases
+      '.no-animation &': {
```
NicholasBoll, line 107: Change to something like `wd-no-animation`

modules/popup/react/lib/useCloseOnOutsideClick.ts
```diff
@@ -25,11 +25,12 @@ export const useCloseOnOutsideClick = <E extends HTMLElement>(
...
     }
   };
+
   React.useLayoutEffect(() => {
-    document.addEventListener('click', onClick);
+    document.addEventListener('mousedown', onClick);
 
     return () => {
-      document.removeEventListener('click', onClick);
+      document.removeEventListener('mousedown', onClick);
```
NicholasBoll, line 33: `mousedown` is needed to integrate into Princess popups...

modules/popup/react/lib/usePopupStack.ts
```diff
@@ -4,27 +4,42 @@ import {PopupStack} from '@workday/canvas-kit-popup-stack';
...
+ * directly with `ReactDOM.createPortal(<YourComponent>, ref.current!)`. This is definitely strange
+ * for React code, but is necessary for the PopupStack to remain framework agnostic and flexible to
+ * integrate with existing popup stack systems
+ * @param target Usually the trigger of a popup. This will fix `bringToTop` and should be provided
+ * by all ephemeral-type popups (like Tooltips, Select menus, etc). It will also add in clickOutside
+ * detection
  */
 export const usePopupStack = <E extends HTMLElement>(
   ref: React.RefObject<E>,
   target?: HTMLElement
```
NicholasBoll, line 22: We should probably allow `target` to be a ref...

modules/popup/react/lib/usePopupStack.ts
```diff
@@ -4,27 +4,42 @@ import {PopupStack} from '@workday/canvas-kit-popup-stack';
...
+ * apply this ref to a React element. Doing so will result in an error. Instead, use this `ref`
+ * directly with `ReactDOM.createPortal(<YourComponent>, ref.current!)`. This is definitely strange
+ * for React code, but is necessary for the PopupStack to remain framework agnostic and flexible to
+ * integrate with existing popup stack systems
+ * @param target Usually the trigger of a popup. This will fix `bringToTop` and should be provided
+ * by all ephemeral-type popups (like Tooltips, Select menus, etc). It will also add in clickOutside
+ * detection
  */
 export const usePopupStack = <E extends HTMLElement>(
   ref: React.RefObject<E>,
```
NicholasBoll, line 21: Should this even be a `RefObject` anymore? The Popup Stack controls this element now. I'm not even sure we should use a ref... Changing this would be an obvious breaking change. Not changing it is a subtle breaking change.

modules/tooltip/react/stories/visual.tsx
```diff
@@ -59,7 +59,6 @@ export const Placements = () => {
         !open ? null : (
           <Popper
             key={placement}
-            containerElement={containerElement}
```
NicholasBoll, line 62: I guess this is proving `containerElement` is just broken.

cypress/integration/Modal.spec.ts
jamesfan: Should we update the `'when the overlay is clicked'` context descriptions to `'when clicking outside the modal'` (or something similar) then?

modules/modal/react/stories/stories_VisualTesting.tsx
```diff
@@ -17,12 +18,33 @@ export default withSnapshotsEnabled({
...
+    >
+      This should be invisible if the Modal is rendering correctly, because the Modal should use the
+      PopupStack which will set a zIndex and place the modal and its overlay _on top_ of the this
+      text. If the text renders on top of the overlay, the text will show up
```
jamesfan, line 32: Remove `"this"` (or `"the"`) in `"on top of the this text"`. Also, `"If the text renders on top of the overlay, the text will show up"` reads a bit odd to me (wasn't sure if it was one of your half-finished thoughts 😆 ). I wonder if `"Otherwise, the text will be visible on top of the overlay."` would read better, or you could just remove the second sentence entirely since it's already implied.

cypress/integration/Modal.spec.ts
NicholasBoll: Yes, thank you!

modules/popup-stack/README.md
```diff
@@ -115,6 +116,13 @@ export interface PopupStackItem {
 }
 ```
 
+### createContainer
+
+Create a HTMLElement as the container for the popup stack item. The returned element reference will
+be the reference to be passed to all other methods. The Popup Stack will control when this element
+is added and removed from the DOM as well as the `z-index` style property. Your content should be
+added to this element.
```
jamesfan, line 124: Should there be a mention of `createAdapter` in the Popup Stack README as well?

NicholasBoll: Yes

PR #788, opened 2020-07-21
DNM: feat(core): Add new experimental type hierarchy to labs core for testing

anicholls:
Several teams want to test some iterations to our type hierarchy. Since the labs hierarchy is being used in production for several products, I've added a "hidden" hierarchy under the namespace `_typeTest`.

```tsx
import {_typeTest as type} from '@workday/canvas-kit-labs-react-core';
// ...
<h1 css={_typeTest.brand}>Brand 1 Header</h1>
```


modules/_labs/core/react/stories/stories.tsx
```diff
@@ -82,3 +82,63 @@ export const Space = () => {
     </Box>
   );
 };
+
+export const TypeTest = () => (
```
anicholls, line 86: This is for testing - will need to be removed before merging

.storybook/preview-head.html
```diff
@@ -2,6 +2,10 @@
   document.title = 'Canvas Kit Storybook';
 </script>
 <style type="text/css">
+  html {
+    font-size: 62.5%;
```
anicholls, line 6: For Sean:

I liked the idea of base 10 (particularly when we were thinking about how the rems would apply to spacing). However, adding this may be more trouble than it's worth. It may not be as big of an issue within product (where pretty much everything has a px font size specified at some level of the css scope tree. However, within our storybook it caused a lot of changes because we have many headers that do not have an explicit font size applied (meaning they use the browser default of rems, but the rem default is not updated to account for this lower percentage).

If we want to keep base 10, in order to not introduce a bunch of changes to storybook we can either remove this (making anything within the CK storybook that uses the test hierarchy 37.5% larger), or add explicit px or rem sizing for all unstyled text in our stories.

seans-cummin: Thanks Alex, would it be cleaner for us to use 16/1rem as our base so we wouldn't need to introduce a percentage for the font size which would introduce that added complexity?

NicholasBoll: I'm personally not a big fan of the `62.5%` trick. The default font-size is `16px`, so `16 * .625 = 10` which sets the default font size to `10px`. The percent is used so that if the user changes their base font size to something other than `16px`, everything will automatically scale. It is a convenience to developers (we can think "the design specifies 16px, so it is easy to convert as 1.6rem").

@anicholls did you try adding `font-size: 1.6rem` to the `body` element below?

The reason I'm not a fan of the the `62.5%` trick is the buy-in required by everyone. If you control everything about an application, including the root template and all CSS, it can work well. In our case we do not control everything. It requires every application to set the `font-size` of the `html` and/or `body` elements to these values. I think it makes adoption more difficult.

I think we can use Emotion and SCSS to our advantage and simply make functions. We can do `rem(16)` for a `16px` equivalent size. Should we care what the output value is if we never really have to see it? The input is `16` and the output will be `1rem`. If we did `rem(1)`, the output would be `0.0625rem` (probably for a 1px border). This gives us the benefit of thinking in terms of pixels but using rems. The same advantage of the 62.5% trick, but with no math on our part and without the downside of explicit by-in from everyone.

anicholls: Yeah I think the big argument for the `62.5%` trick is you see readable values when inspecting. But it seems to cause more problems than it's worth. I will update!

anicholls:
I've been playing around wtih a new API for the new hierarchy. I’m trying to come up with a way that’s less clunky to apply various weights and variants than using the standard `{...hierarchy.level2, ...hierarchy.variant.link}`

@alanbsmith @NicholasBoll Can I get your thoughts on this?

I have 3 different options:

### 1. Basic function call
A basic function call `type(1, ['link', 'caps'], 'bold')`,  where the first prop is the level, the second prop could be an array or string and it is any variants you would like to apply, and the third can override the weight.  
  - Pros: it’s pretty easy to understand (not much magic)
  - Cons: typing is kind of a pain if we want to restrict what variants/weights you can apply to certain levels


### 2. Method chaining that returns a function
Using chaining that returns a function `type(1).bold().link()()`. 
  - Pros: This is super clean and matches my mental model. Depending on the level that gets passed in, we can define what chained functions are available. 
  - Cons: You need to do a double call at the end. This is because the bold, link, etc. functions can’t be included in the css when you just do `type(1)`. Also, the autocomplete is kind of busy because you have all of the function prototype stuff in there too (arguments, caller, length, etc.)

### 3. Method chaining that returns an object
3. Same as #2, but returning an object from `type(1)` instead of a function. This now requires you to use a named function rather than a double call at the end to get the css. E.g. `type(1).bold().link().use()`
  - Pros: The same as 2, but clean autocomplete. 
  - Cons: the `.use()` is awkward particularly on the base case (`type(1).use()`)

---

For 2 and 3, we could do something like `.variant('link', 'caps')` rather than `.link()` and `.caps()` as well, but that’s a minor detail

Here's an example of how this could be implemented: 
image

NicholasBoll:
Why not a chainable API?

```ts
const type = function(level: 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10) {
  let css = hierarchy[`level${level}`]

  return chainableHierarchy(css)
}

function chainableHierarchy (css: CSSProperties) {
  return {...css,
    bold() {
      return chainableHierarchy({...css, fontWeight: 700})
    },
    medium() {
      return chainableHierarchy({...css, fontWeight: 500})
    },
    link() {
      return chainableHierarchy({...css, color: statusColors.error})
    }
  }
}

console.log(type(1)) /* { fontFamily: '"Roboto", "Helvetica Neue", "Helvetica", Arial, sans-serif',
  fontSize: '0.625rem',
  lineHeight: '1rem',
  letterSpacing: '0.025rem',
  fontWeight: 400,
  color: '#494949',
  bold: [λ: bold],
  medium: [λ: medium],
  link: [λ: link] } */
console.log(type(1).bold().link()) /* { fontFamily: '"Roboto", "Helvetica Neue", "Helvetica", Arial, sans-serif',
  fontSize: '0.625rem',
  lineHeight: '1rem',
  letterSpacing: '0.025rem',
  fontWeight: 700,
  color: '#de2e21',
  bold: [λ: bold],
  medium: [λ: medium],
  link: [λ: link] }  */
```

It will have extra functions attached, but would that be an issue? When serialized, the functions will be stripped.

anicholls:
@NicholasBoll I considered that, but figured the chainable functions being included in the resulting CSS would be a problem. I didn't even try it, as I was pretty sure it would be an issue during the serialization. Maybe it will be fine though. Will give it a shot!

modules/_labs/core/react/lib/typeTest.ts
```diff
@@ -0,0 +1,139 @@
...
+  level2: CSSProperties;
+  level1: CSSProperties;
+  [key: string]: CSSProperties;
+}
+
+const broswerDefaultFontSize = 16;
+/**
+ * Converts px values into base 16 rems
+ */
+const rem = (pxValue: number) => {
```
alanbsmith, line 24: This feels very similar to Polish's [rem](https://polished.js.org/docs/#rem) function.

modules/_labs/core/react/lib/typeTest.ts
```diff
@@ -0,0 +1,139 @@
...
+  level6: CSSProperties;
+  level5: CSSProperties;
+  level4: CSSProperties;
+  level3: CSSProperties;
+  level2: CSSProperties;
+  level1: CSSProperties;
+  [key: string]: CSSProperties;
+}
+
+const broswerDefaultFontSize = 16;
```
alanbsmith, line 20: nit: `s/broswerDefaultFontSize/browserDefaultFontSize`

modules/_labs/core/react/lib/typeTest.ts
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Converts px values into base 16 rems
+ */
+const rem = (pxValue: number) => {
+  return `${pxValue / broswerDefaultFontSize}rem`;
```
alanbsmith, line 25: same nit: `s/broswerDefaultFontSize/browserDefaultFontSize`

modules/_labs/core/react/lib/typeTest.ts
```diff
@@ -0,0 +1,139 @@
...
+    letterSpacing: rem(0.4),
+    fontWeight: 400,
+    color: typeColors.body,
+  },
+};
+
+type Variant = keyof typeof beta_Type.variant;
+type TextProps = {
+  level: 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10;
+  variant?: Variant | Variant[];
```
alanbsmith, line 116: I don't know if I would allow multiple variants as you could have some odd style collisions from the [variant list](https://github.com/Workday/canvas-kit/blob/98eed9fb06d0e6b47d8e39beb15e1512471d1e73/modules/core/react/lib/type.ts#L115-L160)

modules/_labs/core/react/lib/typeTest.ts
```diff
@@ -0,0 +1,139 @@
...
+    lineHeight: rem(16),
+    letterSpacing: rem(0.4),
+    fontWeight: 400,
+    color: typeColors.body,
+  },
+};
+
+type Variant = keyof typeof beta_Type.variant;
+type TextProps = {
+  level: 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10;
```
alanbsmith, line 115: I really like the idea of `level`s. I think it helps users get it right. I wonder if `Text` should be limited to levels 1-4 and 5-10 are reserved for Headings to reduce the number of options. 🤔

alanbsmith: Also, any reason to not have these map to h1-h6 and text sizes?

modules/_labs/core/react/lib/typeTest.ts
seans-cummin: If this is something we wish to have perhaps we could expand the variations we have available... say if someone needed a mono-link variation.

modules/_labs/core/react/lib/typeTest.ts
seans-cummin: @alanbsmith there's are number of issues when it comes to mapping our typestyles to HTML semantic naming. From a systems view this isn't a scalable solution, too many type variables are built into these styles, limiting what we can offer. It also doesn't convey the properties of the type effectively.

In my mind we may also run into the same issue with Levels (if we are using levels from style naming that is)

Happy to catch up on this and discuss :)

alanbsmith: Thanks for the context, @seans-cummin. 😀

The main challenge for engineers with scalar props is learning to traverse the scale fluently. Most engineers tend to think in values, not scales (in my experience). Without an existing mental model it's challenging to keep it in your head if you don't think about type scales all the time. And the result can be significant frustration. If an engineer has to reference the documentation often to understand which level to use, it's not really an ergonomic win for them (they'll go back to styling it other ways). And that's fine, but there's no reason to keep the prop.

Some ways you can reduce the burden:
* reduce the number of steps
* make the baseline obvious 
* make a mid-scale baseline (if an engineer can find a mid-point, it's easier to move up or down)
* make the steps guessable
* use meaningful and / or memorable step names
* make step values easily discoverable

You don't need to use all of these, but I've seen combinations of them be effective for engineers to create mental models for type and spacing scales.

modules/_labs/core/react/lib/typeTest.ts
alanbsmith: Yep, variant combinations are challenging to get right. The problem with variant arrays like this they can create unexpected results based on their order in the array. You can avoid that issue by keeping variants limited to high-level, distinct variations, and providing smaller, distinct changes through other props. Let's look at your mono + link suggestion as an example:

````tsx
<Text variant="link" fontFamily="mono">a link with mono text</Text>
````

This makes the api more declarative and reduces the chance for unintended side effects. (Though I would recommend for this particular example we create a semantic Link component that's distinct from Text.)

modules/_labs/core/react/lib/typeTest.ts
seans-cummin: Totally agree!

I'm in the process of formulating a workshop to inform how we should approach the naming convention from a design/engineering perspective so we can form a shared language when we talk about type.

alanbsmith: The more I think about `levels` the more I think they're really just Text variants.

anicholls:
Closing for now until we finish our internal testing

PR #797, opened 2020-07-23
fix(select): Support aria-required

### LINKED ISSUE #781: Labs Select throws an a11y violation when using aria-required
## 🐛 Bug Report

The Labs Select throws an a11y violation with [this test suite](https://ghe.megaleo.com/UIC/uic-common-utils/tree/master/utilities/a11y) when provided with an `aria-required` prop. The Labs Select doesn't perform any special handling for the `aria-required` prop, so it's spread onto the `button` rendered by the component along with the rest of the `elemProps`. However, `aria-required` is not an allowed attribute for a `button` according to the tests run by the referenced suite.

Workday Accessibility suggested we use a `button` for the Labs Select. [This example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html) referenced on the [WAI-ARIA Authoring Practices](https://www.w3.org/TR/wai-aria-practices-1.1/#examples-6) also utilizes a `button`.

## To Reproduce

Follow the installation and usage instructions for the `uic-common-utils-a11y` tests: https://ghe.megaleo.com/UIC/uic-common-utils/tree/master/utilities/a11y

Run the `toHaveNoViolations()` test with `<Select aria-required={true}>` and you should be presented with something similar to the following violation:

```
{
  "id": "aria-allowed-attr",
  "impact": "critical",
  "tags": [
    "cat.aria",
    "wcag2a",
    "wcag412"
  ],
  "description": "Ensures ARIA attributes are allowed for an element's role",
  "help": "Elements must only use allowed ARIA attributes",
  "helpUrl": "https://dequeuniversity.com/rules/worldspace/3.5/aria-allowed-attr?application=attestAPI",
  "nodes": [
    {
      "any": [
        {
          "id": "aria-allowed-attr",
          "data": [
            "aria-required=\"true\""
          ],
          "relatedNodes": [],
          "impact": "critical",
          "message": "ARIA attribute is not allowed: aria-required=\"true\""
        }
      ],
      "all": [],
      "none": [],
      "impact": "critical",
      "html": "<button aria-haspopup=\"listbox\" aria-required=\"true\" type=\"button\" value=\"email\" class=\"css-abphv6 e6ydbno0\">E-mail</button>",
      "target": [
        "button"
      ],
      "failureSummary": "Fix any of the following:\n  ARIA attribute is not allowed: aria-required=\"true\""
    }
  ]
}
```

jamesfan:
## Summary

Fixes #781.

Previously, the Labs Select was throwing an a11y violation when assigned the `aria-required` prop (the Select is rendered using a `button` and `button` isn't allowed to have `aria-required`). This PR adds special handling which passes `aria-required` along to the Select Menu to be applied to the `listbox` in accordance with Workday Accessibility's requirements.

I've also added a corresponding Cypress and unit tests.

### DIFF
--- a/modules/_labs/select/react/spec/Select.spec.tsx
+++ b/modules/_labs/select/react/spec/Select.spec.tsx
@@ -41,6 +41,14 @@ describe('Select', () => {
     });
   });
 
+  describe('when rendered with the required prop', () => {
+    it('should apply aria-required to its menu listbox when the menu is activated', () => {
+      const {getByRole} = render(<Select onChange={cb} options={options} required={true} />);
+      fireEvent.click(getByRole(selectButtonRole));
+      expect(getByRole(listboxRole)).toHaveAttribute('aria-required', 'true');
+    });
+  });
+
   describe('when an option with a different value than the current value of the select is clicked', () => {
     it('should call the onChange callback', () => {
       const {getAllByRole, getByRole} = render(<Select onChange={cb} options={options} />);
--- a/modules/_labs/select/react/spec/Select.spec.tsx
+++ b/modules/_labs/select/react/spec/Select.spec.tsx
@@ -6,6 +6,7 @@ describe('Select', () => {
   const cb = jest.fn();
 
   const selectButtonRole = 'button';
+  const listboxRole = 'listbox';
   const optionRole = 'option';
 
   const options = [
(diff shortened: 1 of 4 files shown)


modules/_labs/select/react/spec/Select.spec.tsx
```diff
@@ -32,6 +33,14 @@ describe('Select', () => {
     });
   });
 
+  describe('when rendered with the aria-required prop', () => {
+    it('should apply aria-required to its menu listbox when the menu is activated', () => {
+      const {getByRole} = render(<Select onChange={cb} options={options} aria-required={true} />);
+      fireEvent.click(getByRole(selectButtonRole));
+      expect(getByRole(listboxRole)).toHaveAttribute('aria-required', 'true');
+    });
+  });
```
NicholasBoll, line 42: I think we should have a Cypress test for this as well.

PR #799, opened 2020-07-23
fix(color-picker): Prevent enter key fowarding to next focus target

### LINKED ISSUE #760: ColorPicker keyDown -> keyPress
## 🐛 Bug? Report

Not exactly a bug, more of a request to change behavior.

The request is to make [this change](https://github.com/Workday/canvas-kit/compare/master...davvidbaker:patch-1) to the color picker hex input behavior. (Use keypress instead of keydown for submission from the hex text input.)

## To Reproduce

The problem with keydown arises in the context of a text editor. We are responding to the `onColorChange` callback by re-focusing the text editor, but the `Enter` keypress event gets fired after the text editor has refocused, resulting in a new line. Changing the color picker text input to respond to the `Enter` keypress instead of keydown would fix this problem.

Here's a gif and sandbox illustrating the problem.
keydown-color

[Sandbox](https://codesandbox.io/s/magical-hooks-g9poq?file=/src/App.js)

If this is not a change y'all would make, do you have a recommended alternative solution?

NicholasBoll:
Fixes #760

## Summary

The original implementation used keydown to detect the "Enter" key and fire the `onColorChange` event. If focus is transferred during the timing of this event, the enter key will actually be forwarded to the next target. This isn't exactly what we want.

This PR moves the custom color input into a `form` element and uses the submit event to handle special functionality like the enter key. This is a more accessible option.

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/_labs/color-picker/react/lib/ColorPicker.tsx
+++ b/modules/_labs/color-picker/react/lib/ColorPicker.tsx
@@ -29,14 +29,19 @@ export interface ColorPickerProps extends React.HTMLAttributes<HTMLDivElement> {
    */
   showCustomHexInput?: boolean;
   /**
-   * The label text of the custom hex input. This is also used as the aria-label
+   * The label text of the custom hex input.
    * @default 'Custom Hex Color'
    */
   customHexInputLabel?: string;
   /**
+   * The label of the custom hex color submit icon button.
+   * @default 'Submit'
+   */
+  submitLabel?: string;
+  /**
    * The function called when the submit icon is clicked.
    */
-  onSubmitClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+  onSubmitClick?: (event: React.FormEvent) => void;
   /**
    * The function called when the color rest button is selected.
    * It is required to be set for the reset button to render.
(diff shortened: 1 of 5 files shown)


modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -117,6 +122,7 @@ const isCustomColor = (colors: string[], hexCode?: string) => {
 const ColorPicker = ({
   colorSet = defaultColors,
   customHexInputLabel = 'Custom Hex Color',
+  submitLabel = 'Submit',
```
NicholasBoll, line 125: I'm not sure why this label was shared with the input. The label didn't describe the functionality any different than the input. This also made it harder to test.

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -29,14 +29,19 @@ export interface ColorPickerProps extends React.HTMLAttributes<HTMLDivElement> {
...
+  /**
+   * The label of the custom hex color submit icon button.
+   * @default 'Submit'
+   */
+  submitLabel?: string;
   /**
    * The function called when the submit icon is clicked.
    */
-  onSubmitClick?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+  onSubmitClick?: (event: React.FormEvent) => void;
```
jamesfan, line 44: This change should also be reflected in the README (duplicate documentation strikes again).

PR #802, opened 2020-07-24
fix(menu): Delay focus to prevent scrolling

### LINKED ISSUE #632: Using Menu with Popper causes a page scroll when popper is attached to document.body
## 🐛 Bug Report

The Popper component portals content to document.body when containerElement is undefined.

When the children is a component like Menu, which sets the focus on a menu item, the browser scrolls to the portal content.

This is probably a timing issue. Popper JS should update the position after the menu is focused.

## To Reproduce

```tsx
	<Popper
		ref={popperRef}
		placement={'left-start'}
		open={isOpen}
		anchorElement={buttonRef.current}
	>
		<Menu
			isOpen={isOpen}
		>
			<MenuItem>
				1
			</MenuItem>
		</Menu>
	</Popper>
```
## Expected Behavior

Popper Js should position the portal correctly so the page doesn't jump.

## Actual Results

The browser jumps to the newly added portal/content. This causes page to scroll.

## Browser (if applicable)

Chrome V80.x

NicholasBoll:
Fixes #632

## Summary

Prevent scrolling the page. The Menu calls `focus` on the `ul` on mount, but this is before PopperJS positions the menu. This causes the browser to scroll to the top of the page (default position of the Popper until positioning). This change delays the `focus` call until after the Popper is positioned.

We'll have a more robust focus utility function later. VoiceOver on iOS has issues with setting focus on newly created DOM too early. See #694

## Additional References

menu-not-scrolling-on-open

### DIFF
--- a/modules/_labs/menu/react/lib/Menu.tsx
+++ b/modules/_labs/menu/react/lib/Menu.tsx
@@ -62,6 +62,7 @@ const List = styled('ul')({
 
 export default class Menu extends React.Component<MenuProps, MenuState> {
   private id = uuid();
+  private animateId: number;
 
   private menuRef: React.RefObject<HTMLUListElement>;
   private firstCharacters: string[];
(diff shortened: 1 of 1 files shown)


modules/_labs/menu/react/lib/Menu.tsx
```diff
@@ -89,16 +90,22 @@ export default class Menu extends React.Component<MenuProps, MenuState> {
     if (this.props.isOpen && !prevProps.isOpen) {
       this.setInitialSelectedItem();
     }
-    if (this.props.isOpen && this.menuRef.current) {
-      this.menuRef.current.focus();
-    }
+    this.animateId = requestAnimationFrame(() => {
```
NicholasBoll, line 93: We delay the focus checking, but track the `requestAnimationId` so we can cancel on unmount just in case the component is mounted before this callback triggers.

PR #827, opened 2020-08-07
refactor(select): Convert more of Select to FC and improve code style

### LINKED ISSUE #759: Improve code style/structure for Labs Select
## Summary

This issue encapsulates a number of suggestions to the Labs Select to improve the overall code style and maintainability of the component.

1. Convert `SelectBase` to a function component and use hooks to handle scrolling behavior.
2. Reduce [burgeoning number of props](https://github.com/Workday/canvas-kit/pull/641#discussion_r439034783) in `SelectBase`. Consolidate props ([reference 1](https://github.com/Workday/canvas-kit/pull/672#discussion_r430622830), [reference 2](https://github.com/Workday/canvas-kit/pull/641#discussion_r439043028)) and consider using [context](https://github.com/Workday/canvas-kit/pull/641#discussion_r438950719).
3. Rename negative variables to affirmative forms ([reference](https://github.com/Workday/canvas-kit/pull/641#discussion_r439536771)).
4. Consider extracting behaviors into reusable hooks ([reference](https://github.com/Workday/canvas-kit/pull/641#discussion_r439035744)).
5. Update README to contain fully working examples.

### LINKED ISSUE #791: Labs Select menu momentarily positioned incorrectly in IE11
## 🐛 Bug Report

When navigating the Labs Select menu in IE11 using the down arrow key, the menu will occasionally disconnect itself from the Select button for a split second before snapping back to the correct position.

## To Reproduce

View the [Scrollable Select story](https://workday.github.io/canvas-kit/?path=/story/labs-select-react-top-label--scrollable) in IE11. Click on the Select to activate the menu and then navigate through the options using the down arrow key. As you navigate down the options, eventually the menu will scroll to keep the newly focused option in view -- occasionally, the menu will shift to the wrong position for a split second before snapping back to the correct position.

If you hold the down arrow (or up arrow) key down to rapidly cycle through the options, the issue will be more noticeable.

## Error Output

ie-menu-position

At the beginning of the capture, I'm holding down the down arrow key to rapidly cycle through the options to reproduce the effect more dramatically. After reaching the bottom of the menu, I navigate up the options at a slower pace to illustrate how the issue looks with more normal usage.

## Browser (if applicable)

IE11. This issue doesn't appear to affect other browsers (or it's imperceptible).

### LINKED ISSUE #872: Labs Select doesn't trigger onChange when renderOption prop is defined in IE 11
## 🐛 Bug Report

Currently, in IE 11, the Labs Select doesn't trigger `onChange` when an option is selected if the `renderOption` prop is defined.

This issue only occurs when an option is selected using the mouse; selecting an option using the keyboard properly triggers `onChange`.

## To Reproduce

1. View the [Labs Select renderOption example](https://workday.github.io/canvas-kit/?path=/story/labs-select-react-top-label--default-with-custom-options-data) in Storybook on IE 11.
2. Select a new option using the mouse.
3. Note how the Select doesn't update to reflect the newly selected option (it should, given that it's being used as a controlled component).

## Expected Behavior

The Select should update to reflect the newly selected option.

## Browser (if applicable)

Only affects IE 11.

jamesfan:
## Summary

Fixes #759.

Also fixes a couple IE 11 issues: 

Fixes #791 and fixes #872.

## Checklist

- [x] Convert `SelectBase` to a function component and use hooks to handle scrolling behavior.
- [x] Rename negative variables to affirmative forms ([reference](https://github.com/Workday/canvas-kit/pull/641#discussion_r439536771)).
- [x] Convert booleans for [menu visibility state](https://github.com/Workday/canvas-kit/pull/672#discussion_r430622830) and [placement](https://github.com/Workday/canvas-kit/pull/641#discussion_r439043028) to string literals.
- [x] Improve code examples in README to be more complete


modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -363,105 +326,133 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
...
+      // IE. Everywhere else (in IE and in other browsers), everything works
+      // fine without rAF.
+      const animateId = requestAnimationFrame(() => {
+        // We cannot use the native Element.scrollIntoView() here because it
+        // doesn't work properly with the portalled menu: when using the keyboard
+        // to advance focus through the options, using scrollIntoView to keep the
+        // newly focused option in view also scrolls the ENTIRE page. Instead, we
+        // call our own scrollIntoViewIfNeeded function.
+        scrollIntoViewIfNeeded(focusedOption, false);
+      });
```
jamesfan, line 352: This scrolling code works fine except for the `Select States Menu On` visual testing story in IE. Without `requestAnimationFrame`, the menu is scrolled slightly further down than it should be, which creates a visual regression. The scrolling code is wrapped in `useLayoutEffect` which should ensure proper measurements, but that doesn't appear to be happening in IE. I'm still investigating.

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -363,105 +326,133 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
...
+    const focusedOption = focusedOptionRef.current;
+
+    // We need to scroll if the menu is either opening or opened in case we decide to
+    // bypass the opening state and jump straight to opened (like in visual testing,
+    // for instance)
+    if (focusedOption && (menuVisibility === 'opening' || menuVisibility === 'opened')) {
+      // TODO: Again, why is rAF necessary here in IE? (see above)
+      const animateId = requestAnimationFrame(() => {
+        scrollIntoViewIfNeeded(focusedOption, true);
+      });
```
jamesfan, line 374: This `rAF` call needs to be present as well in order for `Select States Menu On` to be rendered properly in IE (**both** `rAF` calls need to be present).

modules/_labs/select/react/lib/SelectBase.tsx
jamesfan: Added a bit of extra info in the code comments for this rAF call. Resolving this and adding an updated comment.

modules/_labs/select/react/lib/SelectBase.tsx
jamesfan: Again, added a bit of extra info in the code comments for this rAF call. Resolving this and adding an updated comment.

modules/_labs/select/react/lib/types.ts
```diff
@@ -0,0 +1,33 @@
...
+ * `closing`: The menu is in the process of closing.
+ *
+ * Typically, a menu will transition through states in the following order:
+ * `closed` > `open` > `opening` > `opened` > `close` > `closing` > `closed`
+ *
+ * However, it's possible for a user to open a menu while it's in the process of
+ * closing in which case the menu can transition from `closing` > `opening` (and
+ * vice-versa).
+ */
+export type MenuVisibility = 'closed' | 'open' | 'opening' | 'opened' | 'close' | 'closing';
```
jamesfan, line 33: Recommend you start reviewing here. Most of the interesting work done in this PR is around this new `MenuVisibility` type. Previously, the visibility of the menu was represented using two separate booleans: `isMenuHidden` and `isMenuHiding`. This was problematic for a number of reasons outlined [here](https://github.com/Workday/canvas-kit/pull/641#discussion_r439536771) and [here](https://github.com/Workday/canvas-kit/pull/672#discussion_r430622830).

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -363,105 +326,140 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
...
+      // placement issue in IE (https://github.com/Workday/canvas-kit/issues/791),
+      // so I'm leaving it in for now.
+      const animateId = requestAnimationFrame(() => {
+        // We cannot use the native Element.scrollIntoView() here because it
+        // doesn't work properly with the portalled menu: when using the keyboard
+        // to advance focus through the options, using scrollIntoView to keep the
+        // newly focused option in view also scrolls the ENTIRE page. Instead, we
+        // call our own scrollIntoViewIfNeeded function.
+        scrollIntoViewIfNeeded(focusedOption, false);
+      });
```
jamesfan, line 357: Calling out this specific block of scrolling code wrapped within `requestAnimationFrame`. Code comments should provide the necessary context. The `rAF` call works, but I'm not sure it's the best solution (though I don't have a better one at the moment).

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -363,105 +326,140 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
...
+    // We need to scroll if the menu is either opening or opened in case we decide to
+    // bypass the opening state and jump straight to opened (like in visual testing,
+    // for instance)
+    if (focusedOption && (menuVisibility === 'opening' || menuVisibility === 'opened')) {
+      // TODO: This rAF call is also necessary for the Menu On visual testing
+      // stories to render properly in IE (see above, both rAF calls need to be
+      // present). It has no bearing on the menu placement issue.
+      const animateId = requestAnimationFrame(() => {
+        scrollIntoViewIfNeeded(focusedOption, true);
+      });
```
jamesfan, line 381: This `rAF` call is also necessary for visual testing in IE (both `rAF` calls are needed to avoid a visual regression in IE).

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -136,13 +117,14 @@ const menuListBorderStyles = (theme: EmotionCanvasTheme, error?: ErrorType): CSS
...
 const Menu = styled('div')<
-  Pick<SelectMenuProps, 'error' | 'isHiding' | 'shouldAnimate' | 'theme'> & {width: number}
+  Pick<SelectMenuProps, 'error' | 'theme' | 'visibility'> & {width: number}
 >(
   {
     backgroundColor: colors.frenchVanilla100,
     border: `1px solid ${inputColors.border}`,
     boxSizing: 'border-box',
     position: 'relative',
+    transition: `opacity ${menuAnimationDuration}ms`,
```
jamesfan, line 127: I'm using CSS transitions now (instead of CSS animations) to better support interrupted animations (e.g., opening the menu again while it's closing).

modules/_labs/select/react/lib/types.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * `closed`: The menu is completely closed (not present in the DOM).
+ * `open`: The menu was previously `closed` and has just been instructed to open.
+ *   The menu has been added to the DOM, but it's not yet visible. This state is
+ *   necessary in order to apply an opacity of 0 to the menu before transitioning it
+ *   to an opacity of 1.0 in the opening state.
+ * `opening`: The menu is in the process of opening.
+ * `opened`: The menu is completely open.
+ * `close`: The menu was previously `open` and has just been instructed to close.
+ * `closing`: The menu is in the process of closing.
```
NicholasBoll, line 24: I'm a bit confused on the difference between `close` and `closing`

cypress/integration/SelectLabs.spec.ts
```diff
@@ -337,6 +337,92 @@ describe('Select', () => {
...
+            beforeEach(() => {
+              cy.focused().type('{esc}');
+            });
+
+            context('when the menu is re-opened AFTER it has fully closed', () => {
+              beforeEach(() => {
+                // Wait a short period of time to give the menu time to fully close
+                // (so we don't interrupt the menu's closing animation and cause it
+                // to re-open while it's in the middle of closing)
+                cy.wait(1000);
```
NicholasBoll, line 387: Is it possible to instead explicitly wait for it to close instead? Something like:

```suggestion
                // Wait for menu to fully close before opening again
                cy.findByLabelText('Label')
                  .pipe(h.selectLabs.getMenu)
                  .should('not.exist')

```

modules/_labs/select/react/lib/types.ts
jamesfan: `close` is basically the opposite of `open`. Whereas `open` indicates the menu is _just now_ transitioning from `closed` and starting to open, `close` indicates the menu is _just now_ transitioning from `opened` and starting to close.

I added `close` mostly to keep things symmetrical, but it's not actually necessary. `open` was necessary to provide a starting point where the menu could be added to the DOM but with an opacity of `0` so that it could be transitioned to an opacity of `1` as it opens. However, we can go straight from `opened` to `closing` without adding the intermediate `close` state since the menu already exists in the DOM with the desired starting opacity of `1` in the `opened` state.

cypress/integration/SelectLabs.spec.ts
jamesfan: Nice, I didn't think about using an assertion like this to force Cypress to wait for the assertion to pass before proceeding with the test -- thanks for the tip 👍

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -363,105 +326,140 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
...
+};
+
+SelectBase.defaultProps = {
+  focusedOptionIndex: 0,
+  isMenuFlipped: false,
+  menuPlacement: 'bottom',
+  menuVisibility: 'closed',
+  shouldMenuAutoFlip: true,
+  shouldMenuAutoFocus: true,
+};
```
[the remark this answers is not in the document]
jamesfan: Default parameters sounds great. Looks like we already set the precedent in #525 (Labs Select wasn't included in the refactor since it was still in progress).

Besides, it looks like [`defaultProps` will be deprecated](https://twitter.com/dan_abramov/status/1133878326358171650) at some point.

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -309,39 +358,57 @@ export default class Select extends React.Component<SelectProps, SelectState> {
...
+
   handleOptionSelection = (index: number): void => {
-    // Abort immediately if:
-    // * The menu is in the process of hiding
-    // * A disabled option was clicked (we ignore these clicks)
-    if (this.state.isMenuHiding || this.normalizedOptions[index].disabled) {
+    // Abort if a disabled option was clicked (we ignore these clicks)
+    if (this.normalizedOptions[index].disabled) {
       return;
     }
```
jamesfan, line 413: Snuck in a quick fix to #872 (no longer aborts if the menu is in the process of closing). It's debatable if this check was really needed in the first place -- is it really a problem if an option is selected while the menu is in the middle of fading out?

lychyi (approved):

PR #831, opened 2020-08-12
feat(color-picker): Update and export default color set

jtschult:
## Summary
Adding and exporting an additional Color Picker colorset for consumers to use.

More options are needed than the default color picker palette. We found that multiple teams needed the same palette so instead of duplicating the colorset, we were hoping to export it as apart of the component.

## Checklist

- [x] `yarn test` passes

## Additional References

Screen Shot 2020-08-12 at 10 40 23 AM


modules/_labs/color-picker/react/lib/parts/ColorSets.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export const defaultColorSet = [
+  colors.cinnamon400,
+  colors.peach400,
+  colors.chiliMango400,
+  colors.cantaloupe400,
+  colors.sourLemon400,
+  colors.juicyPear400,
+  colors.kiwi400,
+  colors.greenApple400,
```
NicholasBoll, line 11: It's a little weird that the relationship between rows and columns is implicit (as shown here by newline spacing). I think the final API should have a stronger contract between rows and columns.

jtschult: agreed. We should address this before moving it out of labs 👍

lychyi:
Here's a consideration to think about 🤔 

Canvas Kit tries to make changes based on internal/external-agnostic semantic versioning principles. So it may be true that many teams are using a similar palette within Workday but if a change is needed when this thing is out of Labs, it will require a breaking change. Are we sure this is something we want to bake in this way? We are essentially increasing the API surface of this component.

Alternatively, if we just include this color palette as a gist (with a link it as a comment) or as sample code in the README, users can easily copy and paste it into their app. It's a very similar level of reusability without triggering a breaking change when we need to add colors or change/update the set. Theoretically, the disadvantage with this approach is that when we make updates they won't get it for free... but again, updates will mandate a breaking change here.

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -65,33 +65,69 @@ export interface ColorPickerProps extends React.HTMLAttributes<HTMLDivElement> {
...
+  colors.blueberry100,
+  colors.grapeSoda100,
+  colors.pomegranate100,
+  colors.cinnamon100,
+  colors.cantaloupe100,
+  colors.sourLemon100,
+  colors.greenApple100,
+  colors.jewel100,
+
+  colors.blackPepper600,
```
lychyi, line 123: From our discussion with the design team:

> The only change I have to the existing PR is to use the following colors in that bottom row: Black Pepper 600, 400, 300, 100, French Vanilla 500, 400, 200, 100.

lychyi:
#838 needs to be released in tandem with this feature

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -65,33 +65,69 @@ export interface ColorPickerProps extends React.HTMLAttributes<HTMLDivElement> {
   transformOrigin?: TransformOrigin;
 }
 
-const defaultColors = [
+const defaultColorSet = [
```
lychyi, line 68: Kind of a nit here but I renamed this const to be a more accurate depiction of what these colors are

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -191,4 +227,5 @@ const ColorPicker = ({
   );
 };
 
+ColorPicker.defaultColorSet = defaultColorSet;
```
lychyi, line 230: Exporting the `defaultColorSet` in a namespaced manner here. Thanks @NicholasBoll

Issue #770, opened 2020-07-02
Color Picker doesn't render properly in IE11


jamesfan:
## 🐛 Bug Report

The Color Picker doesn't render properly in IE 11. This is likely due to IE's poor support for CSS Grid.

## To Reproduce

View the Color Picker in Storybook.

![image](https://user-images.githubusercontent.com/146020/86397810-615fc900-bc59-11ea-9441-fcffeb4a25a1.png)

## Expected Behavior

Correct design taken from Chrome:

![image](https://user-images.githubusercontent.com/146020/86397969-ad127280-bc59-11ea-80ce-7cc22a7f49a5.png)

## Browser (if applicable)

IE 11

PR #832, opened 2020-08-18
fix(color-picker): Fix IE rendering for color picker swatchbook

jtschult:
## Summary
Fixes issue: https://github.com/Workday/canvas-kit/issues/770

In order to support grid layout in IE11, -ms-grid-row and -ms-grid-column need to be explicitly set. In addition IE doesn't support gridGap, so adjusted to use margin spacing. Lastly, I abstracted out some variables regarding the number of swatches within a row to possibly help support further future development.

## Checklist

- [x] `yarn test` passes

## Additional References

Screen Shot 2020-08-17 at 9 11 44 PM

### DIFF
--- a/modules/_labs/color-picker/react/lib/ColorPicker.tsx
+++ b/modules/_labs/color-picker/react/lib/ColorPicker.tsx
@@ -3,6 +3,7 @@ import {checkIcon} from '@workday/canvas-system-icons-web';
 import {ColorInput} from '@workday/canvas-kit-react-color-picker';
 import {IconButton} from '@workday/canvas-kit-react-button';
 import {Popup} from '@workday/canvas-kit-react-popup';
+import {TransformOrigin} from '@workday/canvas-kit-react-common';
 import * as React from 'react';
 import FormField from '@workday/canvas-kit-react-form-field';
 import styled from '@emotion/styled';
@@ -57,6 +58,11 @@ export interface ColorPickerProps extends React.HTMLAttributes<HTMLDivElement> {
    * @default 'Reset'
    */
   resetLabel?: string;
+  /**
+   * The origin from which the Toast will animate.
+   * @default {horizontal: 'center', vertical: 'top'}
+   */
+  transformOrigin?: TransformOrigin;
 }
 
 const defaultColors = [
--- a/modules/_labs/color-picker/react/lib/ColorPicker.tsx
+++ b/modules/_labs/color-picker/react/lib/ColorPicker.tsx
@@ -136,6 +136,7 @@ const ColorPicker = ({
   resetColor,
   value = '',
   showCustomHexInput = false,
+  transformOrigin,
   ...elemProps
 }: ColorPickerProps) => {
   const [validHexValue, setValidHexValue] = React.useState('');
@@ -159,7 +160,13 @@ const ColorPicker = ({
   };
 
   return (
-    <Popup width={250} padding={Popup.Padding.s} data-testid="canvas-color-picker" {...elemProps}>
+    <Popup
+      width={250}
+      padding={Popup.Padding.s}
+      transformOrigin={transformOrigin}
+      data-testid="canvas-color-picker"
+      {...elemProps}
+    >
       {onColorReset && resetColor && (
         <ResetButton onClick={onColorReset} resetColor={resetColor} label={resetLabel} />
       )}
--- a/modules/_labs/color-picker/react/lib/ColorPicker.tsx
+++ b/modules/_labs/color-picker/react/lib/ColorPicker.tsx
@@ -59,7 +59,7 @@ export interface ColorPickerProps extends React.HTMLAttributes<HTMLDivElement> {
    */
   resetLabel?: string;
   /**
-   * The origin from which the Toast will animate.
+   * The origin from which the Color Picker will animate.
    * @default {horizontal: 'center', vertical: 'top'}
    */
   transformOrigin?: TransformOrigin;
--- a/modules/_labs/color-picker/react/lib/parts/SwatchBook.tsx
+++ b/modules/_labs/color-picker/react/lib/parts/SwatchBook.tsx
@@ -4,8 +4,6 @@ import {borderRadius, colors, spacing} from '@workday/canvas-kit-react-core';
 import {focusRing, mouseFocusBehavior} from '@workday/canvas-kit-react-common';
 import {ColorSwatch} from '@workday/canvas-kit-react-color-picker';
 
-const numberOfSwatchColumns = 8;
-
 export interface SwatchBookProps {
   colors: string[];
   value?: string;
@@ -14,7 +12,6 @@ export interface SwatchBookProps {
 
 interface SwatchContainerProps {
   isSelected: boolean;
-  index: number;
 }
 
 const accessibilityBorder = `${colors.frenchVanilla100} 0px 0px 0px 2px, ${colors.licorice200} 0px 0px 0px 3px`;
@@ -22,9 +19,6 @@ const accessibilityBorder = `${colors.frenchVanilla100} 0px 0px 0px 2px, ${color
 const SwatchContainer = styled('div')<SwatchContainerProps>(
   {
     display: 'flex',
-    alignItems: 'center',
-    justifyContent: 'center',
-    position: 'relative',
     width: 20,
     height: 20,
     cursor: 'pointer',
@@ -41,10 +35,6 @@ const SwatchContainer = styled('div')<SwatchContainerProps>(
       ...focusRing({separation: 2}),
     },
   },
-  ({index}) => ({
-    '-ms-grid-column': (index % numberOfSwatchColumns) + 1 + '',
-    '-ms-grid-row': Math.ceil((index + 1) / numberOfSwatchColumns) + '',
-  }),
   ({isSelected}) => ({
     boxShadow: isSelected ? accessibilityBorder : undefined,
     ...mouseFocusBehavior({
@@ -63,8 +53,8 @@ const SwatchContainer = styled('div')<SwatchContainerProps>(
 );
 
 const Container = styled('div')({
-  display: ['-ms-grid', 'grid'],
-  gridTemplateColumns: `repeat(${numberOfSwatchColumns}, auto)`,
+  display: 'flex',
+  flexWrap: 'wrap',
   margin: `0px -${spacing.xxs} -${spacing.xxs} 0px`,
 });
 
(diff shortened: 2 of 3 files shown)


mannycarrera4:
This look pretty good, any chance you can add a chromatic test so that we don't regress this? I don't see a visual test for it in the first place

modules/_labs/color-picker/react/lib/parts/SwatchBook.tsx
```diff
@@ -37,6 +41,10 @@ const SwatchContainer = styled('div')<SwatchContainerProps>(
       ...focusRing({separation: 2}),
     },
   },
+  ({index}) => ({
+    '-ms-grid-column': (index % numberOfSwatchColumns) + 1 + '',
+    '-ms-grid-row': Math.ceil((index + 1) / numberOfSwatchColumns) + '',
+  }),
```
NicholasBoll, line 47: Hmm. Lack of autoplacement of grid items makes me wonder why we should use CSS grid for this layout instead of a single `display: flex` or a `display: flex` per row.

Emotion does a lot of stuff behinds the scenes that creates performance issues and this is one of them. For every cell (row/column combination), a new style rule is injected into the DOM containing _all_ styles in this block. Not just a style object with the properties defined here, but all of them. In this example, that's 22 properties - 20 of which are repeats. In this case, we have 24 unique indexes, so 24 * 20 = 480 style rules for a 24 boxes that only have a color difference.

We're doing the same thing for each swatch as well... In that case it's 16 style rules for each unique color where only the background color changes. That's 912 extra style rules that aren't adding any extra value - just space.

modules/_labs/color-picker/react/lib/parts/SwatchBook.tsx
```diff
@@ -4,6 +4,8 @@ import {borderRadius, colors, spacing} from '@workday/canvas-kit-react-core';
 import {focusRing, mouseFocusBehavior} from '@workday/canvas-kit-react-common';
 import {ColorSwatch} from '@workday/canvas-kit-react-color-picker';
 
+const numberOfSwatchColumns: number = 8;
```
NicholasBoll, line 7:
```suggestion
const numberOfSwatchColumns = 8;
```

Adding the `number` widens the type to `number` instead of `8`

modules/_labs/color-picker/react/lib/parts/SwatchBook.tsx
```diff
@@ -55,9 +63,9 @@ const SwatchContainer = styled('div')<SwatchContainerProps>(
 );
 
 const Container = styled('div')({
-  display: 'grid',
-  gridTemplateColumns: 'repeat(8, auto)',
-  gridGap: spacing.xxs,
+  display: ['grid', '-ms-grid'],
```
NicholasBoll, line 66:
```suggestion
  display: ['-ms-grid', 'grid'],
```

It is better to use prefix values first and specification values afterward

modules/_labs/color-picker/react/lib/parts/SwatchBook.tsx
jtschult: Do you advise I remove the grid layout, or enhance the logic to be more performant, or leave as is for this PR?

modules/_labs/color-picker/react/stories/stories_VisualTesting.tsx
```diff
@@ -0,0 +1,44 @@
...
+            showCustomHexInput: true,
+            resetColor: colors.blueberry400,
+            resetLabel: 'Reset',
+            onColorReset: action('Reset Clicked'),
+          },
+        },
+      ]}
+      columnProps={[{label: 'Default', props: {}}]}
+    >
+      {props => <ColorPicker {...props} onColorChange={action('Color Changed')} />}
```
mannycarrera4, line 41: @jtschult since the color picker animates in, you need to add `transformOrigin={null}` as a prop so that the snapshot get covered. Chromatic isn't the greatest when we have things that animate into frame.

mannycarrera4: And then in the `ColorPicker` you can add the `transformOrigin` prop like this: https://github.com/Workday/canvas-kit/blob/master/modules/toast/react/lib/Toast.tsx#L41

https://github.com/Workday/canvas-kit/blob/master/modules/toast/react/lib/Toast.tsx#L92

jtschult: Thank you

modules/_labs/color-picker/react/lib/parts/SwatchBook.tsx
lychyi: At the risk of missing something glaring...

Because the popup width is fixed, could we just make the `SwatchContainer` `display: inline-flex` or make the `Container` a flexbox that specifies `flex-wrap: wrap`. There would be no maths required. It's a little bit of a naive implementation but it should be easier to reason about and not have a perf hit.

jtschult: 👍

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -57,6 +58,11 @@ export interface ColorPickerProps extends React.HTMLAttributes<HTMLDivElement> {
    * @default 'Reset'
    */
   resetLabel?: string;
+  /**
+   * The origin from which the Toast will animate.
```
jtschult, line 62:
```suggestion
   * The origin from which the Toast will animate.
   * The origin from which the Color Picker will animate.
```

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -57,6 +58,12 @@ export interface ColorPickerProps extends React.HTMLAttributes<HTMLDivElement> {
...
+   * The origin from which the Color Picker will animate.
```
jtschult, line 63:
```suggestion
   * The origin from which the Color Picker will animate.
```

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -153,7 +160,13 @@ const ColorPicker = ({
   };
 
   return (
-    <Popup width={250} padding={Popup.Padding.s} data-testid="canvas-color-picker" {...elemProps}>
+    <Popup
+      width={250}
+      padding={Popup.Padding.s}
+      transformOrigin={transformOrigin}
+      data-testid="canvas-color-picker"
```
lychyi, line 167: @jtschult Heads up in case your team uses data-testid to query the `ColorPicker` popup in tests, we generally don't ship data attributes in Canvas Kit so not sure how this slipped through the first time. I'll open a separate PR to get rid of this.

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -130,6 +136,7 @@ const ColorPicker = ({
   resetColor,
   value = '',
   showCustomHexInput = false,
+  transformOrigin,
```
jtschult, line 139: @lychyi - Is this just a style preference. Curious the reason for the addition.

modules/_labs/color-picker/react/stories/stories_VisualTesting.tsx
```diff
@@ -32,15 +34,13 @@ export const ColorPickerStates = () => (
...
+            onColorReset: noop,
           },
         },
       ]}
       columnProps={[{label: 'Default', props: {}}]}
     >
-      {props => (
-        <ColorPicker {...props} transformOrigin={null} onColorChange={action('Color Changed')} />
-      )}
+      {props => <ColorPicker {...props} onColorChange={noop} />}
```
jtschult, line 43: fyi - this was failing without transformOrigin={null} originally because of the animation. If not needed, I'd love to learn why not

modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -153,7 +160,13 @@ const ColorPicker = ({
   };
 
   return (
-    <Popup width={250} padding={Popup.Padding.s} data-testid="canvas-color-picker" {...elemProps}>
+    <Popup
```
lychyi, line 163: Heads up @laurenraddatz and @mannycarrera4 

I think in the (near) future, we need to remove Popup from the ColorPicker. That should be something that the end user composes together not something we ship with this component.

I'm going to track this in  #836

modules/_labs/color-picker/react/stories/stories_VisualTesting.tsx
mannycarrera4: I think per Lynn's comment of shipping the color picker swatch book with popup, the Popup animates into the frame, and Chromatic seems like it takes the snapshot before that happens. If we remove Popup from the color picker per: #836 this shouldn't be needed.

modules/_labs/color-picker/react/lib/ColorPicker.tsx
lychyi: We are generally explicit about destructuring out documented props. The term `...elemProps` refers to the element props (or remainder, rest, etc.) spread onto the containing element (think `id`, `aria-*`, etc.). In this case `transformOrigin` would in fact go to the same place but we're explicitly directing it to go to the `Popup` where it belongs.

It sounds like of like an odd style preference, but we've ran into things in the past when this has bitten us and we weren't aware of overriding props or spreading it to the wrong place until we shipped the code.

modules/_labs/color-picker/react/stories/stories_VisualTesting.tsx
lychyi: Ah thanks for the context, I didn't know. It didn't seem necessary to me and it rendered fine in my local. Ran into the same trap as you (hit us up with an inline comment next time!)

There is a solution to this, I'll push it up in a sec. Basically Chromatic accounts for animations by pausing at the end. Unfortunately, I have to guess and check here, no way to really check this locally before we run snaps.

lychyi: FWIW, it's pretty rare we do this. In fact, we only do it in one other place: https://github.com/Workday/canvas-kit/blob/master/modules/modal/react/stories/stories_VisualTesting.tsx#L12-L16

lychyi: We're in business

image

lychyi:
Thanks for the assist here @jtschult!

PR #837, opened 2020-08-20
fix(select): Prevent disabled options from being keyboard-focused

jamesfan:
## Summary

We manage assistive focus on the Select component using [aria-activedescendant](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_focus_activedescendant); when the keyboard is used to navigate the menu, `aria-activedescendant` is updated to refer to the ID of the currently focused option.

Prior to this fix, assistive focus could be applied to a disabled option using the keyboard (Up/Down/Home/End keys) if that disabled option happened to be the first or last option in the menu. The disabled option was correctly styled without focus styles and it was unselectable, but `aria-activedescendant` was still being set to refer to the disabled option's ID. The Select's internal component state also incorrectly set focus to the disabled option. This meant the user could navigate to a disabled option using the keyboard as long as it was the first or last option (although it wouldn't appear to be focused visually, and they wouldn't be able to select it).

This PR addresses this edge case and prevents all options from receiving assistive focus using the keyboard, even if they're the first or last option in the menu.

I've added Cypress tests to protect against this behavior in the future.

### DIFF
--- a/cypress/integration/SelectLabs.spec.ts
+++ b/cypress/integration/SelectLabs.spec.ts
@@ -353,6 +353,110 @@ describe('Select', () => {
     });
   });
 
+  context('given the "Disabled Options Test" story is rendered', () => {
+    beforeEach(() => {
+      h.stories.load('Testing|React/Labs/Select', 'Disabled Options Test');
+    });
+
+    context('when the menu is opened', () => {
+      beforeEach(() => {
+        cy.findByLabelText('Label (Disabled Options)')
+          .focus()
+          .type('{downarrow}');
+      });
+
+      context('when the down arrow key is pressed', () => {
+        beforeEach(() => {
+          cy.focused().type('{downarrow}');
+        });
+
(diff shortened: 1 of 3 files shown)


jamesfan:
Note that although a user may not use the keyboard to focus a disabled option, it's still possible for a developer to set assistive focus to a disabled option by setting the `value` prop of the Select to a value corresponding to that disabled option. When the menu is displayed, `aria-activedescendant` will refer to the ID of that disabled option. `aria-selected="true"` will also be applied to that disabled option.

Also, if the `value` prop doesn't correspond to any of the provided options, the Select component will automatically set `aria-activedescendant` to the first option in the menu when the menu is displayed, even if that first option is disabled. That first option will also be given `aria-selected="true"` even if it's disabled.

In both cases, once the user uses the keyboard to navigate off of the disabled option, they will no longer be able to navigate back to that (or any other) disabled option.

PR #838, opened 2020-08-20
test(color-picker): Remove data attributes

lychyi:
## Summary

As a policy, we don't add `data-` attributes to our shipped codebase. If we every need them for testing, we'll use them at the `spec` level but never at a component level. This one somehow slipped through so I'm removing them here and changing our Cypress tests accordingly.

Bonus: removed `addon-actions` from the `ColorPicker` story. This is most likely going to freeze browsers like IE11.

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/cypress/integration/ColorPicker.spec.ts
+++ b/cypress/integration/ColorPicker.spec.ts
@@ -1,10 +1,10 @@
 import * as h from '../helpers';
 
 const getColorInput = () => cy.get('[type="text"]');
-const getColorPickerPopup = () => cy.get('[role=dialog]');
-const getOpenButton = () => cy.get('[data-testid=open]');
+const getColorPickerPopup = () => cy.findByRole('dialog');
+const getOpenButton = () => cy.findByLabelText('Select Background Color');
 const getResetButton = () => cy.contains('button', 'Reset');
-const getSubmitButton = () => cy.get('[aria-label=Submit]');
+const getSubmitButton = () => cy.findByLabelText('Submit');
 const getSwatch = (color: string) => cy.get(`div[color="${color}"]`);
 
 const colorInputStory = 'Components|Inputs/Color Picker/Color Input/React/Top Label';
--- a/cypress/integration/ColorPicker.spec.ts
+++ b/cypress/integration/ColorPicker.spec.ts
@@ -146,7 +146,9 @@ describe('ColorPicker', () => {
         beforeEach(() => {
           cy.get(`.wdc-color-picker--color-8660d1`).click();
           cy.get('button').click();
-          cy.get('[data-testid="color-picker-reset"]').click();
+          cy.get('button')
+            .contains('Reset')
+            .click();
         });
 
         it('should set the color picker value to the reset color', () => {
@@ -215,7 +217,9 @@ describe('ColorPicker', () => {
         beforeEach(() => {
           cy.get(`.wdc-color-picker--color-8660d1`).click();
           getColorInput().click();
-          cy.get('[data-testid="color-picker-reset"]').click();
+          cy.get('button')
+            .contains('Reset')
+            .click();
         });
 
         it('should set the color picker value to the reset color', () => {
--- a/cypress/integration/ColorPicker.spec.ts
+++ b/cypress/integration/ColorPicker.spec.ts
@@ -4,12 +4,13 @@ const getColorInput = () => cy.get('[type="text"]');
 
 const getColorPickerPopup = () => cy.get('[role=dialog]');
 
-const expandHex = (hex: string) => {
-  const shorthandRegex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i;
-  return hex.replace(shorthandRegex, function(m: string, r: string, g: string, b: string) {
-    return r + r + g + g + b + b;
-  });
-};
+const getOpenButton = () => cy.get('[data-testid=open]');
+
+const getResetButton = () => cy.contains('button', 'Reset');
+
+const getSubmitButton = () => cy.get('[aria-label=Submit]');
+
+const getSwatch = (color: string) => cy.get(`div[color="${color}"]`);
 
 const colorInputStory = 'Components|Inputs/Color Picker/Color Input/React/Top Label';
 const colorPreviewStory = 'Components|Inputs/Color Picker/Color Preview/React/Top Label';
@@ -117,7 +118,7 @@ describe('ColorPicker', () => {
     context('when the IconButton is clicked', () => {
       beforeEach(() => {
         h.stories.load(colorPickerStory, 'Icon Button Popup');
-        cy.get('button').click();
+        getOpenButton().click();
       });
 
       it('should pass accessibility checks', () => {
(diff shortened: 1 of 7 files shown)


cypress/integration/ColorPicker.spec.ts
```diff
@@ -146,7 +146,9 @@ describe('ColorPicker', () => {
         beforeEach(() => {
           cy.get(`.wdc-color-picker--color-8660d1`).click();
           cy.get('button').click();
-          cy.get('[data-testid="color-picker-reset"]').click();
+          cy.get('button')
+            .contains('Reset')
```
NicholasBoll, line 150:
```suggestion
          cy.contains('button', 'Reset')
```

cypress/integration/ColorPicker.spec.ts
```diff
@@ -215,7 +217,9 @@ describe('ColorPicker', () => {
         beforeEach(() => {
           cy.get(`.wdc-color-picker--color-8660d1`).click();
           getColorInput().click();
-          cy.get('[data-testid="color-picker-reset"]').click();
+          cy.get('button')
+            .contains('Reset')
```
NicholasBoll, line 221:
```suggestion
          cy.contains('button', 'Reset')
```

NicholasBoll:
Do we know if anyone is relying on this `data-testid` attribute? If they are, we should note this as potentially breaking for tests and that they should manually add the attribute if they rely on it.

lychyi:
@NicholasBoll Not that I know of. @jtschult acknowledged this in a [previous PR's comment](https://github.com/Workday/canvas-kit/pull/832/files#r473270264). We'll have to communicate it internally before the release but it should be okay since we don't generally this anyway.

modules/_labs/color-picker/react/lib/parts/SwatchBook.tsx
```diff
@@ -67,15 +67,9 @@ export const SwatchBook = ({colors, value, onSelect}: SwatchBookProps) => (
...
 
-      const formatHex = (value: string) => {
-        return value.replace(/#/g, '').substring(0, 6);
-      };
-      const formattedColor = formatHex(color);
-
       return (
         <SwatchContainer
           key={index + '-' + color}
-          className={`wdc-color-picker--color-${formattedColor}`}
```
lychyi, line 78: This was treated as sort of a `data-testid` so I'm removing it here. There's no css associated with this class name. Hence, the cleaning up of Cypress tests in this PR rather than #831

modules/_labs/color-picker/react/stories/stories_ColorPicker.tsx
```diff
@@ -36,6 +36,7 @@ storiesOf('Labs|Color Picker/React', module)
         <IconButton
           icon={bgColorIcon}
           aria-label="Select Background Color"
+          data-testid="open"
```
lychyi, line 39: This makes grabbing it in Cypress a tad easier.

cypress/integration/ColorPicker.spec.ts
```diff
@@ -1,15 +1,11 @@
...
-const expandHex = (hex: string) => {
-  const shorthandRegex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i;
-  return hex.replace(shorthandRegex, function(m: string, r: string, g: string, b: string) {
-    return r + r + g + g + b + b;
-  });
-};
+const getOpenButton = () => cy.get('[data-testid=open]');
+const getResetButton = () => cy.contains('button', 'Reset');
+const getSubmitButton = () => cy.get('[aria-label=Submit]');
+const getSwatch = (color: string) => cy.get(`div[color="${color}"]`);
```
lychyi, line 8: Still thinking about this one 🤔 

It is kind of brittle because we don't technically need to pass `color` as a prop to the styled `Container` element ([see code](https://github.com/lychyi/canvas-kit/blob/remove-data-attr/modules/color-picker/react/lib/parts/ColorSwatch.tsx#L14-L47)) but we do today so this selector works "as is".

However, one thought that came to mind is that these swatches aren't exactly accessible. The swatches should probably have an `aria-label` attr with the color name or the hex value. This would also solve the selector problem.

However, if we're going to go down that rabbit hole, it'd probably be more prudent to make another PR to address a11y in one shot.

NicholasBoll: Agreed

cypress/integration/ColorPicker.spec.ts
```diff
@@ -129,48 +125,52 @@ describe('ColorPicker', () => {
...
-          cy.get('button').click();
-          cy.get(`.wdc-color-picker--color-0875e1`)
-            .find('.wd-icon')
-            .should('exist');
+          beforeEach(() => {
+            getResetButton()
+              .then($button => {
+                resetColor = $button.find('[color]').attr('color');
+              })
+              .click();
```
lychyi, line 150: Not sure if there's a better way to do this. I want to store the value of the reset color from the yielded button before I click on it. @NicholasBoll any thoughts here?

cypress/integration/ColorPicker.spec.ts
```diff
@@ -194,32 +194,38 @@ describe('ColorPicker', () => {
...
+        const color = '#ff5347';
+
         beforeEach(() => {
-          cy.get(`.wdc-color-picker--color-${color}`).click();
+          getSwatch(color).click();
         });
 
         it('should update the color input to the swatch color', () => {
-          getColorInput().should('have.value', color.toUpperCase());
+          getColorInput().should('have.value', color.slice(1).toUpperCase());
```
lychyi, line 204: I don't love the whole `toUpperCase()` portion but I'm also trying to keep this PR small. If I keep tugging here, I may end up changing the stories that we're testing on.

cypress/integration/ColorPicker.spec.ts
```diff
@@ -194,32 +194,38 @@ describe('ColorPicker', () => {
...
         });
 
         it('should have check icon', () => {
-          getColorInput().click();
-          cy.get(`.wdc-color-picker--color-${color}`)
+          getColorInput().focus();
```
lychyi, line 208: Not sure if this matters too much, it was just odd that we used `focus()` to open the picker and then `click()` to re-open it.

cypress/integration/ColorPicker.spec.ts
```diff
@@ -194,32 +194,38 @@ describe('ColorPicker', () => {
...
         });
-      });
 
-      context('when color reset is clicked', () => {
-        beforeEach(() => {
-          cy.get(`.wdc-color-picker--color-8660d1`).click();
-          getColorInput().click();
-          cy.get('[data-testid="color-picker-reset"]').click();
-        });
+        context('when color reset is clicked', () => {
```
lychyi, line 214: Nested this one a level deep, we shouldn't have to click on a swatch again when we can just put it under the `when swatch is clicked` context

cypress/integration/ColorPicker.spec.ts
```diff
@@ -129,48 +125,52 @@ describe('ColorPicker', () => {
...
         });
-      });
 
-      context('when color reset is clicked', () => {
-        beforeEach(() => {
-          cy.get(`.wdc-color-picker--color-8660d1`).click();
-          cy.get('button').click();
-          cy.get('[data-testid="color-picker-reset"]').click();
-        });
+        context('when color reset is clicked', () => {
```
lychyi, line 141: Nested this one a level deep, we shouldn't have to click on a swatch again when we can just put it under the `when swatch is clicked` context.

Did this below in the "popup" version of this test.

cypress/integration/ColorPicker.spec.ts
NicholasBoll: I see this type of thought a lot for testing, but I tend to disagree. I think if the data is known, we should assert against that known data, removing the need for a variable. A variable means there is logic inside the tests and that logic is often itself untested (although often implicitly tested).

This comes down to the idea that we are making specifications for our components and patterns. A variable is more flexible - the reset color can change without the test changing. But our specifications are about what we expect of our code. If the reset color changes, so has our specifications of the reset color.

I believe variables should almost always be avoided. Most of the use-cases with variables involve a lack of control of the initial state of a system (like seeding database data and getting guids for domain objects, or not being able to accurately seed data and dealing with "random" initial data). I don't think this variable falls into a lack of control of initial state.

A way around a variable in this case is that the reset color value is hard-coded. I don't think the point of a good test is to never fail, but fail only when we expect it to fail and it being easy to understand why and how it fails.

PR #839, opened 2020-08-24
docs(segmented-control): Remove reference to labs

lychyi:


mannycarrera4:
do we need to change this for any previous releases? or is it in the latest that it's the issue?

lychyi:
@mannycarrera4 Just doc changes here. I'm making a (pretty decent?) assumption that people look at the latest docs on master

Issue #779, opened 2020-07-15
text-area: Add Cypress test against placeholder story


NicholasBoll:
We test against these stories:
* https://workday.github.io/canvas-kit/?path=/story/components-inputs-textarea-react-left-label--default
* https://workday.github.io/canvas-kit/?path=/story/components-inputs-textarea-react-left-label--alert
* https://workday.github.io/canvas-kit/?path=/story/components-inputs-textarea-react-left-label--error
* https://workday.github.io/canvas-kit/?path=/story/components-inputs-textarea-react-left-label--disabled

But we don't test against the placeholder story:
* https://workday.github.io/canvas-kit/?path=/story/components-inputs-textarea-react-left-label--placeholder

The test file is found here: https://github.com/Workday/canvas-kit/blob/master/cypress/integration/TextArea.spec.ts

PR #841, opened 2020-08-26
test: Add cypress test for placeholder in input and text area component

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/779

Added test to assert the placeholder attribute exist in the text area component as well as the text input component

### DIFF
--- a/cypress/integration/TextArea.spec.ts
+++ b/cypress/integration/TextArea.spec.ts
@@ -66,5 +66,10 @@ describe('Text Area', () => {
     it('should render a placeholder text', () => {
       getTextArea().should('have.attr', 'placeholder', 'Placeholder');
     });
+
+    it('should reflect the text typed', () => {
+      getTextArea().type('Test');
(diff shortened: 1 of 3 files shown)


cypress/integration/TextArea.spec.ts
```diff
@@ -53,4 +53,18 @@ describe('Text Area', () => {
...
+    });
+
+    it('should pass accessibility checks', () => {
+      cy.checkA11y();
+    });
+
+    it('should render a placeholder text', () => {
+      getTextArea().should('have.attr', 'placeholder', 'Placeholder');
+    });
+  });
```
mannycarrera4, line 69: Curious if we should add another test here to assert when you type into the input or text area, it should display that text? thoughts @NicholasBoll ?

NicholasBoll: Yes, the tests against the other stories include making sure the input gets focus as well as typing in it and making sure the value changes

PR #844, opened 2020-09-01
fix(select): Replace disabled attribute on options with aria-disabled

jamesfan:
## Summary

Previously, the `disabled` attribute was incorrectly being applied to disabled options; options are represented as `<li>` elements, and `disabled` is not a valid attribute for an `<li>`. Workday Accessibility advised that we set `aria-disabled="true"` on the disabled option instead.

This PR also includes a small tweak that prevents `aria-required` from being unnecessarily set to `"false"` on the Select menu if `aria-required` or `required` was never passed as a prop in the first place (i.e., render `<ul role="listbox">` instead of `<ul role="listbox" aria-required="false">`).

## Checklist

### DIFF
--- a/modules/_labs/select/react/lib/SelectBase.tsx
+++ b/modules/_labs/select/react/lib/SelectBase.tsx
@@ -436,7 +436,7 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
           <SelectMenu
             aria-activedescendant={options[focusedOptionIndex].id}
             aria-labelledby={ariaLabelledBy}
-            aria-required={ariaRequired || required}
+            aria-required={ariaRequired || required ? true : undefined}
             buttonRef={buttonRef}
             id={this.menuId}
             error={error}
--- a/modules/_labs/select/react/lib/SelectBase.tsx
+++ b/modules/_labs/select/react/lib/SelectBase.tsx
@@ -333,8 +333,8 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
 
     return options.map((option, index) => {
       const optionProps = {
+        'aria-disabled': option.disabled ? true : undefined,
         'aria-selected': selectedOptionIndex === index ? true : undefined,
-        disabled: option.disabled,
         error,
         focused: focusedOptionIndex === index,
         id: option.id,
--- a/cypress/integration/SelectLabs.spec.ts
+++ b/cypress/integration/SelectLabs.spec.ts
@@ -365,6 +365,14 @@ describe('Select', () => {
           .type('{downarrow}');
       });
 
+      context('the "Carrier Pigeon" option', () => {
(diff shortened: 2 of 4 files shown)


modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -436,7 +436,7 @@ export default class SelectBase extends React.Component<SelectBaseProps> {
           <SelectMenu
             aria-activedescendant={options[focusedOptionIndex].id}
             aria-labelledby={ariaLabelledBy}
-            aria-required={ariaRequired || required}
+            aria-required={ariaRequired || required ? true : undefined}
```
NicholasBoll, line 439: Nice. I was looking to make sure these properties were `undefined` or `true`. The recommendation is to never set to `false`

mannycarrera4: ah, just curious why never false vs undefined?

jamesfan: I'm not sure this is what Nicholas was referring to, but in this case I didn't want `aria-required="false"` to be unnecessarily applied to the `<ul>` rendered by `SelectMenu` (more concise to just omit `aria-required` entirely). Given [the way logical operators work in JavaScript](https://mariusschulz.com/articles/the-and-and-or-operators-in-javascript#using-logical-operators-with-non-boolean-values),  `ariaRequired || required` resolves to `false` if `ariaRequired` is falsy and `required` is explicitly set to `false`.

Therefore, if `ariaRequired` is `undefined` and `required` is `false`, `<SelectMenu aria-required={ariaRequired || required}>` resolves to `<SelectMenu aria-required={false}>`, resulting in `<ul aria-required="false">`. In order to get rid of the unnecessary `aria-required="false"`, I forced the `aria-required` prop to be set to `true` or `undefined` (but never `false`).

Note there are [certain attributes](https://stackoverflow.com/a/31164090) React will omit if they're set to `false` (or any other falsy value), but `aria-required` isn't one of them.

alanbsmith: Right. Undefined removes the attr altogether, which is what you'd want here. Nice work!

NicholasBoll: I wasn't able to find concrete documentation on omitting attributes vs explicit `attribute="false"`. That's a recommendation from accessibility. I think some screen readers get confused and treat `attribute="false"` as a `true` value. Many attributes don't technically have a value, just a presence (like `<input required>`. Attributes can only be strings and all strings resolve as "truthy" in JavaScript.

I think it is just safer to omit attributes for the false case and have `="true"` for the true case unless the specification explicitly says to set it to `"false"` and the accessibility team agrees.

Issue #836, opened 2020-08-19
Separate Popup from ColorPicker in Labs


lychyi:
## 🐛 Bug Report

It's not a bug per se, but we are erroneously shipping a popup with the ColorPicker. That's not really how it should be done. The Popup in [this file](https://github.com/Workday/canvas-kit/blob/master/modules/_labs/color-picker/react/lib/ColorPicker.tsx#L156) should be separate. What if the `ColorPicker` needs to be rendered in another type of container?

This is a labs component so it wouldn't be breaking. But a user could expect to make a change like this for it to continue working:

### From
```tsx
<Button>I open a Color Picker</Button>
<Popper>
  <ColorPicker />
</Popper>
```

### To
```tsx
<Button>I open a Color Picker</Button>
<Popper>
  <Popup>
    <ColorPicker />
  </Popup>
</Popper>
```

PR #846, opened 2020-09-02
fix(color-picker): Separate popup from color picker

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/836

## BREAKING CHANGE
• Remove the popup from color picker
• Color picker now renders just the swatches
• If you want to render the color picker in a popup, you must wrap it in a popup component


modules/_labs/color-picker/react/lib/ColorPicker.tsx
```diff
@@ -94,6 +94,10 @@ const defaultColors = [
   colors.blackPepper400,
 ];
 
+const ColorPickerContainer = styled('div')({
+  width: 216,
```
mannycarrera4, line 98: Not sure how we want to handle this, at this set width the columns align perfectly, should we make this a prop or add an explanation in the code as to why it;s this set width?

modules/_labs/color-picker/react/stories/stories_ColorPicker.tsx
```diff
@@ -5,7 +5,8 @@ import {action} from '@storybook/addon-actions';
 import withReadme from 'storybook-readme/with-readme';
 import {ColorInput} from '@workday/canvas-kit-react-color-picker';
 import {colors} from '@workday/canvas-kit-react-core';
-import {Popper} from '@workday/canvas-kit-react-popup';
+import {TransformOrigin, getTranslateFromOrigin} from '@workday/canvas-kit-react-common';
```
mannycarrera4, line 8: remove

NicholasBoll:
How should we message this change? It would be a breaking change to anyone using the popup pattern.

NicholasBoll (approved):

mannycarrera4:
Should i add something to the readme and more detail to the Pull request info?

NicholasBoll:
We haven't really broken labs much. I think a warning in the pull request description for sure. We'll want to add something to the squash commit message. We don't have an SLA of semantic versioning for labs, but we can still add breaking change messages. People would appreciate the heads-up.

Issue #847, opened 2020-09-03
Update Icon Button to support multiple accents


seans-cummin:
## 🐛 Bug Report

Icon buttons (toggled on) poorly display system icons that feature multiple accents, most notably on active states. Example of an icon with this issue: wd-icon-c-bar-clustered

## To Reproduce

Insert wd-icon-c-bar-clustered into a  toggled state icon button
https://workday.github.io/canvas-kit/?path=/story/testing-react-buttons-button-icon-button--icon-button-states

## Expected Behavior

A system icon with multiple accents like wd-icon-c-bar-clustered should behave like an icon with just a fill.

## Actual Results

A system icon with multiple accents like wd-icon-c-bar-clustered will only display different icon states using the fill, leaving the accent colors unaccounted for (see Image).

<img width="743" alt="Screenshot 2020-09-03 at 11 27 41" src="https://user-images.githubusercontent.com/64029458/92103928-89de7d80-edd8-11ea-9892-e6017bf5db50.png">

## Clarification on the problem
See https://github.com/Workday/canvas-kit/pull/848#discussion_r492398824

mannycarrera4:
This issue is related: https://github.com/Workday/canvas-kit/issues/793

PR #848, opened 2020-09-03
fix(button): Fix accent color on icon button when toggled on

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/847


modules/button/react/lib/parts/ButtonContainer.tsx
```diff
@@ -59,7 +59,14 @@ function getIconColorSelectors(
         fill: fill
           ? color === themePrimary.contrast
             ? themePrimary.main
-            : themePrimary.contrast
+            : themePrimary.contract
+          : color,
+      },
+      '.wd-icon-accent:nth-of-type(2), .wd-icon-accent2:nth-of-type(2)': {
```
mannycarrera4, line 65: This seems to do the trick, but I'm not convinced of the solution, maybe we need to pass in an accent prop?

lychyi:
@mannycarrera4 is this still in draft or can we review?

mannycarrera4:
@lychyi i left it in draft because i wasn't sure of the approach, but I would love a review!

jamesfan:
I have some comments on this PR, just trying to frame my thoughts (it was more complicated than I expected). I'll post them shortly.

modules/button/react/lib/parts/ButtonContainer.tsx
```diff
@@ -55,13 +55,20 @@ function getIconColorSelectors(
...
             : themePrimary.contrast
           : color,
       },
+      '.wd-icon-accent:nth-of-type(2), .wd-icon-accent2, .wd-icon-accent2:nth-of-type(2)': {
+        fill: fill
+          ? color === themePrimary.contrast
+            ? themePrimary.contrast
+            : themePrimary.main
+          : color,
+      },
```
jamesfan, line 71: This PR/Issue has three sources of complexity which made it deceptively challenging to review 🤯 :
1. The complexity of buttons.
2. The complexity of icons due to their multiple layers and the rules for coloring those layers.
3. The complexity of the `nth-of-type` CSS selector, which is being used here to match an element based on its _class_ name rather than its _tag_ name, which seems like [the way the selector is meant to be used](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-of-type). Manny's solution seems to work though (_almost_, more on that later) given [how browsers interpret `.className:nth-of-type`](https://codesandbox.io/s/nth-of-type-bs982?file=/index.html).

The expected behavior in #847 reads:

> A system icon with multiple accents like wd-icon-c-bar-clustered should behave like an icon with just a fill

Put another way (assuming I'm reading it correctly):

> If a system icon has multiple accent layers, the fill and accent layers should all be the same color (when the icon button is toggled on). If the icon only has a single accent layer, leave it the way it is now (the current logic will often set fill and accent to different colors when the icon button is togged on).

The fix above almost works, but it relies on all multi-accent icons having a specific structure: `.wd-icon-accent` has to be the second `<path>` child of its parent. For example:

```
// from cBarClustered (all layers are colored correctly)
<g class="wd-icon-container">
  <path class="wd-icon-fill"></path>
  <path class="wd-icon-accent"></path>
  <path class="wd-icon-accent2"></path>
</g>
```

Almost every multi-accent icon (there aren't many, they all appear to be in the [Charts section](https://design.workday.com/tokens/assets/icons/system-icons)) follows this structure, except for `cWaterFall` where the `.wd-icon-background` throws things off:

```
// from cWaterfall (wd-icon-accent is colored incorrectly)
<g class="wd-icon-container">
  <path class="wd-icon-background"></path>
  <path class="wd-icon-fill"></path>
  <path class="wd-icon-accent"></path>
  <path class="wd-icon-accent2"></path>
</g>
```

Also, this approach breaks several _single_-accent icons where `.wd-icon-accent` happens to be the second `<path>` child of its parent:

```
// from clock (wd-icon-accent is now colored incorrectly)
<g class="wd-icon-container">
  <circle class="wd-icon-background"></circle>
  <path class="wd-icon-fill"></path>
  <path class="wd-icon-accent"></path>
</g>
```

All this to say we need a less fragile solution 😄 . It'd be nice if there was a CSS selector that let us style `.wd-icon-accent` only if `.wd-icon-accent2` was a sibling, but I haven't found such a selector. We need something like [the general sibling selector](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator) which selects the first sibling instead of the second sibling.

mannycarrera4: I couldn't have described it any better! Thanks for taking the time to decipher this and understand it. This is why i kept this PR in draft, because it seemed like a fragile way to address MOST of the problem and even I wasn't sure of it. I originally thought that passing another parameter to getIconButtonColors would do the trick, but it doesn't....

mannycarrera4:
Closing this PR for now. There's some additional complexity in icons and how they're colored that's it's a bit harder than changing a few styles. @jamesfan noted the different layers. Will try to revisit this when I have a bit more time.

PR #853, opened 2020-09-11
fix(form-field): Fix legend overflow bug (IE11)

### LINKED ISSUE #852: Text in FormField legend do not wrap.
## 🐛 Bug Report

Our `FormField` component renders a `legend` element (in place of `label`, when dealing with radio groups for example) which has some interesting properties in IE11. Namely that the text overflows the layout.

Adding `width: 100%` to our `legend` should fix this.

## To Reproduce

* In IE11, create a `FormField` with `useFieldset` `true`.
* Use a really, really long label.

## Expected Behavior

The label text should wrap on IE11.

## Actual Results

The label text does not wrap on IE11.

## Browser (if applicable)

IE11

https://codesandbox.io/s/long-label-on-radio-group-kxgk0 (link doesn't quite work in IE11 but you can open up the iframe output directly in IE11)

lychyi:
## Summary

This fixes #852 . When a `FormField` renders a `legend` as its label (i.e. when the `useFieldset` prop is true) and long-form text overflows the parent container's width. This doesn't happen in Chrome or FF.

Adding `width: 100%` on the `legend` element fixes this bug but can only be applied to top-labeled `FormField`s. If it's applied to the side-labeled fields then it will interfere with the `min-width` property for that label position.

## Checklist

- [x] `yarn test` passes

## Additional References

Does this on IE11:
image

Should do this on all browsers:
image

### DIFF
--- a/modules/form-field/react/lib/Label.tsx
+++ b/modules/form-field/react/lib/Label.tsx
@@ -72,8 +72,12 @@ const RequiredAstrisk = styled('abbr')({
   textDecoration: 'unset',
 });
 
-// Used inside the fieldset component instead of a label for accessible radio groups
-const LegendComponent = styled('legend')<LabelProps>(...labelStyles);
+// Used inside the fieldset component instead of a label for accessible radio groups.
+// Extra `width: 100%` style is to help IE11 wrap text properly in a top-positioned
+// FormField's legend element.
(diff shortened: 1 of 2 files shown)


NicholasBoll (approved):
Will this be okay as a 4.2.0? Or do we need to rebase for a 4.1.2 release?

PR #865, opened 2020-09-25
fix(popup-stack): Move adapter var to popup-stack global namespace

csongnguyen:
## Summary

Previously, if there were multiple imports (from different bundles) of the PopupStack, the adapter would be overwritten each time the PopupStack was imported. While the PopupStack can be overridden, the adapter should not be (otherwise, createAdapter becomes a bit useless 😅 ) 

## Checklist

- [x] `yarn test` passes
      applicable

### DIFF
--- a/modules/popup-stack/lib/PopupStack.ts
+++ b/modules/popup-stack/lib/PopupStack.ts
@@ -111,6 +111,7 @@ interface Stack {
     max: number;
     getValue: typeof getValue;
   };
+  _adapter: Partial<typeof PopupStack>;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


NicholasBoll (approved):
The issue is a design decision that there should only ever be a single stack per page, but there could be multiple versions of an API accessing that stack. This design decision doesn't work well with an adapter present. Without this change, each instance of the PopupStack API would define its own adapter instance vs a global one. That means each instance of the PopupStack API would be adapted separately which is most likely not what we want.

This PR fixes that making the adapter instance shared with all PopupStack API instances. There could be an issue where difference versions of the adapter could compete with each other where the last _run_ call to `createAdapter` would be the actual adapter run on the page.

PR #866, opened 2020-09-28
🚀 New Component: SidePanel (Labs)

### LINKED ISSUE #778: Update SidePanel to match new design specifications
## 🚀 Feature Proposal

Update side panel container to match design updates.

- Side Panel header to include optional accent icon to left of title
- Expand/collapse icon moved from bottom to top to meet A11y requirements
- Expand/collapse and title positioning rules depending on 1 or 2 lines of text
- Update expand/collapse Side Panel animation
- Collapsed side panel should also be Soap 100 on collapse

## Motivation

These updates to the side panel container are to create a more cohesive 

## Example

Screen Shot 2020-07-15 at 8 50 11 AM

lychyi:
## Summary

While, Canvas Kit already has a [Side Panel](https://workday.github.io/canvas-kit/?path=/story/components-containers-side-panel-react--default), we are unifying it as an internal initiative, starting with the container itself. This means that the existing side panel would have had to change significantly. Because of that we've decided to use the Labs namespace to build this new one and eventually deprecate the original one.

### Basics
The side panel is made up of 3 parts:
* The **panel** (a container that is expandable)
* A **control** (a button or something similar that controls the panel)
* A **label** (an arbitrary element that contains the accessible name of the panel)

It also includes a hook that wires the three together `useSidePanel()`. It returns an `expanded` state getter and setter (like any other `useState` hook) and 3 sets of prop objects meant to be spread to their associated parts: `panelProps`, `labelProps`, and `controlProps`.

The hook itself takes a config object that allows you to set: `initialExpanded` (initial state of the **panel**), `panelId`, and `labelId` (if either is defined it will be a string id, otherwise, it is uniquely generated). It can also take no arguments and return a default set of props and expanded state.

The component also ships 1 default sub-component for the **control**: `SidePanel.ToggleButton`. This is an `IconButton` with a default icon and is aware of whether it's a left-handed or right-handed side panel and positions the button accordingly.

In our docs and examples, we recommend that a hidden label be used as an `aria-labelledby` target to make sure all parts of this side panel are accessible and read well on a screen reader.

### Usage
```tsx
const {panelProps, labelProps, controlProps} = useSidePanel({ initialExpanded: false, panelId: 'side-panel-1'});

// somewhere in your JSX
<SidePanel {...panelProps}>
  <SidePanel.ToggleButton {...buttonProps} />
  <span hidden>Accessible Label for Panel</span>
</SidePanel>
```

### Notable differences to the current Side Panel
* Style changes in the form of two variants. Button is located on top rather than on bottom.
* This is _just_ the container, not menu items or anything that goes into a standard side panel. Other components will follow.
* The API is focused on composability via JSX and hooks rather than config props
* A user can set the collapsed and expanded widths but they are configurable
* This side panel can be used in both left and right side orientations. (using the `origin` prop)
* Animations trigger "in-between states". So we give the user two callbacks. One will alert them when the `expanded` prop changes (`onExpandedChange`) while the other is called when the panel transitions between `'collapsed' | 'collapsing' | 'expanded' | 'expanding'` (`onStateTransition`). The latter is useful if the side panel contents need to know when things are `'collapsing'` or `'expanding'`.

### Open Questions
* Not sure what are good names for two variants. One is more prominent than the other internally but there could be more variants in the future. Today it's "standard" and "alternate", not creative, I know.
* I'm using a hack to basically short circuit animation on the first render. This was found in [Material UI](https://github.com/mui-org/material-ui/blob/d9ca53dcdcab99b235ef5954f6d5389438362e69/packages/material-ui/src/Drawer/Drawer.js#L126-L132) but it basically prevents the component from animating when it "mounts" for the first time. Any potential downsides? Is there any way that React forces a re-render without the user specifying it?

## Checklist

- [x] `yarn test` passes
      applicable

## Additional References

[Internal-only Figma Specs](https://www.figma.com/file/pywFr7N2a9zWQH7jK6Jkaq/%F0%9F%9B%A0-Containers---Side-Panel?node-id=76%3A0)

[Original side panel codebase](https://github.com/Workday/canvas-kit/tree/master/modules/side-panel/react)

Closes #778

Blocked by #896


yarn.lock
```diff
@@ -2508,6 +2508,70 @@
     mkdirp "^0.5.1"
     rimraf "^2.5.2"
 
+"@material-ui/core@^4.9.7":
```
NicholasBoll, line 2511: @lychyi I'm not sure why this dependency was added.

modules/common/react/lib/utils/useUniqueId.ts
```diff
@@ -7,8 +7,14 @@ import uuid from 'uuid/v4';
...
-  const [generatedId] = React.useState(() => uuid().replace(/^[0-9]*/gi, ''));
-  return id || generatedId;
+  const [effectiveId] = React.useState(() => {
+    if (typeof id !== 'undefined') {
+      return id;
+    }
+    return uuid().replace(/^[0-9]*/gi, '');
+  });
+
+  return effectiveId;
```
NicholasBoll, line 17: I'm curious about this change. It looks like the only behavior difference is not calling `uuid` unless the `id` is `undefined` vs always calling `uuid` and determining the truthiness of `id`.

yarn.lock
lychyi: good catch, not sure why any of these are added. I'll checkout yarn.lock from master and try again

modules/common/react/lib/utils/useUniqueId.ts
lychyi: Yea, that's the gist of it. It seemed wasteful to compute a uuid even when we want to just pass the id through if it's defined.

alanbsmith (changes requested):
Looks good, @lychyi! Thanks for working on this. 👍  I left some comments and questions and only one change requested. Let me know if anything is unclear. 😄

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,243 @@
...
+  onAnimationStart,
+  onCollapsedChange,
+  onStateChange,
+  origin = 'left',
+  variant = 'standard',
+  width = 320,
+  ...elemProps
+}: SidePanelProps) => {
+  const mounted = React.useRef(false);
+  const [collapsed, setCollapsed, isControlled] = useControllableState<typeof collapsedProp>(
```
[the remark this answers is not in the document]
alanbsmith: You could also call this "oscillation" now that I think about it.

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,243 @@
...
+      ]}
+      onAnimationEnd={handleAnimationEnd}
+      onAnimationStart={handleAnimationStart}
+      {...elemProps}
+    >
+      {typeof collapsedProp === 'undefined' ? (
+        <IconButton
+          variant={IconButton.Variant.CircleFilled}
+          icon={transformationImportIcon}
+          aria-label={`expand or collapse side panel`}
```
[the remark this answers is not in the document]
lychyi: Ah yikes, this was remnants of my code sandbox. I should have had a todo here.

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,243 @@
...
+    }
+  };
+
+  const handleAnimationStart = (event: React.AnimationEvent<HTMLDivElement>) => {
+    if (event.currentTarget === event.target) {
+      if (
+        event.animationName === motion.collapse.name ||
+        event.animationName === motion.expand.name
+      ) {
+        setInternalState(collapsed ? 'collapsing' : 'expanding');
```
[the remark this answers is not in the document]
lychyi: Yup!
image

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,243 @@
...
+            : undefined,
+        },
+        containerVariantStyle[variant],
+      ]}
+      onAnimationEnd={handleAnimationEnd}
+      onAnimationStart={handleAnimationStart}
+      {...elemProps}
+    >
+      {typeof collapsedProp === 'undefined' ? (
+        <IconButton
```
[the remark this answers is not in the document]
lychyi: I was back and forth on this one. But eventually I just figured that this was a good enough first pass and we should get this out there and see what people think.

Here's how it played out in my head. The `IconButton` look (icon, variant, and position) is something we aim to specify and/or standardize as a design system so that was a fixed option. However, a user can most certainly use a custom `IconButton` as one of the children when it's a controlled component.

In other words, because Side Panel consistency has been such a hot topic internally, I wanted to make the standard choice easy, and the non-standard choice doable with ease being less of a factor.

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+export interface SidePanelProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * Specifies the _controlled_ state the side panel is in. Leave undefined if you want to keep this an _uncontrolled_ component.
+   * Changing this prop will fire an animation that resolves to this state. For example, if collapsed is changed from `true` to `false`,
+   * an expand animation will fire and the side panel will eventually be expanded when the animation is finished.
+   *
+   * @default false
+   */
+  collapsed?: boolean;
```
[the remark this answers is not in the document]
lychyi: I hear you on this one. The main problem here is that the `collapsed` state actually denotes a final resting state or any state that will eventually collapse the side panel. It's a tough naming convention.

Another idea is that the states could be called `opened` or `closed` instead. And as an example, `'expanded' -> 'collapsing' -> 'collapsed'` could be the steps it goes through to transition from `opened` to `closed`. It's not semantically accurate but our current `SidePanel` uses `open` as its term today so I don't think it's confusing to anyone.

modules/common/react/lib/utils/useUniqueId.ts
lychyi: Yes, you're right about that. Let me revert this change, I feel like it's not really related to this PR.

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,256 @@
...
+  };
+
+  const handleAnimationEnd = (event: React.AnimationEvent<HTMLDivElement>) => {
+    if (event.currentTarget === event.target) {
+      if (
+        event.animationName === motion.collapse.name ||
+        event.animationName === motion.expand.name
+      ) {
+        setState(expanded ? 'expanded' : 'collapsed');
+      }
```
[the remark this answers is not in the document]
alanbsmith: I also feel a little hesitant about the string names for state, but because the state is typed to `SidePanelTransitionStates`, I think it's okay.

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,256 @@
...
+    right: context.state === 'collapsed' ? 0 : context.origin === 'left' ? spacing.s : undefined,
+    left: context.state === 'collapsed' ? 0 : context.origin === 'right' ? spacing.s : undefined,
+    margin: context.state === 'collapsed' ? 'auto' : undefined,
+    transform:
+      context.state === 'collapsed' || context.state === 'collapsing'
+        ? `scaleX(${context.origin === 'left' ? '1' : '-1'})`
+        : `scaleX(${context.origin === 'left' ? '-1' : '1'})`,
+  });
+
+  // @ts-ignore aria-label type error here. The will decide to use aria-label or aria-labelledby
```
alanbsmith, line 251: RC: I think "The will decide" should be "The hook will decide."

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,256 @@
...
+  alternate: {
+    backgroundColor: colors.frenchVanilla100,
+    ...depth[3],
+  },
+  standard: {
+    backgroundColor: colors.soap100,
+  },
+};
+
+const Panel = styled('section')<Pick<SidePanelProps, 'as'>>({
```
alanbsmith, line 99: Question: Do we want to always use our `styled` function, or only when RTL and theming will be affected?

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,256 @@
...
+  overflow: 'hidden',
+  position: 'relative',
+  boxSizing: 'border-box',
+});
+
+export const SidePanelContext = React.createContext({
+  state: 'expanded',
+  origin: 'left',
+});
```
alanbsmith, line 108: Devil's advocate: 👿 Since this context is only being used in the `ToggleButton`, and `ToggleButton` is already taking `{...controlProps}`, could this just live in the hook props? Is context more helpful here?

alanbsmith (changes requested):
Looks good, @lychyi! I left some small comments and questions, but nothing major. I think this mostly needs Cypress tests, and then it'll be ready to go. 🚀

modules/_labs/side-panel/react/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+Default: `'100%'`
+
+#### `origin: 'left' | 'right'`
+> Which side the side panel is meant to originate from.
+
+Default: `'left'`
+
+#### `onExpandedChange: (expanded?: boolean) => void`
+> The function called when the side panel's `expanded` state changes. States like `'collapsing'` and `'expanding'` are tracked in another callback: `onStateChange`
```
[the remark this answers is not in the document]
lychyi: No conventions here for the weird internal state transition callback so it'll be flexible if we decide to change it. Here's the explanations for the names:
* `onStateTransition` is stating that this cb fires on internal state transitions
* `onExpandedChange` is the general React convention `on<Value>Change` so to disambiguate, I wrote `onStateTransition` rather than `onStateChange`... because `state` isn't a value/prop that's changing on the component necessarily.

modules/_labs/side-panel/react/README.md
```diff
@@ -0,0 +1,131 @@
...
+  <CustomControl {...controlProps} />
+  {/* If you already have an element that can be an accessible name for the panel */}
+  <h3 {...labelProps}>Example Panel</h3>
+</SidePanel>
+```
+
+## Static Properties
+
+#### `ToggleButton`
+> `<SidePanel.ToggleButton>` is a control that is meant to toggle between `expanded = true` and `expanded = false` states. It must be used within the `SidePanel` component as a child. Use in conjunction with `useSidePanel`'s `controlProps`, otherwise it does not come with explicit `onClick` handlers.
```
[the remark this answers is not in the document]
lychyi: Yea it will still take `aria-label`, it's just optional now. Honestly, we should change the base `IconButton` to take either `aria-label` or `aria-labelledby` anyway.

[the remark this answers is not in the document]
lychyi: Tracked here: https://github.com/Workday/canvas-kit/issues/894

modules/_labs/side-panel/react/lib/SidePanel.tsx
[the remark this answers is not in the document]
lychyi: I thought we were going to consider the transition to unions of string literals in favor of enums anyway right?

modules/_labs/side-panel/react/lib/SidePanel.tsx
lychyi: Yup, should definitely use ours

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,256 @@
...
+      max-width: ${normalized.from};
+    } to {
+      width: ${normalized.to};
+      min-width: ${normalized.to};
+      max-width: ${normalized.to};
+    }
+  `;
+};
+
+const containerVariantStyle: {[K in SidePanelVariant]: CSSObject} = {
```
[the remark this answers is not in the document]
lychyi: I think what I was going for was to make sure that this object only had keys of the type `SidePanelVariant` (I also now know I can do this with the `Record` utility type). So in the future if we add a variant, we can make sure it gets accessed appropriately in the code. Not sure if `as const` really plays into this unless we're going for type inference on this object.

modules/_labs/side-panel/react/lib/SidePanel.tsx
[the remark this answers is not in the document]
lychyi: btw, _this_ is why we do code review ^ thank you for that 🥇

modules/_labs/side-panel/react/lib/SidePanel.tsx
```diff
@@ -0,0 +1,256 @@
...
+  };
+
+  return (
+    <Panel
+      as={as}
+      css={[
+        {
+          width: expanded ? expandedWidth : collapsedWidth,
+          maxWidth: expanded ? expandedWidth : collapsedWidth,
+          minWidth: expanded ? expandedWidth : collapsedWidth,
```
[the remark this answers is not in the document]
lychyi: It was probably due to me screwing around with height and widths that are not numerical px values. However, in the spirit of not boiling the ocean...

`collapsedWidth` and `expandedWidth` now map to `minWidth` and `width` only.

As discussed behind-the-scenes, `height` will now just be `height: 100%` without `min` or `max` specified, the user can dictate whether or not they want it to overflow or grow the container that it is in (stories have been updated to reflect this). We won't force a preference here. Also, I removed the `height` prop, it can just be set via `style` or emotion's `css` prop.

cypress/integration/SidePanelLabs.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+        switch (story) {
+          case 'AsAside':
+            cy.findByRole('complementary', {name}).should('exist');
+            break;
+          default:
+            cy.findByRole('region', {name}).should('exist');
+        }
+      });
+
+      // TODO: How to test for aria-controls on the button
```
alanbsmith, line 34: Is this a need-to-have or a nice-to-have test?

lychyi: Good question, this is a nice-to-have a11y requirement at this point so I would probably factor it in as a "nice-to-have".

lychyi: Also, this is covered as a unit test when testing the hook.

cypress/integration/SidePanelLabs.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+                  'id',
+                  buttonAriaControlsValue
+                );
+                break;
+              default:
+                cy.findByRole('region', {name}).should('have.attr', 'id', buttonAriaControlsValue);
+            }
+          });
+        });
+      });
```
NicholasBoll, line 45: Really not a fan of logic in test files, especially `switch` statements. I've done this type of thing before in the `forEach` array instead:
```ts
[
  { story: 'Default', role: 'region' },
  { story: 'AsAside', role: 'complementary' },
  { story: 'AsDiv', role: 'region' }
].forEach(({ story, role }) => {
  // story is the same as before
  // role is now the role you expect from the story
})
```

Another way to think about it is you create a story with your accessibility specifications. Another story can be used for a specification that applies to a different configuration, but not all specifications have to be run on every variant - especially if that requires `if` or `switch` statements.

cypress/integration/SidePanelLabs.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+        beforeEach(() => {
+          cy.tab();
+        });
+
+        context('the button', () => {
+          it('should be focused', () => {
+            cy.findByRole('button', {name}).should('have.focus');
+          });
+        });
+      });
```
NicholasBoll, line 67: This one is a little weird. There is no explicit starting point for focus.

What's the point of it? Are we specifying that the button should be focusable? The modal specs use `cy.tab` after some action where there is a specification for where focus should go and that focus is controlled within the modal, but nothing on the original trigger button.

If the point is to specify that the button should be focusable, maybe we should add a `.should('be.focusable')` that runs checks to verify that the element in question can receive focus explicitly. As this is written, focusability is indirectly measured through `cy.tab`. But without a starting point, `cy.tab` is over verifying.

cypress/integration/SidePanelLabs.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+        switch (story) {
+          case 'AsAside':
+            cy.findByRole('complementary', {name}).should('exist');
+            break;
+          default:
+            cy.findByRole('region', {name}).should('exist');
+        }
+      });
+
+      context('when tabbing', () => {
```
NicholasBoll, line 57:
```suggestion
      context('when the tab key is pressed', () => {
```

Our context blocks explain the user's interaction with the browser. "when tabbing" isn't specific to what interactions is being performed - how many tab presses?

modules/_labs/side-panel/react/lib/SidePanel.tsx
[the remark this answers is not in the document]
NicholasBoll: The WD component team ran into an issue where `CSSProperties` (which we define as `React.CSSProperties & Emotion.CSSObject`) caused performance issues for Typescript compilation. I was able to observe the issue, but not reproduce it. Digging into the types of `CSSObject`, I found this:

```ts
export interface CSSOthersObjectForCSSObject {
  [propertiesName: string]: CSSInterpolation
}

export interface CSSObject
  extends CSSPropertiesWithMultiValues,
    CSSPseudosForCSSObject,
    CSSOthersObjectForCSSObject {}
```

You can see that `CSSObject` is extending from `CSSOthersObjectForCSSObject` which expands all properties to `string`. This gives you autocomplete, but no type safety. `margin` is a suggestion, but `margon` is a perfectly okay mistype.

We think the typecheck slowdown comes from multiple levels of mapped types. `CSSObject` is about 900 properties which could cause slowdowns.

`as const` is a way to narrow values of objects from widened types like `string` to narrower types like `'absolute'`. So `CSSObject` is useful for autocomplete, but will not save you from a mistyped property name.

I asked the WD component team to create an issue with a reproduceable build that shows the slowdown so that we can directly address it. I have experienced periodic slowdowns in the Typescript language service, but I don't know what exactly is causing it. I wouldn't be surprised that mapped types of large interfaces causes performance issues.

A compromise might be to specify which CSSObject properties we intend to use:

cypress/integration/SidePanelLabs.spec.ts
lychyi: We talked about this one offline and created a new story for Cypress that showcases a hypothetical "header" with an Avatar button that could be the last focusable element before going to the Side Panel. When a user hits tab _once_ that should then put focus onto the control button of the panel despite it having other elements within it.

NicholasBoll (approved):

alanbsmith (approved):
🦩

PR #867, opened 2020-09-29
fix(modal): Add exception to modal enter animation

NicholasBoll:
This fixes an issue where the Popup adapter is being used. It will flash because the adapter will remove the containing element for popups and show again. The animation makes detach/attach cycle a visible flashing. This is already added to the Popup.tsx file, we just missed the Modal.

### DIFF
--- a/modules/modal/react/lib/ModalContent.tsx
+++ b/modules/modal/react/lib/ModalContent.tsx
@@ -80,6 +80,7 @@ const Container = styled('div')({
   background: 'rgba(0,0,0,0.65)',
   animationName: `${fadeIn}`,
   animationDuration: '0.3s',
+  // Allow overriding of animation in special cases
   '.wd-no-animation &': {
     animation: 'none',
   },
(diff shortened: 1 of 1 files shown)


modules/modal/react/lib/ModalContent.tsx
```diff
@@ -80,6 +80,9 @@ const Container = styled('div')({
   background: 'rgba(0,0,0,0.65)',
   animationName: `${fadeIn}`,
   animationDuration: '0.3s',
+  '.wd-no-animation &': {
+    animation: 'none',
+  },
```
lychyi, line 85: Do we want any documentation or comments on this like in Popup? This seems like it would be very hard to discover. Or is it meant to be a backdoor that few know?

mannycarrera4:
@NicholasBoll ran me through the code and explained it to me, LGTM 👍

modules/modal/react/lib/ModalContent.tsx
NicholasBoll: It is definitely one of those backdoor things. I added a comment above just like in Popup

mannycarrera4 (approved):

PR #869, opened 2020-10-01
fix: links in README

dschaller:
## Summary

As relative links these resolve within GitHub but from the [storybook page](https://workday.github.io/canvas-kit/?path=/story/welcome-getting-started--page) these all 404.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This project provides a set of components for the Workday Canvas Design System t
 implement user experiences consistent with
 [Workday's design principles](https://design.workday.com/).
 
-<a href="./LICENSE">
+<a href="https://github.com/Workday/canvas-kit/blob/master/LICENSE">
   <img src="https://img.shields.io/badge/license-Apache--2.0-blue.svg" alt="Workday Canvas Kit is released under the Apache-2.0 license" />
 </a>
 <a href="https://travis-ci.org/Workday/canvas-kit">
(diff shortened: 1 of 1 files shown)


willklein:
@dschaller it's so cool you just submitted this, I was just noticing a similar issue with the component status link on the [Welcome -> Getting Started page](https://workday.github.io/canvas-kit/?path=/docs/welcome-getting-started--page). I can see from your PR that it's not just that link, but several that are broken when viewing the Getting Started page.

The approach in your PR would "fix" some of the links so they go directly to the Markdown files, but we should probably figure out how to link to the appropriate page on the Storybook docs as well.

I wish that we could resolve a single link to the appropriate locations on Storybook or the GitHub source files, depending on which context your browsing. For example, there's two usable links for Component Status:

https://workday.github.io/canvas-kit/?path=/docs/welcome-component-status--page
and
https://github.com/Workday/canvas-kit/blob/master/COMPONENT_STATUS.md

I hope this is possible through Storybook, but it may not be there yet. Looks like Canvas Kit is ahead of most and I have to do some more research to see if it's possible. Maybe we need to write an add-on to transform markdown links for the Storybook output, but I'm not sure yet.

If it's not feasible right away, I might lean toward linking the Storybook docs (workday.github.io/canvas-kit/*) over the raw Markdown files on Github.com (github.com/Workday/canvas-kit/blob/master/*).

I'll keep looking into Storybook link capabilities, but what do you think?

NicholasBoll:
Thanks for the PR. The long term plan is to move our documentation off Github/Storybook to a dedicated docs site. We'll still have a Readme with relevant information on what the project is and how to get started and links to the documentation site.

For now, our links are definitely broken when navigating via the Storybook docs and that should be fixed. If we planned to use Storybook long term, we'd probably do something like @willklein suggested (processing in Storybook to rewrite URLs). I'm talking with the team to determine if the URLs should be absolute paths pointing to Github or to Storybook.

PR #875, opened 2020-10-06
feat: Add support for Secondary Outline Button AB test

anicholls:
# Summary

We are going to be AB testing a small color change to our secondary outline buttons, and we need to add some way to override the existing colors. Originally the plan was to add support for the `neutral` palette field in the theme object within the `OutlineButton` component. However, I ran into an issue with this approach:

- The current defaults from `defaultCanvasTheme.palette.neutral` do not map to the colors we currently use within the outline button. 
- If I replace the hardcoded colors within `OutlineButton` (e.g. https://github.com/Workday/canvas-kit/blob/master/modules/button/react/lib/OutlineButton.tsx#L125) with a theme reference (e.g. `theme.canvas.palette.neutral.main`, it would then use the wrong color by default.
- Since the theme has already been "filled" with the default value by this point, there isn't really a clean way to do a check and default to the original hardcoded value.
- **I did not want to update the `neutral` values in the `defaultCanvasTheme` object. While we do not currently use this in any of our components, if another team were using it, it would be a breaking change.**

As an alternative solution, I made use of the scoping we added to our Emotion theme in https://github.com/Workday/canvas-kit/pull/593. I've added hidden support in the secondary outline button for `theme.ABTest.secondaryOutlineButton.*`. This is outside the `canvas` theme scope and is not typed, so consumers should not notice the support unless they're looking at the source code of `OutlineButton`. The maintenance impact is minimal, and it avoids the problems noted above. This can be stripped out once the AB test is complete.

### DIFF
--- a/modules/button/react/lib/OutlineButton.tsx
+++ b/modules/button/react/lib/OutlineButton.tsx
@@ -124,25 +124,25 @@ export const getOutlineButtonColors = (
       return {
         default: {
           background: 'transparent',
-          border: ABTestColors.main || colors.soap500,
+          border: ABTestColors?.main || colors.soap500,
           icon: colors.licorice200,
           label: colors.blackPepper400,
         },
         hover: {
-          background: ABTestColors.dark || colors.licorice500,
-          border: ABTestColors.dark || colors.licorice500,
+          background: ABTestColors?.dark || colors.licorice500,
+          border: ABTestColors?.dark || colors.licorice500,
           icon: themePrimary.contrast,
           label: themePrimary.contrast,
         },
         active: {
-          background: ABTestColors.darkest || colors.licorice600,
-          border: ABTestColors.darkest || colors.licorice600,
+          background: ABTestColors?.darkest || colors.licorice600,
+          border: ABTestColors?.darkest || colors.licorice600,
           icon: themePrimary.contrast,
           label: themePrimary.contrast,
         },
         focus: {
-          background: ABTestColors.dark || colors.licorice500,
-          border: ABTestColors.dark || colors.licorice500,
+          background: ABTestColors?.dark || colors.licorice500,
+          border: ABTestColors?.dark || colors.licorice500,
           icon: themePrimary.contrast,
           label: themePrimary.contrast,
         },
--- a/modules/button/react/lib/OutlineButton.tsx
+++ b/modules/button/react/lib/OutlineButton.tsx
@@ -83,6 +83,8 @@ export const getOutlineButtonColors = (
     canvas: {
       palette: {primary: themePrimary},
     },
+    // TODO: Remove this support after AB test is complete
+    ABTest: {secondaryOutlineButton: ABTestColors},
   } = theme;
 
   switch (variant) {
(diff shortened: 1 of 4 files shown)


utils/storybook/CanvasProviderDecorator.tsx
```diff
@@ -16,6 +17,13 @@ export default makeDecorator({
   wrapper: (storyFn, context, {parameters = {}}) => {
     const theme: PartialEmotionCanvasTheme = {
       canvas: object(label, parameters.theme || defaultCanvasTheme),
+      ABTest: {
+        secondaryOutlineButton: {
+          main: 'orange', //colors.licorice300,
```
anicholls, line 22: I made this super obvious, but commented the intended colors to be used in the ABtest. We can switch to these if preferred, but I was worried the snapshot tests wouldn't catch the small color changes.

NicholasBoll (changes requested):
From my testing this will fail with the following error:
```
TypeError: Cannot read property 'main' of undefined
```

The reason is our Storybook wrapper is providing `ABTest` to the theme object. But a consumer not providing this will have `OutlineButton` fail at the line where it is trying to access `ABTestColors.main`

modules/button/react/lib/OutlineButton.tsx
```diff
@@ -122,25 +124,25 @@ export const getOutlineButtonColors = (
       return {
         default: {
           background: 'transparent',
-          border: colors.soap500,
+          border: ABTestColors.main || colors.soap500,
```
NicholasBoll, line 127:
```suggestion
          border: ABTestColors?.main || colors.soap500,
```

modules/button/react/lib/OutlineButton.tsx
```diff
@@ -83,6 +83,8 @@ export const getOutlineButtonColors = (
     canvas: {
       palette: {primary: themePrimary},
     },
+    // TODO: Remove this support after AB test is complete
+    ABTest: {secondaryOutlineButton: ABTestColors},
```
NicholasBoll, line 87:
```suggestion
    ABTest: {secondaryOutlineButton: ABTestColors} = {},
```

modules/button/react/lib/OutlineButton.tsx
NicholasBoll: Also add this to the rest of the properties

modules/button/react/lib/OutlineButton.tsx
anicholls: Typescript doesn't like this. Your other suggestion should do the trick anyway 👍

NicholasBoll: Unfortunately it isn't as easy as this. Typescript isn't happy with it because `{}` isn't a valid default

NicholasBoll: This or something like it is still needed for runtime.

utils/storybook/CanvasProviderDecorator.tsx
NicholasBoll: Using Storybook doesn't catch the issue where `ABTest` might not be defined on the `theme` object, but because this wraps all our stories, we wouldn't notice that issue. Cypress wouldn't catch this either because Cypress runs in the same context as Storybook.

I think a better place to define this `ABTest` theme override would be in a new `ABTest` story that wraps in a new `<CanvasProvider theme={theme}>` provider. This way the normal `OutlineButton` is proven to work without the override both visually and functionally. The new story can ensure the `ABTest` theme override works correctly both visually and functionally.

The theme override of this wrapper worries me that we might run into issues later. It is nice that unit test catch these errors, but this file make it less obvious there's an issue.

NicholasBoll (approved):

anicholls:
Decided to go with this approach instead:

https://codesandbox.io/s/secondary-outline-button-abtest-forked-m0ep4?file=/src/index.tsx

PR #884, opened 2020-10-12
fix(color-picker): Detect different formats of #ffffff

### LINKED ISSUE #881: Reset Color Swatch doesn't show box-shadow when given different formats of #ffffff (FrenchVanilla100)
## 🐛 Bug Report

If a consumer of colorSwatch passes in #ffffff (colors.frenchVanilla100) a box shadow around the color swatch (https://github.com/Workday/canvas-kit/blob/master/modules/color-picker/react/lib/parts/ColorSwatch.tsx#L30) is created. However, if a user passes #FFFFFF, #fff, or #FFF the box-shadow is not created despite the color being the same. This especially seems strange for the reset color button because if a format other than #ffffff is passed in we do not have any indication of a color swatch next to reset because it all just looks white. 

## To Reproduce

Pass #FFFFFF, #FFF, or #fff into the colorSwatch component and notice that no box-shadow is applied to the swatch.

## Expected Behavior

I would expect that the box shadow should appear when I pass in any format of #ffffff.

## Actual Results

Passing in #FFFFFF, #FFF, or #fff results in a color swatch that blends into the container when the container is also #ffffff.

## Browser (if applicable)

I was on chrome.

### Error Output

image

NicholasBoll:
Fixes #881


modules/color-picker/react/lib/parts/ColorSwatch.tsx
```diff
@@ -11,6 +11,19 @@ export interface ColorSwatchProps extends React.HTMLAttributes<HTMLDivElement> {
...
+/**
+ * Use the browser's built-in mechanism for comparing colors. Works all the way back to IE11. The
+ * browser will take any valid color and convert to something like `rgb(0, 0, 0)` which allows us to
+ * then compare the strings even if the colors are in different formats
+ */
+function compareColors(color1: string, color2: string): boolean {
+  div.style.color = color1;
+  div.style.backgroundColor = color2;
+  return div.style.color === div.style.backgroundColor;
+}
```
NicholasBoll, line 25: This is a clever hack to get the browser to do the color conversion work for us without requiring chromajs. If all we want to do is compare the same color, we can use this method.

modules/color-picker/react/lib/parts/ColorSwatch.tsx
```diff
@@ -11,6 +11,19 @@ export interface ColorSwatchProps extends React.HTMLAttributes<HTMLDivElement> {
   showCheck?: boolean;
 }
 
+// cache a div instance to decrease garbage collection
+const div = document.createElement('div');
```
NicholasBoll, line 15: `document` will not exist in SSR mode

NicholasBoll:
The current approach here will not work since it will brake in SSR. We can't use DOM APIs outside of `use[Layout]Effect` hooks

lychyi:
@NicholasBoll For now, can we just match a `toLowerCase` version of `FFF` or `FFFFFF`? I think that will get this across the finish line and we can think about our own Chroma functions if we want to optimize that (contrast functions aren't too hard to reproduce as utility functions, @mannycarrera4 already did it)

lychyi:
@NicholasBoll Actually now that I think about it, we're using `common` here which comes with chroma-js so we might as well use it and remove it wholesale in a different fix if we think it's problematic.

NicholasBoll:
> @NicholasBoll Actually now that I think about it, we're using `common` here which comes with chroma-js so we might as well use it and remove it wholesale in a different fix if we think it's problematic.

I didn't look into it very hard, but it is possible that a dependency of a package gets tree shaken out from the final bundle. This guarantees dependency. I'm fine with it for now though.

anicholls (approved):

PR #895, opened 2020-10-30
fix(button): Fix accent colors on IconButton when toggled on

### LINKED ISSUE #847: Update Icon Button to support multiple accents

jamesfan:
## Summary

Fixes #847. This is a revised attempt at fixing the issue after PR #848 didn't quite work out.

This addresses an issue where accent layers were being colored incorrectly for a number of icons when an IconButton was toggled on. This issue affected any System Icon with accent layers which did _not_ have a background layer behind them. When the IconButton was toggled on in these cases, the accent layers would be colored to `blueberry400` (assuming the default Canvas theme was in use). This effectively made the accent layers invisible against the `blueberry400` background of the toggled IconButton itself.

If tried to do as much manual verification as I could (there are quite a few System Icons). As far as I can tell, this PR only fixes broken cases -- I didn't notice any regressions 🤞 .

If you'd like to inspect the results: [Grid of IconButtons with fix enabled](https://5d854c26ba934e0020f5e98a-pawfexhaev.chromatic.com/?path=/story/testing-react-buttons-button-icon-button--icon-button-circle-toggleable-grid)

### Before Fix

image

### After Fix

image


modules/button/react/lib/parts/ButtonContainer.tsx
```diff
@@ -49,13 +49,13 @@ function getIconColorSelectors(
...
         fill: color,
       },
       '.wd-icon-background': {
         fill: fill ? color : undefined,
       },
-      '.wd-icon-accent, .wd-icon-accent2': {
+      '.wd-icon-background ~ .wd-icon-accent, .wd-icon-background ~ .wd-icon-accent2': {
         fill: fill
           ? color === themePrimary.contrast
             ? themePrimary.main
```
jamesfan, line 61: Prior to this PR, the coloring logic for accent layers (the nested ternary at the bottom of this code excerpt) only worked when toggling an IconButton on if the accent layer had a background layer behind it to provide contrast. It failed to account for the 50 or so icons which had accent layers but no background layer -- in these cases, the accent layers would be colored to `blueberry400` (assuming default Canvas theme) and blend right into the `blueberry400` of the toggled IconButton itself.

The trick was coloring the accent layers to the same color as the fill layer to start (`frenchVanilla100` in the default Canvas theme). And then only if there's a background layer, do we color the accent layers using the existing logic.

This solution is somewhat fragile in theory since it relies on `.wd-icon-background` being a _prior_ sibling of `.wd-icon-accent` and `.wd-icon-accent2`. I don't believe this should be an issue in practice, however, as I haven't encountered any icons where the background layer follows the accent layers.

modules/button/react/stories/testing/stories_IconButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+  width: '75px',
+
+  p: {
+    fontSize: '12px',
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+  },
+};
+
+export const IconButtonCircleGrid = () => {
```
jamesfan, line 27: I added this manual testing story (no snapshots enabled) to more easily verify the fix was working for all icons. [Link to storybook build](https://5d854c26ba934e0020f5e98a-rrnaqglbcb.chromatic.com/?path=/story/testing-react-buttons-button-icon-button--icon-button-circle-grid).

It may make sense to enable snapshots so we can catch visual regressions automatically. 🤔

jamesfan:
Note: there are a number of icons which still look off when toggling them on. These aren't issues which were introduced with this PR -- IMO, this is because the icons themselves aren't formatted properly:

- `cWaterfall`: background layer doesn't make any sense
- `chart`: background layer is slightly off on the third vertical bar
- `contactCardMatrixManager`: might look better without a background layer
- `deviceDesktop`: might look better without a background layer
- `devicePhone`: might look better without a background layer
- `deviceTablet`: might look better without a background layer
- `packageArrowUp`: has an invalid class name of `wd-icon-fill-accent` on one of its layers
- ... and a handful of other icons I'm sure ...

I'd imagine these issues would be taken care of if/when we take decide to simplify the structure of icons.

mannycarrera4:
lgtm 🥳  awesome work thank you!!!!

modules/button/react/stories/testing/stories_IconButton.tsx
jamesfan: Following up with some additional thoughts @lychyi and I had on this topic.

We were thinking it would make sense to add a snapshot for this story with all buttons toggled on by default. I believe this would be helpful, but there are cases where it would be difficult to resolve visual diffs.

Each icon is an object with the following properties:
```
{
  category: "Profile",
  filename: "wd-icon-academic-appointment-title.svg",
  name: "academicAppointmentTitle",
  svg: "<svg class="wd-icon-academic-appointment-title wd-icon">...</svg>",
  tags: ["schedule", "school", "consult", "consultation", "meeting", "clipboard"],
  type: "system",
}
```

The icons in the grid are sorted alphabetically by `name`. This creates potential issues if brand new icons are added. We're unable to sort the icons in a way that guarantees new icons are added to the end; new icons would be inserted throughout the grid in the proper alphabetical location, likely shifting everything around and creating a very large visual diff in the process.

We could just assume a bunch of new icons were added and approve the diff, though it would be difficult to verify the new icons looked ok in the toggled state given the visual noise of everything shifting around in the grid. Also, if the icon update included changes to existing icons alongside brand new icons, diffs for those changes to existing icons would be very difficult to resolve in the visual noise.

Given how seldomly icons are added (correct me if I'm wrong), I think it still makes sense to add a snapshot here, but I'm curious if anyone else has thoughts.

jamesfan:
I ended up adding visual testing stories with snapshots for [Toggled Circle Icon Buttons](https://5d854c26ba934e0020f5e98a-jmiiridawe.chromatic.com/?path=/story/testing-react-buttons-button-icon-button--icon-button-circle-toggleable-grid) and [Toggled Inverse Icon Buttons](https://5d854c26ba934e0020f5e98a-jmiiridawe.chromatic.com/?path=/story/testing-react-buttons-button-icon-button--icon-button-inverse-toggleable-grid).

mannycarrera4 (approved):

lychyi:
After comparing the diffs, icons that were fixed. We'll need to reference this list in our changelog/release notes:
```
adHocDelivery
alarmClockPlus
alarmClock
boxPlus
boxTextCheck
boxTextPlus
boxTextSearch
boxTextUser
cArea100
cAreaLayered
cAreaStacked
cBar100
cBarClustered
cBarStacked
cBubble
cColumn100
cColumnClustered
cColumnLine
cColumnStacked
cDualLine
cPie
cScatter
calendarUser
cameraPlus
cardView
clipboardBlankCheck
clipboardCheck
commentActive
dashboardExpenses
documentCandidateSearch
documentsCheck
flashAuto
fontSize
highlight
jobInfo
lockPlus
manageDelivery
orderedList
pill
qrCode
receipts
region
reportParameter
rowsCheck
rowsPlus
select
timeOffBalance
time
timelinePerson
unlink
userForward
userPlus
workbook
```

The following are known issues that we should address because they are not rendered incorrectly when inverting the colors.

```
cWaterfall
cart
chart
insertColumnAfter
insertColumnBefore
insertRowAfter
insertRowBefore
mediaClassroom
packageArrowUp
pattern
radioMobile
taskPay
```

Finally there are more than a dozen or so icons that would look a lot better (more accurate) without a background color or with a transparent background.

lychyi (approved):

PR #896, opened 2020-10-30
docs: Remove universal selector for section elements

lychyi:
## Summary

In Storybook, we are doing a universal selector for children of a `<section>` and adding a margin. This was probably for convenience purposes. However, this means that any `section` rendered in a story would universally apply a margin to their children.

For something like `SidePanel` (#866), **it is a section** by default so this would have unintentional layout behaviors in a story.

The fix is to remove the universal selector for `section` and apply the class name `.story` to whatever stories used `section` for this convenience since it does the same thing. All-in-all this smells like some legacy stuff that we just need to get rid of.

Definitely would like to merge this before #866 if possible so I'm going to mark that PR "blocked by" this one.

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -8,7 +8,6 @@
     font-family: sans-serif;
   }
 
-  section > *,
   .story > * {
     margin: 12px !important;
   }
--- a/modules/_labs/core/react/stories/stories.tsx
+++ b/modules/_labs/core/react/stories/stories.tsx
@@ -21,7 +21,7 @@ export default {
 };
 
 export const Type = () => (
-  <section>
+  <section className="story">
     <h1 style={type.brand1}>Brand 1 Header</h1>
     <h2 style={type.brand2}>Brand 2 header</h2>
     <h1 style={type.h1}>H1 Header</h1>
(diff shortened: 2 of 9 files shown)


mannycarrera4 (approved):

NicholasBoll:
Personally I think we should remove `className="story"` from everything as well. I think if we really want a margin added to "stories", we do that at a different layer than the stories themselves - it becomes a layer of noise.

.storybook/preview-head.html
```diff
@@ -8,7 +8,6 @@
     font-family: sans-serif;
   }
 
-  section > *,
```
NicholasBoll, line 11: 👍

modules/_labs/core/react/stories/stories.tsx
```diff
@@ -21,7 +21,7 @@ export default {
 };
 
 export const Type = () => (
-  <section>
+  <section className="story">
```
NicholasBoll, line 24: I don't think we need `className="story"`. I've not been including them and have been removing them when working on stories. If we want to add margins around stories, we can do that at the container level rather than adding the noise to stories.

NicholasBoll: I also think that doesn't need to be done in this PR. I think we can have 1 PR that removes all these story wrappers.

NicholasBoll (approved):

PR #897, opened 2020-11-03
feat: Upgrade Storybook to v6.1.3

anicholls:
## Summary

To make use of Storybook Composition, we need to be on Storybook v6. This PR upgrades CK to use the latest version. This includes:

- Ditching the deprecated style of storybook config (`config.js`, `addons.js`, etc.) for the new style (`main.js`, `manager.js`, `preview.js`). 
- TS is now built in by default, so our webpack/babel config has been simplified and now lives in `main.js`
- The `|` hierarchy separator is gone, and top level sections are just pulled from the string before the first `/`
- Due to [this deprecation](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-support-for-duplicate-kinds), I had to change the organizational structure for all of our theming stories (and had to combine several others into a single story file). You can no longer have multiple default exports with the same name in different files (so all stories in the same storybook folder need to come from the same file). Previously, we usually had one file for visual testing, one file for theming visual testing, both placing a story in the same folder. Unfortunately, if you specify parameters for one exported story it applies to the entire folder of stories, so it was impossible to put themed and unthemed stories in the same file. Since they have to be in separate files, the themed stories had to be put under a `Theming` folder in storybook. I should probably open an issue for the parameter issue...
- I simplified our addons to use `addon-essentials`, which includes Actions (disabled due to poor performance), Backgrounds, Controls, Docs, Viewport, and Toolbars
- There are also several gotchas that I had to address (see below)

## Additional References

Here are some gotchas that I ran into:
- Due to [this issue](https://github.com/storybookjs/storybook/issues/9918), all mdx files have to be prefixed with `.stories`. For example `card.stories.mdx` will work, but `card.mdx` or `stories.mdx` will not.
- Due to [this issue](https://github.com/storybookjs/storybook/issues/11419), static class/function members must be defined after the class/function is defined, but before it is exported. If you define it in both places, this will avoid any Typescript errors. This does not apply for functional components since their static members are always defined after instantiation. For example:

  ```tsx
  class MyComponent extends React.Component<RadioGroupProps> {
    static Foo = Foo;
    // ...
  }
  MyComponent.Foo = Foo;

  export default Foo;
  ```

- Due to [this issue](https://github.com/webpack/webpack/issues/7378) and [this issue](https://github.com/storybookjs/storybook/issues/3346), if you have combined exports for components and types, you need to use the plugin `@babel/plugin-transform-modules-commonjs`. You may still be required to separate some exports. For example:

  ```tsx
  export { MyComponent, MyComponentProps } from "./lib/MyComponent";

  // becomes

  export { MyComponent } from "./lib/MyComponent";
  export type { MyComponentProps } from "./lib/MyComponent";
  ```

- Since we had to combine all stories of the same folder structure into the same file, we now have to conditionally add parameters to some stories but not others. Be careful doing this if the story is a simple re-assignment. Since they have the same functional reference, the parameter will apply to both. Make sure you use a function that returns the original story instead.

  ```tsx
  export const ButtonThemedStates = ButtonStates;
  ButtonThemedStates.parameters = {
    canvasProviderDecorator: {
      theme: customColorTheme,
    },
  };
  
  // Instead, do this:
  
  export const ButtonThemedStates = () => <ButtonStates />;
  ButtonThemedStates.parameters = {
    canvasProviderDecorator: {
      theme: customColorTheme,
    },
  };
  ```
- Due to [this issue](https://github.com/storybookjs/storybook/issues/9052), JSX elements now need a line break between them. Unfortunately divs or React Fragments do not work with Storybook's `Meta` component, so the only solution I've found is adding markdown between JSX tags, or saving the file without formatting.


.github/workflows/ci.yml
```diff
@@ -23,7 +23,8 @@ jobs:
         uses: actions/cache/@v1
         with:
           path: .cache/cypress
-          key: ${{ runner.os }}-cypress-cache-version-${{ steps.cypress-version.outputs.version }}
+          key:
```
anicholls, line 26: @NicholasBoll I was getting a cache error from multiple versions of `Linux-cypress-cache-version` being output here: https://github.com/Workday/canvas-kit/runs/1419196261.

Not sure why this happened all of a sudden, but adding `[0]` seems to have fixed it. Any ideas why this came up? It seems like this caching approach is deprecated, so maybe we should update it?

NicholasBoll: `steps.cypress-version.outputs.version` contains the steps from the step with an id of `cypress-version` which is the output of `node utils/get-cypress-versions.js` which is going through the `yarn.lock` file and getting all versions of `cypress` found.

Cypress v5 is probably coming from the `cypress@*` requirement of `@testing-library/cypress`. So some dependency update forced an upgrade of Cypress to version 5. You could probably verify with a `yarn why cypress`. You could upgrade to `actions/cache/@v2`, but this problem would remain.

NicholasBoll: 2 versions of Cypress causes issues - we probably only want a single version of Cypress included in the `yarn.lock` file

.storybook/manager.js
```diff
@@ -0,0 +1,10 @@
+// TODO: A semi-shoddy hack to enable syntax highlighting for tsx in storybook-readme.
+// Probably should open a PR to put tsx in storybook-readme instead?
+// import 'prismjs/components/prism-tsx';
```
NicholasBoll, line 3: The comments are about this import which is now also commented out. Should we remove 3 lines?

NicholasBoll (approved):

package.json
```diff
@@ -48,6 +50,7 @@
     "babel-loader": "^8.1.0",
     "babel-plugin-emotion": "^10.0.29",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2",
+    "caniuse-lite": "^1.0.30001154",
```
NicholasBoll, line 53: What's with these version numbers? Looks like they are already on `1.0.30001161`

utils/publish-canary.js
```diff
@@ -83,6 +83,7 @@ exec('git diff --name-only HEAD HEAD^')
       `--canary`,
       `--preid ${preid}`,
       `--dist-tag ${distTag}`,
+      `--max-buffer=999999999`,
```
NicholasBoll, line 86: Was this issue fixed with node 12?

anicholls: Not sure, I'm not familiar with why this is here

package.json
anicholls: haha right? We were getting console warnings and the suggested fix was adding this dep. This was the latest when I added it, but I can upgrade again if you want

NicholasBoll: No, just saw it and was wondering which prompted me to look it up. They have about 400 versions in the `1.0.3x` range

utils/publish-canary.js
NicholasBoll: It reminds me of this: https://github.com/Workday/canvas-kit/pull/908/files#diff-53d182f2fc69a23673a933acbee9ac295c91e80b1ccc93e316b87b8f92a55067L90-L95

We were able to remove the max buffer override by changing the version of node we're using in TravisCI from 10 to 12

anicholls: Hmm, I think this may have came in from a rebase/merge so I will remove it.

PR #908, opened 2020-11-16
ci: Use exec instead of node-cmd

lychyi:
Swap `node-cmd` for `child_process.exec` (it's the same underlying function). Clean up unused require statements and update to Node 12 to polish off #903, this increases the buffer size as part of Node updates so we removed that option arg in the codebase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-  - "10"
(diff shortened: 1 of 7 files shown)


NicholasBoll (approved):
LGLTM - One less dependency is one less abstraction to look up.

PR #914, opened 2020-11-20
fix(select): Allow a single option

NicholasBoll:
## Summary

A `Select` should support a single `SelectOption` child.


alanbsmith (approved):
🐎

Issue #901, opened 2020-11-11
Check internal flag for monospaced font family


lychyi:
## 🚀 Feature Proposal

Internal teams are requesting that we do the same changes listed here: https://github.com/Workday/canvas-kit/pull/722/files#diff-95077caaf948584f828f2834ab88bfeef76b54518b5decc190d87d994afa013aR6-R8 but for monospaced fonts.

## Motivation

Determine if a global font is set on the Workday product and fallback to the existing monospaced font family if needed.

PR #916, opened 2020-11-24
feat: Add window config option to inherit or override mono font family

anicholls:
## Summary

Similar to https://github.com/Workday/canvas-kit/pull/553, we also need to support this configuration for mono fonts. This PR achieves that. There are two options two control this, as discussed in the requirements:

- If the component uses only mono fonts, you can set `window.workday.canvas.inheritMonoFontFamily = true` and then specify the font higher up in the CSS hierarchy
- If the component uses both mono and regular fonts, you'll need to explicitly pass the font stack. For example: `window.workday.canvas.inheritMonoFontFamily = '"Noto CJK", "Roboto Mono", "Courier New", Courier, monospace'`

Closes https://github.com/Workday/canvas-kit/issues/901


fullmetal2000 (approved):

mannycarrera4:
lgtm 🚢

NicholasBoll (approved):

PR #917, opened 2020-11-24
fix: Add Emotion dependency to fix transient dep issue

anicholls:
## Summary

Something about #897 has been causing a non-deterministic transient dependency issue in some environments. For some reason, `@emotion/styled-base` is not installed under some circumstances, even though it is a dependency of `@emotion/styled` (which is used in many of our components). We were unable to track down the source of the issue, so I've temporarily added this package as a workspace dependency to get our master builds passing again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "@commitlint/cli": "^8.3.5",
     "@commitlint/config-conventional": "^8.1.0",
     "@emotion/is-prop-valid": "^0.8.2",
+    "@emotion/styled-base": "^10.0.28",
     "@storybook/addon-actions": "^6.1.3",
     "@storybook/addon-docs": "^6.1.3",
     "@storybook/addon-essentials": "^6.1.3",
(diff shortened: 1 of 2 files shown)


willklein (approved):
🙃 Fine yarn, fine.

package.json
```diff
@@ -14,6 +14,7 @@
     "@commitlint/cli": "^8.3.5",
     "@commitlint/config-conventional": "^8.1.0",
     "@emotion/is-prop-valid": "^0.8.2",
+    "@emotion/styled-base": "^10.0.28",
```
NicholasBoll, line 17:
```suggestion
    "@emotion/styled-base": "^10.0.27",
```

If you do this, `@emotion/styled-base` will only be included in the whole dependency tree once instead of multiple times

willklein: I thought about this too, I had trouble testing so I held off mentioning it. It's worth looking at.

NicholasBoll: You couldn't just accept the GH suggestion though, `yarn install` would have to be run again. This will basically clear up the all changes to the `yarn.lock` file.

NicholasBoll (approved):

PR #926, opened 2020-12-09
fix(common): Fix mouseFocusBehavior type issue

### LINKED ISSUE #925: type issue on the hideMouseFocus function
## 🐛 Bug Report

New type on the hideMouseFocus function is causing an error: 

## To Reproduce

```ts
 '&:focus': {
      backgroundColor: red,
    },
    ...mouseFocusBehavior({
      '&:focus': {
        backgroundColor: blue,
      },
    }),
  },
```

## Actual Results

Typescript is throwing the error:
`This spread always overwrites this property.`
`'&:focus' is specified more than once, so this usage will be overwritten.`

Since mouseFocus returns the same type as the selector param. Typescript is complaining about a return value  of 
`{  '&:focus': { backgroundColor: blue }; }`

## Expected Behavior

No typescript errors since the function return does not actually override &:focus

## Workarounds

You can trick typescript by dropping one of the ampersands, i.e. :focus and overriding with &:focus
```ts
 ':focus': {
      backgroundColor: red,
    },
    ...mouseFocusBehavior({
      '&:focus': {
        backgroundColor: blue,
      },
    }),
  },
```

Another workaround is casting the function call with a different type, e.g.
```ts
 ':focus': {
      backgroundColor: red,
    },
    ...(mouseFocusBehavior({
      '&:focus': {
        backgroundColor: blue,
      },
    }) as CSSObject),
  },
```

NicholasBoll:
Fixes #925

I went through many iterations testing with the following code:

```ts
const temp = {
  '&:focus': {
    background: 'red',
  },
  ...mouseFocusBehavior({
    '&:focus': {
      backgroundColor: 'blue',
    },
  }),
};

const Div = styled('div');
const Foo = Div(temp);

const styles = css(
  mouseFocusBehavior({
    '&:focus': {
      outline: 'none',
    },
  })
);
```

I made sure that `mouseFocusBehavior` works for both `styled` and `css`.

### DIFF
--- a/modules/common/react/lib/styles/hideMouseFocus.ts
+++ b/modules/common/react/lib/styles/hideMouseFocus.ts
@@ -1,4 +1,4 @@
-import {CSSObject} from '@emotion/core';
+import {CSSObject, Interpolation} from '@emotion/core';
 
 /**
  * A utility to hide the default canvas style focus ring when using mouse input.
@@ -14,8 +14,6 @@ export const hideMouseFocus: CSSObject = {
   },
 };
 
-type IndexableObject = {[key: string]: object};
-
 /*
  * A utility that simplifies focus selectors since you can't use nested syntax for some reason. Example:
  * [`[data-whatinput="mouse"],
@@ -26,8 +24,10 @@ type IndexableObject = {[key: string]: object};
        }
    },
  */
-export const mouseFocusBehavior = <T extends IndexableObject = IndexableObject>(selectors: T) => {
-  const output: IndexableObject = {};
+export const mouseFocusBehavior = (
+  selectors: Record<string, Interpolation>
+): Record<string, Interpolation> => {
+  const output: Record<string, Interpolation> = {};
 
   Object.keys(selectors).map((selector, index) => {
     selector.split(',').forEach(selectorPart => {
@@ -39,5 +39,5 @@ export const mouseFocusBehavior = <T extends IndexableObject = IndexableObject>(
     });
   });
 
-  return output as T;
+  return output;
 };


NicholasBoll:
I tried:

### Interpolation

```ts
export const mouseFocusBehavior = (selectors: Interpolation): Interpolation => {
```

errors on:
```ts
  ...mouseFocusBehavior({
    '&:focus': {
      backgroundColor: 'blue',
    },
  }),
```

error:
```ts
Spread types may only be created from object types.ts(2698)
```

### Generic on Object Record
```ts
export const mouseFocusBehavior = <T extends Record<string, object>>(selectors: T): T => {
```

errors on:
```ts
const temp = {
  '&:focus': {
    background: 'red',
  },
  ...mouseFocusBehavior({
    '&:focus': {
      backgroundColor: 'blue',
    },
  }),
};
```

error:
```ts
'&:focus' is specified more than once, so this usage will be overwritten.ts(2783)
```

### Object Record
```ts
export const mouseFocusBehavior = (selectors: Record<string, object>):Record<string, object> => {
```

errors on:
```ts
const styles = css(
  mouseFocusBehavior({
    '&:focus': {
      outline: 'none',
    },
  })
);
```

error:
```ts
No overload matches this call.
  Overload 1 of 2, '(template: TemplateStringsArray, ...args: Interpolation<undefined>[]): SerializedStyles', gave the following error.
    Argument of type 'Record<string, object>' is not assignable to parameter of type 'TemplateStringsArray'.
      Type 'Record<string, object>' is missing the following properties from type 'TemplateStringsArray': raw, length, concat, join, and 17 more.
  Overload 2 of 2, '(...args: Interpolation<undefined>[]): SerializedStyles', gave the following error.
    Argument of type 'Record<string, object>' is not assignable to parameter of type 'Interpolation<undefined>'.
      Type 'Record<string, object>' is not assignable to type 'Keyframes'.
        Type 'Record<string, object>' is missing the following properties from type '{ name: string; styles: string; anim: number; toString: () => string; }': name, styles, anim ts(2769)
```

alanbsmith (approved):
🐭

PR #912, opened 2020-11-18
chore: Fix create component scripts

Parker-Ledoux:
## Summary
The create component scripts were still dependent on `node-cmd` which was removed here: https://github.com/Workday/canvas-kit/commit/bc2f1f499736d6b792c3eccc05e7bca93b83c817#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L95

Also, we still need to use the `colors` module for our cli coloring which was removed here: 
https://github.com/Workday/canvas-kit/commit/bc2f1f499736d6b792c3eccc05e7bca93b83c817#diff-cf53da1297de5bd1a9a1f27c12e693097fcad3810a132ab41d49b4a469c0fdf6L4

PR #929, opened 2020-12-10
fix: Fix create component scripts

jamesfan:
## Summary

Our createComponent script was bombing out with `TypeError: exec is not a function` due to incorrect `require` statements for `exec`. Originally, `node-cmd` was used to run commands, but it was removed as a dependency in favor of `exec` in #908. However, the `require` statements in the createComponent scripts weren't updated to use `exec` until #912. This PR adds the necessary destructuring to get those `require` statements working.

It also fixes a path issue with the `stories_VisualTesting` template.

### DIFF
--- a/utils/create-component/templates/react/stories_VisualTesting.js
+++ b/utils/create-component/templates/react/stories_VisualTesting.js
@@ -1,11 +1,11 @@
 // React stories template
 
-module.exports = (storyPath, pascalCaseName, rootPath) => `
+module.exports = (storyPath, pascalCaseName, rootPath, unstable) => `
 /// <reference path="${rootPath}/../typings.d.ts" />
 /** @jsx jsx */
 import {jsx} from '@emotion/core';
 import {StaticStates} from '@workday/canvas-kit-labs-react-core';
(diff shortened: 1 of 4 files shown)


utils/create-component/templates/react/stories_VisualTesting.js
```diff
@@ -1,11 +1,11 @@
 // React stories template
 
-module.exports = (storyPath, pascalCaseName, rootPath) => `
+module.exports = (storyPath, pascalCaseName, rootPath, unstable) => `
 /// <reference path="${rootPath}/../typings.d.ts" />
 /** @jsx jsx */
 import {jsx} from '@emotion/core';
 import {StaticStates} from '@workday/canvas-kit-labs-react-core';
-import {ComponentStatesTable, permutateProps, withSnapshotsEnabled} from '../../../../utils/storybook';
+import {ComponentStatesTable, permutateProps, withSnapshotsEnabled} from '../../../../${unstable ? '../' : ''}utils/storybook';
```
jamesfan, line 8: Adjust path to storybook utils based on whether or not this is a Labs module. Labs modules sit one level deeper in the folder structure since they're placed in the `_labs` folder alongside the non-Labs modules.

NicholasBoll (approved):

PR #937, opened 2021-01-08
docs: Add compound component doc

NicholasBoll:
This doc explains the concept of compound components that is the planned paradigm for building components in Canvas Kit moving forward.

Check out the rendered document here: https://github.com/Workday/canvas-kit/blob/13b61d14c73a48524413fdb7e52d30cdb01961fb/COMPOUND_COMPONENTS.md

### DIFF
--- a/COMPOUND_COMPONENTS.md
+++ b/COMPOUND_COMPONENTS.md
@@ -6,8 +6,8 @@ It is a way of describing a complex component by simultaneously describing sub-c
 
 * Container component
 * Sub-components
-* Shared model
-* Behavior hooks
+* Shared model (optional)
+* Behavior hooks (optional)
 
 A compound component contrasts with a configuration component.
 
@@ -33,11 +33,20 @@ Compound component:
 
 In this example, `Tabs` is the container component and `Tabs.List` is a sub-component.
 
+Compound components always have a container component and sub-components. Components that do not contain sub-components are not considered compound components. Some compound components might not contain state or behavior. An example might be an `IconButton` which is a button that contains an icon. It might be a compound component only for styling purposes, but doesn't contain any special state or behaviors:
+
+```tsx
+<IconButton onClick={onClick}>
+  <IconButton.Icon icon={icon} />
+  <IconButton.Text>Button Text</IconButton.Text>
+</IconButton>
+```
+
 ## Container Components
 
-A container component provides a shared model to sub-components via [React context](https://reactjs.org/docs/context.html). A container component could represent a real DOM element, or just be a context container. For example, the `Pagination` component has a container component that represents a `role=nav` element. The `Tabs` container component, however, does not contain a semantic element.
+A container component is the entry point to a compound component. A container component could represent a real DOM element, or just be a non-element container. For example, the `Pagination` component has a container component that represents a `role=nav` element. The `Tabs` container component, however, does not contain a semantic element.
 
-The container component provides it's own context hook for sub-components. A container component takes props for either the model or configuration for the model. In the `Tabs` compound component example, it might look like this:
+If a compound component contains any state or behavior, it will also provide a shared model to sub-components via [React context](https://reactjs.org/docs/context.html). A container component takes props for either the model or configuration for the model. In the `Tabs` compound component example, it might look like this:
 
 ```tsx
 const TabsModelContext = React.createContext({})
@@ -58,7 +67,7 @@ const Tabs = ({children, model, ...config}) => {
 
 ## Sub-components
 
-A sub-component typically follows ARIA roles. For the `Tabs` example, these are the `tablist`, `tab`, and `tabpanel` roles. A sub-component provides direct access to semantic or key elements of a compound component.
+A sub-component typically follows ARIA roles. For the `Tabs` example, these are the `tablist`, `tab`, and `tabpanel` roles. A sub-component provides direct access to semantic or key elements of a compound component. In the `IconButton` example, the icon is not semantic and might be hidden from screen readers while the `IconButton.Text` content is instead used for a tooltip and as the accessible name while being visibly hidden.
 
 ## Why Compound Components?
 
--- a/COMPOUND_COMPONENTS.md
+++ b/COMPOUND_COMPONENTS.md
@@ -49,10 +49,7 @@ A container component is the entry point to a compound component. A container co
 If a compound component contains any state or behavior, it will also provide a shared model to sub-components via [React context](https://reactjs.org/docs/context.html). A container component takes props for either the model or configuration for the model. In the `Tabs` compound component example, it might look like this:
 
 ```tsx
-const TabsModelContext = React.createContext({})
-
-// This will be used in sub-components
-const useTabsModelContext = () => React.useContext(TabsModelContext)
+export const TabsModelContext = React.createContext({})
 
 const Tabs = ({children, model, ...config}) => {
   // either a model is passed in, or we create one
@@ -283,7 +280,7 @@ In the `Tabs` component example, the Menu doesn't actually exist as component, b
 
 ```tsx
 const TabList = ({children, ...elemProps}) => {
-  const model = useTabsModelContext()
+  const model = React.useContext(TabsModelContext)
   const menu = useMenu(
     model,
     elemProps,
--- a/COMPOUND_COMPONENTS.md
+++ b/COMPOUND_COMPONENTS.md
@@ -9,7 +9,7 @@ Compound components is a pattern in which components are used together in such a
 * [Shared model (optional, advanced)](#models)
 * [Behavior hooks (optional, advanced)](#behavior-hooks)
 
-A compound component contrasts with a configuration component which instead configures from a single interface. A configuration component might be like choosing a desktop computer based on stats - how much RAM or how fast the CPU should be or based on what you want to use it for. A compound component is more like buying the parts individually and assembling yourself.
+A compound component contrasts with a configuration component which instead configures from a single interface, like a configuration object or multiple props in React. A configuration component might be like choosing a desktop computer based on stats - how much RAM or how fast the CPU should be or based on what you want to use it for. A compound component is more like buying the parts individually and assembling yourself.
 
 Configuration component:
 ```tsx
--- a/COMPOUND_COMPONENTS.md
+++ b/COMPOUND_COMPONENTS.md
@@ -44,7 +44,7 @@ Some compound components might not contain state or behavior. An example might b
 
 ## Container Components
 
-A container component is the entry point to a compound component. A container component could represent a real DOM element, or just be a non-element container. For example, the `Pagination` component has a container component that represents a `role=nav` element. The `Tabs` container component, however, does not contain a semantic element.
+A container component is the entry point to a compound component. A container component could represent a real DOM element, or be a non-element container. For example, the `Pagination` component has a container component that represents a `role=nav` element. The `Tabs` container component, however, does not contain a semantic element.
 
 If a compound component contains any state or behavior, it will also provide a shared model to sub-components via [React context](https://reactjs.org/docs/context.html). A container component takes props for either the model or configuration for the model. In the `Tabs` compound component example, it might look like this:
 
--- a/COMPOUND_COMPONENTS.md
+++ b/COMPOUND_COMPONENTS.md
@@ -51,6 +51,7 @@ If a compound component contains any state or behavior, it will also provide a s
 ```tsx
 const TabsModelContext = React.createContext({})
 
+// This will be used in sub-components
 const useTabsModelContext = () => React.useContext(TabsModelContext)
 
 const Tabs = ({children, model, ...config}) => {
--- a/COMPOUND_COMPONENTS.md
+++ b/COMPOUND_COMPONENTS.md
@@ -193,7 +193,7 @@ const useTabsModel = (config = {}) => {
     ...menu.state,
     id,
     activeTab,
-    panels: panels.items,
+    panels: panels.state.items,
   }
   
   const events = {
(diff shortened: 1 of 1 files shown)


COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,336 @@
...
+    <Tabs.Item>Second Tab</Tabs.Item>
+  </Tabs.List>
+  <Tabs.Panel>First Tab Contents</Tabs.Panel>
+  <Tabs.Panel>Second Tab Contents</Tabs.Panel>
+</Tabs>
+```
+
+In this example, `Tabs` is the container component and `Tabs.List` is a sub-component.
+
+Compound components always have a container component and sub-components. Components that do not contain sub-components are not considered compound components. Some compound components might not contain state or behavior. An example might be an `IconButton` which is a button that contains an icon. It might be a compound component only for styling purposes, but doesn't contain any special state or behaviors:
```
NicholasBoll, line 36: I don't like the way this reads:

> Compound components always have a container component and sub-components. Components that do not contain sub-components are not considered compound components.

Those 2 sentences make the same statement. This is also covered by a statement earlier:

> [Compound components are] a way of describing a complex component by simultaneously describing sub-components.

Those 2 sentences do not directly relate to the rest of the paragraph, which talks about not having a model or behavior. Perhaps remove those sentences or move them somewhere else.

```suggestion
Some compound components might not contain state or behavior. An example might be an `IconButton` which is a button that contains an icon. It might be a compound component only for styling purposes, but doesn't contain any special state or behaviors:
```

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,336 @@
...
+## What is a compound component?
+
+It is a way of describing a complex component by simultaneously describing sub-components. A compound component is made up of a few different parts.
+
+* Container component
+* Sub-components
+* Shared model (optional)
+* Behavior hooks (optional)
+
+A compound component contrasts with a configuration component.
```
NicholasBoll, line 12: I think we need a better definition here and to describe a configuration component.

```suggestion
Compound components is a pattern in which components are used together in such a way that describes a higher level of abstraction. A desktop computer is composed of many components that can be serviced and configured individually. Similarly, a compound component allows simultaneous configuration through access to each sub-component. Here are the parts of a compound component:

* Container component
* Sub-components
* Shared model (optional)
* Behavior hooks (optional)

A compound component contrasts with a configuration component which instead configures from a single interface. A configuration component might be like choosing a desktop computer based on stats - how much RAM or how fast the CPU should be or based on what you want to use it for. A compound component is more like buying the parts individually and assembling yourself.
```

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,336 @@
...
+```tsx
+<IconButton onClick={onClick}>
+  <IconButton.Icon icon={icon} />
+  <IconButton.Text>Button Text</IconButton.Text>
+</IconButton>
+```
+
+## Container Components
+
+A container component is the entry point to a compound component. A container component could represent a real DOM element, or just be a non-element container. For example, the `Pagination` component has a container component that represents a `role=nav` element. The `Tabs` container component, however, does not contain a semantic element.
```
NicholasBoll, line 47:
```suggestion
A container component is the entry point to a compound component. A container component could represent a real DOM element, or be a non-element container. For example, the `Pagination` component has a container component that represents a `role=nav` element. The `Tabs` container component, however, does not contain a semantic element.
```

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,336 @@
...
+A behavior hook usually applies to a sub-component and describes the events and attributes that are applied to a sub-component's element. A behavior hook takes in the model and developer-defined HTML attributes and return a merged object of events and HTML attributes.
+
+### Why Behavior Hooks?
+
+A behavior hook allows us to more easily reuse functionality between components with similar sub-components. They also provide another layer of composition to compound components. They are usually the accessibility layer of a sub-component and not the HTML element directly.
+
+For example, the `MenuModel` contains model's internal state and events, but doesn't handle external DOM events directly. The behavior hook is the glue between the model and DOM elements. A `useMenu` behavior hook might look like this:
+
+```ts
+const useMenu const useMenu = (
```
anicholls, line 252:
```suggestion
const useMenu = (
```

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,336 @@
...
+  return {
+    onKeyDown(event) {
+      // if onKeyDown was provided, call it first
+      elemProps.onKeyDown?.(event)
+
+      if (event.key === 'ArrowLeft') {
+         events.previous()
+      }
+      if (event.key === 'ArrowRight') {
+         events.next()
```
anicholls, line 271:
```suggestion
         model.events.previous()
      }
      if (event.key === 'ArrowRight') {
         model.events.next()
```
(or destructure `model` above)

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,336 @@
...
+    >
+      {children}
+    </div>
+  )
+};
+```
+
+There are 2 ways to configure a model.
+
+The easiest way is to pass model configuration directly to the container component. In the `Tabs` example, it might look like this:
```
anicholls, line 304: This example could use a bit of expanding. It's not immediately clear what `onActivateTab` has to do with the model configuration. I think it's just saying something like "props for each model field are generally available for simple customization on components without having to pass in the entire model", etc.

anicholls (commented):
This is excellent! Nice work @NicholasBoll. I find it much more instructional and easier to read than the gist version that came before this. Excited to see more of this in practice 🎉

anicholls:
Slightly unrelated: we're starting to build up a large number of these non-readme documentation files. I wonder if we should put them in a separate folder so the root directory of the repo is cleaner?

COMPOUND_COMPONENTS.md
NicholasBoll: Good catch

NicholasBoll: Also state above... Perhaps I'll destructure then.

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Why Behavior Hooks?
+
+A behavior hook allows us to more easily reuse functionality between components with similar sub-components. They also provide another layer of composition to compound components. They are usually the accessibility layer of a sub-component and not the HTML element directly.
+
+For example, the `MenuModel` contains model's internal state and events, but doesn't handle external DOM events directly. The behavior hook is the glue between the model and DOM elements. A `useMenu` behavior hook might look like this:
+
+```ts
+const useMenu const useMenu = (
+  model,
```
NicholasBoll, line 253:
```suggestion
  { state, events },
```

COMPOUND_COMPONENTS.md
NicholasBoll: I agree. There is a line above that says:
>  A model has configuration for functionality including configuration of events.

But it isn't obvious this is the connection. It should be reiterated.

NicholasBoll: It uses the example above:

https://github.com/Workday/canvas-kit/pull/937/files#diff-a467e1a7dbcaa0ea8af669d262eaab2b94cb1302e12f9d81aaa4117a08872536R185-R214

NicholasBoll: What about the following?

```suggestion
The easiest way is to pass model configuration directly to the container component. Continuing from the `useTabsModel` example above which has a `onActivateTab` action available on the model configuration, it might look like this:
```

NicholasBoll:
> Slightly unrelated: we're starting to build up a large number of these non-readme documentation files. I wonder if we should put them in a separate folder so the root directory of the repo is cleaner?

Agreed. Perhaps a `docs` folder or something? Some of the docs we also use as Storybook pages. We'd have to update URLS as well.

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,336 @@
...
+>
+  ...
+</Tabs>
+```
+
+We expect most use-cases will be covered by the above example, but If you need direct access to the model in your component, you'll have to create the configured model yourself and pass it on to the container component. The allows direct access to state and events of the model. In the `Tabs` example, it might look like this:
+
+```tsx
+const MyTabs = () => {
+  const model = useTabModel({
```
jamesfan, line 318:
```suggestion
  const model = useTabsModel({
```

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,336 @@
...
+  const id = useUniqueId(config.id)
+  const [activeTab, setActiveTab] = React.useState('')
+  const menu = useMenuModel(config)
+  const panels = useListModel(config)
+  
+  const state = {
+    ...menu.state,
+    id,
+    activeTab,
+    panels: panels.items,
```
jamesfan, line 196:
```suggestion
    panels: panels.state.items,
```

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,336 @@
...
+## Container Components
+
+A container component is the entry point to a compound component. A container component could represent a real DOM element, or be a non-element container. For example, the `Pagination` component has a container component that represents a `role=nav` element. The `Tabs` container component, however, does not contain a semantic element.
+
+If a compound component contains any state or behavior, it will also provide a shared model to sub-components via [React context](https://reactjs.org/docs/context.html). A container component takes props for either the model or configuration for the model. In the `Tabs` compound component example, it might look like this:
+
+```tsx
+const TabsModelContext = React.createContext({})
+
+const useTabsModelContext = () => React.useContext(TabsModelContext)
```
NicholasBoll, line 54:
```suggestion
// This will be used in sub-components
const useTabsModelContext = () => React.useContext(TabsModelContext)
```

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,337 @@
...
+```
+
+## Container Components
+
+A container component is the entry point to a compound component. A container component could represent a real DOM element, or be a non-element container. For example, the `Pagination` component has a container component that represents a `role=nav` element. The `Tabs` container component, however, does not contain a semantic element.
+
+If a compound component contains any state or behavior, it will also provide a shared model to sub-components via [React context](https://reactjs.org/docs/context.html). A container component takes props for either the model or configuration for the model. In the `Tabs` compound component example, it might look like this:
+
+```tsx
+const TabsModelContext = React.createContext({})
```
NicholasBoll, line 52: In building a new component, `useTabsModel` and `useTabsModelContext` are very close to each other and return the same type. If you use `useTabsModel` (first autocomplete suggestion) instead of the context version, you'll get a new model instance rather than the shared reference and things don't work the way you expect.

The suggestion is instead to export the `TabsModel` as a context reference and don't export a `useTabsModelContext` hook. This also helps reinforce this is explicitly React Context and not a new model.

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,334 @@
...
+   const state = {}
+   const events = {}
+   
+   return { state, events }
+ }
+ ```
+
+A model has configuration for functionality including configuration of events. An event has two parts:
+* Guard
+* Action
```
NicholasBoll, line 133: Explain more of the of the why of these concepts. Also consider listing actions first - also change "actions" to "callbacks"

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,466 @@
+# Compound Components
+
+## What is a compound component?
+
+Compound components is a pattern in which components are used together in such a way that describes a higher level of abstraction. A desktop computer is composed of many components that can be serviced and configured individually. Similarly, a compound component allows simultaneous configuration through access to each sub-component. Here are the parts of a compound component:
```
willklein, line 5: What do you think about this for an alternative to the opening line here?
```
> Compound components is a pattern where higher level components are composed using smaller components, and you retain control over how those smaller pieces are wired together.
```

willklein: I would even consider formatting it in that way, as a quote, so it stands out and apart from the elaborating paragraph.

COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,466 @@
...
+
+* [Container component](#container-components)
+* [Sub-components](#sub-components)
+* [Shared model (optional, advanced)](#models)
+* [Behavior hooks (optional, advanced)](#behavior-hooks)
+
+A compound component contrasts with a configuration component which instead configures from a single interface. A configuration component might be like choosing a desktop computer based on stats - how much RAM or how fast the CPU should be or based on what you want to use it for. A compound component is more like buying the parts individually and assembling yourself.
```
willklein, line 12: Consider a slight expansion to: "A compound component contrasts with a configuration component which instead configures from a single interface." ...with "... a single interface, like a configuration object or multiple props in React."

willklein: I like the PC analogy!

COMPOUND_COMPONENTS.md
NicholasBoll: I'm not sure we retain control over how the smaller pieces are wired together (at least not from the component API). It allows configuration of both the overall behavior as well as of individual sub-components by providing direct access to semantic elements.

How about:
```suggestion
> Compound components is a pattern where higher level components are composed using smaller components, and you retain access to all the semantic elements of the higher level component.
```

COMPOUND_COMPONENTS.md
NicholasBoll: Do you mean like this?

```suggestion
A compound component contrasts with a configuration component which instead configures from a single interface, like a configuration object or multiple props in React. A configuration component might be like choosing a desktop computer based on stats - how much RAM or how fast the CPU should be or based on what you want to use it for. A compound component is more like buying the parts individually and assembling yourself.
```

COMPOUND_COMPONENTS.md
willklein: Thanks, that's what I was going for.

alanbsmith (approved):
🐼

PR #946, opened 2021-01-21
feat: Add compound component utility functions

NicholasBoll:
* Add `createComponent` helper functions
* Add model types and helper functions
* Update new component templates to use Compound Component helper functions

This PR adds compound component helper functions to help with creating compound components. These utility functions take care of complexity around Typescript, enforce ref forwarding, enable element remapping, etc. Documentation for creating compound components will follow.

## Checklist

- [x] `yarn test` passes
      applicable

Still to be done:
- [ ] Update API pattern guidelines MD file 
- [x] Add documentation for creating compound components #950

### DIFF
--- a/modules/common/react/lib/utils/components.ts
+++ b/modules/common/react/lib/utils/components.ts
@@ -28,7 +28,7 @@ export type Component<T extends React.ElementType, P> = {
 interface RefForwardingComponent<T, P = {}> {
   (
     props: React.PropsWithChildren<P> & {as?: React.ReactElement<any>},
-    ref: T extends undefined
+    ref: T extends null
       ? never
       : React.Ref<T extends keyof ElementTagNameMap ? ElementTagNameMap[T] : T>,
     as: T extends undefined ? never : T
@@ -36,12 +36,21 @@ interface RefForwardingComponent<T, P = {}> {
 }
 
 /**
+ * If no element is passed to `createComponent`, a `NulLElement` will be returned indicating there
+ * is no element type.
+ */
+type NullElement = () => null;
+
+/**
  * Factory function that creates components to be exported. It enforces React ref forwarding, `as`
  * prop, display name, and sub-components, and handles proper typing without much boiler plate. The
  * return type is `Component<element, Props>` which looks like `Component<'div', Props>` which is a
  * clean interface that tells you the default element that is used.
  */
-export const createComponent = <T extends React.ElementType>(as?: T) => <P, SubComponents = {}>({
+export const createComponent = <T extends React.ElementType = NullElement>(as?: T) => <
+  P,
+  SubComponents = {}
+>({
   displayName,
   Component,
   subComponents,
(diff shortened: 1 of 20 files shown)


modules/common/react/lib/utils/components.ts
```diff
@@ -0,0 +1,174 @@
...
+ */
+export type PropsWithAs<P, T extends React.ElementType> = P &
+  Omit<React.ComponentProps<T>, 'as' | 'state' | keyof P>;
+
+/**
+ * Component type that allows for `as` to change the element or component type.
+ * Passing `as` will correctly change the allowed interface of the JSX element
+ */
+export type Component<T extends React.ElementType, P> = {
+  <TT extends React.ElementType>(props: PropsWithAs<P, TT> & {as: TT}): JSX.Element;
```
anicholls, line 24: The double T here (`TT`) is new to me. What does this do??

alanbsmith: It's just another generic, since T is already being used elsewhere.

NicholasBoll: It is used because it is the same intended type as `T`, but is a different type in this instance. `Component` is an overloaded type:

```tsx
const Foo = Component<'div', Props>

<Foo /> // hits the second overload which means it has all the inherit props of HTMLDivElement
<Foo as="button" /> // matches the first overload and means it has all the inherit props of HTMLButtonElement
```

alanbsmith: All that's saying is "whatever component is provided will extend React.ElementType, use its own props + `as`, where `as` is the type of component being provided."

NicholasBoll: The double `TT` is another generic.

modules/common/react/lib/utils/components.ts
```diff
@@ -0,0 +1,174 @@
...
+ * @example
+ * const SubComponent = ({children, model, ...elemProps}: SubComponentProps, ref, Element) => {
+ *   const {state, events} = useModelContext(model, SubComponentModelContext);
+ *
+ *   // ...
+ * }
+ */
+export function useModelContext<T>(context: React.Context<T>, model?: T) {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  return model || React.useContext(context);
```
alanbsmith, line 173: hmm, so context would always be required for all sub-components, even if they were using the model? I'm not sure that's what we want? I'll have to think about it more.

As an example: Lets say I just want to do this:
```tsx
const tabsModel = useTabsModel(config);

<Tabs.List model={tabsModel}>
  <Tab model={tabModel}>One</Tab>
  <Tab model={tabModel}>Two</Tab>
  <Tab model={tabModel}>Three</Tab>
</Tabs.List>
```

alanbsmith (commented):
This looks really good. I have one question for now. We can talk more about it tomorrow though. 😄

modules/common/react/lib/utils/components.ts
NicholasBoll: This adds the flexibility of optionally passing in a model. As far as context always being required... I'm not sure how that's a problem. If a React Context is not found it will return `null`. And in your example, the `React.useContext` would never be hit anyways, so in your example, no Context is created or used.

alanbsmith (approved):
🚀

modules/common/react/lib/utils/components.ts
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * This function is the same as calling `React.useRef(null)` with the added benefit of extracting
+ * the type from the ref passed to it. This means you don't have to provide a generic to the
+ * `useRef` function
+ * @param ref The React ref passed from the `createComponent` factory function
+ */
+export function useComponentRef<T>(ref: React.Ref<T>) {
+  return React.useRef<T>(null);
+}
```
NicholasBoll, line 146: I've been using these together any time I need a local ref for something:

```tsx
  Component: ({name = '', model, children, ...elemProps}: TabProps, ref, Element) => {
    const localRef = useComponentRef(ref);
    const elementRef = useForkRef(ref, localRef);

    // do something with localRef
    return <Element ref={elementRef} ... />
```

Since it is so common to use them together and it feels odd to juggle so many refs, why not a function that combines them?

```tsx
  Component: ({name = '', model, children, ...elemProps}: TabProps, ref, Element) => {
    const { localRef, elementRef } = useLocalRef(ref);

    // do something with localRef
    return <Element ref={elementRef} ... />
```

Issue #751, opened 2020-06-18
Support for optional RTL Icon Mirroring


jpante:
## 🚀 Feature Proposal

Enhance icon components and components that contain icons to allow for icon mirroring. Some icons may not require mirroring so this feature should give control to the implementor to determine if a mirrored icon is required.

## Motivation

Icon components (and buttons that contain icons) in some cases require the icon is flipped for locales that require right to left languages and experiences.

## Example
LTR 
<img width="242" alt="ltr" src="https://user-images.githubusercontent.com/1861673/85058059-6cddcb00-b156-11ea-8f63-af244e45ecac.png">

RTL
<img width="216" alt="rtl" src="https://user-images.githubusercontent.com/1861673/85058062-6e0ef800-b156-11ea-85ee-70ffd08d9aec.png">

In the RTL example, implementor should be able to mirror the arrow or pencil icon to have the correct direction.

PR #949, opened 2021-01-23
feat(icon): Add prop for RTL icon mirroring

stefanuros:
Add enhancement for #751. This change allows for icons and buttons with icons to be mirrored (the icons themselves, not the whole button). This makes RTL transformations a bit easier for icons.

## Checklist

- [x] `yarn test` passes
- [ x code has been documented and, if applicable, usage described in README.md

### DIFF
--- a/modules/button/react/stories/stories_Button.tsx
+++ b/modules/button/react/stories/stories_Button.tsx
@@ -86,6 +86,10 @@ storiesOf('Components/Buttons/Button/React/Standard', module)
           Primary
         </Button>
       </div>
+      <h3>Mirrored Primary</h3>
+      <Button size="medium" variant={Button.Variant.Primary} icon={editIcon} mirrorIcon={true}>
+        Primary
+      </Button>
     </div>
   ))
   .add('Secondary', () => (
@@ -151,6 +155,11 @@ storiesOf('Components/Buttons/Button/React/Standard', module)
       <h3>Growing Secondary</h3>
       <Button size="large" variant={Button.Variant.Secondary} grow={true}>
         Growing Secondary
+      </Button>
+
+      <h3>Mirrored Secondary</h3>
+      <Button size="medium" variant={Button.Variant.Secondary} icon={editIcon} mirrorIcon={true}>
+        Secondary
       </Button>
     </div>
   ));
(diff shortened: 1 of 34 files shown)


modules/icon/react/lib/Svg.tsx
```diff
@@ -32,10 +37,18 @@ export default class Svg extends React.Component<SvgProps> {
...
+                styles,
+                {
+                  display: 'inline-block',
+                  '& svg': {display: 'block'},
+                },
+                mirror
+                  ? {
+                      transform: 'scaleX(-1)',
+                    }
+                  : {}
```
anicholls, line 50: Just a little cleaner:
```suggestion
                  display: 'inline-block',
                  transform: mirror ? 'scaleX(-1)' : undefined,
                  '& svg': {display: 'block'},
                },
```

modules/button/react/stories/stories_Button.tsx
```diff
@@ -86,6 +86,10 @@ storiesOf('Components/Buttons/Button/React/Standard', module)
           Primary
         </Button>
       </div>
+      <h3>Mirrored Primary</h3>
```
anicholls, line 89: Do you mind reverting your additions to all of the button stories and adding them to our visual testing stories instead?  ([see here](https://workday.github.io/canvas-kit/?path=/story/testing-react-buttons-button-button--button-states))

In [modules/button/react/stories/visual-testing/stories_Button.tsx](https://github.com/Workday/canvas-kit/blob/master/modules/button/react/stories/visual-testing/stories_Button.tsx#L33):
```tsx
        mirrorIcon: [
          {value: undefined, label: ''},
          {value: true, label: 'Mirrored Icon'},
        ],
```

You'll probably also want to make use of the [filter function](https://github.com/Workday/canvas-kit/blob/master/utils/storybook/ComponentStatesTable.tsx#L63) accepted by `permutateProps` so that this only adds rows when icon is defined:
```tsx
({icon, mirrorIcon}) => {
  if (mirrorIcon && !icon) {
    return false
  }
}
```

anicholls (commented):
Thanks for the contribution @stefanuros! 

This functionality should also be added to the `StatusIndicator` and `MenuItem` (labs) components. Do you mind adding to those as well?

stefanuros:
@anicholls Yes, I can add that! I won't be able to for the next few days but I will get it in asap.

alanbsmith:
Closing this now as it was addressed in #1246. Thanks for the PR!

PR #950, opened 2021-01-26
docs: Add compound Component Creation Doc

NicholasBoll:
Doc to create new compound components. This document builds on the doc introduced in #937. It is a step-by-step guide to building components from scratch with code sandbox examples between steps.

Rendered docs can be found here: https://github.com/Workday/canvas-kit/blob/67d88f7bf7a21d394e267e3bbefe6276950d9ef7/CREATING_COMPOUND_COMPONENTS.md

### DIFF
--- a/CREATING_COMPOUND_COMPONENTS.md
+++ b/CREATING_COMPOUND_COMPONENTS.md
@@ -390,7 +390,10 @@ The `DisclosureContent` component is in charge of the content. It uses the `open
 
 The working example can be found here: https://codesandbox.io/s/configurable-disclosure-model-components-nvhtv
 
-These components are not fully compliant yet. They do not support `ref`, `as`, or extra props as HTML attributes. The boilerplate to supporting all this gets very complicated. For this reason, a `createComponent` utility function was created to support all this out of the box. `createComponent` takes a default `React.ElementType` which can be an element string like `div` or `button` or a component like `Button`. It also takes a config object containing the `displayName`, `Component`, and any `subComponents`.
+These components are not fully compliant yet. They do not support `ref`, `as`, or extra props as HTML attributes. The boilerplate to supporting all this gets very complicated. For this reason, a `createComponent` utility function was created to support all this out of the box. `createComponent` takes a default `React.ElementType` which can be an element string like `div` or `button` or a component like `Button`. It also takes a config object containing the following:
+- `displayName`: This will be the name of the component when shown by the React Dev tools. By convention, we make that name be the same as typed in a render function. For example `Disclosure.Target` vs `DisclosureTarget`.
+- `Component`: A [forward ref component function](https://reactjs.org/docs/forwarding-refs.html) with an added `Element` property. `Element` is the value passed to the Component's `as` prop. It will default to the provided element.
+- `subComponents`: For container components. A list of sub components to add to the returned component. For example, a sub component called `DisclosureTarget` will be added to the export of `Disclosure` so that the user can import only `Disclosure` and use `Disclosure.Target`. `subComponents` is needed for Typescript because static properties cannot be added to predefined interfaces. `Disclosure.Target = DisclosureTarget` will caused a type error. This property allows the `createComponent` factory function to infer the final interface of the returned component.
 
 Let's convert the Disclosure example to use the `createComponent` utility function to get this extra functionality:
 
@@ -645,7 +648,7 @@ We'll build a behavior hook for the  `DisclosureTarget` component:
 // useExpandableControls.tsx
 import { DisclosureModel } from "./useDisclosureModel"
 
-export const useExpandablControls = (
+export const useExpandableControls = (
   { state }: DisclosureModel,
   elemProps: {}
 ) => {
@@ -664,7 +667,7 @@ Now we can use the behavior hook in the `DiscloseTarget` component:
 
 // ...
 const model = React.useContext(DisclosureModelContext)
-const props = useExpandablControls(model, elemProps)
+const props = useExpandableControls(model, elemProps)
 
 return (
   <Element
@@ -832,10 +835,10 @@ export const TooltipTarget = createComponent("button")({
   displayName: "Tooltip.Target",
   Component: ({ children, ...elemProps }: TooltipTargetProps, ref, Element) => {
     const model = React.useContext(TooltipModelContext)
-    const targetProps = useTooltipTarget(model, elemProps)
+    const props = useTooltipTarget(model, elemProps)
 
     return (
-      <Element ref={ref} {...targetProps}>
+      <Element ref={ref} {...props}>
         {children}
       </Element>
     )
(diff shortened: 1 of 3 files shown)


CREATING_COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,918 @@
...
+      <br />
+      Check the console output
+    </>
+  )
+}
+```
+
+You can see it in action here: https://codesandbox.io/s/basic-configurable-disclosure-model-nuteg
+
+That's a lot of extra boilerplate code for actions and callbacks. Our events don't have any data, but if they did, we'd have to keep the event + guard and callback data types in sync. Also are also creating the `events` object every render. We could use React refs and `React.useMemo` to decrease extra object creation. Luckily the common module has `createEventMap` and `useEventMap` functions to help us reduce boilerplate and reduce possibility of making mistakes.
```
anicholls, line 224:
```suggestion
That's a lot of extra boilerplate code for actions and callbacks. Our events don't have any data, but if they did, we'd have to keep the event + guard and callback data types in sync. We are also creating the `events` object every render. We could use React refs and `React.useMemo` to decrease extra object creation. Luckily the common module has `createEventMap` and `useEventMap` functions to help us reduce boilerplate and reduce possibility of making mistakes.
```

CREATING_COMPOUND_COMPONENTS.md
```diff
@@ -0,0 +1,918 @@
...
+  },
+  callbacks: {
+    onOpen: "open",
+    onClose: "close",
+    onToggle: "toggle"
+  }
+})
+```
+
+This part is a little weird: `createEventMap<DisclosureEvents>()({`. The reason for this is a Typescript issue: https://github.com/microsoft/TypeScript/issues/26242. The gist is a function with generics requires the caller to specify none of the generics or all of them. In this case, the only generic that cannot be inferred is the `DisclosureEvents`. Everything else can be inferred. The only way to separate defined generics vs inferred generics is to have separate functions.
```
anicholls, line 243:
```suggestion
This part is a little weird: `createEventMap<DisclosureEvents>()({`. The reason for this is a Typescript issue: https://github.com/microsoft/TypeScript/issues/26242. The gist is a function with generics requires the caller to specify none of the generics or all of them. In this case, the only generic that cannot be inferred is the `DisclosureEvents`. Everything else can be inferred. The only way to separate defined generics vs inferred generics is to have separate, chained functions.
```

anicholls (approved):
This is amazing and I 100% think it should be a blog post. Great job 👏 

One thing that's been irking me - I feel like we'll have some confusion around the difference between behavioral hooks and components that are named the same thing but intended to do different things. Examples:
-  A `Disclosure` component vs. `useDisclosureModel` that could be used within any component that opens and closes
- A `Menu` component vs. `useMenuModel` that could be used within any component that doesn't remotely look like a menu
- etc.

I can't think of a good way around it, but I feel like it will be a sticking point for some folks.

NicholasBoll:
> One thing that's been irking me - I feel like we'll have some confusion around the difference between behavioral hooks and components that are named the same thing but intended to do different things. Examples:
> 
> * A `Disclosure` component vs. `useDisclosureModel` that could be used within any component that opens and closes
> * A `Menu` component vs. `useMenuModel` that could be used within any component that doesn't remotely look like a menu
> * etc.

Do you mean models and components? Hooks with 'Model' in the name are meant to return models. There are behavior hooks like `useMenu` and a component called `Menu`. I get that. We were trying to decide what a component is vs behavior hooks. 

- 100% mapping of behavior hook to component. Like `useMenu` would provide all the attributes to a menu element.
  - The benefit of this is every component can have a composable interface:
    - `<Menu as={SomeComponent}>`
    - `<Menu>{(props, model) => <SomeComponent {...props} data-some-attribute={model.state.foo}>}</Menu>`
  - There are downsides to this though, mainly that components don't really exist anymore - they are just hooks
- There is a looser relationship between components and behavior hooks.
  - We already use this heavily in our Popup system. For example, `useCloseOnEscape`, `useCloseOnOutsideClick`, etc
  - This has the benefit of greater reuse of behaviors in different components and makes building custom components more like piecing legos together

A good point you make might be the name of the hooks. Like `useMenu` might not be a great name since it is describing the keyboard interaction rather than a menu.

CREATING_COMPOUND_COMPONENTS.md
```diff
@@ -517,7 +521,7 @@ The full code can be found here: https://codesandbox.io/s/configurable-disclosur
 
 ## Model Composition
 
-Our example isn't fully accessible yet. The Disclosure target needs a `aria-controls` attribute to tie the target and content in the accessibility tree. This is done by the use of IDREFs (string IDs that start with a letter). We could add an  `id` to our model, but it is extremely common so let's make a new model and compose from it instead. We'll later use this model in a reusable behavioral hook.
+Our example isn't fully accessible yet. The Disclosure target needs a `aria-controls` attribute to tie the target and content in the accessibility tree. This is done by the use of IDREFs (string IDs that starts with a letter). We could add an  `id` to our model, but it is extremely common so let's make a new model and compose from it instead. We'll later use this model in a reusable behavioral hook.
```
willklein, line 524: I think `start` may have been the appropriate tense?

CREATING_COMPOUND_COMPONENTS.md
```diff
@@ -638,10 +642,10 @@ Behavior hooks allow us to reuse pieces of functionality in difference component
 We'll build a behavior hook for the  `DisclosureTarget` component:
 
 ```tsx
-// useDisclosureTarget.tsx
+// useExpandablControls.tsx
```
willklein, line 645: Spelling: `useExpandableControls`

CREATING_COMPOUND_COMPONENTS.md
```diff
@@ -834,15 +867,15 @@ export const TooltipContent = createComponent("div")({
     Element
   ) => {
     const model = React.useContext(TooltipModelContext)
+    const props = useHidden(model, elemProps)
```
willklein, line 870: I love this.

willklein (approved):
Optional suggestions made!

CREATING_COMPOUND_COMPONENTS.md
NicholasBoll: I think it is "string ID that starts" or "string IDs that start"

CREATING_COMPOUND_COMPONENTS.md
NicholasBoll:
```suggestion
// useExpandableControls.tsx
```

NicholasBoll:
@alanbsmith Some feedback I got was the `subComponents` property in the `createComponent` factory function was confusing. I also want to use it for more than sub components. I'd like to rename it to `staticProperties` instead since that is more accurate. For example, `Button` exports `Size` and `Variant` static properties and that's where those would go.

alanbsmith:
I’m curious what’s confusing about `subComponents`, but I think that’s a fair tradeoff to make it more generic with `staticProperties`.  I don’t think `Variant` and `Size` are good long-term examples as we want to move away from them, but there are other static properties we could add.

PR #953, opened 2021-01-28
Feat/tabs component composable state

### LINKED ISSUE #910: Tabs Phase 2: Single List
Update Tabs to support the following features:

## Overflow
If there isn't enough space in the container, additional tabs should overflow into a menu:
Tabs overflowing into a menu

##  Controllable State
Add ability for developers to observe and/or control state.

NicholasBoll:
Fixes #910 

## Summary

Update the Tabs component to be an official Compound Component using models, behavior hooks, and compound component utility functions.

## Breaking Changes
- `onTabsChange` is now `onActivateTab` and the signature is now:
  ```tsx
  function onActivateTab({ data: { tab: string }, state: TabsState }): void
  ```
- The `<Tabs>` component no longer accepts the `currentTab` property. The Tabs uses a model now. See story for more details


cypress/integration/Tabs.spec.ts
```diff
@@ -4,13 +4,14 @@ describe('Tabs', () => {
...
-  ['Simple', 'NamedKeys'].forEach(story => {
+
+  ['Simple', 'Named Keys'].forEach(story => {
     context(`given the '${story}' story is rendered`, () => {
       beforeEach(() => {
         h.stories.load('Testing/React/Labs/Tabs', story);
       });
 
-      it('should pass accessibility checks', () => {
+      it('should pass axe checks', () => {
```
NicholasBoll, line 14: We should probably change all these to specifically call out "axe" vs "accessibility". We're checking "axe" here. The rest of the file checks accessibility.

cypress/integration/Tabs.spec.ts
```diff
@@ -4,13 +4,14 @@ describe('Tabs', () => {
...
         cy.checkA11y();
```
NicholasBoll, line 15: I might change the name here to `cy.checkAxe`. Axe does not check all accessibility issues, mostly low-hanging fruit or obvious issues.

modules/_labs/tabs/react/index.ts
```diff
@@ -1,7 +1,7 @@
-import Tabs from './lib/Tabs';
-import TabList from './lib/TabList';
-import Tab from './lib/Tab';
-import TabPanel from './lib/TabPanel';
+import {Tabs} from './lib/Tabs';
+import {TabList} from './lib/TabList';
+import {Tab} from './lib/Tab';
+import {TabPanel} from './lib/TabPanel';
```
NicholasBoll, line 4: We haven't talked about this one, but there is so much extra boilerplate in default exports. How are they serving us now? Components like 'Button' have to decide which button export is "default" and we have to remap default exports of almost every file that uses them. Why not always use named exports?

cypress/integration/Tabs.spec.ts
willklein: I agree with this idea.

modules/_labs/tabs/react/lib/Tab.tsx
```diff
@@ -71,59 +88,36 @@ const StyledButton = styled('button')<{isSelected: boolean}>(
...
+      {
+        id: `tab-${state.id}-${tabName}`,
+        tabIndex: !!tabName && state.currentId === tabName ? undefined : -1,
+        'aria-selected': isSelected ? true : undefined,
+        'aria-controls': `tabpanel-${state.id}-${tabName}`,
+        onClick: onSelect,
+        isSelected: isSelected,
+      },
+      elemProps
+    );
```
NicholasBoll, line 115: Every prop list that uses a callback needs to use `mergeProps` to make sure callbacks are handled properly. This example could have use only the `onClick` instead of all the other props... Not 100% on this one.

modules/_labs/tabs/react/lib/TabList.tsx
```diff
@@ -1,16 +1,25 @@
...
 
-export interface TabListProps extends React.HTMLAttributes<HTMLElement> {
-  /**
-   * A list of Tab components.
+export interface TabListProps {
+  children: React.ReactNode;
+  /** Optionally pass a model directly to this component. Default is to implicitly use the same
+   * model as the container component which uses React context. Only use this for advanced use-cases
    */
-  children: Tab | Tab[];
```
NicholasBoll, line 13: We should not restrict children. It doesn't work very well anyways and it is perfectly valid to do the following:
```
<Tabs>
  <div> // Could add padding for the panels
    <Tabs.Panel>
```

modules/_labs/tabs/react/lib/TabList.tsx
```diff
@@ -20,58 +29,52 @@ const TabsListContainer = styled('div')({
...
-        onFocus={resetProgrammaticFocus}
-        {...elemProps}
-      >
-        {children}
-      </TabsListInnerContainer>
-    </TabsListContainer>
+/**
+ * Reset the cursor to the active tab when the tab list looses focus
+ */
+const useResetCursorOnBlur = ({state, events}: TabsModel, elemProps = {}) => {
```
NicholasBoll, line 40: I'm not 100% this is a sharable hook. It depends on the `state.activeTab` which is specific to the Tabs model. If we change `activeTab` to be `activeId` or `activeCursor`, it might apply to something like a Select as well which also resets cursors when focus is lost.

package.json
```diff
@@ -59,10 +59,10 @@
     "cssstats": "^3.3.0",
     "cypress": "^4.4.1",
     "cypress-axe": "^0.5.1",
-    "cypress-enter-plugin": "^1.0.0",
+    "cypress-keyboard-plugin": "^1.0.2",
```
NicholasBoll, line 62: The keyboard plugin includes more than the enter key

modules/_labs/select/react/lib/Select.tsx
```diff
@@ -488,7 +488,8 @@ class Select extends React.Component<SelectProps, SelectState> {
         case 'Enter':
           isShortcut = true;
           if (menuVisibility === 'closed' || menuVisibility === 'closing') {
-            this.toggleMenu(true);
+            // allow the click
+            isShortcut = false;
```
NicholasBoll, line 492: Fixes this error: https://dashboard.cypress.io/projects/odida5/runs/2269/test-results/b8041582-848c-4085-ae92-85e6be37458e

> `event.preventDefault()` was called on this element or a parent element and no additional functionality' simulated default actions. This prevents keyboard accessibility on <button> elements. Read more at https://github.com/cypress-io/cypress/issues/311. Use `{force: true}` to disable this check

cypress/integration/SelectLabs.spec.ts
```diff
@@ -304,7 +304,7 @@ describe('Select', () => {
 
         context('when the space key is pressed', () => {
           beforeEach(() => {
-            cy.findByLabelText('Label').type(' ');
+            cy.findByLabelText('Label').type(' ', {force: true}); // disable event.preventDefault checks
```
NicholasBoll, line 307: Not 100% sure about this. The keyboard plugin does allow disabling of `defaultPrevented` checks if `force: true` is passed. This technically bypasses other checks as well like visibility or interactability, but those checks are covered in other tests. This is required because the Select uses a `<button>` element, but doesn't handle the space bar the same way a button does. A button triggers click on keyup of a spacebar, which a select triggers a click on the keydown of a spacebar key event.

modules/_labs/tabs/react/lib/TabList.tsx
```diff
@@ -20,58 +29,50 @@ const TabsListContainer = styled('div')({
...
+    },
+    elemProps
   );
 };
 
-export default TabList;
+export const TabList = createComponent('div')({
+  displayName: 'Tabs.List',
+  Component: ({children, model, ...elemProps}: TabListProps, ref, Element) => {
+    // const [tabIndicatorRef, setDimensions] = useIndicator(tabsListRef);
```
anicholls, line 66:
```suggestion
    // const [tabIndicatorRef, setDimensions] = useIndicator(tabsListRef);
```

modules/_labs/tabs/react/lib/Tabs.tsx
```diff
@@ -1,210 +1,29 @@
...
-   */
-  initialTab?: string;
-  /**
-   * Callback when a tab changes. The `name` will be the `name` prop passed into the `Tabs.Item` and
-   * `Tabs.Panel` component. If a `name` isn't provided, the value will be a string of the index of
-   * the tab.
-   */
-  onTabChange?: (name: string) => void;
+export interface TabsProps extends TabsModelConfig {
+  model?: TabsModel;
```
anicholls, line 13: docs?

anicholls (commented):
Minor stuff. Looking good 👍  What's the plan for adding a transition to the indicator?

NicholasBoll:
> Minor stuff. Looking good 👍 What's the plan for adding a transition to the indicator?

@anicholls My plan is to use a transition element that is only visible during the animation. This will fix the issue where the indicator is hard to calculate based on font-rendering, RTL, container resizes, window resizes, etc.

modules/_labs/tabs/react/lib/cursor/hooks.tsx
```diff
@@ -0,0 +1,61 @@
...
+    End: 'last',
+  },
+} as const;
+
+/**
+ * Handles the roving focus behavior of a Cursor model. It does not handle the tabIndex. Use
+ * `useRovingTabIndex` for that functionality.
+ * @see https://www.w3.org/TR/wai-aria-practices/#kbd_roving_tabindex
+ */
+export const useRovingFocus = ({state, events}: CursorModel, elemProps = {}) => {
```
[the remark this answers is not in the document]
NicholasBoll: Behavior hooks are tied to models, so this will go where the `CursorModel` goes in the future. Our packages are currently tied to components and not to more abstract concepts (except common). Perhaps `List` and `Cursor` stay abstract concepts and common makes sense.

modules/common/react/lib/utils/index.ts
```diff
@@ -2,8 +2,9 @@ export * from './colorUtils';
 export * from './getTranslateFromOrigin';
 export * from './makeMq';
 export * from './mergeCallback';
-export * from './mergeCallbacks';
```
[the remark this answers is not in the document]
NicholasBoll: `mergeCallbacks` was never documented and only used by Tabs. Should we still add it to the migration guide?

modules/_labs/tabs/react/stories/stories_testing.tsx
```diff
@@ -22,24 +22,26 @@ export default {
...
-    <div css={{marginTop: spacing.m}}>
-      <Tabs.Panel>Contents of First Tab</Tabs.Panel>
-      <Tabs.Panel>Contents of Second Tab</Tabs.Panel>
-      <Tabs.Panel>Contents of Third Tab</Tabs.Panel>
-      <Tabs.Panel>Contents of Fourth Tab</Tabs.Panel>
-      <Tabs.Panel>Contents of Fifth Tab</Tabs.Panel>
-    </div>
-  </Tabs>
-);
+export const Simple = () => {
```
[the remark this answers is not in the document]
NicholasBoll: I don't think our goal is dry test stories, but ensuring components work according to specifications.

We haven't solidified what we'll do with stories yet. We want to have examples that work well with component documentation and stories that are run with tests. Ideally I'd like to test examples to ensure our examples always work, but it blurs the line between where a story goes. Do we try to put all stories in the normal doc location first and only put stories that have specific edge cases that don't match our documentation or would be confusing in a test file instead?

modules/_labs/tabs/react/index.ts
```diff
@@ -1,8 +1,10 @@
...
 
-export {Tabs, Tab, TabList, TabPanel};
+export {Tabs};
 export default Tabs;
 export * from './lib/Tabs';
+
+export * from './lib/TabList';
+export * from './lib/Tab';
+export * from './lib/TabPanel';
+export * from './lib/useTabsModel';
```
alanbsmith, line 10: TODO: Update imports

modules/_labs/tabs/react/stories/stories_testing.tsx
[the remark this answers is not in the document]
anicholls: I guess I was just thinking about this in the context of visual testing, not using this for functional testing. I'm fine either way, just thought that it's less maintenance to re-use if they're identical

.storybook/stories/compound_components.stories.mdx
```diff
@@ -0,0 +1,7 @@
+import {Meta} from '@storybook/addon-docs/blocks';
+import Contributing from '../../COMPOUND_COMPONENTS.md';
+import Markdown from '../../utils/storybook/Markdown.tsx';
+
+<Meta title="Welcome/Compound Components" />
+
+<Markdown content={Contributing} />
```
NicholasBoll, line 7: This adds the compound component docs to Storybook

.storybook/webpack-loader-redirect-mdx-to-github.js
```diff
@@ -1,18 +1,15 @@
...
-    // We can't determine the root directory of the project since everyone might set a different base URL
-    // The most common relative linking is between modules, so we'll split on that
-    const [_, modulePath] = dir.split('modules');
-    const newPath = modulePath
-      ? '/modules' + path.resolve(`${modulePath}`, p2)
-      : path.resolve('/', p2);
+    const newPath = path.relative(basePath, path.resolve(dir, p2));
 
-    return `[${p1}](https://github.com/Workday/canvas-kit/blob/master${newPath})`;
+    return `[${p1}](https://github.com/Workday/canvas-kit/blob/master/${newPath})`;
```
NicholasBoll, line 13: I noticed that some links didn't work and that the logic could be simpler

modules/_labs/tabs/react/README.md
```diff
@@ -4,7 +4,7 @@
   <img src="https://img.shields.io/badge/LABS-beta-orange" alt="LABS: Beta" />
 </a>  This component is work in progress and currently in pre-release.
 
-Composable tab components
+`Tabs` is a [Compound Components](../../../../COMPOUND_COMPONENTS.md).
```
anicholls, line 7:
```suggestion
`Tabs` is a [Compound Component](../../../../COMPOUND_COMPONENTS.md).
```

modules/_labs/tabs/react/stories/Tabs.stories.mdx
```diff
@@ -0,0 +1,174 @@
...
+
+```sh
+yarn add @workday/canvas-kit-labs-react-tabs
+```
+
+## Usage
+
+This component follows the
+[W3 Tabs specification](https://www.w3.org/TR/wai-aria-practices/#tabpanel). `Tabs` is a
+[Compound Components](../../../../COMPOUND_COMPONENTS.md), which means it can be composed in a
```
anicholls, line 30:
```suggestion
[Compound Component](../../../../COMPOUND_COMPONENTS.md), which means it can be composed in a
```

anicholls (approved):
I believe disabled tabs are missing the appropriate style change. Haven't looked at the spec though, maybe I'm wrong. Otherwise this looks good :shipit:

NicholasBoll:
> I believe disabled tabs are missing the appropriate style change. Haven't looked at the spec though, maybe I'm wrong. Otherwise this looks good :shipit:

Good catch. I've updated and added a visual regression with a disabled tab.

modules/_labs/tabs/react/lib/cursor/useCursorModel.tsx
```diff
@@ -0,0 +1,149 @@
...
+  /** Orientation of the list. This property determines what "next" and "previous" mean. This should
+   * be set to match the layout of UI components */
+  orientation: Orientation;
+  /** The id of the list item the cursor is pointing to */
+  cursorId: string;
+};
+
+export type CursorEvents = ListEvents & {
+  /** Directly sets the cursor to the list item by its identifier. Bypasses any checks  */
+  setCursorId(data: {id: string}): void;
```
alanbsmith, line 31: We talked about this being `goTo`, but I don't think we landed on anything. Your call.

modules/_labs/tabs/react/package.json
```diff
@@ -36,11 +36,11 @@
   ],
   "dependencies": {
     "@emotion/core": "^10.0.28",
-    "@emotion/styled": "^10.0.27",
     "@workday/canvas-kit-react-common": "^5.0.0-beta.0",
     "@workday/canvas-kit-react-core": "^5.0.0-beta.0"
   },
   "devDependencies": {
+    "@workday/canvas-kit-react-button": "^5.0.0-beta.0",
```
alanbsmith, line 43: ah, so you did add it.

modules/_labs/tabs/react/stories/Tabs.stories.mdx
```diff
@@ -0,0 +1,181 @@
...
+example has a tab panel with a focusable button in the tab panel.
+
+<ExampleCodeBlock code={AlternativeTabStop} />
+
+### Advanced Usage
+
+Below are some examples for more advanced / custom use cases. We expect most people won't need to
+read this section, but if you're needing to go beyond basic examples or are curious, feel free to
+explore this section. While these examples are not exhaustive, they provide additional insight into
+what's possible.
```
alanbsmith, line 78: This is a great paragraph. 😉

modules/common/react/lib/utils/index.ts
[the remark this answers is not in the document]
alanbsmith: I doubt it was used. You might just put a one-liner saying it's a name change.

alanbsmith (approved):
I left a few comments, but this is good to go. Great work! 🥇

modules/_labs/tabs/react/package.json
NicholasBoll: It hurt

modules/_labs/tabs/react/stories/Tabs.stories.mdx
NicholasBoll: Familiar too... 🤔

modules/_labs/tabs/react/lib/cursor/useCursorModel.tsx
NicholasBoll: Yes. I think `goTo` would work. It is hard to come up with a name that isn't too verbose, yet works in multiple contexts. This event will set the cursor to a specific point in the list, which is used by the Tabs model. In this context, `goTo` will set the focus cursor while `activateTab` will set the active tab.

Should `activateTab` be changed to `activate` and assume the `Tab` context is implied? It can be difficult with model composition to make too many assumptions about the context in which the events will be called.

Should we always or never explicitly specify model context?

Explicit model context:
- `registerListItem` - from `ListModel`
- `goToCursor` - from `CursorModel`
- `activateTab` - from `TabsModel`

Implicit model context:
- `registerItem` - from `ListModel`
- `goTo` - from `CursorModel`
- `activate` - from `TabsModel`

cypress/integration/Popup.spec.ts
```diff
@@ -113,8 +113,7 @@ describe('Popup', () => {
           cy.findByLabelText('Popup 2').should('be.visible');
         });
 
-        // TODO Skip for now until we have a systematic approach to fix this issue
-        it.skip('should close Popup 1', () => {
+        it('should close Popup 1', () => {
```
NicholasBoll, line 116: Apparently it is fixed.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/MenuItemLink.tsx
```diff
@@ -33,7 +33,7 @@ const MenuItemLink = styled('a')({
 export const DropdownMenuItemLink = forwardRef(
   (
     {children, href, ...elemProps}: DropdownMenuItemLinkProps,
-    ref: React.RefObject<HTMLAnchorElement>
+    ref: React.Ref<HTMLAnchorElement>
```
NicholasBoll, line 36: This was caught by `strictFunctionTypes`

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/List/hooks.ts
```diff
@@ -50,7 +50,10 @@ export const useBuildCollapsedList = <E extends HTMLElement>(
     const listItemNodes = listEl.current ? listEl.current.querySelectorAll('li') : [];
     const containerWidth = listEl.current ? listEl.current.clientWidth : 0;
     const listItems: Breadcrumb[] = [];
-    Children.forEach(children, (child: React.ReactElement, index) => {
+    Children.forEach(children, (child, index) => {
+      if (!React.isValidElement(child)) {
+        return;
+      }
```
NicholasBoll, line 56: This was caught by `strictFunctionTypes`. I think this is a valid issue

modules/_labs/pagination/react/lib/Pagination/types.ts
```diff
@@ -9,7 +7,7 @@ export interface PaginationState {
 }
 
 export interface PaginationEvents {
-  setCurrentPage: React.Dispatch<React.SetStateAction<number>>;
+  setCurrentPage: (page: number) => void;
```
NicholasBoll, line 10: This was a valid catch by `strictFunctionTypes`

modules/_labs/menu/react/lib/Menu.tsx
```diff
@@ -132,7 +132,10 @@ export default class Menu extends React.Component<MenuProps, MenuState> {
           ref={this.menuRef}
           {...elemProps}
         >
-          {React.Children.map(children, (menuItem: React.ReactElement<MenuItemProps>, index) => {
+          {React.Children.map(children, (menuItem, index) => {
+            if (!React.isValidElement(menuItem)) {
+              return;
+            }
```
NicholasBoll, line 138: Valid catch by `strictFunctionTypes`

modules/_labs/tabs/react/lib/cursor/hooks.tsx
```diff
@@ -55,15 +55,15 @@ export const useRovingFocus = ({state, events}: CursorModel, elemProps = {}) =>
...
+        (Object.keys(
+          orientationKeyMap[state.orientation]
+        ) as (keyof typeof orientationKeyMap[typeof state.orientation])[]).forEach(key => {
+          if (isRTL ? event.key === rightToLeftMap[key] : event.key === key) {
+            const event = orientationKeyMap[state.orientation][key];
+            focusRef.current = true;
+            events[event]?.();
           }
-        );
+        });
```
NicholasBoll: `Object.keys` returns `string[]` vs `(keyof T)[]` which is why we have to cast. Sometimes I make a simple utility function with the correct signature. Why couldn't they just update `Object.keys`?

modules/_labs/tabs/react/lib/cursor/useCursorModel.tsx
```diff
@@ -114,7 +115,7 @@ export const useCursorModel = (config: CursorModelConfig = {}): CursorModel => {
   const [cursorId, setCursorId] = React.useState('');
   const initialCurrentRef = React.useRef('');
   const list = useListModel({
-    ...config,
+    ...(config as ListModelConfig),
```
NicholasBoll, line 118: Narrowing the type here prevents `strictFunctionTypes` errors on the `...config` object having covariant errors between `CursorModelConfig` and `ListModelConfig` where `config` technically has callback and guard types that accept `CursorState` instead of `ListState`. Bivariance says this is okay since `CursorState` extends from `ListState`, but contravariance says it isn't. This casting prevents the need for bivariance hacks. A type error would occur if the casting was invalid, but I think this is a valid tradeoff.

NicholasBoll: And if you don't fully understand covariance, contravariance, and bivariance, that's okay. I don't either.

modules/common/react/lib/utils/components.ts
```diff
@@ -166,7 +166,7 @@ export function useForkRef<T>(ref1: React.Ref<T>, ref2: React.Ref<T>): (instance
  *   return <Element ref={elementRef} {...elemProps} />
  * }
  */
-export function useLocalRef<T>(ref: React.Ref<T>) {
+export function useLocalRef<T>(ref?: React.Ref<T>) {
```
NicholasBoll, line 169: I made these refs be optional in Typescript because they are technically optional. `strictFunctionTypes` surfaced this need.

modules/_labs/tabs/react/lib/cursor/useCursorModel.tsx
NicholasBoll: More info here: https://www.stephanboyer.com/post/132/what-are-covariance-and-contravariance

modules/common/react/lib/utils/models.ts
```diff
@@ -5,14 +5,17 @@ export type Model<State, Events> = {
   events: Events;
 };
 
+// bivarianceHack is used for force bivariance of the function constraint. Without this, it will fail with `strictFunctionTypes`
+type IEvent = {[key: string]: {bivarianceHack(data?: object): void}['bivarianceHack']};
```
NicholasBoll, line 9: This bivariance hack is the same as what's used in React. This is a case for valid bivariance - I want to set a constraint of an `Events` object. It must be a record of functions that either accept an object or nothing. This is to make callback and guard signatures predictable through the `createEventMap` and `useEventMap` utility functions. `strictFunctionTypes` complain about this bivariance though since a real event will not have a signature of `(data: object) => void` and will look like `(data: { foo: string }) => void` which is a contravariance violation. But it is valid to set up constraints in this case, so a bivariance hack is necessary.

Here's a simplified example of what's going on: [Typescript Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAYgrgOwMZQLxQN4CMCWA3AQwCccDkIAJApAawAoATA4AgfgC4oB7LAKwiTAAlJzxccDKAF8A2gCJchEmSSVqNOQF0AUAHpdoSLEQp0jZgU49+goWgB8UMRKj6oSAnADOEL+64IwEQESqTkTjhcADbMkQhQAO44wAAWUF5BOILwyMBxACrgvtraAGYmeQHuRBDMEDlIADz5UBAAHsAQCAx+DfZ0pQic+XYY2lBQNcBwRPGD2lIlSAEZ1bWdDA1oa3UNdOYsnBhQpVxcnBkkCADm0naojmMTUzMIC0JLK8BQALYgMNRgFwiCBts1Wh0uj1uHwBMB+jgEGA4MBhvdHuNJhBprMoIjkcAFp8EKtOj8wNs-gDBMCQHQMKdzlAAORYYjMqRCIA)

modules/popup/react/lib/Popper.tsx
```diff
@@ -65,10 +65,7 @@ export interface PopperProps extends React.HTMLAttributes<HTMLDivElement> {
 }
 
 export const Popper = React.forwardRef<HTMLDivElement, PopperProps>(
-  ({portal = true, open = true, ...elemProps}: PopperProps, forwardRef) => {
-    const localRef = React.useRef<HTMLDivElement>(null);
-    const ref = (forwardRef || localRef) as React.RefObject<HTMLDivElement>;
-
```
NicholasBoll, line 71: This was a whoops. `forwardRef` could be a function if it was used in a Class component. In this case, `forwardRef` would be type narrowed to a `React.RefObject` (subtype of `React.Ref`). This was caught as a contravariant violation on line 122 which is totally valid. It required me to fix this possible runtime error by refactoring how `usePopupStack` works with refs and accept `React.Ref` instead of only `React.RefObject`.

modules/popup/react/lib/Popper.tsx
```diff
@@ -119,12 +116,12 @@ const OpenPopper = React.forwardRef<HTMLDivElement, PopperProps>(
       containerElement,
       ...elemProps
     },
-    forwardRef: React.RefObject<HTMLDivElement>
+    ref
   ) => {
     const firstRender = React.useRef(true);
     const popperInstance = React.useRef<PopperJS.Instance>();
     const [placement, setPlacement] = React.useState(popperPlacement);
-    const stackRef = usePopupStack(forwardRef, anchorElement as HTMLElement);
```
NicholasBoll, line 127: Possible breaking change. Before, if `forwardRef` was defined, `stackRef` would be the same reference. Now they will always be different references since `usePopupStack` now uses forked refs. This is much more safe.

modules/popup/react/lib/usePopupStack.ts
```diff
@@ -40,23 +41,22 @@ import {PopupStack} from '@workday/canvas-kit-popup-stack';
...
  */
 export const usePopupStack = <E extends HTMLElement>(
-  forwardRef?: React.RefObject<E>,
+  ref?: React.Ref<E>,
   target?: HTMLElement | React.RefObject<HTMLElement>
-): React.RefObject<HTMLDivElement> => {
-  const internalRef = React.useRef<HTMLDivElement>(null);
-  const ref = (forwardRef || internalRef) as React.RefObject<HTMLDivElement>;
+): React.RefObject<HTMLElement> => {
+  const {elementRef, localRef} = useLocalRef(ref);
```
NicholasBoll, line 47: `useLocalRef` is a convenience wrapper around `useForkRef` that extracts types from the provided ref and returns accurate types that are more clear on the intent.

modules/popup/react/lib/usePopupStack.ts
```diff
@@ -40,23 +41,22 @@ import {PopupStack} from '@workday/canvas-kit-popup-stack';
  * useCloseOnEscape(stackRef, closePopup)
  */
 export const usePopupStack = <E extends HTMLElement>(
-  forwardRef?: React.RefObject<E>,
+  ref?: React.Ref<E>,
```
NicholasBoll, line 44: I'm not 100% on what to do here. I really hate that `div` elements are `HTMLDivElement` instead of `HTMLElement` which causes all kinds of type incompatibilities when you use any tag other than a `div`.

The `PopupStack` is in charge of creating the element and it makes a `div`. That ref is never meant to be applied to a React element and this will throw an error if you even attempt it. The `RefObject` returned by this function will be used by all behavior hooks that are meant to be used with stacks and is the `RefObject` used by a React Portal.

modules/common/react/lib/utils/models.ts
mannycarrera4: i really appreciate the description here plus the playground!

modules/popup/react/lib/usePopupStack.ts
```diff
@@ -66,12 +66,12 @@ export const usePopupStack = <E extends HTMLElement>(
...
       : undefined;
-    PopupStack.add({element: ref.current!, owner: targetEl});
-    const element = ref.current!;
+    PopupStack.add({element: localRef.current!, owner: targetEl});
+    const element = localRef.current!;
     return () => {
       PopupStack.remove(element);
     };
-  }, [ref, target, popupRef]);
+  }, [localRef, target, popupRef]);
```
NicholasBoll, line 74: I don't think refs are technically needed anymore by the lint rule if they are locally defined, but it doesn't hurt other than unnecessary reference comparisons.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/MenuItemLink.tsx
alanbsmith: whoops. 😬

modules/popup/react/lib/usePopupStack.ts
```diff
@@ -40,23 +41,22 @@ import {PopupStack} from '@workday/canvas-kit-popup-stack';
  * useCloseOnEscape(stackRef, closePopup)
  */
 export const usePopupStack = <E extends HTMLElement>(
```
NicholasBoll, line 43: I'm not sure I need a generic here at all... This mostly prevents type issues for consumers if they use the wrong ref type...

modules/tooltip/react/spec/Tooltip.spec.tsx
```diff
@@ -1,5 +1,3 @@
-/// <reference types="@testing-library/jest-dom/extend-expect" />
```
NicholasBoll, line 1: Automatically handled now and this manual reference causes TS issues

alanbsmith (approved):
🦅

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/Dropdown/MenuItemLink.tsx
NicholasBoll: No runtime issue, just not the correct type. It was narrowed by the type override. Nothing depending on this `ref` required a `RefObject` or assumed it was. But it was nice `strictFunctionTypes` caught this as a potential problem.

PR #954, opened 2021-01-29
Update Pagination component

alanbsmith:
## Checklist

- [ ] `yarn test` passes
      applicable


modules/_labs/pagination/react/lib/Pagination/Controls.tsx
```diff
@@ -0,0 +1,144 @@
...
+import {
+  chevronLeftSmallIcon,
+  chevron2xLeftSmallIcon,
+  chevronRightSmallIcon,
+  chevron2xRightSmallIcon,
+} from '@workday/canvas-system-icons-web';
+
+import {PaginationModel} from './types';
+import {HStack} from './common/Stack';
+import {useRTL} from './common/utils/useRTL';
```
alanbsmith, line 12: This is not a true hook. It shouldn't use the `use` prefix.

cypress/integration/Pagination.spec.ts
```diff
@@ -1,293 +1,363 @@
 import * as h from '../helpers';
 
-describe('Button', () => {
+function getPaginationNav() {
+  return cy.findByLabelText('Pagination');
+}
```
NicholasBoll, line 5: I'm a huge ball of meh on these types of functions. I personally don't use them as they create an abstraction from the Command Log in a Cypress test making it a bit harder to match up the line that failed. There are 2 problems with this:
1. Command log doesn't match what's in a test. The command log shows `findByLabelText Pagination` while the test shows a `getPaginationNav` function
2. Cypress shows the exact line that fails in a test. It will show this line rather than inside an `it` or `beforeEach` function call

These are definitely not deal breakers, but do aid to slow down the debug process. The only upside of this abstraction is reducing duplication which is not a concern I have in test files.

NicholasBoll: There have been times I've needed to update spec code because a label changed or something. A find/replace still worked.

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,555 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+
+```tsx
+const ControllablePagination = () => {
+  // create the model
+  const model = usePaginationModel({
+  lastPage={100}
+  initialCurrentPage={6}
+  rangeSize={3}
+  onPageChange={pageNumber => console.log(pageNumber)}
+});
```
NicholasBoll, line 243: Spacing is off here:

```suggestion
  const model = usePaginationModel({
    lastPage={100}
    initialCurrentPage={6}
    rangeSize={3}
    onPageChange={pageNumber => console.log(pageNumber)}
  });
```

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,555 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+```
+
+Notice that when providing a model, you no longer need to provide the other props (other than
+`aria-label`). Let's look at a complete example where you might want to use this controllable
+pattern.
+
+First, we'll create a table that will fetch some mock data:
+
+```tsx
+```
```
NicholasBoll, line 260: > This example intentionally left blank

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,555 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+
+## Components
+
+### Pagination
+
+The `Pagination` component is a wrapper for a React context provider and `nav` element. Child
+components such as `Pagination.StepToPreviousButton`, `Pagination.PageList`, etc. subscribe to that
+context.
```
NicholasBoll, line 268: Do we want to talk about this as the `model`? In this doc, this is referred to as "`Pagination`'s context` vs "`PaginationModel` context". The context is the reference to the shared model. The language used here is simpler. I've been trying to carefully use the same language for these things to avoid confusion (i.e. What's the difference between the model and the context?)

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,555 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+    ))
+  }
+</Pagination.PageList>
+```
+
+#### Component Props
+
+| name     | type                                                              | required?  | default | recommended |
+| -------- | ----------------------------------------------------------------- | ---------- | ------- | ----------- |
+| children | `(model: PaginationModel) => React.ReactNode[] | React.ReactNode` | 🚫 `false` | n/a     | n/a         |
```
NicholasBoll, line 421:
```suggestion
| name     | type                                                              | required?  | default | recommended |
| -------- | ----------------------------------------------------------------- | ---------- | ------- | ----------- |
| children | `(model: PaginationModel) => React.ReactNode[] \| React.ReactNode` | 🚫 `false` | n/a     | n/a         |
```
renders:
| name     | type                                                              | required?  | default | recommended |
| -------- | ----------------------------------------------------------------- | ---------- | ------- | ----------- |
| children | `(model: PaginationModel) => React.ReactNode[] \| React.ReactNode` | 🚫 `false` | n/a     | n/a         |

vs:
| name     | type                                                              | required?  | default | recommended |
| -------- | ----------------------------------------------------------------- | ---------- | ------- | ----------- |
| children | `(model: PaginationModel) => React.ReactNode[] | React.ReactNode` | 🚫 `false` | n/a     | n/a         |

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,555 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+```tsx
+<Pagination.GoToLabel>{({state}) => `of ${state.lastPage} results`}</Pagination.GoToLabel>
+```
+
+#### Component Props
+
+| name     | type                                                            | required?  | default | recommended |
+| -------- | --------------------------------------------------------------- | ---------- | ------- | ----------- |
+| htmlFor  | `string`                                                        | 🚫 `false` | n/a     | n/a         |
+| children | `(model: PaginationModel) => React.ReactNode | React.ReactNode` | 🚫 `false` | n/a     | n/a         |
```
NicholasBoll, line 519:
```suggestion
| name     | type                                                            | required?  | default | recommended |
| -------- | --------------------------------------------------------------- | ---------- | ------- | ----------- |
| htmlFor  | `string`                                                        | 🚫 `false` | n/a     | n/a         |
| children | `(model: PaginationModel) => React.ReactNode \| React.ReactNode` | 🚫 `false` | n/a     | n/a         |
```

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,555 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+  {({state}) => `${state.rangeMin}-${state.rangeMax} of ${state.lastPage} results`}
+</Pagination.AdditionalDetails>
+```
+
+#### Component Props
+
+| name              | type                                                            | required?  | default | recommended |
+| ----------------- | --------------------------------------------------------------- | ---------- | ------- | ----------- |
+| shouldHideDetails | `boolean`                                                       | 🚫 `false` | n/a     | n/a         |
+| children          | `(model: PaginationModel) => React.ReactNode | React.ReactNode` | ✅ `true`  | n/a     | n/a         |
```
NicholasBoll, line 548:
```suggestion
| name              | type                                                            | required?  | default | recommended |
| ----------------- | --------------------------------------------------------------- | ---------- | ------- | ----------- |
| shouldHideDetails | `boolean`                                                       | 🚫 `false` | n/a     | n/a         |
| children          | `(model: PaginationModel) => React.ReactNode \| React.ReactNode` | ✅ `true`  | n/a     | n/a         |
```

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,555 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+The Model's `events` describe behaviors that act on `state`. Below are the `events` keys and
+descriptions of their values:
 
-#### previousPageAriaLabel?: string;
+- `jumpToFirst` - sets the current page to the first page
+- `jumpToLast` - sets the current page to the last page
+- `stepToNext` - increments the current page by 1
+- `stepToPrevious` - decrements the current page by 1
+- `setCurrentPage` - sets the current page to a given page number (no safeguards)
+- `goToPage` - sets the current page to a given page number\* (with safeguards)
```
NicholasBoll, line 613: There's some inconsistency with these event names: `jumpTo*`, `stepTo*`, `goTo*`. Tabs is using the `CursorModel` which has simpler named events like `next`, `first`, `last`, `previous`. How do we want to name these types of events?

NicholasBoll: Perhaps we want to align `Pagination` with `Cursor` since the concept is the same - a list of things has a cursor.

modules/_labs/pagination/react/lib/Pagination/Controls.tsx
```diff
@@ -0,0 +1,144 @@
...
+    }
+    onClick?.(e);
+    events.setCurrentPage(state.firstPage);
+  };
+  const {shouldUseRTL} = useRTL();
+  const icon = shouldUseRTL ? chevron2xRightSmallIcon : chevron2xLeftSmallIcon;
+  return (
+    <IconButton
+      aria-disabled={isDisabled || undefined}
+      aria-label={ariaLabel}
```
NicholasBoll, line 46: I've been letting these fall back to `...restProps`. The effect is the same, but less code.

modules/_labs/pagination/react/lib/Pagination/GoTo/useGoToForm.tsx
```diff
@@ -0,0 +1,55 @@
...
+  {inputId, onSubmit, state, events}: UseGoToFormConfig = {} as UseGoToFormConfig
+) => {
+  const [value, setValue] = React.useState<number>();
+
+  React.useEffect(() => {
+    if (value !== undefined && state.currentPage !== value) {
+      setValue(state.currentPage);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [state.currentPage]);
```
NicholasBoll, line 21: The Compound Component Model API would make `onSetCurrentPage` be fired whenever the page changes. This model would then handle state updates instead of using `useEffect` which prevents disabling of `exhaustive-deps`. You could also achieve this by changing `inputProps`:

```tsx
const inputProps = {
  id: uniqueId,
  value: (value !== undefined && state.currentPage !== value) ? state.currentPage : value,
  onChange: handleChange,
}
```

Either will prevent double rendering from the `setValue` call inside a `useEffect`

modules/_labs/pagination/react/lib/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+  children,
+  ...elemProps
+}: PageButtonProps) => {
+  const isCurrentPage = pageNumber === state.currentPage;
+  return (
+    <StyledPageButton
+      aria-current={isCurrentPage ? 'page' : undefined}
+      aria-label={ariaLabel}
+      aria-pressed={undefined}
+      onClick={_ => events.goToPage(pageNumber)}
```
NicholasBoll, line 42: Do you mean for `elemProps.onClick` to override this functionality? If not, you need to extract `onClick` from props and call it here if it exists.

modules/_labs/pagination/react/lib/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,155 @@
...
+Pagination.Controls = (props: ControlsProps) => {
+  return <Controls {...props} />;
+};
+
+Pagination.JumpToFirstButton = (props: IconButtonProps) => {
+  // The linter doesn't recognize the dot syntax, so we're disabling the rule
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const model = React.useContext(PaginationContext);
+  return <JumpToFirstButton {...model} {...props} />;
+};
```
NicholasBoll, line 85: This pattern ends up being a lot of  boilerplate. Tabs has context built into the sub-components to reduce this type of boilerplate. But these components aren't exactly using the Model API. This is doing `{...model}` instead of `model={model}`. Is this okay for now?

modules/_labs/pagination/react/lib/Pagination/GoTo/useGoToForm.tsx
NicholasBoll: I'm not sure that provided example will work. You'd need to do some additional state tracking to know when `state.currentPage` was updated and `value` was not.

modules/_labs/pagination/react/lib/Pagination/common/Box.tsx
```diff
@@ -0,0 +1,46 @@
...
+export const StyledBox = styled('div')<BoxProps>({
+  boxSizing: 'border-box',
+});
+
+export const Box = ({as = 'div', ...props}: BoxProps) => {
+  // TODO: Memoize style props with React.useMemo
+  const boxStyles = getBoxStyles(props);
+  const safeProps = safelySpreadProps(props);
+  return <StyledBox as={as} css={css(boxStyles)} {...safeProps} />;
+};
```
NicholasBoll, line 46: This isn't how refs work in React. Drop the `BoxRef` and `ref` here entirely and do this instead:

```suggestion
type As = keyof JSX.IntrinsicElements;

export interface BoxProps extends SpaceProps, LayoutProps, React.HTMLAttributes<HTMLElement> {
  as?: As;
}

export const StyledBox = styled('div')<BoxProps>({
  boxSizing: 'border-box',
});

export const Box = React.forwardRef<any, BoxProps>(({as = 'div', ...props}, ref) => {
  // TODO: Memoize style props with React.useMemo
  const boxStyles = getBoxStyles(props);
  const safeProps = safelySpreadProps(props);
  return <StyledBox ref={ref} as={as} css={css(boxStyles)} {...safeProps} />;
});
```

This will forward the provided `ref` properly (otherwise it will never ref to the element or anything for that matter).

This is the simplest way, but not the most correct way. To be more correct, you'll have to use something like the `Component` interface:

https://github.com/Workday/canvas-kit/blob/87a8a0179be158c5401cddc790a1d8e2b00f1e98/modules/common/react/lib/utils/components.ts#L19-L36

modules/_labs/pagination/react/lib/Pagination/common/utils/safelySpreadProps.ts
```diff
@@ -0,0 +1,20 @@
...
+import {space} from './space';
+
+const styleProps = {
+  ...flex,
+  ...layout,
+  ...space,
+};
+
+// TODO: Improve this function
+export function safelySpreadProps<T extends object>(props: T) {
```
NicholasBoll, line 12: What does this function do? Is it supposed to only spread style props? There's no JSDoc or tests for it, so I can guess based on the implementation

modules/_labs/pagination/react/lib/Pagination/common/utils/safelySpreadProps.ts
```diff
@@ -0,0 +1,20 @@
...
+  let safeProps = {};
+  for (const key in props) {
+    if (!(key in styleProps)) {
+      safeProps = {...safeProps, [key]: props[key as keyof T]};
```
NicholasBoll, line 16: Mutation is safe within a function's scope if the function owns the variable. It will produce less garbage as well:

```suggestion
      safeProps[key] = props[key as keyof T]};
```

This way the `let safeProps` can be a `const safeProps`

Of course there's tests to make sure such a refactor won't break the specification of this function, right?

modules/_labs/pagination/react/lib/Pagination/usePaginationModel.ts
```diff
@@ -0,0 +1,84 @@
...
+export type UsePaginationModelConfig = {
+  lastPage: number;
+  firstPage?: number;
+  initialCurrentPage?: number;
+  onPageChange?: (pageNumber: number) => void;
+  rangeSize?: number;
+};
+
+export const usePaginationModel = ({
+  firstPage = 1, // Q: Should this be a prop, or leave it out?
```
NicholasBoll, line 15: `firstPage` isn't really state, but it is part of the model. The model is what's shared between components, so it has to exist here. If not, you'll be violating the concept of a model in a compound component and you'll need _another_ context to share this between subcomponents.

modules/_labs/pagination/react/lib/Pagination/usePaginationModel.ts
```diff
@@ -0,0 +1,84 @@
...
+  rangeSize = 5,
+  onPageChange,
+}: UsePaginationModelConfig): PaginationModel => {
+  const [currentPage, setCurrentPage] = React.useState(initialCurrentPage);
+
+  React.useEffect(() => {
+    if (onPageChange) {
+      onPageChange(currentPage);
+    }
+  }, [onPageChange, currentPage]);
```
NicholasBoll, line 27: `useEffect` causes `onPageChange` to be part of another render cycle. `useLayoutEffect` will cause `onPageChange` to be called as part of another state change batch. The compound component utility handles callbacks like this by defining a relationship between events and callbacks that are called in the same event queue vs another event queue.

You could mimic this here by making a function like the following and calling it instead of `setCurrentPage`:
```tsx
const changePage = (page: number) => {
  onPageChange?.(page);

  setCurrentPage(page)
}
```

An example use would look like:
```tsx
  const jumpToFirst = () => {
    changePage(firstPage);
  };
```

NicholasBoll: The theme here is I don't like using `useEffect` for  state changes. It forces multiple renders of a component which is wasteful if it isn't necessary.

modules/_labs/pagination/react/spec/usePaginationModel.spec.ts
```diff
@@ -0,0 +1,131 @@
+import {renderHook, act} from '@testing-library/react-hooks';
```
NicholasBoll, line 1: TIL. I've been testing hooks in the context of a component using them...

modules/_labs/pagination/react/index.ts
```diff
@@ -1,5 +1,4 @@
-import Pagination from './lib/Pagination';
+import {Pagination} from './lib/Pagination';
 
 export default Pagination;
-export {Pagination};
 export * from './lib/Pagination';
```
alanbsmith, line 4: I'm fairly certain `export * from './lib/Pagination';` will cover `export {Pagination};` as it's a named export in `lib/Pagination`.

modules/_labs/pagination/react/lib/Pagination/Controls.tsx
alanbsmith: I'm good with that. Previously, we opted to be more explicit, but I also like less code.

modules/_labs/pagination/react/lib/Pagination/PageButton.tsx
alanbsmith: Good catch. 👍🏻

modules/_labs/pagination/react/lib/Pagination/Pagination.tsx
alanbsmith: > _This pattern ends up being a lot of boilerplate._

Very much agree. This will be resolved with `createComponent` in v5.

> _This is doing `{...model}` instead of `model={model}`_

Yep. This was before we solidified that pattern. I could update it now and that would (potentially) create less churn when users need to update to v5.

modules/_labs/pagination/react/lib/Pagination/Controls.tsx
NicholasBoll: I think we can be explicit with the prop definitions (especially with the JSDoc) of it, but implicitly allow the props to be applied.

modules/_labs/pagination/react/lib/Pagination/Pagination.tsx
NicholasBoll: Up to you. If we don't export `JumpToFirstButton`, I don't think the consumer will see much difference. It will show up in the React render tree as 2 different components. `ref` forwarding is not happening. If it was, we'd have to forward every time we create a new component wrapper. I don't think we're concerned about ref forwarding in v4 though.

modules/_labs/pagination/react/index.ts
NicholasBoll: Yeah. I've been pondering dropping default exports entirely in v5. We have to remap default exports so often I'm just not seeing the benefit.

modules/_labs/pagination/react/lib/Pagination/common/utils/flex.ts
```diff
@@ -0,0 +1,110 @@
...
+  | 'normal'
+  | 'right'
+  | 'stretch'
+  | 'auto'
+  | 'baseline'
+  | 'left'
+  | 'normal'
+  | 'right'
+  | 'stretch';
+type AlignSelfProperty = Globals | SelfPosition | 'auto' | 'baseline' | 'normal' | 'stretch';
```
alanbsmith, line 67: All of this goes away if we use csstype.

modules/_labs/pagination/react/lib/Pagination/usePaginationModel.ts
alanbsmith: Yeah, I agree. I think the reason I did this was because I was passing down `setCurrentPage` in the events, but maybe I don't need to do that. Or I could do something like:

```ts
const events = {
  setCurrentPage: changePage,
  jumpToFirst,
  jumpToLast,
  stepToNext,
  stepToPrevious,
  goToPage,
}
```

modules/_labs/pagination/react/lib/Pagination/Pagination.tsx
alanbsmith: I switched all sub-components to use a model prop instead. 👍🏻

modules/_labs/pagination/react/README.md
alanbsmith: Yeah, I went with `jump` and `step` as they correspond to the button names. I'm okay with switching to `first` `last`, but I'm not sure what you would name `goTo` in that paradigm.

modules/_labs/pagination/react/lib/Pagination/common/utils/safelySpreadProps.ts
alanbsmith: * Added JSDoc
* Added tests

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,555 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+section for details on supporting custom implementations.
+
+#### Pagination with Step Controls
+
+In this example, the component will only display the step controls (next page and previous page) and
+the page list buttons.
+
+```tsx
+import * as React from 'react';
+import Pagination from '@workday/canvas-kit-labs-react-pagination';
```
jamesfan, line 97: The Pagination imports will need to be updated to `import {Pagination}` if there's no longer a default export.

modules/_labs/pagination/react/lib/Pagination/common/utils/safelySpreadProps.ts
alanbsmith: That's a good call out instead of creating a new obj on every iteration. I updated it 👍🏻

modules/_labs/pagination/react/lib/Pagination/usePaginationModel.ts
alanbsmith: I updated this. We're good now. 👍🏻

modules/_labs/pagination/react/README.md
alanbsmith: Probably not. I wrote this before your doc was created. I'll link out to it. Good call. 👍🏻

modules/_labs/pagination/react/README.md
alanbsmith: It's...uh... the essay portion of the exam. Once they've read the docs, they should be able to fill in their own example here, and I'll grade it. 😉

modules/_labs/pagination/react/README.md
alanbsmith: Great catch!

modules/_labs/pagination/react/lib/Pagination/GoTo/useGoToForm.tsx
alanbsmith: > You'd need to do some additional state tracking to know when `state.currentPage` was updated and value was not.

Yeah, that's why the `useEffect` makes sense here, imo. But I understand the double-rendering concern. I'll mess with it.

alanbsmith: I think the model solution is best. I'm not seeing a clear way to track the `state.currentPage` changes.

modules/_labs/pagination/react/README.md
alanbsmith: Updated. Now using `first`, `previous`, `next`, `last`, and `goTo`.

cypress/integration/Pagination.spec.ts
alanbsmith: Updated.

modules/_labs/pagination/react/lib/Pagination/common/utils/flex.ts
```diff
@@ -0,0 +1,110 @@
+// TODO: Consider using CSSType instead
```
anicholls, line 1: Let's definitely do this at some point

modules/_labs/pagination/react/lib/Pagination/common/utils/space.ts
```diff
@@ -0,0 +1,109 @@
+import {spacing} from '@workday/canvas-kit-react-core';
+
+export type Spacing = keyof typeof spacing;
+
+export interface SpaceProps {
```
anicholls, line 5: When we move this to common, we should make sure to deprecate the one in labs-core. We may want to do that as part of v5 TBH

modules/_labs/pagination/react/README.md
```diff
@@ -4,7 +4,67 @@
   <img src="https://img.shields.io/badge/LABS-beta-orange" alt="LABS: Beta" />
 </a>  This component is work in progress and currently in pre-release.
 
-Contains a component for a pagination bar and dispatches for page changes
+`Pagination` is a composable component for handling navigation between pages in a range.
```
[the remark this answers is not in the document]
alanbsmith: Yeah, this was written before we were using the term "compound component." I'll update it. Good catch. 👍🏻

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,555 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+#### Component Props
+
+| name               | type                           | required?  | default | recommended                              |
+| ------------------ | ------------------------------ | ---------- | ------- | ---------------------------------------- |
+| aria-label         | `string`                       | ✅ `true`  | n/a     | "Pagination" (and translated equivalent) |
+| lastPage           | `number`                       | ✅ `true`  | n/a     | n/a                                      |
+| initialCurrentPage | `number`                       | 🚫 `false` | 1       | n/a                                      |
+| rangeSize          | `number`                       | 🚫 `false` | 5       | n/a                                      |
+| firstPage          | `number`                       | 🚫 `false` | 1       | n/a                                      |
+| onPageChange       | `(pageNumber: number) => void` | 🚫 `false` | n/a     | n/a                                      |
```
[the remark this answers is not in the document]
alanbsmith: It needs to be added. The API changed. 👍🏻

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,661 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+`Pagination.StepToNextButton`) that allow you to move to the next page and previous pages.
+
+```tsx
+import * as React from 'react';
+import {
+  Pagination,
+  getLastPage,
+  getVisibleResultsMax,
+  getVisibleResultsMin,
+} from '../lib/Pagination';
```
jamesfan, line 106: I think this was a copy-and-paste error. The imports in the examples should be from `@workday/canvas-kit-labs-react-pagination`.

alanbsmith: Yep. Good catch. 👍🏻

modules/_labs/pagination/react/lib/Pagination/common/utils/space.ts
alanbsmith: Yeah, all of the layout components will be v5 so we can cleanly make that switch.

modules/_labs/pagination/react/lib/Pagination/common/utils/flex.ts
alanbsmith: Agreed. v5!

modules/_labs/pagination/react/lib/Pagination/AdditionalDetails.tsx
```diff
@@ -0,0 +1,43 @@
+/** @jsx jsx */
+import * as React from 'react';
+import {jsx, css} from '@emotion/core';
+import {type, typeColors} from '@workday/canvas-kit-react-core';
+import {accessibleHide} from '@workday/canvas-kit-react-common';
+
+import {PaginationModel} from './types';
+import {Flex, FlexProps} from './common/Flex';
```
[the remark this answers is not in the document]
alanbsmith: I can't wait. 😄

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,661 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+For this example, we'll create a `Pagination` component that handles 128 total items with 10 items
+per page. Normally this would involve some math on your part, but don't worry about that, we've got
+you covered with some handy utilities
+
+```tsx
+import {
+  getLastPage,
+  getVisibleResultsMax,
+  getVisibleResultsMin,
+} from '../lib/Pagination';
```
jamesfan, line 290:
```suggestion
import * as React from 'react';
import {
  Pagination,
  getLastPage,
  getVisibleResultsMax,
  getVisibleResultsMin,
  usePaginationModel,
} from '@workday/canvas-kit-labs-react-pagination';
```
It'll be a lot easier to catch these errors once these examples are actually rendered with Nicholas' docs improvements. :)

modules/_labs/pagination/react/README.md
```diff
@@ -14,82 +74,661 @@ yarn add @workday/canvas-kit-labs-react-pagination
...
+const resultCount = 10;
+const lastPage = getLastPage(resultCount, totalCount);
+
+const ExternalModelPagination = () => {
+  // create the model
+  const model = usePaginationModel({
+    lastPage={lastPage}
+    rangeSize={3}
+    onPageChange={pageNumber => console.log(pageNumber)}
+  });
```
jamesfan, line 302:
```suggestion
  const model = usePaginationModel({
    lastPage,
    rangeSize: 3,
    onPageChange: pageNumber => console.log(pageNumber),
  });
```

NicholasBoll (approved):

PR #958, opened 2021-02-04
test: Add verifyComponent test helper and fix test types

NicholasBoll:
- Added `verifyComponent` to help verify the API of compound components:
  - Verifies `ref` forwarding
  - Verifies `as` prop
  - Verifies extra prop overriding
  - Verifies callback merging
- Update Jest dependencies
- Fix lack of type checking/auto-complete help on Jest spec files 

Compound components should have a jest file like this per component that maps to an element (some container components don't map directly to an element and this should not be done on them):

```tsx
import { MyComponent, useMyComponentModel } from '..';

describe('MyComponent', () => {
  verifyComponent(MyComponent, { modelHook: useMyComponentModel });

  // other specs
});
```

## Additional Details

Our core `tsconfig.json` ignores all spec files. This means each `*.spec.tsx` file is treated by VSCode as generic TS file and it uses a default configuration for the built-in Typescript language service. This includes automated DefinitelyTyped lookups and reading from `@types/*` without actually complaining. One thing that came up recently was an incompatibility of the `jest.Matchers` type. Newer type definitions use 2 generics while older use 1 generic (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39243).

Typescript sees this incompatibility and falls back to `any`. Since the default config for `noImplicitAny` is `false`, it gets silently ignored and we lose auto-completion. `jest-emotion` is the last holdout for this updated type and is not fixed until Emotion 11. To work around this for now, I've held back all `jest` types. `@testing-library/jest-dom` is a dependency that uses the newer `Matchers` shape and I couldn't roll it back I instead created a local type override. I then created a `tsconfig.spec.json` in the root directory and a `tsconfig.json` in all `spec` folders that references this global one.

The net effect is a custom spec TS config for all spec files. Using this, I added `verifyComponent` as a test global. Also all spec files should correctly know the shape of `expect` with the matcher plugins we use.


jest/setupTests.ts
```diff
@@ -3,8 +3,14 @@ import {configure} from 'enzyme';
...
+import {jest} from '@jest/globals';
 
 expect.addSnapshotSerializer(serializer);
 expect.extend(toHaveNoViolations);
 expect.extend(matchers);
 configure({adapter: new Adapter()});
+
+// add convenience variables to the global context
+(global as any).verifyComponent = verifyComponent;
+(global as any).jest = jest;
```
NicholasBoll, line 16: I didn't need to add `jest`, but newer jest versions seemed to drop `jest` as an automatic global.

jest/testing-library__jest-dom.d.ts
```diff
@@ -0,0 +1,529 @@
...
+// Definitions by: Ernesto García <https://github.com/gnapse>
+//                 John Gozde <https://github.com/jgoz>
+//                 Seth Macpherson <https://github.com/smacpherson64>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.8
+
+/// <reference types="jest" />
+
+declare namespace jest {
+  interface Matchers<R> {
```
NicholasBoll, line 14: This is the only difference from the main package. It could be removed when we update to Emotion 11 and all jest types use the new 2 generic `Matcher`

jest/verifyComponent.tsx
```diff
@@ -0,0 +1,85 @@
...
+      expect(container).toContainHTML('<article');
+    });
+
+    it('should forward extra props properly - extra props should overwrite component props', () => {
+      const ref = {current: null};
+      const attrs = {
+        'aria-label': 'MyAriaLabel',
+        id: 'MyID',
+        role: 'MyRole',
+        'data-testid': 'MyTestID',
```
NicholasBoll, line 58: This is not an exhaustive list of attributes. Extra attributes is an infinite set, so I picked common ones.

jest/verifyComponent.tsx
```diff
@@ -0,0 +1,85 @@
...
+        expect(ref.current).toHaveAttribute(key, attrs[key]);
+      });
+    });
+
+    const callbacks = {
+      onClick: 'click',
+      onKeyDown: 'keyDown',
+      onKeyUp: 'keyUp',
+      onMouseDown: 'mouseDown',
+      onMouseUp: 'mouseUp',
```
NicholasBoll, line 71: A limited list of callbacks. `fireEvent` does give me a larger list, but these are the most common ones.

jest/verifyComponent.tsx
NicholasBoll: I made attribute verification the part of the same test since there is enough context in the failure to figure out which attribute is failing.

jest/verifyComponent.tsx
NicholasBoll: I created a separate test for each callback because there is not enough context in the failure to isolate which callback is the problem.

modules/_labs/breadcrumbs/react/lib/Breadcrumbs/List/hooks.ts
```diff
@@ -15,9 +15,9 @@ export const useCollapse = <E extends HTMLElement>(listEl: React.RefObject<E>, m
 };
 
 const getBreadcrumbLink = (breadcrumb: React.ReactElement) => {
-  return Children.toArray(breadcrumb.props.children).filter(child => {
+  return Children.toArray(breadcrumb.props.children).filter((child: React.ReactElement) => {
     return child.type === BreadcrumbLink;
-  })[0];
+  })[0] as React.ReactElement;
```
NicholasBoll, line 20: I was running into an issue with React 17 types. `child` could be a large number of types. The safer fix would be to use a guard type to determine `BreadcrumbLink` children, but it was easier to restrict and cast.

modules/_labs/header/react/spec/SearchBar.spec.tsx
```diff
@@ -12,41 +12,39 @@ describe('Header Search', () => {
...
   });
 
-  test('Searching empty string should focus input', async () => {
+  test('Searching empty string should focus input', () => {
     const inputLabelText = `label`;
-    const {getByLabelText} = render(<SearchBar onSubmit={cb} inputLabel={inputLabelText} />);
+    render(<SearchBar onSubmit={cb} inputLabel={inputLabelText} />);
 
-    fireEvent.submit(await getByLabelText(inputLabelText));
+    fireEvent.submit(screen.getByRole('search', {name: inputLabelText}));
```
NicholasBoll, line 19: `@testing-library/dom` was updated to complain that multiple elements where found with this label. `findByRole` is restricting which element we're talking about.

modules/common/react/spec/components.spec.tsx
```diff
@@ -120,7 +120,7 @@ describe('useForkRef', () => {
   });
 });
 
-describe.only('useLocalRef', () => {
```
NicholasBoll, line 123: Ooops

tsconfig.spec.json
```diff
@@ -0,0 +1,18 @@
...
+  "exclude": [],
+  "compilerOptions": {
+    "declaration": false,
+    "declarationMap": false,
+    "noImplicitReturns": true,
+    "noImplicitThis": true,
+    "preserveConstEnums": true,
+    "allowUnreachableCode": false,
+    "noImplicitAny": false,
+    "strictNullChecks": false,
```
NicholasBoll, line 13: I set these to be false for tests. I've had these on `true` in the past, but it is more overhead in tests... And it isn't actually catching anything useful for us - tests already catch things... These were essentially disabled before since we didn't have any custom config for spec files. The overall effect is more strict typing than before.

utils/check-dependencies-exist.js
```diff
@@ -93,6 +93,10 @@ function formatErrorMessage(pkgFile, errors) {
               const column = errors[key][file].column;
               const message = errors[key][file].formatted;
 
+              if (!line && !column && !message) {
+                return `${file}\n${errors[key][file]}`;
+              }
```
NicholasBoll, line 98: I've been meaning to add this for a while. There are errors that don't have line and column numbers. Parse errors fall in this category. This extra check here gives us useful output on parse errors.

modules/common/react/spec/components.spec.tsx
alanbsmith: hahaha

alanbsmith (approved):
🦘

PR #966, opened 2021-02-10
docs: Fix Story names

NicholasBoll:
Before a story named 'NamedKeys' would show up to Storybook as 'NamedKeys'. This change will present the story as 'Named Keys' like it should.


jamesfan:
@NicholasBoll Looks like the `mdx-code-block-rewrite` loader is being run on plain MD files as well. I'm seeing the Storybook imports in the MD files when viewed on Storybook:

image

image

anicholls (approved):

NicholasBoll:
@jamesfan It should be fixed now.

PR #967, opened 2021-02-11
fix: Add crossorigin to font preloads

NicholasBoll:
Fonts aren't preloading with the following warnings:
> A preload for 'https://design.workdaycdn.com/beta/assets/fonts@1.0.0/roboto/ttf/Roboto-Regular.ttf' is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.

Since preloads are failing, we're also getting the following warnings:
> The resource https://design.workdaycdn.com/beta/assets/fonts@1.0.0/roboto/ttf/RobotoMono-Regular.ttf was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.

This fixes both warnings and makes sure Chromatic is correctly preloading fonts.

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -24,24 +24,29 @@
   href="https://design.workdaycdn.com/beta/assets/fonts@1.0.0/roboto/ttf/Roboto-Light.ttf"
   as="font"
   type="font/ttf"
+  crossorigin
 /><link
   rel="preload"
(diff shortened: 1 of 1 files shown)


NicholasBoll:
I've verified the latest Chromatic build does not show these warnings anymore: https://5d854c26ba934e0020f5e98a-qkqvhxxxet.chromatic.com/

PR #980, opened 2021-02-25
feat(popup): Update Popup behaviors to match a11y specs

### LINKED ISSUE #890: EllipsisTooltip: Add an ellipsis tooltip pattern or component
## 🚀 Feature Proposal

The PoC from #842 added had an example of EllipsisTooltips that many teams want to use for various things. PoC can be found here: https://codesandbox.io/s/xstate-usetooltipmachine-event-creators-mylem

## Motivation

Often text needs to overflow, but for accessibility, the text needs to be visible via keyboard and mouse.

## Example

https://codesandbox.io/s/xstate-usetooltipmachine-event-creators-mylem

In the following example, both buttons have an ellipsis tooltip applied, but the tooltip only opens for the button with an ellipsis present (content is overflowed)

ellipsis_tooltip

### LINKED ISSUE #900: Reopen Tooltip specification a11y suggestions issue
## 💡 Suggestion

Re-opening this issue: https://github.com/Workday/canvas-kit/issues/899#issue-739507824

Tooltips:
- Should appear on focus
  - Disappear on blur
- Should appear on hover
  - Disappear on mouse out - A tooltip should not affect other popup interactions
- Click outside should close any popups that are not tooltips, but should also close any open tooltips
- Should appear on text that is truncated
  - Should not disappear if a user clicks on a non-interactive element (and should disappear on blur/mouse-out)

## Motivation

Tooltips currently behave as every other popup, which doesn't behave the way native tooltips do. This can lead to disruptive behavior especially if these tooltips activate within the context of other popups. At the end of the day, the behavior of a tooltip should not influence the behavior of other popups, especially when it comes to click-outside behavior.

This would enable users to click outside and close the topmost popup with any tooltips included if they are still present in the DOM.

## Example

A popup showing contains a button that has truncated text. When a user focuses on the button that now displays a tooltip, and the user then clicks outside of the popup, the both the tooltip and the popup should close.

### LINKED ISSUE #961: Include option for 'muted' Tooltips (CANVAS-1549)
## Background

Presently we have 3 types of tooltips either in Canvas Kit, or pending.

Label - takes the aria-label from a focusable control and displays it in a tooltip. The tooltip contents are not announced by a screen reader, as the aria-label acts as the label for the button.

```html
<button type="button" aria-label="Open">
  <svg focusable="false" aria-hidden="true">...</svg>
</button>
...
<div>
 Open
</div>
Describe - augments the accessible name of a focusable control by allowing a custom string to be shown in a tooltip. This custom string is announced after a delay by a screen reader.
<button type="button" aria-describedby="hint">
 Open
</button>
...
<div id="hint">
  This is hint text
</div>
```

Truncate - takes truncated text from a focusable control and displays it un-truncated inside a tooltip. The tooltip contents are not announced by a screen reader.

---

## 🚀 Feature **Proposal**

New feature designs are highlighting a desire for a 4th type of tooltip - a muted hint if you like, this would behave as follows -

Allows a custom string to be displayed in a tooltip on a focusable control when, 
that control already has an accessible name 
if the control has a state, this is conveyed to an assistive technology 
it could otherwise be inferred from the context what action the control will perform, the tooltip is a visual hint and not a substitute for a label.
The custom string in the tooltip is not exposed to the accessibility api (thus not announced by a screen reader).

**Example markup - elided**

```html
<button type="button" aria-label="Open panel" aria-expanded="false">
 <svg>...</svg>
</button>
...
<!-- Tooltip -->
<div aria-hidden="true">
 Show stuff
</div>
```

NicholasBoll:
Fixes #900
Fixes #890
Fixes #961 

## Summary

- feat(popup): Add `useAlwaysCloseOnOutsideClick` to always close regardless of position in the stack
  - This behavior adds `data-behavior-click-outside-close="always"` to the stack UI element
- feat(popup): Update `useCloseOnOutsideClick` to only close topmost popup with the behavior applied
  - This behavior adds `data-behavior-click-outside-close="topmost"` to the stack UI element
- feat(tooltip): Add `muted` type to `Tooltip` component
- chore(popup-stack): Update PopupStack story to showcase these behaviors
- docs(tooltip): Update to narrative documentation style
- docs: Add docs package with Specifications table

## Checklist

New a11y specs:
- [x] Should appear on focus
  - [x] Disappear on blur
- [x] Should appear on hover
  - [x] Disappear on mouse out - A tooltip should not affect other popup interactions
- [x] Click outside should close topmost popups that are not tooltips or persistent, but should also close all open tooltips
- [x] Should appear on text that is truncated
  - [x] Should not disappear if a user clicks on a non-interactive element (and should disappear on blur/mouse-out)

## Additional References

Updated behavior for tooltips. Note tooltips are sticking around for 10s to see the effect.

https://user-images.githubusercontent.com/338257/109084369-e7d38a00-76c4-11eb-80e3-d506d9dd3c97.mp4


modules/popup-stack/README.md
```diff
@@ -205,3 +205,26 @@ interact with it.
...
+## Behaviors
+
+The PopupStack does not directly own any behaviors. Behaviors are left up to the implementation, but
+behavior suggestions are listed here as a specification for implementations. Some behaviors are
+aided by a specification of `data-*` attributes to help those behaviors apply globally across
+frameworks.
+
+### Close on outside click
+
+Close only the topmost popup with a click outside the popup when this behavior applied. To determine
```
csongnguyen, line 218: "when this behavior is applied."

modules/popup/react/README.md
```diff
@@ -398,11 +398,28 @@ useCloseOnOutsideClick(stackRef: React.RefObject<HTMLElement>, onClose: () => vo
...
-the `Popper` component since `Popper` uses `usePopupStack` internally.
+the topmost element with this behavior applied in the stack. Adds a `data-behavior-click-outside`
+attribute to track usage of this behavior hook.
 
-This should be used stacked UI elements that are dismissible like Tooltips, Modals, non-modal
-dialogs, dropdown menus, etc.
+The `stackRef` should be the same as the one passed to `usePopupStack` or the `Popper` component
+since `Popper` uses `usePopupStack` internally.
+
+This should be used stacked UI elements that are dismissible like Modals, non-modal dialogs,
```
csongnguyen, line 407: "This should be used to stack UI elements"

"This is used to stack UI elements"?

modules/popup/react/lib/useCloseOnOutsideClick.ts
```diff
@@ -5,29 +5,46 @@ import {PopupStack} from '@workday/canvas-kit-popup-stack';
...
+    () => (event: MouseEvent) => {
+      if (!stackRef.current) {
+        return;
+      }
+      const elements = PopupStack.getElements().filter(e =>
+        e.hasAttribute('data-behavior-click-outside')
+      );
+      if (
+        elements.length &&
+        elements[elements.length - 1] === stackRef.current &&
```
csongnguyen, line 33: Ah so we will always grab the last of the list of elements. This might be worth putting in the method documentation

modules/tooltip/react/stories/Tooltip.stories.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+### Describing an Element
+
+The default mode for a tooltip is to label content via `aria-label`. If a tooltip is meant to
+provide ancilary, but not critical information, the `type` can be set to `describe`. This will add
+`aria-describedby` to the target element.
+
+<ExampleCodeBlock code={DescribeType} />
+
+### Custom CustomContent
```
csongnguyen, line 56: Custom CustomContent 🤔

modules/tooltip/react/stories/Tooltip.stories.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+A tooltip can contain HTML, but should not contain any focusable elements or semantically meaningful
+formatting. The content will loose all semantic meaning when read by a screen reader. If complex
```
csongnguyen, line 59: "will lose"

modules/tooltip/react/stories/Tooltip.stories.mdx
```diff
@@ -0,0 +1,84 @@
...
+for tooltips.
+
+<ExampleCodeBlock code={Placements} />
+
+### The UseTooltip Hook
+
+The `Tooltip` component is a combination of the `TooltipContainer` (a styled element), `Popper`
+(which uses PopperJS and the popup stack), the `useTooltip` hook and some behavior. If custom
+behavior is required, these sub-components can be composed in a custom container element. This
+example uses those parts directly while being functionally equivelent to the original basic example.
```
csongnguyen, line 78: equivalent

modules/popup/react/README.md
NicholasBoll: I think I meant "This should be used with stacked UI elements".

modules/popup/react/lib/useCloseOnOutsideClick.ts
NicholasBoll: The JSDoc says:
```js
/**
 * Registers global listener for all clicks. It will only call the `onClose` callback if the click
 * happened outside the `stackRef` element and its children _and_ the provided `stackRef` element is
 * the topmost element with this behavior applied in the stack. Adds a `data-behavior-click-outside`
 * attribute to track usage of this behavior hook.
 */
```

Do you think this is covered by "the provided `stackRef` element is the topmost element with this behavior applied in the stack"?

modules/tooltip/react/stories/Tooltip.stories.mdx
NicholasBoll: Emphasizes just how custom the content is!

modules/tooltip/react/stories/Tooltip.stories.mdx
NicholasBoll: You can use the suggestion feature of Github:

image

It appears like this:
```suggestion
example uses those parts directly while being functionally equivalent to the original basic example.
```

modules/popup/react/lib/useCloseOnOutsideClick.ts
csongnguyen: I'm not too worried about the wording there, I suppose it's more so for `getElements()`. The list is based on what the adapter passes to the PopupStack, so it's more so a nice-to-have to add explicit expectation of how the getElements() list order should appear (even if we do call this the PopupStack, better to be safe with an extra clarification comment).

NicholasBoll: I'll update the documentation of the Popup stack. It also might make sense for these behaviors to ensure z-index order themselves just in case.

.storybook/main.js
```diff
@@ -21,6 +22,23 @@ module.exports = {
...
+        {
+          loader: require.resolve('string-replace-loader'),
+          options: {
+            search: '[/* SPEC_FILES_REPLACE_BY_WEBPACK */]',
+            replace: JSON.stringify(specs, null, '  '),
+          },
+        },
+      ],
+    });
+
```
NicholasBoll, line 41: This is how specifications in docs works in Webpack. There is a separate process for build to publish specifications.

cypress/integration/PopupStack.spec.ts
```diff
@@ -160,5 +160,43 @@ describe('PopupStack', () => {
...
+            cy.findByRole('tooltip').should('not.be.visible');
+          });
+
+          it('should close the Really Delete Item popup', () => {
+            cy.findByRole('dialog', {name: 'Really Delete Item'}).should('be.not.visible');
+          });
+
+          it('should NOT close the Delete Item popup', () => {
+            cy.findByRole('dialog', {name: 'Delete Item'}).should('be.visible');
+          });
```
NicholasBoll, line 197: New tests to verify only the topmost popup is closing and not all popups.

modules/popup-stack/stories/stories_testing.tsx
```diff
@@ -84,27 +87,54 @@ export const MixedPopupTypes = () => {
...
+                          Really Delete
+                        </DeleteButton>
+                        <Button onClick={onClose}>Cancel</Button>
+                      </>
+                    )}
+                  </TempPopup>
+                  <Button onClick={onClose}>Cancel</Button>
+                </>
+              )}
+            </TempPopup>
```
NicholasBoll, line 136: Nested popup for additional behavior testing

modules/tooltip/react/stories/Tooltip.stories.mdx
```diff
@@ -0,0 +1,130 @@
...
+Here is a basic example of an `IconButton` using a tooltip to label the icon. This is labeling the
+button for both sighted users and for screen readers. A tooltip will provide an `aria-label` to
+child elements for the accessibility tree and a visual tooltip during mouse hover and focus events.
+Mouse over or focus on the x icon below. Click "Show code" to see the source code.
+
+<ExampleCodeBlock code={Default} />
+
+### Describing an Element
+
+The default mode for a tooltip is to label content via `aria-label`. If a tooltip is meant to
```
samcartersmith, line 53: I would consider removing the critical part of this messaging. I propose the following:
The default mode for a tooltip is to label content via `aria-label`. If a tooltip is meant to
provide ancillary information, the `type` can be set to `describe`. This will add
`aria-describedby` to the target element. This will allow screen reader users to hear the name of the control that is being focused and the ancillary tooltip information.

modules/tooltip/react/stories/Tooltip.stories.mdx
```diff
@@ -0,0 +1,130 @@
...
+```
+
+## Usage
+
+This component follows the
+[W3 Tooltip specification](https://www.w3.org/TR/wai-aria-practices/#tooltip). Tooltips are used to
+label icon buttons and provide additional context to elements.
+
+### Basic Example
+
```
samcartersmith, line 43: Could we add a kind of decision flow chart for when to use which type of tooltip here? I've written up this, but happy to take it to the content team if we feel it needs some sprucing up.

When to use tooltips:
If you’re trying to display additional information for users to better understand the purpose, context, or interaction.

When not to use tooltips:
When the visual text will be the exact same as what is visually displayed to the user without the tooltip being visible

Tooltip Usage Process
1. Does this widget need additional context or information
    1. If no, don’t use a tooltip
2. Is the tooltip text different from visual text displayed to users
    1. If there is no text, use aria-label (icon example)
    2. If yes, use aria-describedby
    3. If no, don’t use a tooltip
3. Is the tooltip text useful to screen reader users?
    1. If no, use muted tooltip

modules/specifications/README.md
```diff
@@ -0,0 +1,59 @@
+# Canvas Kit Specifications
+
+A package of all the automated Cypress specifications for use in documentation.
+
+# Specifications
+
+## Installation
+
+or
```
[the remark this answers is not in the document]
NicholasBoll: Oops, nope.

modules/tooltip/react/stories/Tooltip.stories.mdx
NicholasBoll: Thanks! Will update.

modules/popup/react/README.md
```diff
@@ -398,11 +398,28 @@ useCloseOnOutsideClick(stackRef: React.RefObject<HTMLElement>, onClose: () => vo
...
+
+## useAlwaysCloseOnOutsideClick
+
+```ts
+useAlwaysCloseOnOutsideClick(stackRef: React.RefObject<HTMLElement>, onClose: () => void): void
+```
+
+Registers global listener for all clicks. It will only call the `onClose` callback if the click
+happened outside the `stackRef` element and its children regardless of the position in the stack.
+This is useful for Tooltips or hierarchical menus.
```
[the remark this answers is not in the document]
NicholasBoll: Yes. I have the data attributes outlined in the `PopupStack` readme, but should also add here

modules/popup/react/lib/useCloseOnOutsideClick.ts
NicholasBoll: I updated `useCloseOnOutsideClick` to sort just in case.

modules/specifications/lib/specs.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export interface SpecFile {
+  type: 'file';
+  name: string;
+  children: (SpecDescribe | SpecIt)[];
+}
+
+// prettier-ignore
+export const specifications: SpecFile[] = [/* SPEC_FILES_REPLACE_BY_WEBPACK */];
```
NicholasBoll, line 19: Hopefully this is clear enough that Webpack specifically replaces this string with the real specs.

modules/tooltip/react/stories/Tooltip.stories.mdx
NicholasBoll: Perfect! I'll also added the `muted` type and some tests and examples.

NicholasBoll: Is there a spot in that decision tree for a label to provide additional context that a visual user wouldn't need? For example, if a card titled `Dependents` contains a button to add a new dependent, would the button text be `Add New` while the `aria-label="Add New Dependent"`? Or would that be difficult to handle for translations?

It is a balance from the visual being too noisy and buttons not having enough context in isolation (i.e. VO's web rotor lists).

cypress/integration/PopupStack.spec.ts
```diff
@@ -160,5 +160,43 @@ describe('PopupStack', () => {
...
+    context('when the Delete button is clicked', () => {
+      beforeEach(() => {
+        cy.findByRole('button', {name: 'Delete'}).click();
+      });
+
+      it('should open the Really Delete Item popup', () => {
+        cy.findByRole('dialog', {name: 'Really Delete Item'}).should('be.visible');
+      });
+
+      context('when a tooltip is focused', () => {
```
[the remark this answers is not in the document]
NicholasBoll: 👍

modules/modal/react/lib/ModalContent.tsx
```diff
@@ -209,6 +209,11 @@ const ModalContent = ({
   useInitialFocus(stackRef, firstFocusRef);
   useAssistiveHideSiblings(stackRef);
 
+  React.useEffect(() => {
+    // We don't use `useCloseOnOutsideClick` directly, so we add this data attribute manually
```
alanbsmith, line 213: 👍🏻

modules/tooltip/react/stories/Tooltip.stories.mdx
```diff
@@ -0,0 +1,161 @@
...
+<ExampleCodeBlock code={Muted} />
+
+### Custom Content
+
+A tooltip can contain HTML, but should not contain any focusable elements or semantically meaningful
+formatting. The content will lose all semantic meaning when read by a screen reader. If complex
+content or a focusable element is needed by your UI, a tooltip is not a good choice. Consider using
+a dialog instead.
+
+<ExampleCodeBlock code={CustomContent} />
```
alanbsmith, line 97: I love all this content. Super clear and helpful. 👍🏻

modules/popup/react/lib/useCloseOnOutsideClick.ts
```diff
@@ -5,29 +5,46 @@ import {PopupStack} from '@workday/canvas-kit-popup-stack';
...
+      }
+      const elements = PopupStack.getElements()
+        .sort((a, b) => Number(a.style.zIndex) - Number(b.style.zIndex))
+        .filter(e => e.getAttribute('data-behavior-click-outside-close') === 'topmost');
+      if (
+        elements.length &&
+        elements[elements.length - 1] === stackRef.current &&
+        // Use `PopupStack.contains` instead of `ref.current.contains` so that the application can
+        // decide if clicking the target should toggle the popup rather than it toggling implicitly
+        // because the target is outside `ref.current`
```
alanbsmith, line 36: 👍🏻

modules/tooltip/react/lib/OverflowTooltip.tsx
```diff
@@ -0,0 +1,133 @@
...
+   */
+  children: React.ReactElement<any>;
+  /**
+   * Sets the placement preference used by PopperJS.
+   * @default 'top'
+   */
+  placement?: Placement;
+}
+
+function mergeCallbacks<T extends {[key: string]: any}>(
```
alanbsmith, line 78: I assume this is something we can remove in v5, yeah?

alanbsmith (approved):
Looks good! I had a couple questions and nits, but this looks solid overall.

modules/popup-stack/README.md
```diff
@@ -161,7 +161,9 @@ PopupStack.isTopmost(element: HTMLElement): boolean
 
 ### getElements
 
-Returns an array of elements defined by the `element` passed to `add`.
+Returns an array of elements defined by the `element` passed to `add`. This method should return
```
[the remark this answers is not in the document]
NicholasBoll: I was thinking about it as a specification for adapters. If an adapter is present, we cannot guarantee order.

modules/popup-stack/README.md
```diff
@@ -205,3 +207,27 @@ interact with it.
...
+Close only the topmost popup with a click outside the popup when this behavior is applied. To
+determine if this behavior is applied to the global stack, a
+`data-behavior-click-outside-close="topmost"` should be applied. This data attribute should be used
+by all instances of this behavior to determine the "topmost" click-outside behavior to close only
+the topmost popup with this behavior.
+
+### Always close on outside click
+
+Close all popups with this behavior regardless of the position in the stack. A
+`data-behavior-click-outside-close="always"` should be applied to the stack element.
```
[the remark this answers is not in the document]
NicholasBoll: The name of the attribute correctly describes what happens when a click outside occurs. It will always close. The "Close all popups" is describing the net effect.

modules/tooltip/react/lib/OverflowTooltip.tsx
NicholasBoll: Yes. I'd use `mergeProps` from the compound component utility functions.

NicholasBoll: Honestly, the way this works makes no sense to me and I wrote it!

modules/popup-stack/README.md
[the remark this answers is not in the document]
NicholasBoll: Perhaps:

> Always close the popup when the user clicks outside regardless of the position in the stack. A
> `data-behavior-click-outside-close="always"` attribute should be applied to the stack element.

.gitignore
```diff
@@ -26,7 +26,7 @@ junit.xml
 .vscode
 
 # docs
-docs
+/docs
```
NicholasBoll, line 29: We now have a module called 'docs', so we have to limit the scope of this ignore.

modules/tooltip/react/stories/Tooltip.stories.mdx
samcartersmith: Nicholas that is a good call out. Maybe we chat about that one? I was thinking in your example I think we'd just want a muted tooltip. Since the aria-label for the button could be "add new dependent" and the tooltip probably isn't needed at that point. I guess if it was an icon button that had a "+" and then aria-label was "add new dependent" we'd want a muted tooltip since the visual tooltip wouldn't add any additional information. If you think a quick 5 minute chat would be useful let me know.

PR #988, opened 2021-03-04
fix(modal): Update readme with correct import

mannycarrera4:
## Summary

Noticed the import was a bit outdated, so just updated it with the correct import


NicholasBoll (approved):

NicholasBoll:
🔥

Issue #981, opened 2021-02-25
Add position and reposition to `ListModel` for rearranging Tabs


NicholasBoll:
## 🚀 Feature Proposal

The Tabs component uses a `ListModel` which currently doesn't accept an index override. If a `map` was used to create `Tab`s, an update to that list is not reflected properly in the List and can cause inconsistent rendering and behavior.

Provide an optional `index` to the `Tab` prop interface, `index` as an optional input to `registerItem` and add an `updateItemPosition` to rearrange items based on an index.

## Example

```tsx
// some dynamic list of tabs
const tabs = [{
  id: 'first',
  label: 'First',
  content: 'Contents of first Tab'
}]

const MyTabs = ({ tabs }) => {

  return (
    <Tabs>
      <TabsList>
        {tabs.map(({ id, label }, index) => {
          return <Tab id={id} index={index}>{label}</Tab>
        })
      </TabsList>
      // ...
  )
}
```

NicholasBoll:
Closed by #990

PR #990, opened 2021-03-05
feat(tabs): Update Tabs list model to support passing index position

omasrii:
Closes: #981,  These changes will allow for usages where Tabs are asynchronously created/updated and an order must be maintained  

## Summary

* feat(tabs): Add optional `index` prop to Tab interface to allow overriding position of tab item in list for some advanced usages

- Callbacks have always fired during state setting phase in a React render cycle. The callback signature was updated from `{data, state}` to `{data, prevState}` to reflect this. All docs and codesandboxes have also been updated
- Behavior hooks now receive a `ref` in case they need to fork the ref
- `mergeProps` now handles `css` and `style` props
- `ListModel` can now receive `index` for positional updates to lists
- Tabs has a `DynamicTabs` example to showcase adding and removing of tabs from a list. Tests have been added to verify proper focus management in these cases
- tsconfig was fixed for spec files in VSCode

## Checklist

- [x] `yarn test` passes


modules/_labs/tabs/react/lib/list/useListModel.tsx
```diff
@@ -42,6 +45,11 @@ export const listEventMap = createEventMap<ListEvents>()({
      * unspecified state if not managed properly.
      */
     shouldUnregisterItem: 'unregisterItem',
+    /**
+     * Should a list item be unregistered? Use this only for advanced cases. Lists can get into an
+     * unspecified state if not managed properly.
```
NicholasBoll, line 50:
```suggestion
     * Should a list item position be updated? Use this only for advanced cases. Lists can get into an
     * unspecified state if not managed properly.
```

modules/_labs/tabs/react/lib/list/useListModel.tsx
```diff
@@ -58,6 +66,13 @@ export const listEventMap = createEventMap<ListEvents>()({
      * mount/unmount cycles of rendering.
      */
     onUnregisterItem: 'unregisterItem',
+    /**
+     * Called once an item has been unregistered from the list. This is useful to do any `setState`
+     * or side effects. This method is called synchronously during a `setState` batch. Calling state
+     * setters will not cause extra renders. Item registration should be called during the
+     * mount/unmount cycles of rendering.
```
NicholasBoll, line 73:
```suggestion
     * Called once an item's position has been updated in the list. This is useful to do any `setState`
     * or side effects. This method is called synchronously during a `setState` batch. Calling state
     * setters will not cause extra renders. Item registration should be called during the
     * mount/unmount cycles of rendering.
```

modules/_labs/tabs/react/lib/Tab.tsx
```diff
@@ -120,11 +124,19 @@ const StyledButton = styled('button')<{isSelected: boolean} & StyledType>(
...
-    const tabName = useRegisterItem({state, events}, localRef, name);
+    const tabName = useRegisterItem({state, events}, localRef, name, index);
+
+    if (
+      index &&
+      state.items.length &&
+      !state.items.find((item, i) => item.index === i && item.index === index && item.id === name)
+    ) {
+      events.updateItemPosition({id: name, index});
+    }
```
NicholasBoll, line 139: `events` shouldn't be called directly in a render function since `events` typically call `React.setState` internally. This is also common functionality for any component using a list, so we should make a hook out of this. This would be a `ListModel` behavior, so it goes with `useRegisterItem`. Or perhaps it is rolled into `useRegisterItem` (since you pass the index anyways).

I'm also unsure if the model needs to record the index at all. It isn't really a concern of the model and the `state.items.find` can be expensive per render.

My suggestion is to move this logic into the `useRegisterItem` hook and add a `useLayoutEffect` section that tracks changes to the `index` and when the index changes, call `events.updateItemPosition`. It might look like this:

```tsx
export const useRegisterItem = (
  {state, events}: ListModel,
  localRef: React.RefObject<any>,
  id = '',
  index?: number
): string => {
  const [localId, setLocalId] = React.useState(id);
  const firstRender = React.useRef(true);

  useMountLayout(() => {
    const defaultId = state.indexRef.current;
    const itemId = id || String(defaultId);
    events.registerItem({item: {id: itemId, ref: localRef, index}});
    setLocalId(itemId);
    firstRender.current = false;

    return () => {
      events.unregisterItem({id: itemId});
    };
  });

  useLayoutEffect(() => {
    // Don't update on first render
    if (firstRender.current) {
      return;
    }
    
    events.updateItemPosition({id: name, index})

  }, [index])

  return localId;
};
```

modules/react/common/lib/utils/components.ts
```diff
@@ -239,49 +239,51 @@ export function useModelContext<T>(context: React.Context<T>, model?: T): T {
...
-  hook1: (model: M, props: P) => O1,
-  hook2: (model: M, props: P) => O2
-): (model: M, props: P) => P & O1 & O2;
-export function composeHooks<M, P extends {}, O1 extends {}, O2 extends {}, O3 extends {}>(
-  hook1: (model: M, props: P) => O1,
-  hook2: (model: M, props: P) => O2,
-  hook3: (model: M, props: P) => O3
-): (model: M, props: P) => P & O1 & O2 & O3;
+export function composeHooks<M, R, P extends {}, O1 extends {}, O2 extends {}>(
+  hook1: (model: M, props: P, ref: React.Ref<R>) => O1,
```
NicholasBoll, line 243: @alanbsmith This is where ref forwarding for hooks is added

modules/react/common/lib/utils/mergeProps.ts
```diff
@@ -12,9 +12,20 @@ export function mergeProps<T extends object, S extends object>(
...
+      } else if (key === 'style' && targetProps.hasOwnProperty('style')) {
+        // merge style objects
+        // @ts-ignore TS complains that key might not be on targetProps, but typeof returns
+        returnProps[key] = {...targetProps[key], ...sourceProps[key]};
+      } else if (
+        typeof sourceProps[key] === 'function' &&
+        // @ts-ignore TS complains that key might not be on targetProps, but typeof returns
+        // `undefined` in that case, so this is safe
+        typeof targetProps[key] === 'function'
+      ) {
```
NicholasBoll, line 28: @alanbsmith Added support for `css` and `style` attributes when merging props

cypress/integration/Tabs.spec.ts
```diff
@@ -6,7 +6,7 @@ describe('Tabs', () => {
   });
 
   ['Simple', 'Named Keys'].forEach(story => {
-    context(`given the '${story}' story is rendered`, () => {
+    context(`given the [Labs/Tabs/React, ${story}] story is rendered`, () => {
```
NicholasBoll, line 9: Updated to get Spec table to link to example

modules/labs-react/tabs/lib/TabPanel.tsx
```diff
@@ -60,7 +60,7 @@ export const TabPanel = createComponent('div')({
       setTabName(tabName);
 
       return () => {
-        events.unregisterItem({id: tabName});
+        events.unregisterPanel({id: tabName});
```
NicholasBoll, line 63: Total oops right here

modules/labs-react/tabs/lib/cursor/hooks.tsx
```diff
@@ -69,3 +73,50 @@ export const useRovingFocus = ({state, events}: CursorModel, elemProps = {}) =>
...
   );
 };
+
+/**
+ * Handles the roving focus behavior of a Cursor model. It should be added to the element
+ * representing a list item. It does not handle the tabIndex.
+ *
+ * @see https://www.w3.org/TR/wai-aria-practices/#kbd_roving_tabindex
+ */
+export const useRovingFocusItem = ({state}: CursorModel, elemProps = {}) => {
```
NicholasBoll, line 83: @alanbsmith This hook is how focus management is working on deleted list items. No subscriptions needed.

modules/labs-react/tabs/lib/useTabsModel.tsx
```diff
@@ -100,32 +101,40 @@ export const useTabsModel = (config: TabsModelConfig = {}): TabsModel => {
   const initialTabRef = React.useRef(config.initialTab);
   const [activeTab, setActiveTab] = React.useState(initialTabRef.current || '');
 
-  const menu = useCursorModel({
+  const cursor = useCursorModel({
```
NicholasBoll, line 104: Missed this variable when we renamed "menu" to "cursor"

tsconfig.spec.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["./jest/testing-library__jest-dom.d.ts", "./jest/verifyComponent.d.ts", "**/*.spec.ts?(x)"],
+  "include": ["./jest/testing-library__jest-dom.d.ts", "./jest/verifyComponent.d.ts", "**/*.spec.tsx"],
```
NicholasBoll, line 3: This fixes the issue where spec files weren't being properly type checked in VSCode. `verifyComponent` was registering as undefined and `expect` was returning `any`

modules/react/common/lib/utils/components.ts
```diff
@@ -239,49 +239,51 @@ export function useModelContext<T>(context: React.Context<T>, model?: T): T {
...
+  hook4: (model: M, props: P, ref: React.Ref<R>) => O4,
+  hook5: (model: M, props: P, ref: React.Ref<R>) => O5
+): (model: M, props: P, ref: React.Ref<R>) => P & O1 & O2 & O3 & O4 & O5;
+export function composeHooks<M, R, P extends {}, O extends {}>(
+  ...hooks: ((model: M, props: P, ref: React.Ref<R>) => O)[]
+): (model: M, props: P, ref: React.Ref<R>) => O {
+  return (model: M, props: P, ref: React.Ref<R>) => {
     return hooks.reverse().reduce((props: any, hook) => {
-      return hook(model, props);
+      return hook(model, props, props.ref || ref);
```
NicholasBoll, line 286: @alanbsmith Here's how `ref` is reduced when composing hooks. Refs are first passed to functions. If the hook returns `ref` in the prop list, it will be used to continue passing down, otherwise it falls back to `ref` passed to composition.

NicholasBoll (approved):

PR #992, opened 2021-03-09
Slash Imports

anicholls:
## Summary

Spike on refactoring Canvas Kit to use a slash import system rather than individual modules/packages for each component.
Opening a draft PR because I'd like to get eyes on this quickly. The sooner it goes in the better because the longer we wait, the more merge conflicts we'll have to deal with.

Before:
```tsx
import {ActionBar, Banner, BannerVariant} from '@workday/canvas-kit-react';
```
After:
```tsx
import { ActionBar } from "@workday/canvas-kit-react/action-bar";
import { Banner, BannerVariant } from "@workday/canvas-kit-react/banner";
```

- [x] Reorganize all modules into 7 top level modules
- [x] Remove each components package.json and create a new one for each package as needed
- [x] Add a new codemod CLI package
- [x] Add a codemod transform to convert the previous imports to the new ones
- [x] Update package.json dependencies to reference new packages
- [x] Working correctly in storybook
- [x] Get builds working
- [x] Get tests working
- [x] Update lint rule to restrict imports from `lib/` folder
- [x] Update create-component utility
- [x] Update documentation
- [x] Audit additional tooling to see if we'll have any issues
- [x] Make sure CSS modules still work without slash imports

**I'm proposing we do these items in a follow up PR to get this in ASAP:**
- [ ] Roll up components so you can't import from the `lib/` folder
- [ ] Test with tree shaking 
- [ ] Consider recategorizing storybook
- [ ] Consider switching back to independent versioning

---

### Codemod package

Once published, the codemod should be runnable via npx:
```sh
> npx @workday/canvas-kit-codemod <transform> [path]
```

For now you need to run it locally:
```sh
> node modules/codemod/index.js v5 [path]
```

```sh
canvas-kit-codemod <transform> [path]

Commands (transforms):
  canvas-kit-codemod v5 [path]  Canvas Kit v4 > v5 migration transform

Options:
  -v, --version  Show version number                                   [boolean]
  -h, --help     Show help                                             [boolean]
```

Note: These codemods only work on .js, .jsx, .ts, and .tsx extensions. You may need to make some manual changes in other file types (.json, .mdx, .md, etc.).

Note: You may need to run your linter after executing the codemod, as it's resulting formatting (spacing, quotes, etc.) may not match your project's styling.

To run the codemod tests:
```sh
> yarn test modules/codemod/lib/v5/spec/codemod-v5.spec.ts
``` 
For some reason there's a transpilation issue when running them from the `codemod/` folder. Haven't figured this out yet.

---

### Questions:

- Since we're leaving CSS as separate modules and putting it into maintenance mode, do we want to recategorize the storybook in a future PR? (note: this will break existing links)


utils/check-dependencies-exist.js
```diff
@@ -145,6 +145,13 @@ function formatErrorMessage(pkgFile, errors) {
 const modulePath = process.cwd();
 const packageName = require(path.join(modulePath, 'package.json')).name;
 
+// TODO: Figure out why we need this.
+if (modulePath.endsWith('/modules/react')) {
+  depCheckOptions.ignoreMatches.push('@workday/canvas-kit-react');
+} else if (modulePath.endsWith('/modules/labs-react')) {
+  depCheckOptions.ignoreMatches.push('@workday/canvas-kit-labs-react');
+}
```
anicholls, line 153: I'm not quite sure why this is needed. Oddly enough, `@workday/canvas-kit-react` only seems to complain about `@workday/canvas-kit-react/core` imports (i.e. `@workday/canvas-kit-react/icon` works fine), but `@workday/canvas-kit-labs-react` complains about any internal imports from `@workday/canvas-kit-labs-react` (e.g. `@workday/canvas-kit-labs-react/menu` or `@workday/canvas-kit-labs-react-color-picker`)

modules/codemod/lib/v5/spec/codemod-v5.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+  //   transform,
+  //   {},
+  //   `
+  // import * as foo from '@workday/canvas-kit-react';
+  //     `,
+  //   `
+  // import * as foo from "@workday/canvas-kit-react/core";
+  //     `,
+  //   'Changes namespaced bundle import to namespaced core import'
+  // );
```
anicholls, line 42: This will be a headache to account for as we need to convert this into separate imports from all of the mapped modules, then assign them all to an object called `foo`. Thoughts on whether it's worth handling this case? I can't imagine it's too common.

NicholasBoll: I suppose if someone files an issue, we could try updating based on their usage.

NicholasBoll:
Manually merged via https://github.com/Workday/canvas-kit/commit/340c8635710a04e4ec8a581b6f0bd5bd41e04287

Manual reduced the changed lines by manually overriding the `renamelimit` in git, otherwise Github would create a commit with 40K lines changed instead of around 3K!

PR #994, opened 2021-03-09
ci: Update setup-node action and remove custom matcher

NicholasBoll:
Remove custom GH Actions matcher code now that https://github.com/actions/setup-node/issues/152 is resolved.


anicholls (approved):

modules/tooltip/react/lib/Tooltip.tsx
```diff
@@ -62,7 +62,7 @@ function mergeCallbacks<T extends {[key: string]: any}>(
 }
 
 export const Tooltip = ({
-  type = 'label',
+  type = 'label1',
```
NicholasBoll, line 65:
```suggestion
  type = 'label',
```

PR #995, opened 2021-03-13
feat: Add basic primary theme support to CSS components

aaronanderson:
Updated the canvas-kit-css SASS files to support new theme variables like the canvas-kit-react theme feature so that the default main color of blueberry can be overwritten.

## Summary

Currently in the canvas-kit-css modules there is no way to override the default blueberry color on buttons,  text inputs, checkboxes, etc. This update creates new SASS variables:

$wdc-color-lightest: $wdc-color-blueberry-100 !default;
$wdc-color-light: $wdc-color-blueberry-200 !default;
$wdc-color-main: $wdc-color-blueberry-400 !default;
$wdc-color-dark: $wdc-color-blueberry-500 !default;
$wdc-color-darkest: $wdc-color-blueberry-600 !default;

and all direct updates to the blueberry color have been updated to these new overridable variables. Now it is possible to use SASS variables to create CSS files with a different theme color.

## Checklist

- [ ] `yarn test` passes
      applicable?path=/story/components-inputs-textarea-css-left-label--default


modules/core/css/lib/variables/colors.scss
```diff
@@ -9,3 +9,8 @@ $wdc-color-alert-light: $wdc-color-cantaloupe-200;
 $wdc-color-alert-dark: $wdc-color-cantaloupe-500;
 $wdc-color-shadow-light: rgba(82, 97, 115, 0.14);
 $wdc-color-shadow: rgba(82, 97, 115, 0.18);
+$wdc-color-lightest: $wdc-color-blueberry-100 !default;
+$wdc-color-light: $wdc-color-blueberry-200 !default;
+$wdc-color-main: $wdc-color-blueberry-400 !default;
+$wdc-color-dark: $wdc-color-blueberry-500 !default;
+$wdc-color-darkest: $wdc-color-blueberry-600 !default;
```
anicholls, line 16: Could you update these to include `primary` so they match the react theme colors? (e.g. `$wdc-color-primary-lightest`)

anicholls (changes requested):
Thanks for the contribution @aaronanderson! 🎉 

Just out of curiosity, could you share your use case for this functionality? As a heads up, due to the infrequent use of our CSS modules, they will be going into maintenance mode in Canvas Kit v5. We will still support them with bug fixes and significant visual updates, but they generally won't be receiving new components, additional features, etc. We're currently targeting the end of May for that release.

modules/core/css/lib/variables/colors.scss
aaronanderson: Thanks for the feedback. I updated the SASS theme variable names with a primary identifier. 

I work for a Workday Services Partner and I am currently using the Canvas-Kit CSS components in a LitElement Web Component based [supplemental library](https://github.com/aaronanderson/lwdc) that I use in internal applications that interface with Workday. 

Thanks for the heads up on the CSS module deprecation. While I am dismayed that traditional CSS and VanillaJS will no longer be supported I realize that React is the primary target JS framework. Given there is no overlap between the React and CSS modules and the component library continues to grow it is understandable that it is no longer feasible to maintain duplicate components.

In light of this new information I will reassess my approach to building applications with Canvas Kit and determine if I can incorporate the React components or maintain my own component variants like I have done for component that do not have CSS variants.

anicholls: Thanks for the info and the update! It's very cool to have external partners using and contributing to our kits. Thanks for sharing your library as well :) 

> traditional CSS and VanillaJS will no longer be supported
It's not that they won't be supported, we just won't be adding new functionality. The reason we're putting them in maintenance mode and not deprecating them completely is for people like yourself.

It's probably also worth mentioning that we plan to invest some time into exploring compile-time frameworks that will allow us to provide a framework-agnostic version of all our components at some point in the future (e.g. Svelte, Stencil, etc.). Unfortunately we just don't have the bandwidth for it at this time. But keep your eyes out for this down the road!

anicholls (approved):

modules/core/css/lib/variables/colors.scss
aaronanderson: Thanks for the encouraging news on potential plug-and-play framework-agnostic components in the future. I look forward to using them if/when they become available.

PR #998, opened 2021-03-18
docs: Relocate supplemental docs to new docs module

anicholls:
## Summary

Moves migration guides, code style, etc. docs into new `@workday/canvas-kit-docs` module so they can be imported into our new documentation site.


.storybook/main.js
```diff
@@ -39,6 +40,12 @@ module.exports = {
       ],
     });
 
+    // Update @storybook/addon-docs webpack rules to load all .mdx files in storybook
+    const mdxRule = config.module.rules.find(rule => rule.test.toString() === /\.mdx$/.toString());
+    mdxRule.use.find(loader => loader.loader.includes('@mdx-js')).options['compilers'] = [
+      createCompiler({}),
+    ];
```
anicholls, line 47: This change makes it so that every `.mdx` file found in the `stories` field above will be processed for Storybook and rendered within Storybook. This means that every `.mdx` file will require a `<Meta />` tag otherwise you will get an error in Storybook (something about a default export with no title). Previously, storybook would only render `.mdx` files with the `.story` or `.stories` prefix.

To get around this, anything you'd like visible in github, but not in Storybook should be `.md`:

- Github only: `.md`
- Storybook + Github: `.mdx` with `<Meta />`

modules/docs/mdx/changelog.stories.mdx
```diff
@@ -1,6 +1,6 @@
 import {Meta} from '@storybook/addon-docs/blocks';
-import Changelog from '../../CHANGELOG.md';
-import Markdown from '../../utils/storybook/Markdown.tsx';
+import Changelog from '../../../CHANGELOG.md';
```
NicholasBoll, line 2: Does it make sense for this file to go here? It will be published with our docs package, but won't work in the Canvas site because the `CHANGELOG.md` file will not be published. So importing in the Canvas Site will result in a "module not found" error.

Maybe it has to be? Or it gets left where it was and becomes a Storybook-only MDX file - along with the "Getting Started" since that references the README.md file which also is not published.

anicholls: Hmm... That's a good point. We could leave it at the root and change it to `.mdx` - github can render them just fine, it will just skip any jsx tags. 

However, we actually may want to render the changelog on the Canvas Site. I left it at the root because it felt weird for a project to have that buried in a module folder. What do you think?

NicholasBoll: It is a common convention to have a `CHANGELOG.md` file in the root of a project. If we want to render on the Canvas Site, we may have to copy the CHANGELOG.md file to the doc module directory as part of the build process and make sure the file is deployed to npm. I'm not sure if at that point this `MDX` wrapper is helpful to the Canvas Site since the Storybookisms have to be stripped away.

I think our Readme is going to be reduced to a much simpler version and have a link to more official documentation anyways. I'm not sure how useful that will be for the Canvas Site. I think having these MDX files are fine here even if not used by the Canvas Site to avoid too much odd configuration.

NicholasBoll (approved):

PR #1009, opened 2021-03-24
fix: Allow refs to be passed to createComponent components

NicholasBoll:


modules/react/common/spec/components.test-d.tsx
```diff
@@ -0,0 +1,51 @@
...
+    },
+  });
+});
+
+it('should allow a valid ref when wrapping components', () => {
+  const Component = createComponent('button')({Component: (props: {}) => null});
+
+  createComponent('button')({
+    Component: (props: {}, ref, Element) => {
+      return <Component ref={ref} />;
```
NicholasBoll, line 48: This is where the issue was. According to `PropsWithAs`, `ElementType` was a Component Class for Components created with `createComponent`. Passing a ref of an element caused signatures to not match. I had to extract the element type in `PropsWithAs` to fix this.

modules/react/common/lib/utils/components.ts
```diff
@@ -29,7 +49,7 @@ export type PropsWithAs<
          * If `as` is a component, the reference will be to that component (or element if the component
          * uses `React.forwardRef`).
          */
-        ref?: React.Ref<ElementType>;
+        ref?: ExtractRef<ElementType>;
```
NicholasBoll, line 52: Here's the fix. The ref type must be extracted from the `ElementType` generic which could be `undefined`, an `Element`, or an `ElementComponent<'someElementType', SomeProps>`

modules/react/common/spec/components.test-d.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import {expectTypeOf} from 'expect-type';
```
NicholasBoll, line 2: [expect-type](https://github.com/mmkal/ts/tree/master/packages/expect-type) is a simple no-dependency package that has no runtime and instead uses Typescript to static analyze with an expect-style matching. It isn't perfect, but it works without a complicated setup. Other contenders were [dtslint](https://github.com/microsoft/dtslint) and [tsd](https://github.com/SamVerschueren/tsd) both which were complicated to set up and had a large amount of dependencies.

It would be nice if `expect-type` failed more gracefully. It can be used as more of a regression-style test. Failures result in some complicated type errors or don't fail correctly. It worked well for enough cases to be useful. It is a very small dependency (basically a 135 TS file) and uses `tsc` to work.

modules/react/common/spec/components.test-d.tsx
```diff
@@ -0,0 +1,51 @@
...
+});
+
+it('should add sub-components to the signature', () => {
+  const component = createComponent('div')({
+    Component: (props: {foo: 'bar'}) => null,
+    subComponents: {
+      Foo: 'bar',
+    },
+  });
+  expectTypeOf(component).toEqualTypeOf<ElementComponent<'div', {foo: 'bar'}> & {Foo: number}>();
```
NicholasBoll, line 18: This is actually something that `expect-type` fails at. Not sure why. `Foo` is actually a different type here...

modules/react/common/spec/components.test-d.tsx
```diff
@@ -0,0 +1,51 @@
...
+  });
+});
+
+it('should assign ref and Element correctly for createComponent components', () => {
+  const component = createComponent('article')({Component: (props: {}) => null});
+
+  createComponent(component)({
+    Component: (props: {}, ref, Element) => {
+      expectTypeOf(ref).toEqualTypeOf<React.Ref<HTMLElement>>();
+      expectTypeOf(Element).toEqualTypeOf<ElementComponent<'article', {}>>();
```
NicholasBoll, line 37: This is kind of a cool way to verify what gets returned by functions in an explicit way.

modules/react/common/spec/components.test-d.tsx
anicholls: Nit: This one seems a bit out of place without context since there's no `expect` call. Do you mind adding a comment referencing this PR or something?

anicholls (approved):

modules/react/common/spec/components.test-d.tsx
NicholasBoll: Hmm. I think this is more complicated than it needs to be. We really are ensuring `ref` can be sent to `<Component>`. It doesn't need another `createComponent` wrapper.

NicholasBoll: I'm cleaning it up.

PR #1016, opened 2021-04-01
fix(select): Fix undesired scrolling when activating the menu

jamesfan:
## Summary

Fixes an issue where the page was scrolling to the top after activating the Select Menu. This Codesandbox provided by the reporter of the issue illustrates the problem: https://codesandbox.io/s/blur-render-forked-d03kk

The Menu is given focus when it's displayed, which can cause issues when focus is applied too soon before the Menu has been positioned and visually attached to its corresponding Select Button. I've included additional detail in [the test](https://5d854c26ba934e0020f5e98a-pqyzxzvepr.chromatic.com/?path=/story/testing-react-labs-select--portal-test) I created to protect against this type of behavior in the future (scroll to **Blur Test** at the bottom).

I've also removed visual snapshots for a number of Select stories which were never meant to have snapshots in the first place (they're used for Cypress tests and manual testing).


modules/_labs/select/react/lib/Select.tsx
```diff
@@ -418,16 +404,16 @@ class Select extends React.Component<SelectProps, SelectState> {
     this.fireChangeEvent(this.normalizedOptions[index].value);
   };
 
-  handleMenuBlur = (event: React.FocusEvent): void => {
-    this.toggleMenu(false);
-  };
```
jamesfan, line 423: With `preventDefault` removed on the mousedown handler on the Select Button, we can no longer hide the Menu on blur (I was hiding the Menu on blur to handle its click-outside-to-dismiss behavior). Hiding the Menu on blur prevents you from clicking on the Select Button to dismiss the Menu while its open given the way the event handlers interact.

Clicking the Button:
1. Gives focus to the Button, which triggers the Menu blur handler, which causes the Menu to start closing... 
2. And triggers the Button's click handler which sees the Menu closing and immediately re-opens it again (this is intended behavior: if you activate the Menu while it's closing, it should re-open).

Fortunately, I was able to use Popup's `useCloseOnOutsideClick` hook as a replacement (see below).

modules/_labs/select/react/lib/SelectMenu.tsx
```diff
@@ -258,14 +268,7 @@ const SelectMenu = ({
...
   useCloseOnEscape(popupRef, () => onCloseOnEscape?.());
-
-  // TODO: use the useCloseOnOutsideClick hook after it's been modified to handle
-  // multiple things happening on the same click. Currently, the hook doesn't
-  // support the use case where one Select Menu is activated by clicking on its
-  // Select Button while another Select Menu is already active (both Menus are
-  // visible instead of the original Menu dismissing). In order to support the
-  // correct behavior for now, we're dismissing the Menu on blur instead of
-  // using the hook.
+  useCloseOnOutsideClick(popupRef, () => onCloseOnOutsideClick?.());
```
jamesfan, line 271: I was able to use Popup's `useCloseOnOutsideClick` hook to replace the functionality lost by removing the Menu's blur handler (previously, `useCloseOnOutsideClick` could leave multiple Select Menus open if a page contained multiple Selects -- that issue has since been resolved).

`useCloseOnOutsideClick` also has the additional benefit of making it much easier to inspect the Menu (previously, the Menu would remove itself from the DOM as soon as you clicked anywhere else, including the browser developer tools).

modules/_labs/select/react/stories/stories_visualTesting.tsx
```diff
@@ -163,159 +157,6 @@ export const SelectStatesOption = () => (
...
-    <Container>
-      <p>Disabled options may not be assistively focused using the keyboard.</p>
-      <FormField label="Label (Disabled Options)" inputId="select-disabled-options">
-        {controlComponent(<Select name="select-disabled-options" options={disabledOptions} />)}
-      </FormField>
-    </Container>
-  </div>
-);
-
-export const PortalTest = () => (
```
jamesfan, line 251: There were a number of stories in `stories_visualTesting.tsx` that don't require snapshots and never should have been there in the first place (stories used for Cypress and manual testing). I've moved them to `stories_testing.tsx`.

modules/_labs/select/react/stories/stories_CypressTesting.tsx
```diff
@@ -1,167 +1,20 @@
...
-    label: option.label || option.value,
-    value: option.value,
-  };
-});
+import {manyOptions, options} from './stories';
 
-export default withSnapshotsEnabled({
-  title: 'Testing/React/Labs/Select',
+export default {
+  title: 'Testing/React/Labs/Select/Cypress',
```
jamesfan, line 15: There were a number of stories in `stories_VisualTesting.tsx` that don't require snapshots and never should have been there in the first place (stories used for Cypress and manual testing). I've moved them to `stories_CypressTesting.tsx.` Following [the convention](https://workday.github.io/canvas-kit/?path=/story/testing-react-labs-side-panel--side-panel-states) established by the Labs Side Panel, these Cypress stories are contained in a separate `Cypress` folder in Storybook (one level down from the visual testing stories).

modules/_labs/select/react/lib/SelectBase.tsx
```diff
@@ -312,6 +312,9 @@ const SelectBase = ({
         ...(onOptionSelection
           ? {
               onClick: (event: React.MouseEvent) => {
+                // TODO: Figure out why this preventDefault call exists.
+                // Removing it doesn't have any obvious consequences,
+                // but need to do more careful testing.
                 event.preventDefault();
```
NicholasBoll, line 318: My best guess as to why this is here is if the menu option is a link or a button without an explicit `type="button"` which would cause the browser to navigate (if a link) or submit a form (if a button). Ideally, `event.preventDefault()` is kept to a minimum and only for specific issues as we might be accidentally preventing valid [default actions](https://www.w3.org/TR/uievents/#default-action) from occurring.

NicholasBoll (approved):

PR #1017, opened 2021-04-02
chore: Refactor Button to use createComponent

NicholasBoll:
## Summary

- Update Migration guide
- Add codemod
- Update codemods to handle multiple transforms
- Type check spec files
- Remove default export from Button
- Remove `beta_Button`


modules/codemod/lib/v5/index.ts
```diff
@@ -1,192 +1,22 @@
...
+
+import {API, FileInfo, Options} from 'jscodeshift';
+
+export default function transformer(file: FileInfo, api: API, options: Options) {
+  // These will run in order. If your transform depends on others, place yours after dependent
+  // transforms
+  const fixes = [slashImports, removeDefaultImports, removeButtonEnums, renameButtonRefs];
+
+  let src = file.source;
+  fixes.forEach(fix => {
```
willklein, line 14: Consider `reduce` for this but that's a completely optional suggestion.

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -69,3 +76,25 @@ Components are transitioning to [Compound Components](./COMPOUND_COMPONENTS.md)
...
+    variant: ButtonVariant
+  }
+  // after
+  <Button variant="primary" />
+  interface Props {
+    variant: 'primary' | 'secondary'
+  }
+  ```
+- Buttons now use `createComponent` utility from the `common` module which now forwards `ref` and allows `as` to change the underlying element.
+  - 🤖 `buttonRef` was renamed to `ref` provided by `createComponent`. `buttonRef` is no longer valid. The codemod will rewrite all instances of `buttonRef` to `ref`.
```
willklein, line 98: May be worth a warning/note/blockquote to call out that this could affect other code unintentionally.

modules/codemod/lib/v5/index.ts
NicholasBoll: That's fair. In general, I'm not happy with iteration support in Typescript. `Object.keys` give `string` as an item type and building objects in a `reduce` creates a lot of type overriding. This case will work well for a `reduce` though.

modules/labs-react/pagination/lib/Pagination/Controls.tsx
```diff
@@ -62,8 +63,8 @@ export const StepToPreviousButton = ({onClick, model, ...restProps}: ControlButt
   return (
     <IconButton
       aria-disabled={isDisabled || undefined}
-      variant={IconButton.Variant.Square}
-      size={IconButton.Size.Small}
+      variant={'square'}
```
anicholls, line 66: @NicholasBoll why do some of these turn into `variant="square"` (line 43) and others `variant={'square'}`?

modules/react/button/lib/DeleteButton.tsx
```diff
@@ -1,31 +1,20 @@
...
 } from '@workday/canvas-kit-react/common';
 
-export interface DeleteButtonProps
-  extends React.ButtonHTMLAttributes<HTMLButtonElement>,
-    Themeable,
-    GrowthBehavior {
+export interface DeleteButtonProps extends Themeable, GrowthBehavior {
   /**
    * The size of the Button.
-   * @default 'medium'
```
anicholls, line 17: Why remove the default?

modules/codemod/lib/v5/spec/removeButtonEnums.spec.ts
```diff
@@ -0,0 +1,390 @@
+import {expectTransformFactory} from './expectTransformFactory';
+import transform from '..';
+
+const expectTransform = expectTransformFactory(transform);
```
NicholasBoll, line 4: The `expectTransformFactory` seems to be just enough abstraction to make tests easier to write and understand. The signature of `expectTransform` is `input, output, options?`.

modules/codemod/lib/v5/spec/removeButtonEnums.spec.ts
```diff
@@ -0,0 +1,390 @@
...
+        ButtonSize,
+      } from '@workday/canvas-kit-react/button';
+    `;
+
+    const expected = `
+      import { Button } from '@workday/canvas-kit-react/button';
+    `;
+
+    expectTransform(input, expected);
+  });
```
NicholasBoll, line 27: This way of writing codemod tests aligns more with all our other tests of having well-defined `describe` and `it` blocks. The `input` and `expected` constants have enough of a visual separation to more easily understand via syntax highlighting where the input string and expected strings start and end. The `it` block also allows for `.only` for test isolation.

modules/react/button/lib/DropdownButton.tsx
```diff
@@ -1,68 +1,51 @@
...
 import {caretDownIcon} from '@workday/canvas-system-icons-web';
-import {GrowthBehavior, useTheme, Themeable} from '@workday/canvas-kit-react/common';
-import {DropdownButtonVariant, ButtonIconPosition, ButtonOrAnchorComponent} from './types';
+import {
+  GrowthBehavior,
+  useTheme,
+  Themeable,
+  createComponent,
+} from '@workday/canvas-kit-react/common';
+import './types';
```
anicholls, line 9: Is this necessary?

modules/codemod/package.json
```diff
@@ -4,8 +4,7 @@
   "license": "Apache-2.0",
   "version": "0.0.0",
   "description": "A collection of codemods for use on Workday Canvas Kit packages.",
-  "main": "dist/commonjs/index.js",
-  "module": "dist/es6/index.js",
+  "main": "index.js",
```
NicholasBoll, line 7: I removed the build step for the codemods since jscodeshift supports Typescript files and this simplifies development and packaging.

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -21,6 +22,21 @@ about the update.
 
 Due to the infrequent use of our CSS modules, they will be going into maintenance mode in Canvas Kit v5. We will still support `@workday/canvas-kit-css` with bug fixes and significant visual updates, but it generally won't be receiving new components, additional features, etc. This will allow us to provide more focused support, and dedicate our efforts to making bigger and better improvements to our most used components: Canvas Kit React. If you have questions or concerns, please [let us know](https://github.com/Workday/canvas-kit/issues/new?labels=&template=question.md).
 
+### Codemod
```
NicholasBoll, line 25: Codemod is now it's own section with code mod info specific to each breaking change.

modules/labs-react/drawer/lib/DrawerHeader.tsx
```diff
@@ -12,7 +12,7 @@ export interface DrawerHeaderProps extends React.HTMLAttributes<HTMLDivElement>
   /**
    * The function called when the DrawerHeader close button is clicked.
    */
-  onClose?: IconButtonProps['onClick'];
+  onClose?: React.MouseEventHandler<HTMLButtonElement>;
```
NicholasBoll, line 15: Button props no longer extend from a concrete interface since `createComponent` returns an `ElementComponent<'button', ButtonProps>` type that defined Props only in usage. The presence of an `as` will change this interface which is why `ButtonProps` contains no element interface specific to any one element. Uses of `Button` will have to manually determine interfaces used if required. Usage of `mergeProps` and Compound Component APIs in general reduce the need for this though.

The `DrawHeader` component is an example of a complex component that doesn't use the Compound Component API and thus does not give direct access to elements and therefore has a fractured component interface directly creating "prop holes" to access underlying elements. This will still happen in application components where Canvas Kit components are wrapped in configuration APIs. I don't have a direct way to codemod this though...

anicholls:
Button `README.md` needs to be updated.

modules/labs-react/drawer/spec/Drawer.spec.tsx
```diff
@@ -20,7 +16,7 @@ describe('Drawer', () => {
   });
 
   test('should not render a close icon button', async () => {
-    const {container} = render(<DrawerHeader headerText={'Header Title'} />);
+    const {container} = render(<DrawerHeader title={'Header Title'} />);
```
NicholasBoll, line 19: I fixed type checking in spec files in #958, but I didn't add type checking to the CLI. So type errors in spec files were only visible in an IDE. I've added a spec file typecheck to CI now, so these types of type errors will now be caught.

modules/labs-react/pagination/lib/Pagination/PageButton.tsx
```diff
@@ -17,14 +17,15 @@ const StyledPageButton = styled(IconButton)<{toggled?: boolean}>(
...
+    return toggled ? toggledStyles : {};
   }
 );
 
-export interface PageButtonProps extends IconButtonProps {
-  model: PaginationModel;
-  pageNumber: number;
-}
+export type PageButtonProps = IconButtonProps &
+  React.ButtonHTMLAttributes<HTMLButtonElement> & {
```
NicholasBoll, line 25: When `PageButton` uses `createComponent`, the props will no longer need to extend from `React.ButtonHTMLAttributes`

modules/labs-react/side-panel/spec/SSR.spec.tsx
```diff
@@ -7,7 +7,7 @@ import SidePanel from '../';
 
 describe('SidePanel', () => {
   it('should render on a server without crashing', () => {
-    const ssrRender = () => renderToString(<SidePanel />);
+    const ssrRender = () => renderToString(<SidePanel touched={false} />);
```
NicholasBoll, line 10: Another type error not caught in a spec file

modules/react/button/lib/DeleteButton.tsx
```diff
@@ -1,31 +1,20 @@
...
   size?: 'small' | 'medium' | 'large';
-  /**
-   * The ref to the button that the styled component renders.
-   */
-  buttonRef?: React.Ref<HTMLButtonElement>;
-  /**
-   * The alternative container type for the button. Uses Emotion's special `as` prop.
-   * Will render an `a` tag instead of a `button` when defined.
-   */
-  as?: 'a';
```
NicholasBoll, line 28: Both `buttonRef` and `as` are handled by `createComponent`

modules/react/button/spec/Button.spec.tsx
```diff
@@ -9,25 +9,28 @@ import {
...
+  Button,
+  DeleteButton,
+  DropdownButton,
+  HighlightButton,
+  OutlineButton,
+  TextButton,
+  DeprecatedButton,
+  // We need to cast as `any` and cast as a specific button because TS will complain about no call signatures...
+] as any[]).forEach((ButtonComponent: ElementComponent<'button', ButtonProps>) => {
+  console.log('displayName', ButtonComponent.displayName);
```
NicholasBoll, line 25: Remove this

modules/react/button/stories/stories_Button.tsx
```diff
@@ -1,155 +1,119 @@
...
-import {jsx} from '@emotion/core';
+
+import React from 'react';
 import {storiesOf} from '@storybook/react';
 import withReadme from 'storybook-readme/with-readme';
 
 import {editIcon, playCircleIcon, activityStreamIcon} from '@workday/canvas-system-icons-web';
 
-import {Button} from '../index';
+import {Button} from '@workday/canvas-kit-react/button';
```
NicholasBoll, line 9: All stories should have package name imports. Also this file was processed by codemods which requires imports from CK.

modules/react/button/stories/visual-testing/stories_Button.tsx
```diff
@@ -50,9 +51,4 @@ export const ButtonStates = () => (
   </StaticStates>
 );
 
-export const ButtonThemedStates = () => <ButtonStates />;
-ButtonThemedStates.parameters = {
-  canvasProviderDecorator: {
-    theme: customColorTheme,
-  },
-};
+export const ButtonThemedStates = () => <ButtonStates theme={{canvas: customColorTheme}} />;
```
NicholasBoll, line 54: This change is subtle, but Docs mode doesn't handle story parameter changes like this. I thought it made more sense to pass the theme to the story instead. This way docs mode renders the theme story correctly.

modules/react/color-picker/spec/ColorInput.spec.tsx
```diff
@@ -31,7 +31,7 @@ describe('ColorInput', () => {
       test('should render a ColorInput with a value and check', () => {
         const {container, getByRole} = render(<ColorInput value={value} showCheck={true} />);
 
-        expect(getByRole('textbox').value).toBe(value);
+        expect(getByRole('textbox')).toHaveValue(value);
```
NicholasBoll, line 34: The type error said the `.value` is possibly `undefined`. Since I fixed the type definitions for spec files, `toHaveValue` is defined on the `expect` interface and is the correct matcher in this case.

modules/react/common/spec/components.spec.tsx
```diff
@@ -173,29 +173,29 @@ describe('composeHooks', () => {
   });
 
   it('should merge properties from both hooks', () => {
-    const props = composeHooks(hook1, hook2)(model, {});
+    const props = composeHooks(hook1, hook2)(model, {}, null);
```
NicholasBoll, line 176: New type errors showed the test didn't pass the 3rd parameter. Lack of type checking makes it so easy to miss these subtle changes.

modules/react/common/spec/focusRing.spec.ts
```diff
@@ -3,12 +3,12 @@ import {memoizedFocusRing} from '../lib/styles/focusRing';
...
-      ringWidth: 3,
-      separationWidth: 1,
+      width: 3,
+      separation: 1,
       animate: true,
       inset: 'outer',
-      innerShadowColor: '#333333',
-      outerShadowColor: '#666666',
+      innerColor: '#333333',
+      outerColor: '#666666',
```
NicholasBoll, line 11: Same issue here. The test passed because it only made sure the function returned the same reference, but missed type interface changes since spec files were never type checked.

package.json
```diff
@@ -145,10 +145,11 @@
     "commitmsg": "commitlint --edit $HUSKY_GIT_PARAMS",
     "build-storybook": "build-storybook -c .storybook -o docs",
     "lint": "node utils/check-lockfile.js && node utils/check-mismatched-dependencies.js && eslint -c ./.eslintrc.js --ext=jsx,ts,tsx .",
-    "typecheck": "npm-run-all --parallel typecheck:src typecheck:cypress typecheck:types",
+    "typecheck": "npm-run-all --parallel typecheck:src typecheck:cypress typecheck:types typecheck:specs",
```
NicholasBoll, line 148: Here's where specs are explicitly type checked in our CI

tsconfig.json
```diff
@@ -20,5 +20,9 @@
     "types": [],
     "incremental": true
   },
+  "ts-node": {
+    "transpileOnly": true,
+    "module": "commonjs"
+  },
```
NicholasBoll, line 26: I don't remember why this was necessary...

tsconfig.spec.json
```diff
@@ -1,11 +1,11 @@
 {
   "extends": "./tsconfig.json",
   "include": ["./jest/testing-library__jest-dom.d.ts", "./jest/verifyComponent.d.ts", "**/*.spec.tsx", "**/*.spec.ts"],
-  "exclude": [],
+  "exclude": ["cypress"],
```
NicholasBoll, line 4: Cypress has an explicit type check command. Leaving this out means that Cypress spec files are checked without defining the Cypress type interface which errors on `cy is not defined`.

tsconfig.spec.json
```diff
@@ -1,11 +1,11 @@
...
   "compilerOptions": {
     "declaration": false,
     "declarationMap": false,
-    "noImplicitReturns": true,
+    "noImplicitReturns": false,
```
NicholasBoll, line 8: I'm not sure why this is necessary. Leaving it to `true` cause the `spec` typecheck to find return errors in source files which is confusing. There's a separate type check process for source files. I also don't think implicit return checks are necessary for Jest spec files. You can see the TS config file here is more lax than the config for source files. I don't consider jumping through extra hoops to keep the type system happy is productive in spec files.

modules/codemod/package.json
```diff
@@ -4,8 +4,7 @@
...
   "sideEffects": false,
   "bin": {
     "canvas-kit-codemod": "dist/commonjs/index.js"
```
anicholls, line 10:
```suggestion
    "canvas-kit-codemod": "index.js"
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -21,6 +22,21 @@ about the update.
...
+We've provided a [new codemod package](./modules/codemod) that you can use to update the majority of these breaking code changes. Simply run:
+
+```sh
+> npx @workday/canvas-kit-codemod v5 [path]
+```
+> Note: This codemod only work on .js, .jsx, .ts, and .tsx extensions. You may need to make some manual changes in other file types (.json, .mdx, .md, etc.).
+
+> Note: You may need to run your linter after executing the codemod, as it's resulting formatting (spacing, quotes, etc.) may not match your project's styling.
+
+Breaking changes covered by this codemox will be marked with a 🤖.
```
anicholls, line 36:
```suggestion
Breaking changes accounted for by this codemod will be marked with a 🤖.
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -38,16 +54,7 @@ import canvas from "@workday/canvas-kit-react/core";
...
-
-```sh
-> npx @workday/canvas-kit-codemod v5 [path]
-```
-> Note: This codemod only work on .js, .jsx, .ts, and .tsx extensions. You may need to make some manual changes in other file types (.json, .mdx, .md, etc.).
-
-> Note: You may need to run your linter after executing the codemod, as it's resulting formatting (spacing, quotes, etc.) may not match your project's styling.
-
-Please [let us know](https://github.com/Workday/canvas-kit/issues/new?labels=bug&template=bug.md) if you have any troubles or missed use cases with this codemod. The `@workday/canvas-kit-codemod` package will help us maintain additional codemod transforms to make future migrations easier.
+This change is covered by the codemod, but doesn't currently run against `package.json` files. Those files will have to be updated manually.
```
anicholls, line 57:
```suggestion
This change is covered by the codemod, but it doesn't currently run on files that don't use the .js, .jsx, .ts, and .tsx extensions. Those files will have to be updated manually.
```

modules/react/button/lib/DeleteButton.tsx
NicholasBoll: Default is handled by prop tables. I was tempted to get rid of the documentation of props in READMEs, but I wanted to do that with the migration to MDX and I didn't want to do that as part of this PR. If you feel strongly, we can revert the changes to `@default` in JSDoc

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -69,3 +76,25 @@ Components are transitioning to [Compound Components](./COMPOUND_COMPONENTS.md)
 PR:
 
 - https://github.com/Workday/canvas-kit/pull/953
+
+### Button
+
+- `beta_Button` export was removed.
```
anicholls, line 82: Shouldn't the codemod handle this too? 🤖

anicholls (approved):
Left some comments, but once those are addressed, this LGTM 👍 Thanks!

modules/labs-react/pagination/lib/Pagination/Controls.tsx
NicholasBoll: The codemod does `{"square"}` since I'm really adjusting a `MemberExpression` to a `Literal`, but the node is still inside a `JSXExpressionContainer`. It is a stylistic preference to turn them into `JSXLiteral` instead. I was doing this manually with a replace-all regex post-processing. It probably happened as a result of a file being saved (turning `"` into `'`) before the Regex was run. I'll search for these patterns and adjust them for consistency.

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
NicholasBoll: Yes. Definitely. I see 102 instances on GHE alone

modules/labs-react/pagination/lib/Pagination/Controls.tsx
anicholls: But these are both the same context: returning a JSX element with `variant={IconButton.Variant.Square}`. The codemod has updated both, but one gets `"square"`, the other gets `{'square'}` - the inconsistency is what I'm wondering about.

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
NicholasBoll: All cases I could find looked like:

```tsx
import { beta_Button as Button } from '@workday/canvas-kit-react-button;
```

NicholasBoll: The previous will now be rewritten as:
```tsx
import { Button } from '@workday/canvas-kit-react-button;
```

Also renaming is supported:
```tsx
// before
import { beta_Button as BetaButton } from '@workday/canvas-kit-react-button';

// after
import { Button as BetaButton } from '@workday/canvas-kit-react-button';
```

modules/react/button/lib/DropdownButton.tsx
NicholasBoll: heh. Oops. All invalid imports were removed, but nothing checks an empty import. This could happen if only Variants are imported, but this is very unlikely in application code. It won't hurt if this happens, but I'll remove that here.

modules/labs-react/pagination/lib/Pagination/Controls.tsx
NicholasBoll: Fixed

modules/react/button/lib/DeleteButton.tsx
NicholasBoll: Ugh. I'll reverse all of these. It turns out React docgen doesn't understand this very well.

modules/react/tooltip/spec/useTooltip.spec.tsx
```diff
@@ -1,5 +1,3 @@
-/// <reference types="@testing-library/jest-dom/extend-expect" />
```
NicholasBoll, line 1: These are no longer needed

modules/react/button/stories/visual-testing/stories_TextButton.tsx
```diff
@@ -9,30 +9,31 @@ import {
...
           ],
           allCaps: [
             {value: false, label: ''},
             {value: true, label: 'All Caps'},
           ],
           size: [
-            {value: TextButton.Size.Small, label: 'Small'},
-            {value: TextButton.Size.Large, label: 'Large'},
+            {value: 'small', label: 'Small'},
+            {value: 'medium', label: 'Medium'},
```
NicholasBoll, line 36: Not 100% sure why the visual tests were testing 'Large' when the component states only `small` and `medium` are allowed. The visual tests look the same except "Medium" is a longer word and sometimes requires more lines to render the label which is a noisy diff.

PR #1018, opened 2021-04-05
build: Docs mdx build

anicholls:
## Summary

Add a simple build process that strips out the storybook imports and `<Meta />` tags for mdx files in the `@workday/canvas-kit-docs` module.


modules/docs/utils/build.js
```diff
@@ -0,0 +1,56 @@
...
+  var targetFolder = path.join(target, path.basename(source));
+  if (!fs.existsSync(targetFolder)) {
+    fs.mkdirSync(targetFolder);
+  }
+
+  if (fs.lstatSync(source).isDirectory()) {
+    files = fs.readdirSync(source);
+    files.forEach(file => {
+      const curSource = path.join(source, file);
+      console.log(curSource);
```
NicholasBoll, line 41: This could be removed
```suggestion
```

modules/docs/package.json
```diff
@@ -28,6 +28,7 @@
   "scripts": {
     "watch": "yarn build:es6 -w",
     "clean": "rimraf dist && rimraf ts3.5 && rimraf .build-info && mkdirp dist && mkdirp ts3.5/dist",
+    "build:mdx": "node ./utils/build.js mdx/",
```
NicholasBoll, line 31: I couldn't suggest on the `build` script line, but currently nothing is calling this. Should this script be added to the `build` script below?

NicholasBoll (approved):

PR #1022, opened 2021-04-09
fix: Scope buttonRef codemod to only Button components

NicholasBoll:
Fixes `renameButtonRef` to only include named imports from the CK button module. Handles renames. Tested on our codebase.

### DIFF
--- a/modules/codemod/lib/v5/renameButtonRefs.ts
+++ b/modules/codemod/lib/v5/renameButtonRefs.ts
@@ -1,12 +1,30 @@
 import {API, FileInfo, Options, JSXIdentifier} from 'jscodeshift';
 import {getImportRenameMap} from './getImportRenameMap';
 
+const buttonNames = [
+  'Button',
+  'DeleteButton',
+  'deprecated_Button',
+  'DropdownButton',
+  'HighlightButton',
+  'Hyperlink',
+  'IconButton',
+  'OutlineButton',
+  'TextButton',
+  'ToolbarDropdownButton',
+  'ToolbarIconButton',
+];
+
 export default function transformer(file: FileInfo, api: API, options: Options) {
   const j = api.jscodeshift;
 
   const root = j(file.source);
 
-  const {containsCanvasImports} = getImportRenameMap(j, root, '@workday/canvas-kit-react/button');
+  const {containsCanvasImports, importMap} = getImportRenameMap(
+    j,
+    root,
+    '@workday/canvas-kit-react/button'
+  );
 
   if (!containsCanvasImports) {
     return file.source;
@@ -14,8 +32,13 @@ export default function transformer(file: FileInfo, api: API, options: Options) 
 
   root
     .find(j.JSXIdentifier, (value: JSXIdentifier) => value.name === 'buttonRef')
-    .replaceWith(() => {
-      return j.jsxIdentifier('ref');
+    .replaceWith(nodePath => {
+      // only replace via one of our Button imports
+      const elementName = nodePath.parent?.parent?.value?.name?.name;
+      if (buttonNames.includes(elementName) || buttonNames.includes(importMap[elementName])) {
+        return j.jsxIdentifier('ref');
+      }
+      return nodePath.value;
     });
 
   return root.toSource();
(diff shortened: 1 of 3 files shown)


modules/codemod/lib/v5/spec/renameButtonRefs.spec.ts
```diff
@@ -19,4 +19,102 @@ describe('renameButtonRefs', () => {
 
     expectTransform(input, expected);
   });
+
+  it('should replace "buttonRef" with "ref" on renamed Button components', () => {
```
alanbsmith, line 23: 👍🏻

modules/codemod/lib/v5/spec/renameButtonRefs.spec.ts
```diff
@@ -19,4 +19,102 @@ describe('renameButtonRefs', () => {
...
+      import {SomeOtherButton} from '../Button'
+
+      <SomeOtherButton buttonRef={ref} />
+    `;
+
+    const expected = `
+      import {Button} from '@workday/canvas-kit-react/button'
+      import {SomeOtherButton} from '../Button'
+
+      <SomeOtherButton buttonRef={ref} />
```
alanbsmith, line 51: Nice.

modules/codemod/lib/v5/spec/renameButtonRefs.spec.ts
```diff
@@ -19,4 +19,102 @@ describe('renameButtonRefs', () => {
...
+        <DropdownButton ref={ref} />
+        <HighlightButton ref={ref} />
+        <Hyperlink ref={ref} />
+        <IconButton ref={ref} />
+        <OutlineButton ref={ref} />
+        <TextButton ref={ref} />
+        <ToolbarDropdownButton ref={ref} />
+        <ToolbarIconButton ref={ref} />
+      </>
+    `;
```
alanbsmith, line 116: Awesome.

modules/codemod/lib/v5/renameButtonRefs.ts
```diff
@@ -1,21 +1,44 @@
...
     .find(j.JSXIdentifier, (value: JSXIdentifier) => value.name === 'buttonRef')
-    .replaceWith(() => {
-      return j.jsxIdentifier('ref');
+    .replaceWith(nodePath => {
+      // only replace via one of our Button imports
+      const elementName = nodePath.parent?.parent?.value?.name?.name;
+      if (buttonNames.includes(elementName) || buttonNames.includes(importMap[elementName])) {
+        return j.jsxIdentifier('ref');
+      }
+      return nodePath.value;
```
alanbsmith, line 41: Ok, the logic here is basically:
• If a file is importing from `/button`
• and they're using one of the CK Buttons listed above
• and there's a `buttonRef` prop
• transform it to `ref`

Cool. That covers the edge cases pretty well, I think.

modules/codemod/lib/v5/renameButtonRefs.ts
```diff
@@ -1,21 +1,44 @@
...
     return file.source;
   }
 
   root
     .find(j.JSXIdentifier, (value: JSXIdentifier) => value.name === 'buttonRef')
-    .replaceWith(() => {
-      return j.jsxIdentifier('ref');
+    .replaceWith(nodePath => {
+      // only replace via one of our Button imports
+      const elementName = nodePath.parent?.parent?.value?.name?.name;
```
alanbsmith, line 37: This chaining feels tenuous, but that's been my experience with parsing AST, so it's probably fine.

alanbsmith (approved):
Looks good! Thanks for the fix!

modules/codemod/lib/v5/renameButtonRefs.ts
NicholasBoll: The alternative would be to find all `JSXOpeningElement` nodes and filter based on a match, then walk through all `JSXAttributes` looking for the right `JSXIdentifier` and replace the name. This felt simpler. `nodePath.parent` returns `any`, but I didn't bother to set up proper types. Technically those parents and values could be null, so I just added existential operators.

PR #1028, opened 2021-04-16
chore(card): Convert to a Compound Component and create codemod

### LINKED ISSUE #1023: Convert Card into a Compound Component
## 🚀 Feature Proposal

Convert Card into a Compound Component. `heading` and `headingId` props can be removed and instead export a `Card.Header` sub component.

## Motivation

Other components that use Card and need to grant element-level access to the Header will need this. Without this, `Popup` and `Modal` will have to bypass the props and re-create a Header sub component. For example, Modal needs to make the Header an `h1`, but the header in card is a hard-coded `h3`.

## Example

```tsx
// before
<Card heading={'Card Header'} headingId={id}>Card Body</Card>

// after
<Card>
  <Card.Header id={id} as="h1">Card Header</Card.Header>
  <Card.Body>Card Body</Card.Body>
</Card>
```

NicholasBoll:
- Add unit tests for each component
- Create codemod to handle rewriting to new API
- Update documentation
Fixes #1023

### DIFF
--- a/modules/codemod/lib/v5/compoundCard.ts
+++ b/modules/codemod/lib/v5/compoundCard.ts
@@ -1,0 +1,121 @@
+import {API, FileInfo, Options, JSXElement, JSXAttribute, JSXSpreadAttribute} from 'jscodeshift';
+
+import {getImportRenameMap} from './getImportRenameMap';
+
+function filterDefined<T>(input: any): input is T {
+  return !!input;
+}
+
+export default function transformer(file: FileInfo, api: API, options: Options) {
+  const j = api.jscodeshift;
+
+  const root = j(file.source);
+
+  const {containsCanvasImports, importMap, styledMap} = getImportRenameMap(
+    j,
+    root,
+    '@workday/canvas-kit-react/card'
+  );
+
+  if (!containsCanvasImports) {
+    return file.source;
+  }
+
+  root
+    .find(
+      j.JSXElement,
+      (value: JSXElement) =>
+        value.openingElement.name.type === 'JSXIdentifier' &&
+        (value.openingElement.name.name === importMap.Card ||
+          value.openingElement.name.name === styledMap.Card)
+    )
+    .forEach(nodePath => {
+      // bail early if we find a `Card.Body` as a child
+      for (const child of nodePath.value.children || []) {
+        if (
+          child.type === 'JSXElement' &&
+          child.openingElement.name.type === 'JSXMemberExpression' &&
+          child.openingElement.name.object.type === 'JSXIdentifier' &&
+          child.openingElement.name.object.name === 'Card' &&
+          child.openingElement.name.property.type === 'JSXIdentifier' &&
+          child.openingElement.name.property.name === 'Body'
+        ) {
+          return;
+        }
+      }
+
+      const findAttribute = (name: string) => (item: JSXAttribute | JSXSpreadAttribute) => {
+        if (
+          item.type === 'JSXAttribute' &&
+          item.name.type === 'JSXIdentifier' &&
+          item.name.name === name
+        ) {
+          return true;
+        }
+        return false;
+      };
+      const attributes = nodePath.value.openingElement.attributes;
+
+      if (attributes) {
+        const header = attributes.find(findAttribute('heading')) as JSXAttribute | undefined;
+        const id = attributes.find(findAttribute('headingId')) as JSXAttribute | undefined;
+        const children = nodePath.value.children;
+
+        // remove these attributes from the list - we'll add them as sub components
+        nodePath.value.openingElement.attributes = attributes.filter(item => {
+          if (
+            item.type === 'JSXAttribute' &&
+            item.name.type === 'JSXIdentifier' &&
+            ['heading', 'headingId'].includes(item.name.name)
+          ) {
+            return false;
+          }
+          return true;
+        });
+
+        const CardHeaderJSX = j.jsxMemberExpression(
+          j.jsxIdentifier(importMap.Card),
+          j.jsxIdentifier('Header')
+        );
+
+        const CardBodyJSX = j.jsxMemberExpression(
+          j.jsxIdentifier(importMap.Card),
+          j.jsxIdentifier('Body')
+        );
+
+        const convertAttributeToChildren = (input: JSXAttribute) => {
+          switch (input.value?.type) {
+            case 'JSXExpressionContainer':
+              // unwrap JSXElements
+              if (input.value.expression.type === 'JSXElement') {
+                return [input.value.expression];
+              }
+              return [input.value];
+            default:
+              return [input.value!];
+          }
+        };
+
+        // Add header back as children
+        nodePath.value.children = [
+          !!header
+            ? j.jsxElement(
+                j.jsxOpeningElement(
+                  CardHeaderJSX,
+                  id ? [j.jsxAttribute(j.jsxIdentifier('id'), id.value)] : []
+                ),
+                j.jsxClosingElement(CardHeaderJSX),
+                convertAttributeToChildren(header)
+              )
+            : undefined,
+          j.jsxElement(
+            j.jsxOpeningElement(CardBodyJSX),
+            j.jsxClosingElement(CardBodyJSX),
+            children
+          ),
+        ].filter<JSXElement>(filterDefined);
+      }
+    });
+
+  return root.toSource();
+}
--- a/modules/codemod/lib/v5/compoundCard.ts
+++ b/modules/codemod/lib/v5/compoundCard.ts
@@ -73,9 +73,9 @@ export default function transformer(file: FileInfo, api: API, options: Options) 
           return true;
         });
 
-        const CardHeaderJSX = j.jsxMemberExpression(
+        const CardHeadingJSX = j.jsxMemberExpression(
           j.jsxIdentifier(importMap.Card),
-          j.jsxIdentifier('Header')
+          j.jsxIdentifier('Heading')
         );
 
         const CardBodyJSX = j.jsxMemberExpression(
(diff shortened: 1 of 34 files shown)


modules/codemod/lib/v5/compoundCard.ts
```diff
@@ -0,0 +1,121 @@
...
+    return file.source;
+  }
+
+  root
+    .find(
+      j.JSXElement,
+      (value: JSXElement) =>
+        value.openingElement.name.type === 'JSXIdentifier' &&
+        (value.openingElement.name.name === importMap.Card ||
+          value.openingElement.name.name === styledMap.Card)
```
NicholasBoll, line 30: This will match `Card` imports, renamed `Card` imports, and styled `Card`.

modules/codemod/lib/v5/compoundCard.ts
```diff
@@ -0,0 +1,121 @@
...
+          item.name.name === name
+        ) {
+          return true;
+        }
+        return false;
+      };
+      const attributes = nodePath.value.openingElement.attributes;
+
+      if (attributes) {
+        const header = attributes.find(findAttribute('heading')) as JSXAttribute | undefined;
```
NicholasBoll, line 60: Returned types are `JSXAttribute | JSXSpreadAttribute`, but `find` will never return a `JSXSpreadAttribute`

modules/react/card/lib/Card.tsx
```diff
@@ -60,28 +58,21 @@ const Box = styled('div', {
...
+  Component: (
+    {depth = depthValues[2], children, padding = space.l, ...elemProps}: CardProps,
+    ref,
+    Element
+  ) => {
     return (
-      <Box {...elemProps} depth={depth} padding={padding}>
-        {heading && <Header id={headingId}>{heading}</Header>}
-        <Body>{this.props.children}</Body>
+      <Box ref={ref} as={Element} depth={depth} padding={padding} {...elemProps}>
```
anicholls, line 69: Nit: is it worth renaming this due to the upcoming release of `Box` from @alanbsmith?

modules/react/card/stories/card.stories.mdx
```diff
@@ -9,7 +9,9 @@ import {Depth} from './examples/Depth';
 
 # Canvas Kit Card
 
-Generic Canvas Card component
+Generic Canvas Card component. The card is a presentational
+[compound component](../../../../modules/docs/mdx/COMPOUND_COMPONENTS.mdx) is a building block for
+other components, like Popups and Modals.
```
anicholls, line 14:
```suggestion
Generic Canvas Card component. The card is a presentational
[compound component](../../../../modules/docs/mdx/COMPOUND_COMPONENTS.mdx) used as a building block for
other components, like Popups and Modals.
```

modules/react/card/stories/card.stories.mdx
```diff
@@ -19,21 +21,74 @@ Generic Canvas Card component
...
 
 <ArgsTable of={Card} />
+
+### Card.Header
+
+The `Card.Header` component is an optional sub-component that is meant to describe the card. Since
+the `Card` is a non-semantic presentational component, the `Card.Header` does not automatically have
+any semantic meaning. If your use-case requires the header to label the card, you must do so
+manually. For example, the `Modal` component uses a `Card` and adds a `aria-labelledby` and a `role`
+to the `Card` contaienr component and adds an `id` to the `Card.Header` component:
```
anicholls, line 64:
```suggestion
manually. For example, the `Modal` component uses a `Card`. It adds an `aria-labelledby` and a `role`
to the `Card` container component, then adds an `id` to the `Card.Header` component:
```

modules/react/card/stories/examples/Depth.tsx
```diff
@@ -4,8 +4,9 @@ import {depth} from '@workday/canvas-kit-react/tokens';
 
 export const Depth = () => {
   return (
-    <Card heading="Title" depth={depth.inset}>
-      No Padding
+    <Card depth={depth.inset}>
+      <Card.Header>Title</Card.Header>
+      <Card.Body>No Padding</Card.Body>
```
anicholls, line 9: I know this will cause a visual change, but we generally never use inset depth on cards so it's weird to show as an example
```suggestion
    <Card depth={depth[1]}>
      <Card.Header>Title</Card.Header>
      <Card.Body>Less depth</Card.Body>
```

modules/react/card/lib/Card.tsx
```diff
@@ -60,28 +58,21 @@ const Box = styled('div', {
...
+        {children}
       </Box>
     );
-  }
-}
+  },
+  subComponents: {
+    Header: CardHeader,
```
anicholls, line 75: Any reason why you chose `CardHeader` instead of `CardHeading`? Technically, `heading` is the correct term and it's what was used for the prop before.

modules/react/tooltip/stories/stories_testing.tsx
```diff
@@ -4,8 +4,15 @@ import {TooltipContainer, Tooltip} from '@workday/canvas-kit-react/tooltip';
 import {Card} from '@workday/canvas-kit-react/card';
 import {withSnapshotsEnabled} from '../../../../utils/storybook';
 
+const fontDelay = 150; // best guess for the font delay to prevent incorrect Chromatic regressions
```
anicholls, line 7: Lol - my sanity thanks you 👏

modules/codemod/lib/v5/spec/compoundCard.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    expectTransform(input, expected);
+  });
+
+  it('should restructure Card usages with no props', () => {
+    const input = stripIndent`
+      import {Card} from '@workday/canvas-kit-react/card';
+
+      <Card>
+        No Padding
```
anicholls, line 29: Nit: this text doesn't really make sense in this context. It's used in all the tests.

anicholls:
Also need to update the migration guide

modules/react/card/lib/Card.tsx
NicholasBoll: I figured this would be refactored to use Box when it comes out.

modules/codemod/lib/v5/spec/compoundCard.spec.ts
NicholasBoll: I prefer explicit testing vs implicit testing. If this test fails, it is explicitly for the lack of props where the other tests have additional failure points

modules/react/card/lib/Card.tsx
NicholasBoll: Good point. I like "Heading" better. I had issues with using both "Heading" and "Header", but "Heading" is more accurate. I'll update.

NicholasBoll:
> Also need to update the migration guide

Do you mean this? https://github.com/Workday/canvas-kit/pull/1028/files#diff-fb39a339ed5fc8e471331456e72c6f4b6d40332d4d129a7659b6a1b0c1ca445fR251

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -247,3 +247,61 @@ PR:
...
+</Card>
+
+// after
+<Card>
+  <Card.Heading id="header-id">Card Title</Card.Heading>
+  <Card.Body>Card Body</Card.Body>
+</Card>
+```
+
+The codemod will do it's best to rewrite the JSX to match the new API. It will work if you rename
```
alanbsmith, line 269:
```suggestion
The codemod will do its best to rewrite the JSX to match the new API. It will work if you rename
```

modules/react/card/lib/Card.tsx
alanbsmith: Yeah, agree. Eventually Card will be built with Box. This is fine for now.

modules/codemod/lib/v5/spec/compoundCard.spec.ts
alanbsmith: I think I see what you're saying, @NicholasBoll, but doesn't the default Card have padding though?
https://github.com/Workday/canvas-kit/pull/1028/files#diff-c9e6ab3d13769a201aedc2ac98abcf34e296920d943bc799a372f392f8323174R64

alanbsmith: Maybe the text should read "No props" instead?

modules/codemod/lib/v5/removeButtonEnums.ts
```diff
@@ -14,6 +14,7 @@ const importsToRemove = [
   'ButtonVariant',
   'ButtonIconPosition',
   'ButtonSize',
+  'ButtonIconSize',
```
alanbsmith, line 17: Just double-checking the remove button enums are an intentional addition to the PR. I didn't see anything about this in the PR description.

alanbsmith (approved):
Looks good. Thanks for working on this! I left a couple comments and made a small suggestion for a typo update, but nothing significant enough to hold this PR up, imo.

modules/codemod/lib/v5/removeButtonEnums.ts
NicholasBoll: It is an enum that was supposed to be removed, but missed. Thanks for calling this out

modules/codemod/lib/v5/spec/compoundCard.spec.ts
anicholls: Yes, I was just referring to the text inside the card, not the test itself

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -247,3 +247,61 @@ PR:
       onClick?: React.MouseEventHandler<HTMLButtonElement>;
     }
     ```
+
+### Card
```
anicholls, line 251: Not sure how I missed this haha. Still needs to be added to the table of contents though

NicholasBoll: I was working on it during your review and pushed it right before your comment about it missing.

PR #1034, opened 2021-04-20
refactor: Button recategorization

### LINKED ISSUE #987: Buttons for v5
### Overview

Implement breaking changes for v5 to improve developer ergs and ensure the component is cohesive with new API approaches.

**Rename buttons based on new taxonomy / structure:**
- `Button` > `SecondaryButton`
- Primary `Button` > `PrimaryButton`
- `HighlightButton` > `SecondaryButton`
- `OutlineButton` > `SecondaryButton`
- Primary `OutlineButton` > `SecondaryButton` (include note to users that if they want the primary color back, they should use `PrimaryButton`)
- `TextButton` > `TertiaryButton`
- `DropdownButton` > `SecondaryButton` or `PrimaryButton` with correct props

Additional: 
- Add `iconPosition` to all buttons

## Related Issues
#970 Convert Button family to use `createComponnt` helper

anicholls:
## Summary

There has been common confusion around the large number of buttons Canvas supports and when each
should be used. To improve the usability of our design system, we have been working to recategorize
and simplify our button offering. To align with the changes recently made in our Figma libraries, we
have reorganized our buttons, renaming a few and removing some others. The majority of button use
cases have been simplified into three different components: `PrimaryButton`, `SecondaryButton`, and
`TertiaryButton`, each level representing its emphasis and hierarchy in a UI. We hope this makes
your usage of our buttons more intentional and clear. We have provided a codemod to make all of
these changes automatic updates.

Closes #987 

**Renamed:**

- 🤖 `Button` has been split into `PrimaryButton` and `SecondaryButton` (depending on the `variant`
  prop).
- 🤖 `OutlineButton` (`secondary`) is now `SecondaryButton`.
- 🤖 `OutlineButton` (`inverse`) is now `SecondaryButton` with an `inverse` variant.
- 🤖 `TextButton` is now `TertiaryButton`.

**Removed:**

- 🤖 `HighlightButton`. Use `SecondaryButton` instead.
- 🤖 `DropdownButton`. This can be achieved simply using `PrimaryButton` or `SecondaryButton` with
  an `icon` prop and `iconPosition="right"`.

## Note

**This change is going to break the [existing design site](https://design.workday.com/components/buttons/buttons) when we merge to master since the links to our button stories are hardcoded there**


modules/codemod/lib/v5/recategorizeButtons.ts
```diff
@@ -0,0 +1,226 @@
...
+   * 2. Find `Button`
+   *  - If it has variant="primary"
+   *    - Swap to `PrimaryButton` and remove variant prop
+   *    - Add `PrimaryButton` import if it doesn't exist.
+   *  - If not
+   *    - Swap to `SecondaryButton` and remove variant prop
+   *    - Add `SecondaryButton` import if it doesn't exist
+   */
+  root
+    .find(j.JSXElement, {openingElement: {name: {name: 'Button'}}})
```
anicholls, line 43: Note: These changes are based purely on the JSX element (therefore not informed by the import). This means there's a possibility of affecting the incorrect components. For example:

```
import {MyButton as Button} from 'my-project';
import {Button as CKRButton} from '@workday/canvas-kit-react';
```

It caught me on our `import {deprecated_Button as Button}` usage in the deprecated button stories.

It's going to add quite a bit of complexity to try to address this case. Do we think it's worth doing?

NicholasBoll: There's a `getImportRenameMap` that both checks for Canvas-base imports and what the mappings are. You can do the following to check if the element was an imported `Button` from Canvas Kit:

```tsx

const {containsCanvasImports, importMap} = getImportRenameMap(
  j,
  root,
  '@workday/canvas-kit-react/button'
);

if (!containsCanvasImports) {
  return file.source
}

root
  .find(j.JSXElement, (node: JSXElement) => {
    // the following line might need additional `.type` checks
    if (node.openingElement.name.name === importMap.Button) {
      return true
    }
    return false
  })
```

If you want to even check for instances of `styled(Button)`, there's also a `styledMap` that detects styled renames, but only from CK imports.

Both are available from my PR: https://github.com/Workday/canvas-kit/pull/1028/files#diff-dc51f739fd52e024d8c2d94f0d87c81d9416fcd78c796376705c7af12fdaffa9R9

anicholls: @NicholasBoll After looking into it, it's still pretty complicated with that helper. It basically doubles the amount of logic in the codemod because I have to make different changes to both the element and the import depending on whether they use a namespaced import or not.

I can add it, but it's just a lot of complexity for something I'm not sure how many people are using.

NicholasBoll: With the changes made in the Card PR, it isn't much extra code. Basically what I pasted minus the extra type checks. After the Card PR, all maps are any import names. `import.Button` will equal whatever the import name is. Default is `Button`. If they renamed it, it would be whatever the renamed value is. That was the reason for the update.

NicholasBoll:
Overall question about this change: Why components over variants? `<PrimaryButton />` vs `<Button variant="primary">`. We're moving away from the enforcement game (the reason `DeleteButton` was created), so curious about this change.

NicholasBoll:
Also, I thought Outline was becoming Secondary due to the low-contrast background of Secondary.

anicholls:
Note to self: move system icon dependency to devDeps

anicholls:
There's been some questions/discussion around the differences between the work in this PR for v5 and the ideal state that we'll have coming in v6. I'm hoping this comment clarifies some of that and serves as a reference for future discussions.

## Design & Variants

In v5, we are only making organizational/taxonomical changes so that Canvas Kit matches the conventions used in our Figma library. This means that the buttons boil down to `PrimaryButton`, `SecondaryButton` and `TertiaryButton`. We are currently working on some visual updates to our buttons for improved accessibility, but that is still in progress so it is being held for v6. The primary focus of these visual updates is replacing the "filled" style of the Secondary button with an "outline" style (currently two different types of buttons in v4). This is because the "filled" style does not pass accessibility requirements.

For v5, We were planning to deprecate everything except for the Primary, Secondary, Tertiary and Delete buttons (including the existing Outline button). However, due to the fact that we're not making any visual changes, the Secondary button is still the "filled" style. We are hesitant to remove the more accessible "outline" style button, so we've included it as a variant on the Secondary button (e.g. `<SecondaryButton variant="outline" />`. **We considered replacing the "filled" secondary button with the "outline" one, but our UI-Platform team needs to be given more of a heads up before we do this. However, there has been some confusion around the continued inclusion of both the "outline" and "filled" style secondary buttons - so maybe this is a discussion we should revisit and we should push for this replacement to happen in v5**.

## API

There have been questions about whether moving away from a single `Button` component with multiple variants is the right approach. In this PR, this has been separated into the `PrimaryButton`, `SecondaryButton`, and `TertiaryButton` components. Generally we do this only if the APIs differ between these components, which in this case they do not. The reason I made this change was twofold: 

- _Technically_ we don't consider Primary, Secondary and Tertiary to be different variants. They are different types of buttons.
- Because the variants ended up feeling a bit random:

  - `<Button variant={*} />`
    - `primary`
    - `secondary`
    - `secondary-outline`
    - `secondary-inverse`
    - `tertiary`
    - `tertiary-inverse`
  
  **vs.**

  - `<PrimaryButton />`
    - no variants
  - `<SecondaryButton variant={*} />`
    - `undefined`
    - `outline`
     - `inverse`
  - `<TertiaryButton variant={*} />`
    - `undefined`
    - `inverse`

I think the biggest influence on this decision should be what we expect the ultimate API to look like once the visual changes are in place in v6:

- Primary
  - default
  - inverse
- Secondary
  - default
  - inverse
- Tertiary
  - default
  - inverse

Since we don't consider the different types as variants, I would recommend sticking with `PrimaryButton`, `SecondaryButton` and `TertiaryButton` with either a `variant: 'inverse' | undefined` prop or an `inverse: boolean` prop. The only technical downside in going with this approach is if a developer needs to change a button from primary to secondary, it will re-render the entire button causing it to lose focus. I do not feel this is something we need to worry about.

---

Anyway, open to suggestions/discussion on all of this - you know where to find me 🙂

anicholls:
After discussion, it was agreed that we would leave the existing deprecated secondary as is, and just remove the existing filled secondary button in favor of the new outline styling.

NicholasBoll (approved):

PR #1038, opened 2021-04-22
fix: Fix slash import paths

### LINKED ISSUE #1030: Broken paths in version 5
## 🐛 Bug Report

In CK v5 Beta 1 we moved to "slash imports" to allow the CK main packages to be under a single version.

[Slash Import PR: #992](https://github.com/Workday/canvas-kit/pull/992)

The problem is the new paths don't actually exist when packaged.

## To Reproduce

Install Beta 1:
```
npm i @workday/canvas-kit-react@5.0.0-beta.1
```

Set up a TSX file with the following:
```tsx
import { Button } from "@workday/canvas-kit-react/button";
```

## Expected Behavior

The path should be valid.

## Actual Results

An error occurs saying the module could not be found

### Error Output

image

The following path does work, however:

```tsx
import { Button } from "@workday/canvas-kit-react/dist/es6/button";
```

We will have to fix the way the packages are published to support the valid path:

### Extra info

We currently support 2 targets: `es6` and `commonjs`. Bundlers typically support both and will prefer the `es6` path because of the `module` property in the `package.json` file.

Slash import paths will not work with more than one target since the path needs to be exposed as-is in the package that is published to npm. A default target must be chosen and additional targets can be supported via a nested path. For example, Material-UI v5 supports ES modules by default and commonjs via a `node` path:

```ts
// v5, this is a ES file
import {Button} from '@material-ui/core/Button'

// v5, this is the CommonJS file
import {Button} from '@material-ui/core/node/Button'
```

### Questions
- What should the default module target be? `es6` or `commonjs`?
  - Webpack supports both and prefers ES modules (`module` will be chosen over `main`, although if there's a path in the import, neither will be used)
  - ES module syntax is supported via specification, works in modern browsers and node 13+ (14 is the current LTS)

NicholasBoll:
- Fixes #1030
- Add source code back to packages for type source maps
- fix `spec/tsconfig.json` file in the `create-component` script

## Summary

This change adds `package.json` files to all the paths inside `react` and `labs-react` packages during the `prepack` phase and removes them again during the `postpack`. This allows slash imports to work correctly again. This change also adds back the source files for correct TypeScript types source maps (go to definition goes to source files rather than type definition files).

### DIFF
--- a/modules/labs-react/breadcrumbs/package.json
+++ b/modules/labs-react/breadcrumbs/package.json
@@ -1,5 +1,0 @@
-{
-  "main": "../dist/commonjs/breadcrumbs",
-  "module": "../dist/es6/breadcrumbs",
-  "sideEffects": false
-}
--- a/modules/labs-react/breadcrumbs/package.json
+++ b/modules/labs-react/breadcrumbs/package.json
@@ -1,0 +1,5 @@
+{
+  "main": "../dist/commonjs/breadcrumbs",
+  "module": "../dist/es6/breadcrumbs",
+  "sideEffects": false
+}
--- a/modules/labs-react/package.json
+++ b/modules/labs-react/package.json
@@ -36,6 +36,8 @@
     "build:rebuild": "npm-run-all clean build",
     "build:downlevel-dts": "yarn run downlevel-dts dist ts3.5/dist",
     "build": "npm-run-all --parallel build:cjs build:es6 --sequential build:downlevel-dts",
+    "prepack": "node ../../utils/publish.js pre labs-react",
+    "postpack": "node ../../utils/publish.js post labs-react",
     "depcheck": "node ../../utils/check-dependencies-exist.js"
   },
   "keywords": [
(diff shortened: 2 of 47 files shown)


modules/labs-react/package.json
```diff
@@ -13,6 +13,9 @@
     "url": "http://github.com/Workday/canvas-kit/tree/master/modules/canvas-kit-labs-react"
   },
   "files": [
+    "*/package.json",
+    "*/lib/*",
+    "*/index.ts",
```
NicholasBoll, line 18: Add source code support for "Go to definition"

modules/labs-react/breadcrumbs/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "main": "../dist/commonjs/breadcrumbs",
+  "module": "../dist/es6/breadcrumbs",
+  "sideEffects": false
```
alanbsmith, line 4: TIL: `"sideEffects": false`

alanbsmith (approved):
Looks good! Thanks for the quick fix!

anicholls (approved):

PR #1039, opened 2021-04-23
chore: Remove package.json from create component script

NicholasBoll:
This was accidentally left in from when the `package.json` was in source control.


alanbsmith (approved):
🐅

PR #1041, opened 2021-04-26
feat(useThemeRTL): Add hook to make RTL easier

vibdev:
## Summary

As discussed here:

https://github.com/Workday/canvas-kit/discussions/1012

This is a way to automatically transform CSS, applied using either Emotion's css attribute or by a
style attribute, when there is a RTL theme in use. Instead of authoring two sets of CSS objects, one
for each language direction. Now you can author the LTR version and this hook will automatically
create the RTL counterpart for you on the fly. This works for both object and string template
Emotion methods.

For convenience we also export a ComponentStyles type if you want to organize your styles all
together.

## Checklist

- [x] branch has been rebased on the latest pre-release commit
- [x] module has been added to `labs` exports


modules/labs-react/common/lib/theming/useThemeRTL.ts
```diff
@@ -0,0 +1,20 @@
+import {SerializedStyles} from '@emotion/core';
+import {ContentDirection, EmotionCanvasTheme, useTheme} from '@workday/canvas-kit-react/common';
+import {CSSProperties} from '@workday/canvas-kit-react/tokens';
+import rtlCSSJS from 'rtl-css-js';
+
+type TemplateOrObjectStyles = CSSProperties | SerializedStyles;
```
vibdev, line 6: CSSProperties for object notation, SerializedStyles for string template

modules/labs-react/common/lib/theming/useThemeRTL.ts
```diff
@@ -0,0 +1,20 @@
...
+const getThemedRTL = (theme: EmotionCanvasTheme, ...styles: TemplateOrObjectStyles[]) => {
+  return theme.canvas.direction === ContentDirection.RTL ? rtlCSSJS(styles) : styles;
+};
+
+export function useThemeRTL(...cssObject: TemplateOrObjectStyles[]) {
+  const theme = useTheme();
+
+  const styles = getThemedRTL(theme, ...cssObject);
+
+  return styles.reduce((first, second) => ({...first, ...second}), {}) as CSSProperties;
```
vibdev, line 19: need to flatten out the object so it works with style attributes

modules/labs-react/common/lib/theming/useThemeRTL.ts
```diff
@@ -0,0 +1,23 @@
...
+  const theme = useTheme();
+
+  const themeRTL = useMemo(() => {
+    return (...cssObject: CSSProperties[]) => {
+      const styles = getThemedRTL(theme, ...cssObject);
+      return styles.reduce((first, second) => ({...first, ...second}), {}) as CSSProperties;
+    };
+  }, [theme]);
+
+  return {themeRTL, theme};
```
alanbsmith, line 22: Why does this also return the `theme` here? Is that for convenience?

modules/labs-react/common/lib/theming/useThemeRTL.ts
```diff
@@ -0,0 +1,23 @@
...
+import {CSSProperties} from '@workday/canvas-kit-react/tokens';
+import {useMemo} from 'react';
+import rtlCSSJS from 'rtl-css-js';
+
+export type ComponentStyles = Record<string, CSSProperties>;
+
+const getThemedRTL = (theme: EmotionCanvasTheme, ...styles: CSSProperties[]) => {
+  return theme.canvas.direction === ContentDirection.RTL ? rtlCSSJS(styles) : styles;
+};
+
```
alanbsmith, line 11:
```suggestion

/**
 * A helpful hook for supporting bidirectional styles. 
 * * Read below for more detail or [view the docs](https://github.com/Workday/canvas-kit/blob/master/modules/labs-react/common/README.md#useThemeRTL).
 * 
 * @returns
 * * `themeRTL` - a function to transform bidirectional styles
 * * `theme` - the Canvas theme object (optional, provided for convenience)
 *
 * `themeRTL` allows you to support bidirectionality with a single set of styles and pass them along to a component.
 * It accepts CSS object styles and transforms CSS attributes based on the content direction set in the theme (LTR or RTL).
 *
 * @example
 * import { type } from '@workday/canvas-kit-react/tokens';
 * import { useThemeRTL } from '@workday/canvas-kit-labs-react/common';
 *
 * const ErrorText: React.FC = (props) => {
 *   const { themeRTL, theme } = useThemeRTL();
 *   // `borderLeft` will be converted to `borderRight`
 *   // when the theme direction is `RTL`.
 *   // All other styles will remain unchanged.
 *   const errorTextStyles = themeRTL({
 *     ...type.levels.subtext.medium,
 *     color: theme.canvas.palette.error.main,
 *     borderLeft: `solid 2px ${theme.canvas.palette.error.main}`,
 *   });
 *
 *   return <span css={errorTextStyles} {...props} />;
 * }
 */
```

alanbsmith: Adding JSDoc comments ^^

modules/labs-react/common/lib/theming/useThemeRTL.ts
```diff
@@ -0,0 +1,23 @@
+import {ContentDirection, EmotionCanvasTheme, useTheme} from '@workday/canvas-kit-react/common';
```
alanbsmith, line 1:
```suggestion
import {EmotionCanvasTheme, useIsRTL, useTheme} from '@workday/canvas-kit-react/common';
```
I'd like to reuse `useIsRTL` in `getThemedRTL` below.

modules/labs-react/common/lib/theming/useThemeRTL.ts
```diff
@@ -0,0 +1,23 @@
...
+import {CSSProperties} from '@workday/canvas-kit-react/tokens';
+import {useMemo} from 'react';
+import rtlCSSJS from 'rtl-css-js';
+
+export type ComponentStyles = Record<string, CSSProperties>;
+
+const getThemedRTL = (theme: EmotionCanvasTheme, ...styles: CSSProperties[]) => {
+  return theme.canvas.direction === ContentDirection.RTL ? rtlCSSJS(styles) : styles;
```
alanbsmith, line 9:
```suggestion
  return useIsRTL(theme) ? rtlCSSJS(styles) : styles;
```

modules/react/status-indicator/lib/StatusIndicator.tsx
```diff
@@ -21,10 +20,10 @@ export enum StatusIndicatorEmphasis {
 }
 
 export interface StatusIndicatorGenericStyle extends GenericStyle {
-  styles: CSSObject & React.CSSProperties;
+  styles: CSSProperties;
   variants: {
     [statusType in StatusIndicatorType]: {
-      [statusEmphasis in StatusIndicatorEmphasis]?: CSSObject & React.CSSProperties;
+      [statusEmphasis in StatusIndicatorEmphasis]?: CSSProperties;
```
alanbsmith, line 26: Sorry, I'm not following. Are these changes related to the hook?

modules/labs-react/common/README.md
```diff
@@ -16,3 +16,41 @@ yarn add @workday/canvas-kit-labs-react/common
...
+const styles: ComponentStyles = {
+  panel: {
+    margin: '8px',
+  },
+  top: {
+    top: '8px',
+  },
+  red: {
+    color: 'red',
+  },
```
alanbsmith, line 46: import {space} from '@workday/canvas-kit-react/tokens';

```suggestion
const styles: ComponentStyles = {
  panel: {
     // `themeRTL` will transform this to `marginRight` when the direction is `RTL`
    marginLeft: space.xxs,
  },
  top: {
    top: space.xxs,
  },
  red: {
    color: 'red',
  },
```

This is a good, simple example, but I'd like to make a couple slight alterations:

* use space tokens just to encourage the idea of using token values
* switch `margin` to `marginLeft` as that attr will be updated for RTL

modules/labs-react/common/README.md
```diff
@@ -16,3 +16,41 @@ yarn add @workday/canvas-kit-labs-react/common
...
+};
+export const HelloWorld = () => {
+  const {themeRTL} = useThemeRTL();
+  return (
+    <div css={themeRTL(styles.panel, styles.top)} style={themeRTL(styles.red)}>
```
alanbsmith, line 51:
```suggestion
  const {themeRTL, theme} = useThemeRTL();
  const panelStyles = {
    ...themeRTL(styles.panel),
    backgroundColor: theme.canvas.palette.neutral.main, 
  };
  // You can also pass `panelStyles` to the `style` prop, if preferred.
  return (
    <div css={panelStyles} style={{...styles.top, ...styles.red}}>
```

A couple things here as well:

* show `theme` being provided as well and why you might want to use it so it's not hidden
* passing bidirectional styles to themeRTL and separating out the rest


I'm very open to feedback on this usage though. If it makes more sense to leave it as:

```tsx
<div css={themeRTL(styles.panel, styles.top)} style={themeRTL(styles.red)}>
```

I'm good with that too. 👍🏻

alanbsmith (changes requested):
Looks great, @vibdev! Thanks for contributing! I made a few suggestions, but nothing major. And thank you for the robust test coverage. 🙏🏻

modules/labs-react/common/README.md
vibdev: > passing bidirectional styles to themeRTL and separating out the rest

I might avoid doing this, seems error prone and then you have to think about what needs to be transformed. Like the example of using theme

modules/labs-react/common/lib/theming/useThemeRTL.ts
vibdev: can do will need a small refactor to make hooks happy 👍

modules/react/status-indicator/lib/StatusIndicator.tsx
vibdev: I was looking at what CSSProperties was, for use in my example. Looks like this file missed the conversion whenever y'all created it. Happy to pull this out if you think it is too unrelated.

modules/labs-react/common/lib/theming/useThemeRTL.ts
```diff
@@ -1,23 +1,53 @@
...
+ *     color: theme.canvas.palette.error.main,
+ *     borderLeft: `solid 2px ${theme.canvas.palette.error.main}`,
+ *   });
+ *
+ *   return <span css={errorTextStyles} {...props} />;
+ * }
+ */
 export function useThemeRTL() {
   const theme = useTheme();
+  const direction = useIsRTL(theme);
```
alanbsmith, line 43: 👍🏻

alanbsmith (approved):
Great work! Thanks for the PR! 🎉

PR #1045, opened 2021-04-27
feat(checkbox): Add aria-checked for accessibility and testability

svagi:
## Summary

- Add [`aria-checked`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role) for accessibility and easier testability with [`toBeChecked`](https://github.com/testing-library/jest-dom#tobechecked) / [`toBePartiallyChecked`](https://github.com/testing-library/jest-dom#tobepartiallychecked)

## Checklist

- [x] `yarn test` passes

### DIFF
--- a/modules/checkbox/react/lib/Checkbox.tsx
+++ b/modules/checkbox/react/lib/Checkbox.tsx
@@ -310,6 +310,7 @@ export const Checkbox = ({
           type="checkbox"
           value={value}
           error={error}
+          aria-checked={indeterminate ? 'mixed' : checked}
           {...elemProps}
         />
         <CheckboxRipple />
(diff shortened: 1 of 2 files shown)


mathalete:
Hey Jan,

Placing `aria-checked` on a native `<input type="checkbox"` unfortunately isn't exposed the accessibility api in Safari and Internet Explorer 11.

This is not a bug in Canvas Kit, but more that Webkit and IE11 are not sending this information to the accessibility tree to then be announced by a screen reader.

Test case located here >
https://cdpn.io/lefthandev/debug/3c609124b725254937a0ae320f96e6e1

A more widely supported approach across the Workday browser and screen reader matrix would be to go against the first rule of aria and to use a `<div>` with a `role="checkbox"` and to use `aria-checked="mixed"` on that to convey the state.

`<div role="checkbox" aria-checked="mixed" aria-label="Apples" tabindex="0" /></div>`

A related problem was highlighted in material design last year https://github.com/mui-org/material-ui/issues/20476

NicholasBoll (approved):
I think we can merge for now and add a more accessible checkbox later.

PR #1054, opened 2021-05-05
fix(popup): Allow popups to be owners of each other

csongnguyen:
## Summary

Prior to this PR, if a hidable popup contained a target element that opened a non-hidable popup, clicking on the close button of the non-hidable popup would cause both the hidable popup and non-hidable popup to close. 

To fix this, we check in the contains() method to ensure that popups can be owners of other popups, so when useCloseOnOutsideClick() is activated, any click action upon a popup that is opened from a hidable popup would not be considered an outside click.

## Checklist

- [x] `yarn test` passes
      applicable


cypress/integration/Popup.spec.ts
```diff
@@ -117,6 +117,37 @@ describe('Popup', () => {
...
+
+            cy.findByLabelText(Popup3Title).should(h.popup.beOnTopOfLabelledByText('Popup 2'));
+          });
+
+          context('then the close button in Popup 3 is clicked', () => {
+            beforeEach(() => {
+              cy.findByLabelText(Popup3Title)
+                .parentsUntil('[data-popper-placement]')
+                .find('[data-close]')
+                .click();
```
NicholasBoll, line 139: This selector group is a bit weird since too much non-semantic DOM info must be known. Cypress testing library gives us some good selectors for what we need. What about the following:
```suggestion
              cy.findByRole('dialog', {name: Popup3Title})
                .findByRole('button', {name: 'Close'})
                .click();
```

The other specs should be updated to use `findByRole` as well. `name` wasn't a scoping option at the time these specs were written, but I think it makes it more clear.

NicholasBoll (approved):
Looks good!

PR #1056, opened 2021-05-06
Popup accessible example

csongnguyen:
## Summary
Created an accessible popup story to include for popup examples.

## Checklist

- [ ] `yarn test` passes
      applicable


modules/popup/react/README.md
```diff
@@ -10,7 +10,7 @@ Hooks.
 If you are building your own custom stacked UI components, use the [Popper](#popper) component along
 with our [hooks](#hooks). The `Popper` component and hooks work with the stack management system for
 correct rendering and accessibility behavior. If you cannot use `Popper`, use the
-[usePopupStack](#usepoupstack) hook to properly register and deregister the popup at the correct
```
NicholasBoll, line 13: "poup stack"? Sounds messy

modules/tooltip/react/stories/Tooltip.stories.mdx
```diff
@@ -118,7 +118,7 @@ white-space: nowrap;
 
 **Note**: Text truncation should be avoided if possible. A user should not have to activate a
 tooltip to access important content. If user-generated content is being truncated, the following
-situation might occur which is a bad user experence. Consider the following list:
+situation might occur which is a bad user experience. Consider the following list:
```
NicholasBoll, line 121: I made sure to enable spell check in MDX files. Apparently that was the only file type not covered by my spell checker.

modules/popup/react/stories/stories.tsx
```diff
@@ -79,3 +80,110 @@ export const RTL = () => {
...
+        const elem = getFirstFocusableElement(popupRef.current);
+        if (elem) {
+          elem.focus();
+        }
+      }
+      return () => {
+        if (handlerRef) {
+          handlerRef.focus();
+        }
+      };
```
NicholasBoll, line 137: This isn't exactly what we want. This callback will be called when any dependencies have changed. Since `popupRef` will never change, that leaves `popupOpen` which changes both when the popup is opened and when it closes. It happens to work in our simple example, but what actually happens when the `popupOpen` goes from `false` to `true`:

- `handlerRef` is focused
- A new layout effect callback is set up and `elem.focus` is called.

When the state goes from true to false, a new handler ref captures the currently focused element and the check for open fails and the callback is done. I think what you really want to do is capture the changing state of `popupOpen`:

```suggestion
      if (popupRef.current && popupOpen) {
        const elem = getFirstFocusableElement(popupRef.current);
        if (elem) {
          elem.focus();
        }
      } else if (handlerRef && !popupOpen) {
        handlerRef.focus();
      }
```

NicholasBoll (approved):

PR #1065, opened 2021-05-11
refactor(popup): Convert Popup to a compound component

### LINKED ISSUE #932: Stop Restricting Toast Children to String
## 💡 Suggestion

[Canvas Kit > Popups > Toast](https://workday.github.io/canvas-kit/?path=/story/components-popups-toast-react--success) should allow `React.ReactNode` children as usual, not restrict to string children.

## Motivation

Toasts require children to be a string, presumably to avoid "bad behavior", people putting in images or something crazy. But we discussed (in an [internal thread](https://workday-dev.slack.com/archives/C8X0W5AQH/p1607387329388800)) and decided that it's reasonable to include semantics (more markup) in these Toast children.

## Example

```<Toast>New user <em>Chano Fernandez<em> successfully added.</Toast>```

### LINKED ISSUE #1029: Convert Popup into a Compound Component
## 🚀 Feature Proposal

Convert Popup into a Compound Component. `heading` and `headingId` props can be removed and instead export a `Popup.Header` sub component. Popup is based on Card, which is now a compound component.

## Motivation

Other components that use Popup and need to grant element-level access to the Header will need this. Without this, `Modal` will have to bypass the props and re-create a Header sub component. For example, Modal needs to make the Header an `h1`, but the header in card defaults to an `h3`.

## Example

```tsx
// before
<Popup heading={'Popup Header'} headingId={id}>Popup Body</Popup>

// after
<Popup>
  <Popup.Header id={id} as="h1">Popup Header</Popup.Header>
  <Popup.Body>Popup Body</Popup.Body>
</Popup>
```

NicholasBoll:
Fixes #1029
Fixes #932

## Checklist

- [ ] `yarn test` passes
      applicable


modules/labs-react/combobox/lib/Combobox.tsx
```diff
@@ -421,7 +421,7 @@ const Combobox = ({
           ? getOptionId(componentId, selectedAutocompleteIndex)
           : undefined,
       onClick: handleClick,
-      onChange: handleSearchInputChange,
+      onChange: handleSearchInputChange,y
```
anicholls, line 424:
```suggestion
      onChange: handleSearchInputChange,
```

modules/react/popup/lib/hooks/useCloseOnEscape.ts
```diff
@@ -0,0 +1,52 @@
...
+  console.log('close on escape');
+  const onKeyDown = React.useCallback(
+    (event: KeyboardEvent) => {
+      console.log('useCloseOnEscape', event.key, model.state.stackRef);
+      if (
+        (event.key === 'Esc' || event.key === 'Escape') &&
+        PopupStack.isTopmost(model.state.stackRef.current!)
+      ) {
+        if (model.state.willReturnFocus.current) {
+          console.log('change focus');
```
anicholls, line 23:
```suggestion
```

modules/react/popup/lib/hooks/useCloseOnEscape.ts
```diff
@@ -0,0 +1,52 @@
...
+      model.events,
+    ]
+  );
+
+  // `useLayoutEffect` for automation
+  React.useLayoutEffect(() => {
+    if (!model.state.visible) {
+      return;
+    }
+    console.log('useCloseOnEscape useLayoutEffect');
```
anicholls, line 43:
```suggestion
```

modules/react/popup/lib/hooks/useCloseOnEscape.ts
```diff
@@ -0,0 +1,52 @@
...
+    document.addEventListener('keydown', onKeyDown);
+    return () => {
+      console.log('useCloseOnEscape unlisten');
```
anicholls, line 46:
```suggestion
```

modules/react/popup/lib/hooks/useAlwaysCloseOnOutsideClick.ts
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+
+import {PopupStack} from '@workday/canvas-kit-popup-stack';
+import {PopupModel} from '../usePopupModel';
+
+/**
+ * Registers global listener for all clicks. It will only call the `PopupModel` `hide` event if the
+ * click happened outside the model's `stackRef` element and its children regardless of the position
+ * in the stack. This is useful for Tooltips or hierarchical menus.
```
anicholls, line 9: Would be good to add a note about when to use this vs. `useCloseOnOutsideClick`

modules/react/popup/lib/hooks/useCloseOnOutsideClick.ts
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
+import {PopupStack} from '@workday/canvas-kit-popup-stack';
+import {PopupModel} from '../usePopupModel';
+
+/**
+ * Registers global listener for all clicks. It will only call the PopupModel's `hide` event if the
```
anicholls, line 7: Would be good to add a note about when to use this vs. `useAlwaysCloseOnOutsideClick`

anicholls (commented):
Taking a break for lunch 🥪 . This is what I've got so far haha.

anicholls:
I appreciate that you called this PR a chore 😂

modules/react/popup/stories/stories.tsx
```diff
@@ -15,67 +15,53 @@ import {
 import README from '../README.md';
 
 export default {
-  title: 'Components/Popups/Popup/React',
+  title: 'Components/Popups/Popup/React/Old',
```
anicholls, line 18: Can we just remove these? Again - I don't think we should document the old way of doing it.

modules/react/toast/lib/Toast.tsx
```diff
@@ -34,11 +33,6 @@ export interface ToastProps {
    * The text of the Toast action.
    */
   actionText?: string;
-  /**
-   * The origin from which the Toast will animate.
-   * @default {horizontal: 'center', vertical: 'top'}
-   */
-  transformOrigin?: TransformOrigin;
```
anicholls, line 41: I feel like I'm missing something, but why is the `transformOrigin` prop being removed? This is still supported on `Popup.Card`

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+<Meta title="Components/Popups/Popup/React" component={Popup} />
+
+# Canvas Kit Popups
+
+A "popup" is a classification for a type of stacked UI element that appears "on top" of statically
+positioned content. Tooltips, Modals, Dropdown menus, etc are all examples of "popups". Canvas Kit
+has a "stack manager" system for managing these stacked UIs. Different types of popups have
+different requirements of behavior for UX and accessibility - we can call them behaviors,
+capabilities, or traits. Canvas Kit comes with a number of [behaviors](#hooks) in the form of React
+Hooks.
```
anicholls, line 25: Nit: If the link is to a hooks section, the hooks reference should be included in the link

```suggestion
capabilities, or traits. Canvas Kit comes with a number of [behaviors in the form of React
Hooks](#hooks).
```

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+
+You should use the most semantic component for your use-case before using `Popup` directly, like
+`Modal`, which already has the correct behaviors built-in. If no component already exists, you can
```
anicholls, line 28:
```suggestion
`Modal`, which already has the correct behaviors built-in. If no component already exists that matches your use case, you can
```

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+use `Popup` and use our [hooks](#hooks). The `Popup` component comes with a `Popup.Popper` component
```
anicholls, line 29:
```suggestion
use `Popup` and use our [hooks](#hooks). The `Popup` component comes with a `Popup.Popper` sub-component
```

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+that positions a popup using [PopperJS](https://popper.js.org/) that registers a popup with the
+`PopupStack` automatically and sets the popup model's `placement` property. `Popup.Popper` component
+and hooks work with the stack management system for correct rendering and accessibility behavior. If
+you cannot use `Popup.Popper`, use the [usePopupStack](#usepoupstack) hook to properly register and
+deregister the popup at the correct time. If you cannot use our hooks, consider upgrading your
+component to use Hooks. If you cannot do that, you'll have to look up the `PopupStack` package for
```
anicholls, line 35: > `PopupStack` package

Link?

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-react
+```
+
+## Usage
+
+The `Popup` component is a low-level compound component that is used to build popup UIs that are not
```
anicholls, line 48:
```suggestion
The `Popup` component is a generic catch-all compound component that is used to build popup UIs that are not
```

I don't know if I'd call it "low-level"

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+already covered by Canvas Kit.
+
+### Basic
+
+The Popup has no pre-defined behaviors already registered, therefore the `usePopupModel` must always
```
anicholls, line 53: > The Popup has no pre-defined behaviors already registered

Not really sure what you're trying to say with this.

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+and `useReturnFocus` behaviors. You can read through the [hooks](#hooks) section to learn about all
+the popup behaviors. For accessibility, these behaviors should be included most of the time.
+
+<ExampleCodeBlock code={Simple} />
+
+### Without a model
+
+This example shows how a `Popup.Card` can be used outside a `Popup` component, but accessibility is
+lost. If you use the Popup this way, you'll have to manually add a `aria-labelledby` to the
+`Popup.Card` component and an `id` to the `Popup.Heading` component.
```
anicholls, line 66: When/why would you want/need to do this?

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+an intermediate way to describe Popups with a more incremental upgrade path from version 4.
+
+<ExampleCodeBlock code={ManualControl} />
+
+### Initial focus
+
+If you want focus to move to a specific element, set the `initialFocusRef` of the model. Check with
+accessibility before doing this. The following example sets the focus on the "OK" button with an
+`aria-describedby` pointing to the model's `id` state so screen readers properly announce the
+message of the popup when focus is changed to the button.
```
anicholls, line 88:
```suggestion
message of the popup when focus is changed to the button. By default, focus will be placed on the close button when the modal is opened.
```

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+the model visibility set to `true`. This may be required for legacy uses or where the entire `Popup`
+component API is undesired. This example has the same functionality as the [Basic](#basic) example,
+but breaks up control of a Popup in a way that is common to the way Popups are done in v4. This is
+an intermediate way to describe Popups with a more incremental upgrade path from version 4.
+
+<ExampleCodeBlock code={ManualControl} />
+
+### Initial focus
+
+If you want focus to move to a specific element, set the `initialFocusRef` of the model. Check with
```
anicholls, line 85:
```suggestion
If you want focus to move to a specific element when the modal is opened, set the `initialFocusRef` of the model. Check with
```

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+<Popup.Target>Show Popup</PopupTarget>
+
+// using props instead
+const popupTargetButtonProps = usePopupTargetButton(model);
+<Button {...popupTargetButtonProps}>Show Popup</Button>
+```
+
+## Popup.Popper
+
+A `Popup.Popper` is a [Popper](#popper) component that is hooked up to the `PopupModel`
```
anicholls, line 154: There's no `#popper` heading

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+
+**Note:** If you're using `Popper`, you do not need to use this hook directly.
+
+This hook will add the `stackRef` element to the Popup stack on mount and remove on unmount. If you
+use `Popper`, the popper `stackRef` is automatically added/removed from the Popup stack. The Popup
+stack is required for proper z-index values to ensure Popups are rendered correct. It is also
+required for global listeners like click outside or escape key closing a popup. Without the Popup
+stack, all popups will close rather than only the topmost one.
+
+If `forwardRef` is provided, it will be the same as `stackRef`. If `forwardRef` is not provided`,
```
anicholls, line 190:
```suggestion
If `forwardRef` is provided, it will be the same as `stackRef`. If `forwardRef` is not provided,
```

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,318 @@
...
+  </>
+);
+```
+
+### usePopup
+
+Convenience hook for common Popups used as non-modal dialogs. It provides props to mix into
+composite parts of the Popup pattern.
+
+> Note: This is **deprecated**. Please upgrade to using `usePopupModel` and the `Popup` compound
```
anicholls, line 220: I'm still confused about `usePopup` and `usePopupOld` if they're both deprecated haha. I still say let's just drop this. People can always use CK v5 with `@worday/canvas-kit-react-popup@4.x` if they need the old stuff temporarily.

modules/react/modal/stories/stories.tsx
```diff
@@ -13,7 +13,7 @@ import README from '../README.md';
 import {controlComponent} from '../../../../utils/storybook';
 
 export default {
-  title: 'Components/Popups/Modal/React',
+  title: 'Components/Popups/ModalOld/React',
```
anicholls, line 16: Let's drop the old stuff.

modules/react/modal/lib/ModalContent.tsx
```diff
@@ -169,27 +137,22 @@ const useWindowSize = (): {width: number; height: number} => {
...
-      elem.focus();
-    }
+const useReturnFocus = (model: PopupModel) => {
+  React.useEffect(() => {
+    const element = (model.state.returnFocusRef || model.state.targetRef).current;
     return () => {
-      if (handlerRef) {
-        handlerRef.focus();
-      }
+      console.log('useReturnFocus', element);
```
anicholls, line 144:
```suggestion
```

anicholls (commented):
Looks like modal docs and migration guide are still on their way. Haven't had a chance to look at the codemod, but here's another round of feedback!

modules/react/popup/lib/Popper.tsx
```diff
@@ -50,6 +50,12 @@ export interface PopperProps extends React.HTMLAttributes<HTMLDivElement> {
    * @default bottom
    */
   placement?: Placement;
+  /**
+   * A callback function that will be called whenever PopperJS chooses a placement that is different
+   * from the provided `placement` preference. If a `placement` preference doesn't fit, PopperJS
+   * will choose a new one and call this callback.
+   */
+  updatePlacement?: (placement: Placement) => void;
```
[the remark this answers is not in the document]
NicholasBoll: I like that better

modules/react/popup/lib/Popup.tsx
```diff
@@ -136,15 +147,12 @@ const CloseIconContainer = styled('div')<Pick<PopupProps, 'closeIconSize'>>(
...
-export default class Popup extends React.Component<PopupProps> {
-  static Padding = PopupPadding;
-
-  private id = uuid();
-  private closeButtonRef = React.createRef<any>();
-
-  public render() {
-    const {
-      padding = Popup.Padding.l,
+export const PopupOld = createComponent('div')({
```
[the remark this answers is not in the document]
NicholasBoll: The event is `updatePlacement`, but the callback should be `onPlacementChange` for this component

modules/react/popup/lib/Popup.tsx
```diff
@@ -184,19 +198,60 @@ export default class Popup extends React.Component<PopupProps> {
...
 }
 
+export const Popup = createComponent('div')({
+  displayName: 'Popup',
+  Component: ({children, model, ...config}: PopupProps) => {
+    const value = useDefaultModel(model, config, usePopupModel);
+    return <PopupModelContext.Provider value={value}>{children}</PopupModelContext.Provider>;
+  },
+  subComponents: {
+    Padding: PopupPadding,
```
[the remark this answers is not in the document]
NicholasBoll: 👍 

I was waiting for the Box component updates and use `padding` from that.

modules/react/popup/lib/PopupCloseIcon.tsx
```diff
@@ -0,0 +1,65 @@
...
+      requestAnimationFrame(() => {
+        if (state.willReturnFocus?.current) {
+          changeFocus((state.returnFocusRef || state.targetRef).current);
+        }
+      });
+    },
+  };
+});
+
+export const PopupCloseIcon = createComponent('button')({
```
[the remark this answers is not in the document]
NicholasBoll: We use element strings as the default when the component handles its own styling. We use components as the default when the component doesn't handle styling (behavior-only components).

I guess this is kind of middle ground. It does seem like it should be `IconButton`, but the component does add absolution positioning to get the icon in the right spot on the Popup. Before this was handled by a `div` element that did the positioning. I didn't want to do the `div` route because it meant nobody could change the location if they really needed to.

I guess the question is what would people expect to happen why they pass an `as`? Would they want to change the rendered element or the rendered component? `IconButton` still works perfectly fine since we're really rendering a `StyledIconButton` and the `as` is passed down to `styled`. `StyledIconButton` is still a `styled(IconButton)`, so changing the rendered component doesn't change how it looks. For example, `<Popup.CloseIcon as={DeleteButton}>` still renders as an `IconButton`. Emotion will merge the styles together and the `IconButton` styles will win (this example makes for a long css prop list).

modules/react/popup/lib/PopupTarget.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import {
+  createComponent,
+  useModelContext,
+  createHook,
+  useForkRef,
+  composeHooks,
+  composeHooks2,
+  composeHooks3,
+  mergeProps,
```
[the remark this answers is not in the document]
NicholasBoll: WIP testing. I pushed what I had to move using a remote SSH environment

modules/react/popup/lib/usePopupModel.ts
```diff
@@ -0,0 +1,126 @@
...
+  callbacks: {
+    ...disclosureEventMap.callbacks,
+    /**
+     * Called whenever the placement changes. This callback only fires when placement changes. For
+     * example, if the preferred placement of a Popper is 'bottom' and there's enough room to place
+     * a Popup on the bottom, the Popup will render below a target element without firing this
+     * callback. If there is not enough room and PopperJS chooses to update, this callback will be
+     * called to inform of the change in placement.
+     */
+    onUpdatePlacement: 'updatePlacement',
```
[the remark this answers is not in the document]
NicholasBoll: Haha. Yes. I named the event `updatePlacement`, but callback names should be `onUpdatePlacement`. I fixed this so it sounds more correct. events !== callbacks, so the naming conventions of events should not follow naming conventions of callbacks.

modules/react/popup/stories/examples/OpenWithoutModel.tsx
```diff
@@ -0,0 +1,36 @@
...
+        padding={PopupOld.Padding.s}
+        handleClose={() => null}
+      >
+        <div style={{marginBottom: '24px'}}>
+          Are you sure you'd like to delete the item titled 'My Item'?
+        </div>
+
+        <DeleteButton style={{marginRight: '16px'}}>Delete</DeleteButton>
+        <Button>Cancel</Button>
+      </PopupOld>
```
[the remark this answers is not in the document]
NicholasBoll: I left this here for eyeball comparison since the dom structure was changing.

modules/react/toast/lib/Toast.tsx
NicholasBoll: The transform origin is determined by the `placement` in the `PopupModel`. This means the developer doesn't have to set this correctly, it will be handled by the chosen placement. This is an improvement for most popups since the Popup card will now animate away from the target. This effectively has no changes for Toast since they animate down by default and `bottom` is the default placement.

NicholasBoll: The animation is also applied as a CSS style. It is still possible to override the animation by passing in a `css` prop or using `styled(Toast)` to override animations.

modules/react/popup/stories/Popup.stories.mdx
NicholasBoll: What about "behavioral hooks"?

modules/react/popup/stories/Popup.stories.mdx
NicholasBoll: "sub-component" or "subcomponent"?

modules/react/popup/stories/Popup.stories.mdx
NicholasBoll: I like "generic", but not "catch-all"

modules/react/popup/stories/Popup.stories.mdx
NicholasBoll: Behaviors like closing on escape key, closing when clicking outside, focus management, etc. I realize now that "behavior hook" is overloaded with compound components. "Popup behaviors" existed before "behavior hooks" were defined in our compound component doc...

Hooks like `useCloseOnEscape` are still behaviors, but they aren't directly tied to subcomponents like the compound component doc lays out. `useCloseOnEscape` operates only on the model and not on a direct element. I guess technically they operate on the `stack` element which isn't owned by React, but instead owned directly by the `PopupStack` and storied in the `PopupModel`.

modules/react/popup/stories/Popup.stories.mdx
NicholasBoll: I don't think people should do this, but it is the easiest way to upgrade. I'm not sure where this type of transition documentation should be, but it will be questions we'll commonly get. I wouldn't want v5 adoption to be blocked by Popups being too difficult to migrate to for some teams.

modules/react/popup/stories/Popup.stories.mdx
NicholasBoll: Hmm. That's fair. `usePopupOld` is a duplicate during testing.

modules/react/popup/stories/Popup.stories.mdx
anicholls: Ah - I see. I think it was the term "registered" that confused me. Maybe just something like:
```suggestion
The Popup has no pre-defined behaviors built in, therefore the `usePopupModel` must always
```

modules/react/popup/stories/Popup.stories.mdx
anicholls: Maybe we put that in the migration guide rather than in our storybook docs then?

modules/react/popup/lib/hooks/useFocusRedirect.ts
```diff
@@ -0,0 +1,46 @@
...
+ * "Trap" or "loop" focus within a PopupModel's `stackRef` element. This is required for
+ * accessibility on modals. If a keyboard users hits the Tab or Shift + Tab, this will force
+ * "looping" of focus. It effectively "hides" outside content from keyboard users. Use an overlay to
+ * hide content from mouse users and `useAssistiveHideSiblings` to hide content from assistive
+ * technology users.
+ */
+export const useFocusRedirect = (model: PopupModel, elemProps = {}) => {
+  const onKeyDown = React.useCallback(
+    (event: KeyboardEvent): void => {
+      if (model.state.stackRef.current && event.key === 'Tab') {
```
anicholls, line 16: This breaks if you shift + tab while on the last focusable element in the popup

modules/react/popup/README.md
```diff
@@ -285,150 +135,154 @@ stack, all popups will close rather than only the topmost one.
...
+      : null}
+  </>
+);
 ```
 
 ## useAssistiveHideSiblings
 
 ```ts
-useAssistiveHideSiblings(ref: React.RefObject<HTMLElement>): void
+useAssistiveHideSiblings(model: PopupModel): {}
```
anicholls, line 163: Should we just link to the storybook mdx/docs for this stuff so we don't have it duplicated? That's what we've done elsewhere

modules/react/popup/stories/Popup.stories.mdx
```diff
@@ -0,0 +1,483 @@
...
+```sh
+yarn add @workday/canvas-kit-react
+```
+
+## Usage
+
+The `Popup` component is a low-level compound component that is used to build popup UIs that are not
+already covered by Canvas Kit.
+
+### Basic
```
anicholls, line 51: Nit: The story is called `Simple`, but the heading is `Basic`. Let's align on one.

modules/react/popup/stories/examples/Simple.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <Popup model={model}>
+      <Popup.Target as={DeleteButton}>Delete Item</Popup.Target>
+      <Popup.Popper placement={'bottom'}>
+        <Popup.Card width={400} padding="16px">
+          <Popup.CloseIcon aria-label="Close" />
+          <Popup.Heading>Delete Item</Popup.Heading>
+          <Popup.Body>
+            <p>Are you sure you'd like to delete the item titled 'My Item'?</p>
+
+            <Popup.CloseButton as={DeleteButton}>Delete</Popup.CloseButton>
```
anicholls, line 31: Nit: `HStack` for spacing?

modules/react/modal/stories/stories.tsx
```diff
@@ -39,50 +39,58 @@ export const Default = () => {
...
 
   return (
     <>
       <DeleteButton {...targetProps}>Delete Item</DeleteButton>
-      <Modal data-testid="TestModal" heading={'Delete Item'} {...modalProps}>
+      <Modal heading={'Delete Item'} {...modalProps}>
         <p>Are you sure you'd like to delete the item titled 'My Item'?</p>
-        <SecondaryButton {...popup1.targetProps}>Open Popup to reach Delete button</SecondaryButton>
-        <SecondaryButton {...popup2.targetProps}>Non-hidable Popup</SecondaryButton>
+        <Popup.Target model={popup1}>Open Popup to reach Delete button</Popup.Target>
```
anicholls, line 52: Nit: Can we add some spacing to this story in both popups to make things look a little nicer?

modules/react/modal/stories/stories.tsx
```diff
@@ -39,50 +39,58 @@ export const Default = () => {
...
+          <Popup.CloseIcon model={popup2} aria-label="Close" />
+          <Popup.Heading model={popup2}>Does Not Hide On Click Outside</Popup.Heading>
+          <Popup.Body>
+            Pressing 'Delete' will close the modal
+            <Tooltip
+              placement="left"
+              title={'Really, Really, Really, Really, Really sure'}
+              type={'muted'}
+            >
+              <DeleteButton style={{marginRight: '16px'}} onClick={closeModal}>
```
anicholls, line 87: It's odd that clicking this doesn't close itself as well.

modules/react/modal/stories/stories.tsx
```diff
@@ -39,50 +39,58 @@ export const Default = () => {
...
+          <Popup.CloseIcon model={popup1} aria-label="Close" />
+          <Popup.Heading model={popup1}>Really Delete</Popup.Heading>
+          <Popup.Body>
+            Pressing 'Delete' will close the modal
+            <Tooltip
+              placement="left"
+              title={'Really, Really, Really, Really, Really sure'}
+              type={'muted'}
+            >
+              <DeleteButton style={{marginRight: '16px'}} onClick={closeModal}>
```
anicholls, line 69: It's odd that clicking this doesn't close itself as well.

anicholls (commented):
Do you plan on adding a modal mdx file in this PR? or will that come in a future update?

modules/react/popup/stories/Popup.stories.mdx
NicholasBoll: I think we should link between documents, but we don't have a way to do that well yet. Do we link to other stories? If yes, how do we know when those links break?

modules/react/popup/README.md
NicholasBoll: I think so... But where to link? What's the Storybook URL? Do we just link to Github Pages? This would mean the link doesn't work until we merge to master.

modules/react/popup/lib/hooks/useFocusRedirect.ts
NicholasBoll: Excellent catch! I'll fix and add Cypress specifications.

modules/docs/lib/Specifications.tsx
```diff
@@ -1,11 +1,11 @@
 /* eslint-disable no-param-reassign */
 import React from 'react';
 import {toId} from '@storybook/csf';
-import { Table, TableRow } from "@workday/canvas-kit-react/table";
-import { Hyperlink } from "@workday/canvas-kit-react/button";
+import {Table, TableRow} from '@workday/canvas-kit-react/table';
```
NicholasBoll, line 4: Prettier running on the file. I'm okay to leave in

modules/docs/lib/Specifications.tsx
```diff
@@ -74,25 +74,31 @@ export const Specifications = ({file, name}: SpecificationsProps) => {
   }
 
   const renderGiven = (input: string) => {
+    if (!input) {
+      return `Could not find a "given". Check the spec file.`;
+    }
+
```
NicholasBoll, line 80: This error is more helpful than the spec table crashing.

cypress/integration/Popup.spec.ts
```diff
@@ -151,4 +132,243 @@ describe('Popup', () => {
...
+            it('should NOT close the "Delete Item" popup', () => {
+              cy.findByRole('dialog', {name: 'Delete Item'}).should('be.visible');
+            });
+          });
+        });
+      });
+    });
+  });
+
+  context('given the [Components/Popups/Popup/React, FocusRedirect] story is rendered', () => {
```
NicholasBoll, line 316: The following are new tests that verify `useFocusRedirect`

cypress/integration/Popup.spec.ts
```diff
@@ -151,4 +132,243 @@ describe('Popup', () => {
       });
     });
   });
+
+  context('given the [Testing/React/Popups/Popup, MixedPopupTypes] story is rendered', () => {
```
NicholasBoll, line 136: These specs used to live in `PopupStack.spec.ts`, but I deleted that file and move the story it was testing. All the functionality requires Popup hooks, so it makes more sense for the stories and specs to live in `Popup` instead of `PopupStack`

cypress/support/commands.ts
```diff
@@ -30,7 +48,9 @@ Cypress.Commands.overwrite('tab', (originalFn, subject) => {
 
   log.snapshot('before', {next: 'after'});
 
-  return Cypress.Promise.try(() => originalFn(subject))
+  return Cypress.Promise.try(() => {
+    return originalFn(subject, options);
```
NicholasBoll, line 52: This fixes an issue where `cy.tab({ shift: true })` didn't work, because the override didn't pass `options`

cypress/support/cypress-plugin-tab.d.ts
```diff
@@ -0,0 +1,5 @@
+declare namespace Cypress {
```
NicholasBoll, line 1: Added because of https://github.com/Bkucera/cypress-plugin-tab/issues/47

cypress/support/index.ts
```diff
@@ -1,5 +1,5 @@
 import 'cypress-pipe';
-import 'cypress-plugin-tab';
+import './cypress-plugin-tab';
```
NicholasBoll, line 2: https://github.com/Bkucera/cypress-plugin-tab/issues/47

jest/verifyComponent.tsx
```diff
@@ -59,7 +59,7 @@ export function verifyComponent(
         role: 'MyRole',
         'data-testid': 'MyTestID',
       } as const;
-      render(<Test ref={ref} {...attrs} {...callbacks} />);
+      render(<Test ref={ref} {...attrs} />);
```
NicholasBoll, line 62: `...callbacks` did nothing but throw a bunch of warnings in Jest. I'm not sure why it was there.

modules/react/modal/stories/stories.tsx
NicholasBoll: Hmm. This is a testing story

jamesfan (commented):
Fixing a couple leftover issues from resolving merge conflicts.

modules/react/modal/stories/stories.tsx
NicholasBoll: I fixed this and moved to a testing story

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
   component documentation for more details.
+
+### Popper
+
+Popper used to render an empty `div` element as a child of the element created by the `PopupStack`.
+`ref` and `elemProps` (extra props) applied to that element. Now `ref` applies to the element
+created by the `PopupStack` and `elemProps` don't get applied to anything. The `PopupStack` is not
+React-specific and there is no easy way to spread extra props onto this element like other
+components do. The `ref` still points to that element, so you can add anything extra to that `ref`
+via DOM APIs if necessary. There is no codemod for this change.
```
jamesfan, line 674:
```suggestion
In v4, Popper rendered an empty `div` element as a child of the element created by the `PopupStack`
and applied `ref` and `elemProps` (extra props) to that `div` element.

We've updated Popper in v5 to instead apply `ref` directly to the element created by the
`PopupStack`. The `PopupStack` is not React-specific; there is no easy way to spread extra props to
this element as we do for other components, so we've discarded `elemProps`. If necessary, you can
still target the element using `ref` and modify it using DOM APIs.

There is no codemod for this change.
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+
+### Popups
+
+`Popup` has transitioned to a compound component, along with all Popup-based behavior hooks. What
+was a `Popup` prior to v5 is now a `Popup.Card` in v5. The target button and `Popper` components
+have been moved to subcomponents of a `Popup`.
```
jamesfan, line 680:
```suggestion
Popup has transitioned to a [compound component](./COMPOUND_COMPONENTS.mdx), along with all
Popup-based behavior hooks. What was a `Popup` in v4 is now a `Popup.Card` in v5. The target button
and `Popper` components have also been converted to subcomponents of `Popup`.
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+
+**Before**
```
jamesfan, line 682:
```suggestion
#### v4
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+          <DeleteButton onClick={onDeleteClick}>Delete</DeleteButton>
+          <Button onClick={closePopup}>Cancel</Button>
+        </Popup>
+      </Popper>
+    </>
+  );
+};
+```
+
+**After**
```
jamesfan, line 728:
```suggestion
#### v5
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+      </Popup.Popper>
+    </Popup>
+  );
+};
+```
+
+The most notable changes is `Popup` is now a container component that takes a `PopupModel` and has
+several subcomponents like `Popup.Target` and `Popup.CloseButton`. These components are hooked up to
+the `PopupModel` via React context and have access to state and events. `Popup.Card` is what the v4
+`Popup` once was.
```
jamesfan, line 780:
```suggestion
Most notably, `Popup` is now a container component that takes a `PopupModel` and has several
subcomponents like `Popup.Target` and `Popup.CloseButton`. These components are hooked up to the
`PopupModel` via React context and have access to state and events. `Popup.Card` is what the v4
`Popup` once was.
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+
+All behavior hooks, like `useCloseOnEscape` now take a `model` instead of variable parameters. This
+allowed us to fix some subtle bugs. Using the `PopupModel` means all hooks have access to all Popup
+state and events without passing in many parameters.
+
+#### usePopup
```
jamesfan, line 786:
```suggestion
#### usePopup and usePopupModel
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+
+In the above example, `usePopup` has been removed. `usePopupModel` should be used instead. All
+subcomponents have an associated behavior hook. For example, `Popup.Target` uses a hook called
+`usePopupTarget`. If you need to use your own components for any reason, these hooks are available.
+`Popup.Target` and `Popup.CloseButton` do not include any styling. They both use `SecondaryButton`
+as the default component rendered. You can change this via the `as` prop. For example, the following
+will render an unstyled button:
```
jamesfan, line 793:
```suggestion
As shown in the example above, `usePopupModel` should now be used instead of `usePopup`. All
subcomponents have an associated behavior hook. For example, `Popup.Target` uses a hook called
`usePopupTarget`. If you need to use your own components for any reason, these hooks are available.
`Popup.Target` and `Popup.CloseButton` do not include any styling. They both render
`SecondaryButton` by default. You can change this via the `as` prop. For example, the following will
render an unstyled button:
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+
+```tsx
+<Popup.Target as="button">Show</Popup.Target>
+```
+
+Pass a `css` prop, or a styled button instead to have a custom styled button. You could even pass
+`IconButton` if you need an icon button to show a popup instead!
+
+If you were using `usePopup`, here's a list of equivalent APIs:
```
jamesfan, line 802:
```suggestion
Pass a `css` prop or a styled button instead to have a custom styled button. You could even pass
`IconButton` if you need an icon button to show a Popup instead!

If you were using `usePopup` before, here's a list of equivalent APIs:
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+
+#### New Focus management
+
+A common theme we noticed in uses of Popup in the wild was focus management. Developers were
+manually passing a `ref` to the target button element and manually returning focus to it when
+closing a Popup through whatever means it was closed. This use-case should be handled now by a new
+hook: `useReturnFocus`. By default, `useReturnFocus` will return focus to the `targetRef` in the
+model, which is set by `Popup.Target`. This can be overridden via `returnFocusRef` passed to the
+model on creation. This `returnFocusRef` might make some migrations easier if `Popup.Target` cannot
+be used for whatever reason.
```
jamesfan, line 821:
```suggestion
#### New Focus Management

A common theme we noticed in uses of Popup in the wild was focus management. Developers were
manually passing a `ref` to the target button element and manually returning focus to it when
closing the Popup. This use case should now be handled by the new `useReturnFocus` hook. By default,
`useReturnFocus` will return focus to the `targetRef` in the model, which is set by `Popup.Target`.
This can be overridden by passing `returnFocusRef` to the model on creation. `returnFocusRef` should
make your migration easier if `Popup.Target` cannot be used for whatever reason.
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+  returnFocusRef: buttonRef, // only use if you cannot use `Popup.Target`
+});
+
+useReturnFocus(model);
+```
+
+Another common focus thing to do is to focus something within a popup when the popup is shown. The
+`useInitialFocus` was created for this purpose. `useInitialFocus` will set focus on the first
+focusable element when the popup becomes visible. This behavior can be overridden by setting the
+`initialFocusRef` in the model.
```
jamesfan, line 844:
```suggestion
Another common use case involved focusing something within the Popup when the Popup was shown. The
`useInitialFocus` hook was created for this purpose. `useInitialFocus` will set focus to the first
focusable element when the Popup becomes visible. This behavior can be overridden by passing
`initialFocusRef` to the model.
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+  initialFocusRef: someRef, // only use if you want to explicitly focus on something. Could be useful for an input.
+});
+
+useInitialFocus(model);
+```
+
+#### Using your own positioning
+
+If you use your own positioning and only need to use the original `Popup`, you can use `Popup.Card`
+instead. Without the model and behaviors, the following is equivalent:
```
jamesfan, line 869:
```suggestion
#### Managing Positioning

If you'd prefer to manage positioning yourself, you can use `Popup.Card` on its own. Without the
model and behaviors, the following is equivalent:
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+
+```tsx
+// before
```
jamesfan, line 872:
```suggestion
// v4
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+<Popup width={width} handleClose={onClose} heading="Popup Heading">
+  Popup Content
+</Popup>
+
+// after
```
jamesfan, line 877:
```suggestion
// v5
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+<Popup.Card with={width}>
+  <Popup.CloseIcon aria-label="Close" onClick={onClose} />
+  <Popup.Heading>Popup Heading</Popup.Heading>
+  <Popup.Body>Popup Content</Popup.Body>
+</Popup.Card>
+```
+
+`Popup.Card` is using `Card`, which is now using `Box`. This means changes to some props:
```
jamesfan, line 885:
```suggestion
`Popup.Card` uses `Card`, which is now using `Box`. Consequently, the following props have changed:
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+
+| Before                         | After                                      |
+| ------------------------------ | ------------------------------------------ |
+| `padding={Popup.Padding.zero}` | `padding="zero"` or `padding={space.zero}` |
+| `depth={depth[0]}`             | `depth={0}`                                |
+| `popupRef={ref}`               | `ref={ref}`                                |
+
+#### Transitioning
+
+We noticed Popups were used in 2 different ways: always rendering and conditional rendering.
```
jamesfan, line 895:
```suggestion
We noticed Popups were used in two different ways: always rendering and conditional rendering.
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+
+```tsx
+// always rendering
```
jamesfan, line 898:
```suggestion
// Always rendering
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+        <Popup>
+          {/* content */}
+          <button onClick={handleClose}>Close</button>
+        </Popup>
+      </Popper>
+    </>
+  )
+}
+
+// conditional rendering
```
jamesfan, line 923:
```suggestion
// Conditional rendering
```

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -621,3 +663,346 @@ we've made a few updates to the component in v5:
...
+```
+
+The difference between the two is subtle, but when always rendering example, the `usePopup` hook
+runs on every render. In the conditional rendering example, the `usePopup` hook only runs when
+`MyPopup` renders it. This means hooks like `useCloseOnEscape` need to function properly in both
+cases, but `open` is not passed to the hook. This caused subtle bugs, and now `useCloseOnEscape` is
+passed a `PopupModel` which does have access to the popup's visible state. `useCloseOnEscape` will
+now only run when the popup is visible, but this means the conditional rendering example will have
+to do extra work because the `target` is out of scope of the `MyOpenPopup` component. The following
+is equivalent to the example in v5:
```
jamesfan, line 969:
```suggestion
The difference between the two is subtle, but in the always rendering example, the `usePopup` hook
runs on every render. In the conditional rendering example, the `usePopup` hook only runs when
`MyPopup` renders it. This means hooks like `useCloseOnEscape` need to function properly in both
cases, but `open` is not passed to the hook. This caused subtle bugs. Now, `useCloseOnEscape` is
passed a `PopupModel` which has access to the popup's visible state. `useCloseOnEscape` will now
only run when the popup is visible, but this means the conditional rendering example will have to do
extra work because the `target` is out of scope of the `MyOpenPopup` component. The following is
equivalent to the example in v5:
```

jamesfan (commented):
Did a bit of wordsmithing and fixed a few grammar nits. Feel free to use as much or as little of this as you'd like, particularly if I've misrepresented you in any of these changes.

modules/react/popup/README.md
NicholasBoll: I did a link similar to Breadcrumbs - a link to the master storybook. I'd like to have a more robust solution later.

NicholasBoll:
> Do you plan on adding a modal mdx file in this PR? or will that come in a future update?

Modal will be another PR

anicholls (approved):

PR #1068, opened 2021-05-12
feat: Forward refs for input components

### LINKED ISSUE #1050: Forward refs for Inputs
- Move **InputRef > Ref**
- Implement Code mod

Target Components
- Text Input
- Text Area
- Radio
- Checkbox
- Switch
- Select One in Labs

jamesfan:
## Summary

Fixes #1050.

Updates the following input components (using `createComponent`) to forward refs using the `ref` attribute instead of the `inputRef` prop:

- Checkbox
- Color Input
  - Not in the original scope, but I updated it since it wraps Text Input
- Color Preview
  - Not in the original scope, but I updated it since it wraps Color Input
- Radio
- Select (Main)
  - Not in the original scope, but I updated it anyway it to maintain parity with the rest of the input components in the Main package
  - Didn't support `inputRef` to begin with
- Select (Preview/Labs)
  - Didn't support `inputRef` to begin with
- Switch
- Text Input
- Text Area

Also updates Combobox and Search Bar to work with the updated Text Input (Combobox uses Text Input and Search Bar uses Combobox).

## Checklist

- [x] Implement codemod for converting `inputRef` to `ref`
- [x] Update component docs to replace `inputRef` with `ref`
- [x] Update 5.0 Migration Guide
- [x] Update unit tests to use `verifyComponent` to test ref forwarding (among other things)


modules/labs-react/combobox/lib/Combobox.tsx
```diff
@@ -184,11 +184,19 @@ const Combobox = ({
...
+  // forwarded to the TextInput clone.
+  const inputRef: React.RefObject<HTMLInputElement> = useRef(null);
+  // We need access to the original TextInput's ref _property_ (not prop) so we
+  // can combine it with the internal inputRef using useForkRef. ref isn't
+  // listed in the ReactElement interface, but it's there, so we cast children
+  // to satisfy TS.
+  const elementRef = useForkRef(
+    (children as React.ReactElement<TextInputProps> & {ref: React.Ref<HTMLInputElement>}).ref,
+    inputRef
+  );
```
jamesfan, line 199: Calling out this funky bit of casting I had to do to get TS to stop complaining about the lack of a `ref` property.

@NicholasBoll We discussed this code a bit earlier. I ended up not using `PropsWithAs` in favor of adding `ref` to the casted interface explicitly.

modules/labs-react/combobox/lib/Combobox.tsx
```diff
@@ -409,12 +417,13 @@ const Combobox = ({
         paddingRight: space.xl,
       };
     }
-    const newTextInputProps: Partial<TextInputProps> = {
+    const newTextInputProps: Partial<TextInputProps &
+      React.InputHTMLAttributes<HTMLInputElement> & {ref: React.Ref<HTMLInputElement>}> = {
```
jamesfan, line 421: I had to modify this type definition to get TS to stop complaining about the `ref` below. I don't think this definition is necessary (TS seems to be able to infer just fine without it), however, so I may remove it.

modules/labs-react/select/spec/Select.spec.tsx
```diff
@@ -49,6 +49,17 @@ describe('Select', () => {
...
+  // TODO: Replace ref test with verifyComponent after fixing Select to handle
+  // consumer-provided onKeyDown event handler. Currently fails the onKeyDown
+  // callback test in verifyComponent.
+  describe('when rendered with a ref', () => {
+    it('should set the ref to the button element', async () => {
+      const ref: React.RefObject<HTMLButtonElement> = React.createRef();
+      const {findByRole} = render(<Select onChange={cb} options={options} ref={ref} />);
+      expect(await findByRole(selectButtonRole)).toEqual(ref.current);
+    });
+  });
```
jamesfan, line 61: The Labs Select doesn't handle a consumer-provided onKeyDown handler correctly, so it fails the callback test in `verifyComponent`. Addressing this in a later PR and testing ref forwarding separately in the meantime.

jamesfan (commented):
I haven't implemented the codemod yet, but everything else in this PR is ready for review.

@NicholasBoll I haven't updated the unit tests to use `screen` yet. Will look into this later (or save it for a later PR).

modules/labs-react/combobox/spec/Combobox.spec.tsx
```diff
@@ -90,6 +90,18 @@ describe('Combobox', () => {
...
+  describe('when the child input is rendered with a ref', () => {
+    it('should set the ref to the child input element', async () => {
+      const ref: React.RefObject<HTMLInputElement> = React.createRef();
+      const {findByRole} = render(
+        <Combobox>
+          <TextInput ref={ref} />
+        </Combobox>
+      );
+      expect(await findByRole('combobox')).toEqual(ref.current);
+    });
```
NicholasBoll, line 102: The rest of this file uses the `screen` interface. Also, why did you need to await? Would the following work?
```suggestion
    it('should set the ref to the child input element', () => {
      const ref: React.RefObject<HTMLInputElement> = React.createRef();
      render(
        <Combobox>
          <TextInput ref={ref} />
        </Combobox>
      );
      expect(screen.getByRole('combobox')).toEqual(ref.current);
    });
```

modules/labs-react/select/spec/Select.spec.tsx
NicholasBoll: This could probably be done pretty easily. Without using `mergeProps`, you instead do something like the following:

```tsx
const { onKeyDown } = props

const handlers = {
  onKeyDown(event) {
    onKeyDown?.(event)
    // other stuff
  }
}
```

NicholasBoll (approved):

modules/labs-react/combobox/spec/Combobox.spec.tsx
jamesfan: @NicholasBoll The `async` and `await` are there from me copy-and-pasting [an old input ref test](https://github.com/Workday/canvas-kit/pull/1068/files#diff-00d14d17114b0b458110cb29abeca60151e11e4ff55e6b31522f84eeb9b9c688L58) and using it here. I thought we might be using `findByRole` due to `ref.current` not being set immediately, but the test seems to work fine if I remove `async/await` and use `getByRole` instead.

Regardless, I updated the test to use `screen`. Instead of importing `screen`, the Combobox tests set `screen` to the return value of `render` and run queries against it -- your code above works as long as I set `const screen = render(<Combobox>...</Combobox>)`.

modules/react/color-picker/lib/ColorInput.tsx
```diff
@@ -109,29 +116,52 @@ const SwatchTile = styled(ColorSwatch)({
...
+    Element
+  ) => {
+    const formattedValue = formatValue(value);
+
+    const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+      const value = formatValue(e.currentTarget.value);
+
+      if (onChange) {
+        onChange(e);
+      }
```
alanbsmith, line 151: Comment (not an RC) I _think_ `createComponent` will handle merging this `onChange` for you. I'd need to poke at the code a bit to be sure though. Maybe something to look at. Your call.

modules/react/color-picker/spec/ColorInput.spec.tsx
```diff
@@ -12,6 +12,8 @@ describe('ColorInput', () => {
     cb.mockReset();
   });
 
+  verifyComponent(ColorInput, {});
+
```
alanbsmith, line 16: This is a nice test clean up. 👍🏻

modules/react/select/README.md
```diff
@@ -15,8 +15,8 @@ yarn add @workday/canvas-kit-react
 #### Simple Example
 
 **Note:** While a base select component is provided in this package, it is **not accessible** when
-used as is. It should be used in tandem with [`FormField`](../../form-field/react) to be
-made fully accessible (see below).
+used as is. It should be used in tandem with [`FormField`](../../form-field/react) to be made fully
+accessible (see below).
```
alanbsmith, line 19: 👏🏻 Good call out here.

alanbsmith (approved):
Looks great! I had one question, but nothing to hold this PR up. Thanks for the docs improvements as well. 👏🏻

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -166,6 +167,57 @@ Components are transitioning to [Compound Components](./COMPOUND_COMPONENTS.md)
...
+- Color Preview
+- Select
+- Select (Preview)
+- Text Input
+- Text Area
+
+Passing `ref={ref}` to any of these components in v4 would have set `ref.current` to the mounted
+instance of the entire component
+([source](https://reactjs.org/docs/refs-and-the-dom.html#accessing-refs)) rather than the underlying
+HTML element represented by the component. This is no longer the case in v5.
```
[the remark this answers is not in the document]
jamesfan: Thanks!

modules/labs-react/pagination/lib/Pagination/GoTo/TextInput.tsx
```diff
@@ -4,6 +4,7 @@ import {TextInput, TextInputProps} from '@workday/canvas-kit-react/text-input';
 
 export type GoToTextInputProps = TextInputProps & {
   'aria-label': string;
+  value?: string | number;
```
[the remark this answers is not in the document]
jamesfan: It's related to the changes in this PR, specifically the use of `createComponent` for `TextInput`. If you look at [the changes to `TextInputProps`](https://github.com/Workday/canvas-kit/pull/1068/files#diff-de6c3686ab8bc1cfd55f609d7988122166a7249cc92fa00829fc91668d2243c1R13) in this PR, you'll see that it no longer extends from `React.InputHTMLAttributes<HTMLInputElement>`. This is because that portion of the `TextInput` interface (the HTML attributes of the underlying element being represented by `TextInput`) is now being handled by `createComponent` rather than being specified in the `TextInputProps` definition.

Without `value` in `GoToTextInputProps`, TS complains about `value` not existing in `GoToTextInputProps` when it's referenced later in this file (because `value` is no longer in `TextInputProps`).

modules/labs-react/combobox/spec/Combobox.spec.tsx
NicholasBoll: That's fine. React Testing Library uses `act` internally which synchronously flushes rendering updates. This is why `get*` works almost always. I really like the queries in React Testing Library, but honestly the differences between `get*`, `find*` and `query*` are kind of a mess. I didn't understand the differences until I started contributing to the project...

modules/react/color-picker/lib/ColorInput.tsx
NicholasBoll: `createComponent` doesn't take care of merging props, `mergeProps` does. Something like:

```tsx
const inputProps = mergeProps({
 onChange: handleChange,
  // other props you want
}, elemProps)
```

Then on `<CustomHexInput>`, you'd do something like:
```tsx
<ColorHexInput
  dir="ltr"
  // ...
  {...inputProps}
>
```

I'm not sure I'd worry about it though. I think it would be better to turn this into a compound component with behavior hooks that are exported which handle prop merging better.

If you wanted, this line could be simplified to:

```tsx
onChange?.(e)
```

alanbsmith: Oh I was under the impression it was merging the css prop and other event handler functions. I’ll look back at the source again.

modules/codemod/lib/v5/renameInputRefs.ts
```diff
@@ -0,0 +1,79 @@
...
+          // (a) elementName was the name used in a default import of the current
+          //     input being processed (e.g., `import TextInput ...`)
+          // (b) elementName was renamed from a named import of the current input
+          //     being processed (e.g., `import {TextInput as CanvasTextInput} ...`)
+          // (c) elementName was created by calling `styled` on the current input
+          //     being processed (e.g., `const StyledTextInput = styled(TextInput)`)
+          if (
+            defaultImports.includes(elementName) ||
+            inputNames.map(n => importMap[n]).includes(elementName) ||
+            inputNames.map(n => styledMap[n]).includes(elementName)
```
NicholasBoll, line 67: Glad the `getImportRenameMap` utility function was useful! Makes for simpler codemods

NicholasBoll (approved):

modules/codemod/lib/v5/renameInputRefs.ts
jamesfan: It was such a time-saver -- thanks for putting it together!

PR #1074, opened 2021-05-18
fix(popup): Allow tooltips to close alongside modals on click outside

csongnguyen:
## Summary

Before the fix, when a tooltip is opened on top of a modal, it required two clicks to close the tooltip and then the modal. With the fix, a single click will close a tooltip and the modal underneath. 

## Checklist

- [x] `yarn test` passes
      applicable

### DIFF
--- a/cypress/integration/Modal.spec.ts
+++ b/cypress/integration/Modal.spec.ts
@@ -262,7 +262,6 @@ describe('Modal', () => {
 
           context(`when clicking outside the modal`, () => {
             beforeEach(() => {
-              // cy.findByRole('dialog', {name: 'Delete Item'}).click();
               cy.get('body').click('top');
             });
 
--- a/cypress/integration/Modal.spec.ts
+++ b/cypress/integration/Modal.spec.ts
@@ -162,6 +162,132 @@ describe('Modal', () => {
     });
   });
 
+  context(`given the 'With Tooltips' story is rendered`, () => {
+    beforeEach(() => {
+      h.stories.load('Components/Popups/Modal/React', 'With Tooltips');
+    });
+
+    context('when the modal is open', () => {
+      beforeEach(() => {
+        cy.findByText('Delete Item').click();
+      });
+
+      it('should open the modal', () => {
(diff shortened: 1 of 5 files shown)


cypress/integration/Popup.spec.ts
```diff
@@ -113,8 +113,7 @@ describe('Popup', () => {
           cy.findByRole('dialog', {name: 'Popup 2'}).should('be.visible');
         });
 
-        // TODO Skip for now until we have a systematic approach to fix this issue
-        it.skip('should close Popup 1', () => {
+        it('should close Popup 1', () => {
```
csongnguyen, line 116: I noticed this test was passing, so I removed the skip method here. Let me know if this should be reverted.

cypress/integration/Modal.spec.ts
```diff
@@ -162,6 +162,132 @@ describe('Modal', () => {
...
+                'be.visible'
+              );
+            });
+
+            it(`should close the modal`, () => {
+              cy.findByRole('dialog', {name: 'Delete Item'}).should('not.be.visible');
+            });
+
+            // TODO: Manually, this test case works, but cypress click on body does not close the tooltip.
+            it.skip(`should close the 'Delete' tooltip`, () => {
```
csongnguyen, line 280: I could not get around this error and could not reproduce manually, but this is the error that appears in the DOM when the test runs: `Failed to execute 'removeChild' on 'Node'`

If someone has an idea of how to get around this error, that would be really helpful 🙏

cypress/integration/Popup.spec.ts
NicholasBoll: This is passing in v5 as well, so I removed the `.skip` there. I think we ended up fixing the issue! 🎉

cypress/integration/Modal.spec.ts
csongnguyen: Update from offline conversation with Nicholas: this was a legitimate issue caught by cypress, and this will be fixed in v5, so for now, keeping this test in here and skipping it until we upgrade to v5.

NicholasBoll (approved):
Looks good!

Issue #131, opened 2019-08-19
Display package version in Storybook


lychyi:
## 🚀 Feature Proposal

To get a better understanding of the Storybook we're looking at we should be displaying a version in an obvious location.

## Motivation

This will make looking at a storybook and using a component stay in sync mentally if features or bug fixes got added.

## Example

For example, it would be nice to know what package version was represented when you looked at an `Input` or `Button` component.

NicholasBoll:
This is fixed by #1082

PR #1082, opened 2021-05-26
feat: Add package version in Storybook

mihaelamiches:
## Summary

Display the version number on storybook to make it obvious what's the version of the documentation we're looking at.

Related to #131

Not sure if in the issue the intent was to display the version number separately for each module, but I figured since lerna is set to use fixed mode for versions, it will be safe to display the version number on the sidebar.

## Checklist

- [x] `yarn test` passes

## Additional References

screenshot of storybook displaying the version number under the CanvasKit title


NicholasBoll (approved):
I think this will work for now. Thank you!

mihaelamiches:
office-dance

PR #1086, opened 2021-05-28
chore: Upgrade Card to use Box

NicholasBoll:
- Remove docgen processing that Storybook does so we don't double-process files
- Filter out `Box`, `Flex`, etc style props from prop tables other than `PropTables.tsx` example files
- Add `ExtractProps` utility type


NicholasBoll:
The state tables reveal a regression in depth settings.

PR #1091, opened 2021-06-07
Refactor/modal compound component

### LINKED ISSUE #56: Modal has 1px gray card border
## 🐛 Bug Report

Modals are placed on a dark background, but have a gray border from the nested `Card` 
component. `Modal` renders a `Popup`, which then renders the `Card`. This gray border makes the edge look blurry. This could potentially be an issue when placing cards on any dark or colored background (i.e. the design.workday.com).

Screen Shot 2019-07-31 at 4 11 35 PM

## Expected Behavior

While we want the gray border on Cards and Popups, there should be no border on Modals.

### LINKED ISSUE #398: Modal handleClose prop is being used to determine multiple things
## 🐛 Bug Report

https://workday.github.io/canvas-kit/?path=/story/components-popups-modal-react--without-close-icon shows an example of a modal without the `X` icon in the top right but the documentation doesn't clearly state how it's intended to be configured that way. Passing `handleClose={undefined}` seems to do the trick but then setting `closeOnEscape={true}` no longer works.

## Expected Behavior

I expected the modal to have a boolean prop such as `showCloseIcon` to help with discoverability.

### LINKED ISSUE #493: Support Custom Heading Elements for Modal
## 🚀 Feature Proposal
By default, `Modal` seems to render the header provided to it inside an `h3`. In some cases, we have complex headings that would benefit from richer semantic markup for accessibility purposes. (Typically we'd probably use some pairing of a heading + subheading, but there are other uses, too.)

It would be great if `Modal`, as part of CKR4, understood the non-string case to be a _totally custom_ header, and wrapped the content in non-semantic markup, like a `div` or whatnot 🙂 

## Motivation
Currently, complex headers are not as accessible as they could be, as any header provided will be rendered as a child of an `h3`.

## Example
```tsx
const Header = () => (
  <>
    <h3>My Header</h3>
    <h4>For Item 43</h4>
  </>
)

const Header2 = () => (
  <>
    <SystemIcon icon={/* ... */} />
    <h3>My Header</h3>
  </>
)

<Modal header={Header} /* ... */ />
```

### LINKED ISSUE #694: Modal is not sending focus back to the target button, in responsive

### LINKED ISSUE #756: Modal - allow aria-labeldBy prop to be passed
## 🚀 Feature Proposal

We are using Modal with a custom title, and we need to pass down aria-labeldBy to the dialog.
```
<Modal aria-labeldby="title">
  <h1 id="title">Title</h1>
</Modal>
```

Currently the modal accepts only `heading` prop which will create a `h3` title, and `aria-label`, but we are missing aria-labeldby to target our custom title.

### LINKED ISSUE #962: Modal: Convert to Compound Component
## 🚀 Feature Proposal

Evolve modal to the compound component pattern (CCP). 

## Motivation

Modal is one of the most requested components for variants, evolving the modal to support the CCP will allow for increased flexibility and usability of the CK Modal component

## Related Issues

#905 feat(modal): Support scrolling entire modal
#756 Modal - allow aria-labeldBy prop to be passed
#786 Modal - full width and full height support
#681 More Modal variants
#790 Modal: Support scrolling entire modal / overlay for overflowing content.

NicholasBoll:
- Add Dialog component
- Update Modal documentation
- Fixes #962
- Fixes #756
- Fixes #694
- Fixes #493
- Fixes #398
- Fixes #56


modules/react/modal/lib/ModalHeading.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export interface ModalHeadingProps extends ExtractProps<typeof Popup.Heading> {}
+
+export const ModalHeading = createComponent('h2')({
+  displayName: 'Modal.Heading',
+  Component: ({children, model, ...elemProps}: ModalHeadingProps, ref, Element) => {
+    const localModel = useModelContext(PopupModelContext, model);
+
+    const props = useModalHeading(localModel, elemProps, ref);
+    console.log('props', ref, props);
```
NicholasBoll, line 16: Remove

modules/react/modal/lib/ModalOverlay.tsx
```diff
@@ -0,0 +1,109 @@
...
+const OpenModalOverlay = createComponent('div')({
+  displayName: 'Modal.OpenOverlay',
+  Component: ({model, children, ...elemProps}: ModalOverlayProps, ref, Element) => {
+    const props = useModalOverlay(model!, elemProps, ref);
+    const windowSize = useWindowSize();
+
+    const content = (
+      <Container as={Element} {...props}>
+        <CenteringContainer
+          style={{width: windowSize.width % 2 === 1 ? 'calc(100vw - 1px)' : '100vw'}}
```
willklein, line 95: Would it be worth breaking this style out into a named object to succinctly describe its purpose?

jamesfan (approved):

modules/react/modal/lib/ModalOverlay.tsx
NicholasBoll: Added a comment

modules/react/modal/lib/hooks/useCloseOnOverlayClick.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+export const useCloseOnOverlayClick = (model: PopupModel, elemProps = {}) => {
+  const onClick = React.useCallback(
+    (event: MouseEvent) => {
+      if (!model.state.stackRef.current) {
+        return;
+      }
+      const elements = PopupStack.getElements()
+        .sort((a, b) => Number(a.style.zIndex) - Number(b.style.zIndex))
+        .filter(e => e.getAttribute('data-behavior-click-outside-close') === 'topmost');
```
willklein, line 19: Minor/optional: consider filtering before sorting.

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -970,6 +971,139 @@ const MyPopup = () => {
...
+
+Most notably, `Modal` is now a container component that takes a `ModalModel` and has several
+subcomponents. `Modal` looks much like the structure of [Popups](#popups), except `Modal` has a
+`Modal.Overlay` subcomponent instead of a `Popup.Popper` component. The `Modal.Overlay` is the
+component in charge of adding an element to the `PopupStack`.
+
+v5 introduced focus management behaviors like `useInitialFocus` and `useReturnFocus` that was
+originally hard-coded into the `Modal`, but now `Modal` uses the hooks from the `popup` module. We
+noticed some application code that do custom focus management. Most of the special focus management
+code could be removed when using v5 the `Modal`.
```
jamesfan, line 1062: Is this supposed to say `v4 introduced focus management behaviors...`?

I'm going to suggest some alternate wording for this paragraph, but I won't use the GH suggestion feature in case I'm getting any details wrong - you can use however much of this makes sense to you.

```
v4 introduced focus management behaviors such as `useInitialFocus` and `useReturnFocus` that
were hard-coded into the `Modal`. In v5, `Modal` uses hooks from the `@workday/canvas-kit-react/popup`
module instead. These behaviors are no longer hard-coded and can be customized by passing
your own `PopupModel` to the `Modal` container component (see the [handleClose example](#handleclose)
below).

During our investigation of how `Modal` was currently being used, we noticed several instances
where custom focus management was necessary. Most of this special focus management should
no longer be necessary after migrating to v5.
```

NicholasBoll: `v5` introduced the hooks available for export. The functionality was internal to Modal before that. In `v5`, the functionality was generalized to work for any popup.

NicholasBoll: I'm planning to open another PR with doc updates. I wanted the PR mergable just in case while I worked more on docs.

PR #1078, opened 2021-05-25
feat: Remove Preview/Tokens & Upgrade CKR Labs Type Hierarchy

alanbsmith:
## Summary

Resolves #311 

- **Remove preview-react/tokens** 
  - Move Labs Type tokens from preview-react/tokens to tokens
  - Move Labs StaticStates from preview-react/tokens to common
  - Create a codemod to migrate imports
    - run codemod against canvas-kit to update all instances
- **Update the type tokens in react/tokens**
  - update the type hierarchy structure
  - add type properties
  - reduce and simplify variants
  - update `Hyperlink` component to contain deprecate `link` variant styles
  - add codemod to migrate legacy and beta type + variants
  - run codemod against canvas-kit to update all instances
  - Update migration guide and tokens documentation

## Checklist

- [ ] branch has been rebased on the latest master commit
- [ ] tests are changed or added
- [ ] `yarn test` passes
- [ ] all (dev)dependencies that the module needs is added to its `package.json`
- [ ] code has been documented and, if applicable, usage described in README.md
- [ ] module has been added to `canvas-kit-react` and/or `canvas-kit-css` universal modules, if
      applicable
- [ ] design approved final implementation
- [ ] a11y approved final implementation
- [ ] code adheres to the [API & Pattern guidelines](../API_PATTERN_GUIDELINES.md)

PR #1096, opened 2021-06-08
fix: Type cleanup

anicholls:
A few follow-up cleanup items on #1078 to prep us for the v5 release!

- [x] Corrects some of the manual migration done to the tokens stories in #1078 
- [x] Fixes missing logic in type codemod that was preventing the font weight from being preserved for h3 and h4 headings that were specifying the type object in a `styled()` argument
- [x] Run codemod update (or manually update the issues)
- [x] BREAKING: Removes unintended `0.01rem` letter spacing from `body.m` and `body.l`
- [x] Upsize toast text to 14px because 12px was too small


anicholls:
Moved these fixes to #1098

PR #1110, opened 2021-06-12
fix(tokens): Fixes color token exports

### LINKED ISSUE #1109: TS Error with Color Tokens
## 🐛 Bug Report

The problem is [here](https://github.com/Workday/canvas-kit/blob/master/modules/react/tokens/index.ts#L25).
TLDR: The semantic colors has a colors key, and we’re exporting the default of colors as colors so it’s always being overwritten.

## To Reproduce

* Using TS ^3.9.9
* Look at `modules/react/tokens/index.ts`
* See the error

## Expected Behavior

Color tokens shouldn't be overwriting one another

## Actual Results

TS is angry.

## Browser (if applicable)

n/a

n/a

### Error Output

```bash
27   colors,
node_modules/@workday/canvas-kit-react/tokens/index.ts:27:3 - error TS2783: 'colors' is specified more than once, so this usage will be overwritten.
33   ...semanticColors,
node_modules/@workday/canvas-kit-react/tokens/index.ts:33:3
This spread always overwrites this property.
```

or

alanbsmith:
## Summary

Resolves #1109

Previously we were parsing out colors from `canvas-colors-web` like this:

```ts
const {default: colors, gradients, ...semanticColors} = canvasColorsWeb;
```

And exporting them from a default `canvas` object like this:

```ts
const canvas = {
  colors,
  depth,
  space,
  type,
  fontFamily,
  monoFontFamily,
  ...semanticColors,
};
```

TS throws an error because `...semanticColors` has a `color` key which is overwrites the default export. This happened to work because the colors in the default export and the `colors` key were identical.

To resolve this, we're removing the default export and explicitly import each color key.

## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/react/tokens/index.ts
+++ b/modules/react/tokens/index.ts
@@ -1,4 +1,15 @@
-import * as canvasColorsWeb from '@workday/canvas-colors-web';
+import {
+  buttonColors,
+  chartingColorOffsets,
+  chartingColors,
+  colors,
+  commonColors,
+  gradients,
+  iconColors,
+  inputColors,
+  statusColors,
+  typeColors,
+} from '@workday/canvas-colors-web';
 
 import {borderRadius, CanvasBorderRadius, CanvasBorderRadiusKeys, CanvasBorderRadiusValues} from './lib/radius';
 import {BrandingColor, CanvasColor} from './lib/colors.types';
@@ -22,15 +33,21 @@ import {type,
 } from './lib/type';
 import {CSSProperties} from './lib/types';
 
-const {default: colors, gradients, ...semanticColors} = canvasColorsWeb;
 const canvas = {
+  buttonColors,
+  chartingColorOffsets,
+  chartingColors,
   colors,
+  commonColors,
   depth,
-  space,
-  type,
   fontFamily,
+  iconColors,
+  inputColors,
   monoFontFamily,
-  ...semanticColors,
+  space,
+  statusColors,
+  type,
+  typeColors,
 };
 
 export * from '@workday/canvas-colors-web';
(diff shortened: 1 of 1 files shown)


modules/react/tokens/index.ts
```diff
@@ -22,15 +33,21 @@ import {type,
...
-  type,
   fontFamily,
+  iconColors,
+  inputColors,
   monoFontFamily,
-  ...semanticColors,
+  space,
+  statusColors,
+  type,
+  typeColors,
```
alanbsmith, line 50: I also alphabetized this list to make it easier to find things.

modules/react/tokens/index.ts
```diff
@@ -1,4 +1,15 @@
...
+  chartingColorOffsets,
+  chartingColors,
+  colors,
+  commonColors,
+  gradients,
+  iconColors,
+  inputColors,
+  statusColors,
+  typeColors,
+} from '@workday/canvas-colors-web';
```
alanbsmith, line 12: It seemed odd to me that `gradients` is the only color key excluded from tokens `canvas` default export. Was this intentional? Was it a mistake? I left it out of the export, but we could add it back without it being a breaking change.

alanbsmith: E.g. Why is `canvas.chartingColorOffsets` a thing, but not `canvas.gradients`?

anicholls: Yeah, I'd say it's worth adding 👍  Not sure why it got left out

anicholls (approved):

PR #1115, opened 2021-06-16
fix(popup): Remove ref fowarding requirement in Popup.Target

### LINKED ISSUE #1114: Remove the requirement for forwarding refs for custom popup targets
## Motivation

It is common to use custom target elements for `Popup`, `Dialog`, and `Modal`. But for correct accessibility, ref forwarding is required which is not very common outside of library code. The required `ref` can instead be set using `onClick`. The only reason we require ref forwarding now is for the edge case where the application programmatically open a Popup without the user interacting first. In this case, without the `ref`, the popup appears at the top-left of the screen since PopperJS doesn't have a reference to an element to appear around.

## Example

```tsx
const MyTarget = ({...props}: React.HTMLButtonAttributes<HTMLButtonElement>) => {
  <button {...props}>Open</button>
}

const App = () => {
  return (
    <Modal>
      <Modal.Target as={MyTarget} />
      ...
    </Modal>
  )
}
```

Without this feature, `MyTarget` would have to use `React.forwardRef`

NicholasBoll:
- Remove ref forwarding requirement in `Popup.Target`, `Modal.Target`, and `Dialog.Target`
- Add examples to show using custom targets for `Popup` and `Modal`
- Add Cypress specifications for new examples

Fixes #1114

### DIFF
--- a/modules/react/modal/stories/examples/CustomTarget.tsx
+++ b/modules/react/modal/stories/examples/CustomTarget.tsx
@@ -1,0 +1,26 @@
+import React from 'react';
+
+import {Modal} from '@workday/canvas-kit-react/modal';
+
+interface MyTargetProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  label: string;
+}
+
+const MyTarget = ({label, ...props}: MyTargetProps) => {
+  return <button {...props}>{label}</button>;
+};
+
+export const CustomTarget = () => {
+  return (
+    <Modal>
+      <Modal.Target as={MyTarget} label="Open" />
+      <Modal.Overlay>
+        <Modal.Card>
+          <Modal.CloseIcon aria-label="Close" />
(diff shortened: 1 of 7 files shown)


modules/react/modal/stories/examples/CustomTarget.tsx
```diff
@@ -0,0 +1,26 @@
...
+      <Modal.Overlay>
+        <Modal.Card>
+          <Modal.CloseIcon aria-label="Close" />
+          <Modal.Heading>Modal</Modal.Heading>
+          <Modal.Body>Contents</Modal.Body>
+        </Modal.Card>
+      </Modal.Overlay>
+    </Modal>
+  );
+};
```
mannycarrera4, line 26: YASSS thank you!

mannycarrera4 (approved):
Thanks for working through this with me and coming up with a solution!

PR #1117, opened 2021-06-16
Popup adapter transfer ownership

csongnguyen:
## Summary

Previously, if a new adapter is used and there happened to be existing popups on the page, the new adapter loses reference of those existing popups and has no way to interact with them. This fix allows a transfer of ownership from the old adapter to the new one.

## Checklist

- [x] `yarn test` passes
      applicable


csongnguyen:
Coming back to this.... will attempt implementing this in-house rather than on the API itself.

PR #1127, opened 2021-06-25
docs: Convert Buttons category stories to use mdx

anicholls:


anicholls:
The Cypress tests are failing because the delete button story is not loading. I can't figure out why that story's URL has a `-story` suffix, when none of the others do (see [here](https://5d854c26ba934e0020f5e98a-htppkwlbge.chromatic.com/?path=/story/components-buttons-button-react--delete-story)). I don't see how I'm treating that story any differently in `Buttons.stories.mdx`

jamesfan:
> The Cypress tests are failing because the delete button story is not loading. It's failing because the URL generated by the cypress helpers does not match the story. I can't figure out why that story's URL has a `-story` suffix, when none of the others do (see [here](https://5d854c26ba934e0020f5e98a-htppkwlbge.chromatic.com/?path=/story/components-buttons-button-react--delete-story)). I don't see how I'm treating that story any differently in `Buttons.stories.mdx`. Anyone have ideas?

I think this has something to do with the story name (`Delete`) being a reserved keyword in JavaScript. I noticed the same thing in [my PR](https://github.com/Workday/canvas-kit/pull/1128) for the Input docs. Storybook is adding a `-story` suffix to all of the `Default` stories (e.g., [`Default` Checkbox story](https://5d854c26ba934e0020f5e98a-xfghbacnjj.chromatic.com/?path=/story/components-inputs-checkbox-react--default-story))

If you replace [L61](https://github.com/Workday/canvas-kit/pull/1127/files#diff-e649fc48ed6e44b1341277ac90521fc9ec5e8bf100d0f9f77518f61641f85b3eR61) on `Button.stories.mdx`:

```
<ExampleCodeBlock code={Delete} />
```

with:

```
<Canvas>
  <Story
    name="Delete"
    parameters={{ storySource: { source: Delete.__RAW__ } }}
  >
    <Delete />
  </Story>
</Canvas>

// also, add Canvas and Story to your import from '@storybook/addon-docs/blocks'
```

(which is what the `mdx-code-block-rewrite` loader is doing) and change `name="Delete"` to `name="ANY_RESERVED_WORD_IN_JS"`, you'll see it always adds `-story` to the end.

Haven't figured out a way around this yet.

anicholls:
@jamesfan Makes sense - good find. I will try to update the cypress helper to account for this 👍

cypress/integration/Button.spec.ts
```diff
@@ -17,13 +17,13 @@ describe('Button', () => {
...
         .first()
-        .should('contain', 'Primary');
+        .should('contain', 'Click Me');
     });
   });
 
   context('given delete buttons are rendered', () => {
     beforeEach(() => {
-      h.stories.load('Components/Buttons/Button/React', 'Delete');
+      h.stories.load('Components/Buttons/Button/React', 'DeleteStory');
```
anicholls, line 26: @jamesfan had a [good writeup](https://github.com/Workday/canvas-kit/pull/1127#issuecomment-869947485) of why this is necessary. After doing some digging in the Storybook [source code](https://github.com/storybookjs/storybook/blob/master/addons/docs/src/mdx/mdx-compiler-plugin.js#L23) it looks like they sanitize/rename stories if the name of the story 1. starts with a number, or 2. uses a reserved name in javascript (e.g. `delete`, `default`, `const`, etc.). I started to put some fixes into [`h.stories.load`](https://github.com/Workday/canvas-kit/blob/master/cypress/helpers/stories.ts#L10), but I quickly realized this is probably something that should go in [`cypress-storybook`](https://github.com/NicholasBoll/cypress-storybook).

@NicholasBoll I started to work on a PR in `cypress-storybook` for this, but ran into some issues due to how Storybook treats CSF vs. MDX stories. Unfortunately it seems like Storybook treats these two inconsistently:

For CSF stories:
- Stories with reserved keyword names (e.g. `Delete`) are not touched (URL is unaffected)
- Stories with number prefixed names have `Story` appended (resulting in `-story` to be appended in the URL)

For MDX stories:
- Stories with reserved keyword names (e.g. `Delete`) have `Story` appended (resulting in `story` being appended to the URL)
- Stories with number prefixed names have `_` prepended. 

I started to add logic similar to https://github.com/storybookjs/storybook/blob/master/addons/docs/src/mdx/mdx-compiler-plugin.js#L23 to [`loadStory`](https://github.com/NicholasBoll/cypress-storybook/blob/411184e81a6555ea1d6cac6572408aa45c8e5fa7/cypress.js#L7). However, in Storybook this logic is only run for MDX, so I realized that this would cause problems due to the inconsistency of treatment between CSF & MDX. AFAIK, `cypress-storybook` has no way to know whether the story is MDX or CSF based, so I got stuck. Pretty frustrating on Storybook's part, to be honest. 

I decided to just append `Story` in this instance for now and we can address this in `cypress-storybook` once we figure out the correct approach.

modules/react/button/stories/IconButton.stories.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-react
+```
+
+## Usage
+
+Here is a basic example of a Primary Button. Buttons can be customized with several props to change
```
jamesfan, line 31:
```suggestion
Here is a basic example of an Icon Button. Buttons can be customized with several props to change
```

modules/react/segmented-control/stories/SegmentedControl.stories.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+<ExampleCodeBlock code={Default} />
+
+## Properties
+
+<ArgsTable of={SegmentedControl} />
+
+## Specifications
+
+<Specifications file="SegmentedControl.spec.ts" name="SegmentedControl" />
```
jamesfan, line 37: No SegmentedControl.spec.ts file, so this isn't rendering anything.

jamesfan (approved):
I'm guessing you're still incorporating changes based on the [MDX Standardization doc](https://github.com/Workday/canvas-kit/discussions/1131), but I went ahead and added a few comments below.

A couple additional notes:

- I wiped out the bulk of the README content in my [Input MDX PR](https://github.com/Workday/canvas-kit/pull/1128). Each README now only contains a link to the Storybook docs and a link to the corresponding page on the current Design Site (if one exists). Should probably do the same for the Buttons READMEs.
- We had different approaches for organizing our `stories` folders for modules with multiple components (for me it was Color Input and Color Preview in the `color-picker` module). See [this comment](https://github.com/Workday/canvas-kit/pull/1128#discussion_r662777951). You created sub-folders in `stories/examples` whereas I created sub-folders one level up in `stories`. Not sure if this matters too much from a technical level, but might be nice if we were consistent. Feel free to move files around in my PR if you want to stick to your approach.

anicholls:
@jamesfan 

> As you noticed, I wiped out the bulk of the README content in my Input MDX PR. Each README now only contains a link to the Storybook docs and a link to the corresponding page on the current Design Site (if one exists). Should probably do the same for the Button READMEs.

I was hesitant to do this until the prop tables are fixed. If I removed the READMEs right now, there would be undocumented props. Going to leave them in for now. 

---

> We had different approaches for organizing our stories folders for modules with multiple components (for me it was Color Input and Color Preview in the color-picker module). See this comment. You created sub-folders in stories/examples whereas I created sub-folders one level up in stories. Not sure how much this matters from a technical perspective, but it would probably be nice if we were consistent. Feel free to move files around in my PR if you want to stick to your approach.

I've reorganized the stories to follow a similar pattern. I've used `icon-button` and `button`. Technically each of the components in `button` (e.g. `PrimaryButton`, `SecondaryButton`, etc.) are their own component, but I did it to align with the MDX files.

Issue #1126, opened 2021-06-25
createComponent components no longer work with react-docgen-typescript


anicholls:
## 🐛 Bug Report

We're starting to rely more heavily on MDX (and automatically generated prop tables). At some point in the development of v5, the compatibility with react-docgen-typescript has been broken. Now, the prop table only displays the props that `createComponent` accounts for (`ref` and `as`)

## To Reproduce

Render a component in the prop table (e.g. `<ArgsTable of={IconButton} />`)

## Expected Behavior

All props should be rendered.

## Actual Results

Only `ref` and `as` are listed.

PR #1128, opened 2021-06-26
docs: Add MDX docs and examples for Input components

jamesfan:
## Summary

Migrates MDX docs and code examples for the following Input components from the Canvas Site:

- Checkbox
- Form Field
- Radio
- Select
- Switch
- Text Area
- Text Input

Also adds MDX docs with code examples for the following Input components:

- Color Input
- Color Preview

I've removed most of the content from the READMEs for these components -- they now consist only of a link to the Storybook documentation and a link to the corresponding component page on the Design Site (if it exists).


modules/react/checkbox/README.md
```diff
@@ -1,133 +1,7 @@
...
-#### Accessible Example
-
-```tsx
-import * as React from 'react';
-import Checkbox from '@workday/canvas-kit-react/checkbox';
-import FormField from '@workday/canvas-kit-react/form-field';
-
-<FormField label="My Field" inputId="my-checkbox-field">
-  <Checkbox disabled={false} checked={checked} onChange={this.handleCheck} id="my-checkbox-field" />;
-</FormField>;
```
anicholls, line 37: Any reason why we dropped this in the MDX docs? All of the other inputs note `FormField` usage for a11y.

modules/react/checkbox/README.md
```diff
@@ -1,133 +1,7 @@
...
-| Type  | Description                     |
-| ----- | ------------------------------- |
-| Error | Red outline with error icon.    |
-| Alert | Yellow outline with alert icon. |
-
-Default: `undefined`
-
----
-
-#### `indeterminate: boolean`
```
anicholls, line 128: We should have an indeterminate example.

modules/react/radio/README.md
```diff
@@ -1,211 +1,7 @@
 # Canvas Kit Radio
 
-Input of type radio
-
-Coming soon:
-
-- Error/Alert handling
```
anicholls, line 7: LOL this is pretty old 😅

modules/react/radio/stories/examples/Error.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import {FormField} from '@workday/canvas-kit-react/form-field';
+import {Radio, RadioGroup} from '@workday/canvas-kit-react/radio';
+
+export const Error = () => {
+  const [value, setValue] = React.useState<string | number>('thin');
```
anicholls, line 6: Nit: I feel like this should be set to `deep-dish` (same as the alert example).

```suggestion
  const [value, setValue] = React.useState<string | number>('deep-dish');
```

modules/react/select/stories/Select.stories.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+Set the `error` prop of the wrapping Form Field to `FormField.ErrorType.Alert` and provide an
+explanation using the `hintText` prop to set the Select to the Alert state.
+
+<ExampleCodeBlock code={Alert} />
+
+The Alert state may be applied directly to the Select using the static `Select.ErrorType.Alert`
+value if Form Field is not being used.
+
+<ExampleCodeBlock code={AlertStandalone} />
```
anicholls, line 101: I'm not sure we need all of these standalone stories for props that are available on both `FormField` and the base component. They make the sidebar quite busy. I think the sentence above is enough and it doesn't really require an extra example.

```suggestion
The Alert state may be applied directly to the `Select` component directly using the static `Select.ErrorType.Alert`
value if `FormField` is not being used.

<ExampleCodeBlock code={AlertStandalone} />
```

modules/react/select/stories/Select.stories.mdx
```diff
@@ -0,0 +1,113 @@
...
+import {LabelPosition} from './examples/LabelPosition';
+import {RefForwarding} from './examples/RefForwarding';
+import {Required} from './examples/Required';
+import {Standalone} from './examples/Standalone';
+
+<Meta title="Components/Inputs/Select/React" component={Select} />
+
+# Canvas Kit Select
+
+Select Drop-downs allow users to choose one option from a list of items in a Drop-down Menu.
```
anicholls, line 23: Since the majority of people are starting to use the Select component in preview, it would be nice to link to it here (and maybe in the readme).

anicholls (commented):
Can you add prop tables so they'll appear when https://github.com/Workday/canvas-kit/issues/1126 is resolved? It seems like you've leaned toward creating an example for each prop. We should probably discuss our approach for examples as a team. 

We've lost the static properties documentation with this change. Do you think it's okay to just leave it to the examples to illustrate those? I would say so (since we're moving towards string literals from enums), but I just wanted to call it out.

modules/react/text-input/stories/TextInput.stories.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-react
+```
+
+## Default Example
+
+Text Input should be used in tandem with Form Field to meet accessibility standards.
```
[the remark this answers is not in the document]
jamesfan: Sure, I've linked "Form Field" throughout the MDX docs to the Form Field Docs tab in Storybook (just the first occurrence of Form Field in each article, though).

I didn't convert "Form Field" to `FormField` yet since it seems like we would then need to convert all component mentions to the `code` format throughout the docs (e.g., "Text Input" to `TextInput`). But maybe we should establish this as a convention now? Alan seems to have done it for [Box](https://workday.github.io/canvas-kit/?path=/docs/labs-common-box--as) and the [Layout components](https://workday.github.io/canvas-kit/?path=/docs/labs-layout-flex--flex-card).

modules/react/checkbox/README.md
jamesfan: I left it out as I was rewriting the Checkbox docs on `canvas-site` since it seemed redundant to have a FormField on a Checkbox given that it already has a label. Meant to look into this further, but haven't had time -- I've gone ahead and added FormField back to the Checkbox examples.

I didn't add it to the [Indeterminate example](https://5d854c26ba934e0020f5e98a-duglscczpo.chromatic.com/?path=/docs/components-inputs-checkbox-react--indeterminate), though, because I couldn't think of a good label for my example with the pizza toppings. Any ideas?

modules/react/checkbox/README.md
jamesfan: This one slipped through the cracks as I as migrating docs. I've added it back in.

modules/react/select/stories/Select.stories.mdx
jamesfan: Agreed, I've removed all Standalone stories (`Standalone`, `Alert Standalone`, `Error Standalone`, etc.) throughout all the Input component MDXes.

modules/react/select/stories/Select.stories.mdx
jamesfan: Done. I've added it to [the top of the Select docs](https://5d854c26ba934e0020f5e98a-duglscczpo.chromatic.com/?path=/docs/components-inputs-select-react--default-story) -- let me know if you have any suggestions on wording.

jamesfan:
@anicholls I've added `ArgsTables` to the components and addressed most of your remaining feedback (follow-up questions are in the original threads).

> It seems like you've leaned toward creating an example for each prop. We should probably discuss our approach for examples as a team.

Agreed.

> We've lost the static properties documentation with this change. Do you think it's okay to just leave it to the examples to illustrate those? I would say so (since we're moving towards string literals from enums), but I just wanted to call it out.

I think it's ok to leave it to the examples, too.

---

How strongly do people feel about preserving how stories for Input components are currently organized in Storybook?

Right now, every Input has a `Left Label` section and a `Top Label` section with the same stories displayed in each section (the only difference is the `labelPosition` passed into `FormField`). Here's what Text Input looks like, for example:

old-text-input-stories

I didn't organize my examples with this hierarchy in mind while writing the updated documentation on `canvas-site`, so I ended up with this structure when I migrated the docs for Input components over:

new-text-input-stories

Thoughts on the new structure? It's a bit quicker to navigate since there's one less level to drill down. We're no longer duplicating stories between `Left Label` and `Top Label`, but I think that's ok?

If we end up keeping this new structure, I'll need to update the Cypress tests to work with the new story paths, as well as the Storybook links on the current Design Site.

modules/react/text-input/stories/TextInput.stories.mdx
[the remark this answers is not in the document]
anicholls: Hmm that's a good point. In the context of Storybook, the code formatting makes sense, but maybe not as much on the site...

modules/react/checkbox/README.md
anicholls: Good point. The checkbox label has a `for` attribute, so it should be fine a11y-wise. Maybe let's leave it out for all the examples, but either mention that it can be used with FormField or provide an example for it?

anicholls:
Can you update the ColorPreview and ColorInput files with similar changes (particularly with removing the top/left label categorization)

modules/react/checkbox/README.md
jamesfan: Done. I've included a mention that Checkbox can be used with or without Form Field, and there are a few examples which use Form Field: Label Position, Required, and the Error States.

jamesfan (commented):
@anicholls I've addressed all of your remaining feedback, including adding MDXes and examples for Form Field, Color Input, and Color Preview (with no more Top/Left Label categorizations). I left a few minor callouts below, but nothing which should block this PR -- everything should be good to merge.

modules/react/form-field/README.md
```diff
@@ -1,233 +1,5 @@
...
-> required for accessibility on radio groups. If you're using a `RadioGroup` inside of FormField,
-> make sure you set this to true.
-
-Default: `false`
-
----
-
-# Hint
-
-> Component implementing hint text styling for forms.
```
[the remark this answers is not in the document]
anicholls: Yeah it's more of an internal subcomponent - so this is the right call 👍

anicholls (approved):

PR #1132, opened 2021-07-01
fix(tooltip): Fix overflow ellipsis detection

### LINKED ISSUE #1125: OverflowTooltip not displayed for TextButton with an icon
## 🐛 Bug Report

The OverflowTooltip fails to display when wrapping a TextButton with an icon.

## To Reproduce

[Reproduction Codesandbox](https://codesandbox.io/s/overflow-tooltip-textbutton-icon-68qzi?file=/src/App.js).

Mouse over the first TextButton (no icon). Note how the tooltip displays properly. Now, mouse over the second TextButton with the icon (`<TextButton icon={resetIcon}>`) -- no tooltip is displayed.

overflow-tooltip

## Expected Behavior

A tooltip of `Text Button with OverflowTooltip` should also be displayed when mousing over the second TextButton with the icon.

NicholasBoll:
Fixes #1125

The intent of the overflow tooltip is to first find an ellipsis element, then fallback to finding an overflow element if no ellipsis element was found. The `findEllipsisElement` is intended to be a recursive function, but accidentally called `findOverflowElement` instead of `findEllipsisElement`.

### DIFF
--- a/modules/tooltip/react/lib/OverflowTooltip.tsx
+++ b/modules/tooltip/react/lib/OverflowTooltip.tsx
@@ -8,7 +8,7 @@ import {TooltipContainer} from './TooltipContainer';
 import {useTooltip} from './useTooltip';
 
 /**
- * Look for an element within the tree for a `text-overflow` CSS property of `ellipsis`.
+ * Look for an element within the tree for an overflow element (auto, scroll, clip, or hidden).
  * This could be the passed element, or a descendant. If no element is found, `null` is returned.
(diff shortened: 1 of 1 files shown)


jamesfan (approved):
Fix looks good 👍.

Should probably move [this comment](https://github.com/Workday/canvas-kit/blob/master/modules/react/tooltip/lib/OverflowTooltip.tsx#L11) to `findEllipsisElement ` as well.

PR #1135, opened 2021-07-07
fix(modal): Fix focus trap when Modal contains an iframe

### LINKED ISSUE #1097: Focus Trap does not work with iframes
## 🐛 Bug Report

focus-trap-js library does not support iframes.
it was designed for maximum performance, it is used in workday grids 2.0, however supporting iframes will make it less performant which is not desirable. 

We can replace it with react-focus-lock instead, which supports iframes
https://www.npmjs.com/package/react-focus-lock

## To Reproduce

Iframe within modal
https://codesandbox.io/s/holy-architecture-qrl1b

## Expected Behavior

Focus should be trapped

NicholasBoll:
Fixes #1097

The library code has been copied locally using the code from https://github.com/alexandrzavalii/focus-trap-js/pull/11

This is fine for v4, but will require some rework in v5. v5 introduces `useFocusRedirect` with similar logic. It makes sense to move code from the `focus-trap-js` library into the common module that can handle both focus trapping and focus redirecting.

### DIFF
--- a/cypress/integration/Modal.spec.ts
+++ b/cypress/integration/Modal.spec.ts
@@ -620,4 +620,49 @@ describe('Modal', () => {
       });
     });
   });
+
+  context(`given the 'Iframe Test' story is rendered`, () => {
+    beforeEach(() => {
+      h.stories.load('Testing/React/Popups/Modal', 'Iframe Test');
+    });
+
+    context('when the modal is opened', () => {
+      beforeEach(() => {
+        cy.contains('button', 'Delete Item').click();
+      });
+
+      context('when Shift + Tab key is pressed', () => {
+        beforeEach(() => {
+          cy.focused().tab({shift: true});
+        });
+
+        it('should focus in the iframe', () => {
+          cy.get('iframe').should('have.focus');
+        });
+
+        it('should focus on the last button in the iframe', () => {
+          cy.get('iframe')
+            .its('0.contentDocument.body')
+            .then(cy.wrap)
+            .contains('button', 'iframe button 2')
+            .should('have.focus');
+        });
+
+        // skipping because the cy.tab plugin isn't capable of starting inside an iframe. We have to test this manually
+        context.skip('when the Tab key is pressed', () => {
+          beforeEach(() => {
+            cy.get('iframe')
+              .its('0.contentDocument.body')
+              .then(cy.wrap)
+              .focused()
+              .tab();
+          });
+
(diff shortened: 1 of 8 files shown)


cypress/integration/Modal.spec.ts
```diff
@@ -620,4 +620,49 @@ describe('Modal', () => {
...
+        it('should focus on the last button in the iframe', () => {
+          cy.get('iframe')
+            .its('0.contentDocument.body')
+            .then(cy.wrap)
+            .contains('button', 'iframe button 2')
+            .should('have.focus');
+        });
+
+        // skipping because the cy.tab plugin isn't capable of starting inside an iframe. We have to test this manually
+        context.skip('when the Tab key is pressed', () => {
```
NicholasBoll, line 652: Perhaps there's a way to get around this in a future version of `cypress-plugin-tab`, but skipping for now.

cypress/support/commands.ts
```diff
@@ -16,7 +16,7 @@ Cypress.Commands.add('injectAxe', () => {
 });
 
 // Add better logging to cy.tab
-Cypress.Commands.overwrite('tab', (originalFn, subject) => {
+Cypress.Commands.overwrite('tab', (originalFn, subject, options) => {
```
NicholasBoll, line 19: This is fixed in v5, but I could not Shift + Tab without it

modules/docs/utils/get-specifications.js
```diff
@@ -35,6 +35,7 @@ function getSpecifications() {
               cb();
               children = childrenBefore;
             };
+            describe.skip = noop;
```
NicholasBoll, line 38: Added for the `context.skip`. Without this line, Storybook would fail to build

modules/popup/react/lib/focus-trap-js.js
```diff
@@ -0,0 +1,122 @@
+/* eslint-disable */
+// refactor for v5
```
NicholasBoll, line 2: This is the `focus-trap-js` library inlined into v4 with changes included from https://github.com/alexandrzavalii/focus-trap-js/pull/11

NicholasBoll: v5 introduces `useFocusRedirect` which hard-codes similar logic from this library, but without the hard-coded focus trap logic. This should all be refactored for v5

alanbsmith (approved):
🐼

PR #1136, opened 2021-07-08
feat: Add storybook utils to labs common

vibdev:
## Summary

CKR has some really great internal tools for showing variations of components in Storybook. 

One of the great things about CKR is being able to build up or theme our own components based on base CKR components. 

However, it is arduous to test our variations. 

Well good news! Now you can get your very own storybook utils right here in CKR Labs Common.

## Checklist

- [x] `yarn test` passes


modules/labs-react/common/lib/storybook-utils/ComponentStatesTable.tsx
```diff
@@ -0,0 +1,79 @@
...
+    paddingRight: 16,
+    textAlign: 'left',
+  },
+});
+
+export const ComponentStatesTable = ({
+  rowProps,
+  columnProps,
+  children,
+}: ComponentStatesTableProps) => {
```
vibdev, line 45: Changed this to a FC from a class. Also dropped the default props, I couldn't find any module actually using them, so figured it was better to just make them required.

modules/labs-react/common/lib/storybook-utils/permutateProps.ts
```diff
@@ -0,0 +1,45 @@
+export const permutateProps = (
```
vibdev, line 1: split this out to its own file

utils/spec/tsconfig.json
```diff
@@ -1,3 +0,0 @@
-{
-  "extends": "../../tsconfig.spec.json"
```
anicholls, line 2: Any reason you removed this guy?

vibdev: I removed the folder since there were no specs left

anicholls: Hah - thanks 😅

NicholasBoll (approved):
The only weird thing is seeing these components imported from a `common` location. A component state table could definitely be used in a production setting, but most likely would be used for testing. I wonder if `test-utils` or something would be a more appropriate location...

anicholls (approved):

anicholls:
@NicholasBoll You mean adding a new sub-folder inside the labs module? I'm fine either way.

vibdev:
I think common was chosen to align with the StaticStates export.

NicholasBoll:
> @NicholasBoll You mean adding a new sub-folder inside the labs module? I'm fine either way.

Yes, but I'm fine if it is in common. I'm not sure we meant to export `StaticStates` in common, but I'm if it's there. It just looks weird if I'm importing and in the autocomplete is Storybook utility components and functions.

PR #1147, opened 2021-07-09
feat(common): Add HTML attribute interface to ExtractProps

### LINKED ISSUE #1145: When extending component props for an interface, the consumer has to redefine the props based on the element.
## 🐛 Bug Report

In `v4`  as a consumer I could extend a component props and add on to it.

```tsx
   import {TextInputProps} from '@workday/canvas-kit-react-text-input'
   export interface MyComponentProps extends TextInputProps {
      myExtraProp: boolean
   }
```

Where TextInputProps has attributes like `placeholder` and `disabled`

In `v5` this is not possible and we have to redefine those props. I believe `createComponent` handle adding those attributes based on the element when it uses the `as` prop however, as a consumer we lose all those props and have to redefine them.

When I try something like this:

```tsx
   import {TextInputProps} from '@workday/canvas-kit-react-text-input'
   export interface MyComponentProps extends TextInputProps {
      myExtraProp: boolean
   }
```

`MyComponent` doesn't have access to the base props for TextInput like disabled or placeholder

It would be great to have those props exported someone instead of redefining them

NicholasBoll:
Fixes #1145 

```tsx
ExtractProps<typeof Card>; // React.HTMLAttributes<HTMLDivElement> & CardProps
ExtractProps<typeof Card, 'aside'>; // React.HTMLAttributes<HTMLElement> & CardProps
ExtractProps<typeof Card, never>; // CardProps
```

Usage:
```tsx
import {ExtractProps} from '@workday/canvas-kit-react/common';
import {TextInput} from '@workday/canvas-kit-react/text-input';

export interface MyComponentProps extends ExtractProps<typeof TextInput> {
  myExtraProp: boolean
}

const MyComponent = ({ myExtraProp, ...props}: MyComponentProps) => {
  // do something with `myExtraProp`
  return <TextInput {...props} />
}
```

### DIFF
--- a/modules/react/common/README.md
+++ b/modules/react/common/README.md
@@ -14,6 +14,9 @@ Includes:
 - [InputProvider](#inputprovider)
 - [Theming](#theming)
 - [Bidirectionality](#bidirectionality)
+- [Component Functions](#component-functions)
+  - [createComponent](#createcomponent)
+  - [ExtractProps](#extractprops)
 
 ## CanvasProvider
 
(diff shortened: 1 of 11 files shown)


modules/react/common/README.md
```diff
@@ -237,3 +240,130 @@ Theming documentation has its own README. You can find it [here](./lib/theming/R
...
+This is used for compound components that do not require an element for accessibility reasons and
+only provide a common model for the subcomponents.
+
+### `ExtractProps`
+
+`ExtractProps` can be used in place of importing a prop interface directly. If you wish to extend a
+Canvas Kit component by extending it's props, use `ExtractProps`.
+
+```tsx
+interface MyNewComponentProps extends ExtractProps<typeof MyComponent> {
```
mannycarrera4, line 296: i would write this as `interface MyNewComponentProps extends ExtractProps<typeof Card> {`

so it's clear that it's a canvas kit component. super nit

NicholasBoll: `MyComponent` is referring to the component defined further up the doc on line 259.

jhensley:
Thanks @NicholasBoll! Appreciate the quick turn around)

PR #1154, opened 2021-07-13
fix: ChangeFocus takes any input and Dialog/Modal accepts model config

NicholasBoll:
- Update `changeFocus` to take any input and explicitly test for `HTMLElement`
- Update `Dialog` and `Modal` types to take optional model config.

### DIFF
--- a/modules/react/common/lib/utils/changeFocus.ts
+++ b/modules/react/common/lib/utils/changeFocus.ts
@@ -2,19 +2,23 @@
  * Programmatically change focus in a way that causes a focus ring around the new element.
  * @param element Element that should be focused
  */
-export const changeFocus = (element: HTMLElement | null) => {
+export const changeFocus = (element: unknown) => {
   // Normally the browser will ignore calls to `focus` on an element that already has focus, but we
   // don't want to accidentally show the focus ring in this case, so we'll detect and bail early.
   // We'll also bail if there is no element
   if (document.activeElement === element || !element) {
     return;
   }
-  // Dispatch an unidentified keyboard event for an input provider prior to a focus change so that
-  // the ring will appear.
-  if (typeof KeyboardEvent === 'function') {
-    const event = new KeyboardEvent('keydown', {bubbles: true, key: 'Unidentified'});
-    element.dispatchEvent(event);
-  }
 
-  element.focus();
+  if (element instanceof HTMLElement) {
+    // Dispatch an unidentified keyboard event for an input provider prior to a focus change so that
+    // the ring will appear.
+    if (typeof KeyboardEvent === 'function') {
+      const event = new KeyboardEvent('keydown', {bubbles: true, key: 'Unidentified'});
+
+      element.dispatchEvent(event);
+    }
+
+    element.focus();
+  }
 };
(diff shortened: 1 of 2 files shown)


modules/react/common/lib/utils/changeFocus.ts
```diff
@@ -2,19 +2,23 @@
  * Programmatically change focus in a way that causes a focus ring around the new element.
  * @param element Element that should be focused
  */
-export const changeFocus = (element: HTMLElement | null) => {
+export const changeFocus = (element: unknown) => {
```
NicholasBoll, line 5: It is possible to be given non-HTMLElement types. The use-case was a non-ref forwarding `MenuItem` component. So the ref was a `MenuItem` class instead of an `HTMLElement`. It is safer to just assume anything will be passed and deal with it in the function.

modules/react/dialog/lib/Dialog.tsx
```diff
@@ -1,13 +1,18 @@
...
+  PopupModel,
+  PopupModelConfig,
+} from '@workday/canvas-kit-react/popup';
 
 import {DialogPopper} from './DialogPopper';
 import {DialogCard} from './DialogCard';
 import {useDialogModel} from './hooks';
 
-export interface DialogProps {
+export interface DialogProps extends PopupModelConfig {
```
NicholasBoll, line 15: Both `Dialog` and `Modal` can receive `PopupModelConfig` and it works at runtime. This just adds the types so that Typescript is happy with it.

modules/react/common/lib/utils/changeFocus.ts
alanbsmith: I refreshed my understanding of the difference between `unknown` and `any`, but is there a reason to opt for it here? I don't have a preference, I'm just curious.

 You're not calling any actions on `element` until after checking that it's an instance of `HTMLElement`. Using `unknown` does ensure you don't get a false positive calling `focus`, and I suppose if someone modified this function later, it would be a little safer. Is that the reasoning here?

alanbsmith (approved):
Thanks! I have one question, but nothing that would block the PR. :shipit:

modules/react/common/lib/utils/changeFocus.ts
NicholasBoll: `unknown` should almost always be preferred over `any`.

The main difference between `unknown` and `any`:

```tsx
const foo: any = null
foo.bar() // totally legal according to Typescript

const foo: unknown = null;
foo.bar(); // Object is of type 'unknown'. ts(2571)

if (typeof foo === 'object') {
  // object | null
  if (foo) {
    // object

    // define a guard
    function isFoo(input: unknown): input is {bar: Function} {
      return (
        typeof input === 'object' &&
        !!input &&
        input.hasOwnProperty('bar') &&
        typeof (input as any).bar === 'function'
      );
    }
    if (isFoo(foo)) {

      // { bar: Function }
      foo.bar();
    }
  }
}
```

`unknown` means the type is unknown and you must prove the type with runtime checks what the type is. `typeof` and `instanceof` are guard functions that determine the correct type. `unknown` is safer because of the runtime requirement to test what the type is. `any` will strip away any type checks.

PR #1157, opened 2021-07-15
fix(menu): Fix Menu item icon colors

### LINKED ISSUE #975: Accent and background icon layers aren't colored properly in hovered/focused Menu items
## 🐛 Bug Report

The MenuItem component of the Menu accepts `icon` and `secondaryIcon` props if you wish to display a System Icon before and/or after the content of the MenuItem. All layers of the icon (fill, accent, and background) should change to ensure proper contrast if the MenuItem is hovered over or focused using the keyboard.

Currently, only the fill layer is being colored properly on hover/focus. The accent and background layers are not being colored properly on hover/focus.

## To Reproduce

See: https://codesandbox.io/s/menu-icon-colors-nvrbc?file=/src/index.tsx

Hover over each item in the menu, or click on the menu and navigate the items using the keyboard. Notice how the accent layers of the icons (the dot in `tagIcon`, the speech lines in `activityStreamIcon`, and the silhouette in `taskContactIcon`) fail to change color and remain gray on hover/focus.

image

## Expected Behavior

The accent and background layers of `icon` and `secondaryIcon` should change colors based on the hover/focus state of the MenuItem.

It would make sense to use the same logic we use to color [toggleable buttons](https://workday.github.io/canvas-kit/?path=/story/components-buttons-button-react-icon-button--circle) since the logic isn't quite as simple as coloring the entire icon white on focus. The icon colors for keyboard-focused MenuItems should match the icon colors for toggled buttons (and the same for hover).

sahlhoff:
## Summary

Fixes #975.


jamesfan (commented):
@sahlhoff: FYI, I added `Fixes #975` to the PR summary to link this PR to its corresponding issue.

I haven't taken a super-close look at this as I remember coloring icons to be deceptively complicated, but I think we still need to fix how icon background layers are colored in MenuItems. Glancing over `MenuItem.tsx`, it doesn't look like it has any styles for `.wd-icon-background`.

Assuming we want the icon colors for a focused MenuItem to match the colors for a [toggled Icon Button](https://workday.github.io/canvas-kit/?path=/story/testing-react-buttons-button-icon-button--icon-button-circle-toggleable-grid) (this would make sense to me), the background layer needs to be colored to `iconColors.inverse` if the MenuItem is focused. Ideally, we could reuse the [icon coloring logic in ButtonContainer.tsx](https://github.com/Workday/canvas-kit/blob/master/modules/react/button/lib/parts/ButtonContainer.tsx#L41) used to color Icon Buttons, but this logic and the surrounding code can be quite confusing. I'm not sure how difficult it would be to factor it out -- it may be easier just to keep the styles/logic for coloring menu icons in `MenuItem.tsx` as you've done here.

Note that you'll also need to color the accent layer differently depending on whether or not the icon has a background layer to prevent the two layers from blending into each other. To better illustrate this:

https://codesandbox.io/s/icon-structure-v5-34fx1

Click on the Icon Buttons at the bottom. Note how the accent layer is colored differently in `adHocDelivery` than it is in `taskContact` since `taskContact` has a background layer. I handled this in the past when fixing Icon Buttons by [using the general sibling selector  (~)](https://github.com/Workday/canvas-kit/pull/895#discussion_r515421647).

NicholasBoll:
Perhaps a visual test should be added to validate this change as it seems to be a CSS-only change, but there's no Chromatic change.

jamesfan (approved):
A spot check of the icons looks good, ship it 👍 

I think the styles in `MenuItem.tsx` could be cleaned up (it's a bit confusing how we have icon styles split between `Item = styled('li')` and `setIconProps`), but my head hurts too much from juggling all the combinations of row states and icon layers to think about this too deeply, haha. 🤯 

This seems to work. I say we ship it and clean it up later, possibly when Menu is converted to a compound component.

sahlhoff:
@jamesfan thank you for the review and sorry I never followed up on the first comment...You are right it is overly complex! 😂 😭  I was thinking instead of over specifying the icon selectors for these states in the component, maybe we could update the SystemIcon props to include `Hover` and `Focus` states? Just my two cents. Hoping we can merge this fix in as is and reexamine at a later time.

jamesfan:
@sahlhoff Agreed, the biggest issue I have is that icon styles are often spread out in the code (both here in Menu and in Icon Button) -- combine that with the combinations of different states applied to icons (hover, focus, disabled, etc.) and it becomes very difficult to determine which styles are being applied to an icon at a given moment.

Merging this in, we can take a closer look at the underlying issues with icon styling later.

PR #1159, opened 2021-07-15
test(popup): Add visual tests for RTL in Modal and Popup

### LINKED ISSUE #1158: Support RTL in Popups
## 🐛 Bug Report

A Popup inside an RTL `<CanvasProvider>` doesn't render correctly.

## To Reproduce

Go to https://codesandbox.io/s/1gj73

## Expected Behavior

The heading of RTL text should be on the right as well as the body text. Also there should be space between the "OK" and "Cancel" buttons.

image

## Actual Results

Heading and text are on the left and there is no space between the buttons.

image

Go to https://codesandbox.io/s/1gj73

If you click the `Fix` button, it will add a `dir="rtl"` to the `Modal.Card` which will fix all rendering issues.

## Suggestion

Add `dir` to the `usePopupStack` hook so that direction properly follows portals. React theme context already does which is why the 'X' icon flips correctly, but nothing else does.

NicholasBoll:
Adds visual tests for `Modal` and `Popup` for RTL.

Modal:
Modal RTL Screenshot


modules/react/modal/stories/stories_VisualTesting.tsx
```diff
@@ -53,3 +53,22 @@ const TestModal = () => {
 };
 
 export const ModalSmallWidth = withSnapshotsEnabled(() => <TestModal />);
+
+export const ModalRTL = withSnapshotsEnabled(() => {
```
NicholasBoll, line 57: New Story to make sure Modals render RTL correctly

modules/react/popup/stories/stories_VisualTesting.tsx
```diff
@@ -108,30 +108,21 @@ export const PopupStates = withSnapshotsEnabled(() => (
...
-              heading: 'למחוק פריט',
-              hasCloseIcon: true,
-              width: 300,
-            },
-          },
-        ]}
-        columnProps={[{label: 'Default', props: {}}]}
-      >
-        {({heading, hasCloseIcon, closeIconSize, ...props}) => (
-          <Popup.Card style={{animation: 'none'}} {...props}>
```
NicholasBoll, line 128: This didn't test rendering through portals, so we'll render a portal instead.

NicholasBoll:
Blocked on #1183

alanbsmith (approved):
🐏

PR #1183, opened 2021-07-26
fix(popup): Fix RTL rendering for popups

### LINKED ISSUE #1158: Support RTL in Popups

NicholasBoll:
Fixes #1158
- Add `dir` to `usePopupStack` to automatically set direction for portalled popups

Before:
bug: RTL Popup with LTR content

After:
fix: RTL Popup with RTL content

### DIFF
--- a/modules/popup/react/lib/usePopupStack.ts
+++ b/modules/popup/react/lib/usePopupStack.ts
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import {PopupStack} from '@workday/canvas-kit-popup-stack';
+import {useIsRTL} from '@workday/canvas-kit-react-common';
 
 /**
  * This hook should not be used directly. Use the `Popper` component instead.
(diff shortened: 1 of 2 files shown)


alanbsmith (approved):
👍🏻

PR #1188, opened 2021-07-28
fix: Misc. MDX fixes

anicholls:
Fix some misc. MDX issues:

- Since we convert example exports to `default`, we also have to update the imports in the mdx files. These imports were broken
- Converted `ActionBar` `fixed` example to a code snippet so the `position:fixed` div won't go beyond it's example container bounds (if not rendered in an iframe)
- Updates Button MDX and readme. We held off from finishing this and stripping stuff out of the README until the prop tables were fixed. 
  - Note: I started to add examples for `size`, `icon` and `grow` to the button mdx file, but this caused them to show up in the sidebar as stories. Unfortunately, there is no way to categorize these without creating separate MDX files for each button component, which I think we should avoid (for duplication reasons, and increased complexity when using them in our doc site).


jamesfan (approved):
Looks good 👍 

This doesn't need to block this PR, but regarding the examples you wanted to add for `size`, `icon` and `grow`: given how brief the markup is for Buttons, you could just add multiple Buttons in each of the example files. For example, `Primary.tsx` would have `<PrimaryButton size="small">` and `<PrimaryButton size="large">` in addition to the default medium `<PrimaryButton>`. You could add `<PrimaryButton icon={}>` to that same file as well.

This is similar to what I did for the [Basic Example for Status Indicator](https://workday.github.io/canvas-kit/?path=/docs/components-indicators-status-indicator-react--basic) (and even [the older stories for Buttons](https://5d854c26ba934e0020f5e98a-llrxodeyez.chromatic.com/?path=/story/components-buttons-button-react--delete)) where I thought it would be helpful to see different variations of the same component without having a separate example file for each one.

Also, is there supposed to be a `Hyperlink` story? I noticed it's still in the README along with `ToolbarIconButton` and `ToolbarDropdownButton`, but it's not in the Storybook sidebar anywhere with the exception of a visual states table.

NicholasBoll (approved):

alanbsmith (approved):
🐅

PR #1189, opened 2021-07-29
fix(common): Remove @ts-ignore from ExtractProps

NicholasBoll:
Users were reporting that `ExtractProps` was giving compile errors if `skipLibCheck` was set to `true`. It turns out we used `@ts-ignore` in an exported type, but `//` comments do not get translated to the definition files, so the `@ts-ignore` was missing and users were getting the error.

The TS error we were getting was because of how we tested if a generic was a `never` using a trick from https://github.com/microsoft/TypeScript/issues/23182. The problem was this was a hack and Typescript didn't narrow the generic from the conditional.

I fixed this by rearranging the conditional types to have `never` be the last possibility of the conditional checks. Extending a key was checked first, then `undefined`, then the default was `never`. This allowed me to remove `ts-ignore` without any hacks.


modules/react/common/lib/utils/components.ts
```diff
@@ -78,26 +76,26 @@ type ExtractHTMLAttributes<
...
-    ? [TElement] extends [never] // test if user passed `never` for the `TElement` override. https://github.com/microsoft/TypeScript/issues/23182
-      ? P // if `TElement` was `never`, return only the inferred props `P`
+    ? TElement extends keyof JSX.IntrinsicElements
+      ? ExtractHTMLAttributes<JSX.IntrinsicElements[TElement]> & P // `TElement` was defined, so we'll return the HTML attribute interface + inferred `P` props
       : TElement extends undefined // else test if TElement was defined
       ? ExtractHTMLAttributes<JSX.IntrinsicElements[E]> & P // `TElement` wasn't explicitly defined, so let's fall back to the inferred element's HTML attribute interface + props `P`
-      // TS thinks `TElement` cannot be an index type because of the `undefined`, but we've already ensured that can't happen, so we'll ignore that error
-      // @ts-ignore
-      : ExtractHTMLAttributes<JSX.IntrinsicElements[TElement]> & P // `TElement` was defined, so we'll return the HTML attribute interface + inferred `P` props
+      : P // `TElement` was `never`, return only the inferred props `P`
```
NicholasBoll, line 92: The conditional output is the same, but I reordered the checks.

Before it checked if `TElement` was:
1. `never`
2. `undefined`
3. default (`keyof JSX.IntrinsicElements`)

Now it is checking the reverse order:
1. `keyof JSX.IntrinsicElements`
2. `undefined`
3. default (`never`)

modules/react/common/lib/utils/components.ts
```diff
@@ -78,26 +76,26 @@ type ExtractHTMLAttributes<
...
     : P // E isn't in JSX.IntrinsicElements, we don't know what it is, so return the inferred props `P`
   : TComponent extends Component<infer P> // test if `TComponent` is a `Component`, while inferring props `P`
   ? P // it was a `Component`, return inferred props `P`
-  : {}; // `TComponent` does not extend either `ElementComponent` or `Component`, return an empty object
+  : TComponent extends React.ComponentType<infer P> // test if `TComponent` is a `React.ComponentType` (class or functional component)
+  ? P // it was a `React.ComponentType`, return inferred props `P`
+  : {}; // We don't know what `TComponent` was, return an empty object
```
NicholasBoll, line 98: I'm sneaking this feature in. `ExtractProps` can now extract the props of any React component. Why not?

NicholasBoll, line 98: I'm sneaking this feature in. `ExtractProps` can now extract the props of any React component. Why not?

modules/react/common/lib/utils/components.ts
```diff
@@ -148,10 +146,11 @@ interface RefForwardingComponent<T, P = {}> {
  * return type is `Component<element, Props>` which looks like `Component<'div', Props>` which is a
  * clean interface that tells you the default element that is used.
  */
-export const createComponent = <T extends React.ElementType | undefined = undefined>(as?: T) => <
-  P,
-  SubComponents = {}
->({
+export const createComponent = <
+  E extends keyof JSX.IntrinsicElements | React.ComponentType | undefined = undefined
```
NicholasBoll, line 150: I extracted out `React.ElementType` to the component parts to avoid confusion. The type is the same though.

modules/react/common/lib/utils/components.ts
```diff
@@ -177,40 +176,46 @@ export const createComponent = <T extends React.ElementType | undefined = undefi
    *   )
    * }
    */
-  Component: RefForwardingComponent<T, P>;
+  Component: RefForwardingComponent<E, P>;
```
NicholasBoll, line 179: `E` made more sense than `T`. It is a element or component.

modules/react/common/lib/utils/components.ts
```diff
@@ -177,40 +176,46 @@ export const createComponent = <T extends React.ElementType | undefined = undefi
...
+      E extends undefined ? React.FC : E,
       P
     >) &
   SubComponents => {
-  const ReturnedComponent = React.forwardRef<T, P & {as?: React.ElementType}>(
+  const ReturnedComponent = React.forwardRef<E, P & {as?: React.ElementType}>(
     ({as: asOverride, ...props}, ref) => {
       return Component(
         props as P,
-        // @ts-ignore React considers no passed ref as `null`, Typescript considers this
```
NicholasBoll, line 199: This `ts-ignore` is an internal type and won't be in the definition file, but I think we should prefer `any` over `ts-ignore` as a bail to avoid issues for users.

modules/react/common/lib/utils/components.ts
```diff
@@ -58,16 +58,14 @@ type ExtractHTMLAttributes<
...
 /**
  * Extract props from any component that was created using `createComponent`. It will return the
  * HTML attribute interface of the default element used with `createComponent`. If you use `as`, the
  * HTML attribute interface can change, so you can use an override to the element you wish to use.
  * You can also disable the HTML attribute by passing `never`:
  *
- * - `ExtractProps<typeof Card>`: `React.HTMLAttributes<HTMLDivElement> & CardProps`
- * - `ExtractProps<typeof Card, 'aside'>`: `React.HTMLAttributes<HTMLElement> & CardProps`
+ * - `ExtractProps<typeof Card>`: `CardProps & React.HTMLAttributes<HTMLDivElement>`
+ * - `ExtractProps<typeof Card, 'aside'>`: `CardProps & React.HTMLAttributes<HTMLElement>`
```
NicholasBoll, line 68: I ended up reordering these. It made more sense.

modules/react/common/lib/utils/components.ts
```diff
@@ -78,26 +76,43 @@ type ExtractHTMLAttributes<
...
+type ExtractPropsFromComponent<TComponent> = TComponent extends ElementComponent<infer E, infer P> // test if `TComponent` is an `ElementComponent`, while inferring both default element and props associated
+  ? E extends keyof JSX.IntrinsicElements // test if the inferred element `E` is in `JSX.IntrinsicElements`
+    ? P & ExtractHTMLAttributes<JSX.IntrinsicElements[E]> // return inferred props `P` + HTML attributes of inferred element `E`
+    : P // `E` wasn't a key of `JSX.IntrinsicElements`, so just return the inferred props `P`
   : TComponent extends Component<infer P> // test if `TComponent` is a `Component`, while inferring props `P`
   ? P // it was a `Component`, return inferred props `P`
-  : {}; // `TComponent` does not extend either `ElementComponent` or `Component`, return an empty object
+  : TComponent extends React.ComponentType<infer P> // test if `TComponent` is a `React.ComponentType` (class or functional component)
+  ? P // it was a `React.ComponentType`, return inferred props `P`
+  : {}; // We don't know what `TComponent` was, return an empty object
```
NicholasBoll, line 115: Complete refactor of this type. I've added some new features while fixing the bug:
- `ExtractProps` now supports component overrides: `ExtractType<ElementComponent<'div', Props>, typeof SomeComponent`. We support `as` for other components, `ExtractProps` should as well

modules/react/common/lib/utils/components.ts
```diff
@@ -148,10 +163,15 @@ interface RefForwardingComponent<T, P = {}> {
...
-export const createComponent = <T extends React.ElementType | undefined = undefined>(as?: T) => <
-  P,
-  SubComponents = {}
->({
+export const createComponent = <
+  E extends
+    | keyof JSX.IntrinsicElements
+    | React.ComponentType
+    | ElementComponent<any, any>
+    | undefined = undefined
```
NicholasBoll, line 171: This was probably the source of Typescript errors when trying to use other `ElementComponent`s as an input to another component: `createComponent(Box)({...})`. While there's not always a good use-case for doing so, this removes the Typescript error.

modules/react/common/lib/utils/components.ts
NicholasBoll: I ran into this issue when refactoring: https://www.typescriptlang.org/play?#code/C4TwDgpgBAKhDOwbggHgFBVlCAPYEAdgCbxQCuJEAZgJaETFQA+UDAbhAE5QC8FVOg2LoAfH2x4CJMpWI16jKAHplOADbxoBRFFrUoAAxiGoAdwCGZeUMaYoAfigBvAIwAuKIi70A5gF8sVSMTcytCAHJgHFwwdVoAY1pgdRAoG0URLE9nACZPbz9AlTVjU0syQw5uQwAaKB9fAAtgJxgmi2AIsmAm6AB7QlSoXr8odRpgADoZqABRTkJonWAxvDjE5OHqfp4qiE4uUz6uaGJ+hEjos12AaymoAElorQgAWx7+qF8IaPV+xAzKbodCgSBQABi-X6rgkcEQyHBwTcBWAPkIAXqCV2pwSwFBKEh0NycIQSBQqGqXHEwSp9XgTX65HUTAARtA8qj0QEQcF4eTwRZfBZ6PVWeRrskmlALIQYhsktF5AQ8bRBlB+gZ9odTAB9dk7U66oxyBTCQwE8H8xEQXIYLAwGLSUgCDLCFhsA7cCSm2wicT8ADaMAAuk6iC7A1Sw8EIJptGS9FrQhUjFTDA8AMoQaAtYBgeDuVS+KXkVlTbFvZRvRJcAGa4DKG2ZhI+MCN2jweDkBDKXIAZlcAA5cvYnJyvGiiiUQuUrGmvUd7DkPJPuYFgmUoG9yLp2SbBJkLZboAAhCxcWH8a0oEnI1eFDH+LE4iB4k9Qc9cEnXsk2u1UjSagTo+mJQNiXC4sAQA

modules/react/common/lib/utils/components.ts
vibdev: woot!

modules/react/common/lib/utils/components.ts
```diff
@@ -78,26 +76,43 @@ type ExtractHTMLAttributes<
  * @example
  * interface MyProps extends ExtractProps<typeof Card.Body> {}
  *
- * ExtractProps<typeof Card>; // React.ClassAttributes<HTMLDivElement> & React.HTMLAttributes<HTMLDivElement> & CardProps
- * ExtractProps<typeof Card, 'aside'>; // React.ClassAttributes<HTMLElement> & React.HTMLAttributes<HTMLElement> & CardProps
```
NicholasBoll, line 82: I ended up removing `React.ClassAttributes` from the type which removes `ref` and `key`. `ref` is added by `ElementComponent` and depends on the `as` prop, and `key` is already available via JSX typings and cannot be accessed internally anyways.

alanbsmith (approved):
Looks good. Thanks for the update. 👏🏻

Issue #845, opened 2020-09-02
Reconsider how we handle required form fields


jamesfan:
_While adding support for `aria-required` to the Labs Select (#797), a number of other questions came up regarding how we handle required form fields in Canvas Kit in general. This issue summarizes those questions._

## 🔍 Spike / Exploration

Currently, FormField accepts a `required` prop. If `required` is set to `true`, FormField renders `<abbr title="required">*</abbr>` inside of its `<label>`. This is _not_ a recommended approach for signaling to assistive technology that this field is required. From @mathalete:

> The `<abbr>` tag is used to expand the definition of an abbreviated word, this approach isn't recommended. For example - VoiceOver simply does not announce the title attribute content on a abbr
>
> if the ask here is to robustly include the word 'required' in the `<label>` then adding visually hidden text would be the way to do so

Peter suggested the following markup (where `.sr` is the accessible hiding class):
```html
<label for="foo">Name <span aria-hidden="true">*</span> <span class="sr">Required</span></label>
```

On a related note, we also need to determine if we need to add special handling for `aria-required` to FormField. MDN states `aria-required` as a [useful fallback](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute) for user agents that do not yet support HTML5, although a cursory Google search suggests that `required` may be sufficient ([source 1](https://developer.paciellogroup.com/blog/2019/02/required-attribute-requirements/), [source 2](https://stackoverflow.com/a/37975985)).

## What questions would be answered once this spike was completed?

1. What was the original reasoning behind adding `<abbr title="required">` for required FormFields? If its sole purpose is to signal to assistive technology that the FormField is required, should we instead implement the suggested fix above where "required" is visually hidden? If not, what would be the recommended solution?
2. Does FormField need to handle `aria-required` as well? If so, do we need to handle it any differently than `required`? 
3. ~Regardless of whether or not we add support for `aria-required`, if we implement the suggested fix referenced in (1) (or whatever solution we arrive at), do we need to retain the code we added to the Labs Select in #797 to special-case handling of `aria-required` or is it no longer needed?~ _Edit: The Labs Select is a special case among input components. Investigate this separately._
4. Do we need any additional handling of `required` and/or `aria-required` in the input components themselves?

NicholasBoll:
Is there any serious objections to the following?

```html
<label for="foo">Name <span aria-hidden="true">*</span></label>
[...]

PR #1201, opened 2021-08-06
feat(labs): Add new compound component for text inputs

### LINKED ISSUE #1197: Convert TextInput into a Compound Component
## 🚀 Feature Proposal

Convert TextInput into a compound component.

vibdev:
## Summary

Resolves #1197

This PR creates a new labs component for text inputs. Combines all the best from the old school FormField and TextInput components and wraps them up in a nice and fancy compound component.

category

## Release Note

## Breaking Changes

Optional breaking changes message. If your PR includes breaking changes. It is extremely rare to put breaking changes outside a `prerelease/v*` branch. Anything in this section will show up in release notes. Remove this section if no breaking changes are present.

---

## Checklist

- [x] `yarn test` passes
      applicable


cypress/integration/TextInputLabs.spec.ts
```diff
@@ -0,0 +1,79 @@
+import * as h from '../helpers';
```
vibdev, line 1: These are copied from other Text Input, as a follow up we should identify additional tests, e.g. that the model applies ids, for, and aria-* correctly.

modules/labs-react/text-input/lib/TextInput.Label.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import {TextInputModel} from './useTextInputModel';
+import {TextInputModelContext} from './TextInput';
+
+export interface TextInputLabelProps extends Themeable {
+  model?: TextInputModel;
+  /**
+   * If the field is required, provide the title required label.
+   */
+  isRequiredLabel?: string;
```
vibdev, line 22: Seemed redundant to have required and required label, but wondering if it would make more sense to make 'isRequired' part of the model state and make this field have a default value of "required"

modules/labs-react/text-input/lib/useTextInputModel.tsx
```diff
@@ -0,0 +1,78 @@
...
+  hasError?: ErrorType;
+  inputId?: string;
+  hintId?: string;
+};
+
+type TextInputEvents = {
+  addError(): void;
+  removeError(): void;
+  addAlert(): void;
+  removeAlert(): void;
```
vibdev, line 22: Thinking about this more I wonder if Alert even really needs to be a built in state. It only controls the error ring color so you could really just manually add that yourself around TextInput.Field. In that case we could simplify, make hasError a boolean. And then just add example of how to set the color to anything, Alert Yellow, Green success, etc. Thoughts?

modules/labs-react/text-input/lib/TextInput.Label.tsx
```diff
@@ -0,0 +1,79 @@
...
+export const TextInputLabel = createComponent('label')({
+  displayName: 'TextInput.Label',
+  Component: (
+    {isRequiredLabel, isVisuallyHidden, theme, model, children, ...elemProps}: TextInputLabelProps,
+    ref
+  ) => {
+    const {state} = useModelContext(TextInputModelContext, model);
+
+    return (
+      <StyledLabel
```
vibdev, line 65: Kinda wanted this to be a `<Box as='label' ...` but that doesn't seem to be allowed when using styled:

```
Type 'HTMLDivElement' is missing the following properties from type 'HTMLLabelElement': control, form, htmlForts(2769)
```

https://codesandbox.io/s/admiring-feistel-u601g?file=/src/App.tsx

vibdev: `const StyledBox = styled(Box as typeof TextInputLabel)` fixes it, but is a bit weird to have to use

vibdev:
😕 not sure why my last example code is failing:
```
Error: modules/labs-react/text-input/stories/examples/LoginForm.tsx(3,25): error TS2307: Cannot find module 'formik'.
Error: modules/labs-react/text-input/stories/examples/LoginForm.tsx(4,22): error TS2307: Cannot find module 'yup'.
```

They are in dev dependencies and are only used in a story?

cypress/integration/TextInputLabs.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+        h.stories.load('Labs/Text Input/React', story);
+      });
+
+      it('should not have any axe errors', () => {
+        cy.checkA11y();
+      });
+
+      context('when clicked', () => {
+        beforeEach(() => {
+          getTextInput().click();
```
NicholasBoll, line 23: I think these helpers are fine, but I've been preferring Cypress Testing Library helpers:
```ts
cy.findByRole('textbox', { name: 'Email' })
```

Sometimes it is fine to not specify the accessible name:
```ts
cy.findByRole('textbox')
```

The downside is repeating, but the upside is a direct correlation to Cypress command log. `getTextInput` will not show on the Cypress command log while `findByRole` with it's arguments will.

NicholasBoll:
> 😕 not sure why my last example code is failing:
> 
> ```
> Error: modules/labs-react/text-input/stories/examples/LoginForm.tsx(3,25): error TS2307: Cannot find module 'formik'.
> Error: modules/labs-react/text-input/stories/examples/LoginForm.tsx(4,22): error TS2307: Cannot find module 'yup'.
> ```
> 
> They are in dev dependencies and are only used in a story?
> 
> Noticed the same pattern in preview story, so must be missing something:
> 
> https://github.com/Workday/canvas-kit/blob/e9db585994fe910538899fc8e0fb22be9817821e/modules/preview-react/package.json#L63
> 
> https://github.com/Workday/canvas-kit/blob/fa0f4d9226ada17cbd9f293eb914313d94d966d8/modules/preview-react/menu/stories/stories.tsx#L9

@vibdev You forgot to update/commit the `yarn.lock` file. You'll have to temporarily remove the `--frozen-lockfile` line on the `.yarnrc` file and then run `yarn install` and commit the updated lockfile.

modules/labs-react/text-input/lib/TextInput.Field.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
```
NicholasBoll, line 1: lol. I didn't even think to put a `.` in the file name. `TextInput.Field.tsx` instead of `TextInputField.tsx`. I'm not sure which I like better. The exported component variable is still `TextInputField`.

modules/labs-react/text-input/lib/TextInput.Field.tsx
```diff
@@ -0,0 +1,92 @@
...
+  StyledType,
+  Themeable,
+  useModelContext,
+} from '@workday/canvas-kit-react/common';
+import {Box} from '@workday/canvas-kit-labs-react/common';
+
+import {TextInputModelContext} from './TextInput';
+import {TextInputModel} from './useTextInputModel';
+
+export interface TextInputFieldProps extends ExtractProps<typeof Box, 'input'>, Themeable {
```
NicholasBoll, line 19: Other components do not include the element interface since it is added again in use. Change the `'input'` to `never` to not automatically add them. What happens is TS will add all the element attributes in this interface, then remove them all only to add them all again.

```suggestion
export interface TextInputFieldProps extends ExtractProps<typeof Box, never>, Themeable {
```

`ExtractProps` was made to add them again if someone wants to extend our components with a baked interface. For example:

```ts
type SomeProps = ExtractProps<typeof TextInput.Field> // TextInputFieldProps & React.InputHTMLAttributes<HTMLInputElement>
```

The `React.InputHTMLAttributes<HTMLInputElement>` attributes are added by `ExtractProps` based on the default element passed into `createComponent`.

modules/labs-react/text-input/lib/TextInput.Field.tsx
```diff
@@ -0,0 +1,92 @@
...
+  {
+    ...type.levels.subtext.large,
+    border: `1px solid ${inputColors.border}`,
+    display: 'block',
+    backgroundColor: inputColors.background,
+    borderRadius: borderRadius.m,
+    boxSizing: 'border-box',
+    height: 40,
+    transition: '0.2s box-shadow, 0.2s border-color',
+    padding: spaceNumbers.xxs, // Compensate for border
```
NicholasBoll, line 35: Something we should think about and be careful of:

`Box` has `padding` on it, so `padding` is part of the accepted props. However, if a property is defined using `styled` or `css`, it will _always_ override a passed attribute. As it currently stands, the following will be allowed by Typescript, but not work at runtime:

```tsx
<TextInput.Field padding="l" />
```

The `padding` will be silently overridden by the `styled(Box)`. The following will work, however:

```tsx
<TextInput.Field css={{ padding: spacing.l }} />
```

modules/labs-react/text-input/lib/TextInput.Hint.tsx
```diff
@@ -0,0 +1,58 @@
...
+  StyledType,
+  Themeable,
+  useModelContext,
+} from '@workday/canvas-kit-react/common';
+import {space, type} from '@workday/canvas-kit-react/tokens';
+import {Box} from '@workday/canvas-kit-labs-react/common';
+import {TextInputModelContext} from './TextInput';
+import {TextInputModel} from './useTextInputModel';
+
+export interface TextInputHintProps extends ExtractProps<typeof Box>, Themeable {
```
NicholasBoll, line 17: Same here, remove the element attributes:
```suggestion
export interface TextInputHintProps extends ExtractProps<typeof Box, never>, Themeable {
```

modules/labs-react/text-input/lib/TextInput.Label.tsx
```diff
@@ -0,0 +1,83 @@
...
+  useModelContext,
+} from '@workday/canvas-kit-react/common';
+import {type} from '@workday/canvas-kit-react/tokens';
+import {HStack} from '@workday/canvas-kit-labs-react/layout';
+import {Box} from '@workday/canvas-kit-labs-react/common';
+
+import {TextInputModel} from './useTextInputModel';
+import {TextInputModelContext} from './TextInput';
+
+export interface TextInputLabelProps extends ExtractProps<typeof Box, 'label'>, Themeable {
```
NicholasBoll, line 19:
```suggestion
export interface TextInputLabelProps extends ExtractProps<typeof Box, never>, Themeable {
```

modules/labs-react/text-input/lib/TextInput.Label.tsx
```diff
@@ -0,0 +1,83 @@
...
+        as="label"
+        ref={ref}
+        htmlFor={state.inputId}
+        theme={theme}
+        isVisuallyHidden={isVisuallyHidden}
+        {...elemProps}
+      >
+        <HStack spacing="xxxs">
+          <span>{children}</span>
+          {!!isRequiredLabel && <RequiredAsterisk title={isRequiredLabel}>*</RequiredAsterisk>}
```
NicholasBoll, line 78: https://github.com/Workday/canvas-kit/issues/845

This should be:

```tsx
<RequiredAsterisk aria-hidden="true">*</RequiredAsterisk>
```

Also, the default component should be a `span` instead of an `abbr`.

I'm also on the fence about if `RequiredAsterisk` is a hidden component within this subcomponent. The other option is to make it a subcomponent: `<TextInput.Required />`.

modules/labs-react/text-input/lib/useTextInputModel.tsx
NicholasBoll: Typically in form libraries or form usage, the error is a property of the form model and not each field. It is a prop. If you add something like `hasError` or `errorType` to `TextInputModelConfig`, it could be simply forwarded to the state. Using events adds complexity to integrating with libraries. I can see it in the example:

https://github.com/Workday/canvas-kit/pull/1201/files#diff-8bfe0cebce955a2e1f5d234f29843603d8a03b04e376bdad11bddf10aec11d12R52-R63

NicholasBoll:
After reviewing much of this PR, I was wondering if what we really want is `FormField` to be the compound component and not `TextInput`:

```tsx
<FormField>
  <FormField.Label>Label</FormField.Label>
  <FormField.Field as={TextInput} />
</FormField>
```

This also means any other field could be used:
```tsx
<FormField>
  <FormField.Label>Label</FormField.Label>
  <FormField.Field as={SelectOne} items={["One", "Two"]} />
</FormField>
```

Otherwise we'll have to make a compound component that involves all the field stuff for each field type. As it stands now, `TextInput.Field` contains all styling information, so you couldn't pass `SelectOne` without messing up styling.

NicholasBoll:
What would a `TextInput` compound component look like? Maybe contain things like a `TextInput.ClearButton` to clear the input (or just add styling)?

vibdev:
> After reviewing much of this PR, I was wondering if what we really want is `FormField` to be the compound component and not `TextInput`:
> Otherwise we'll have to make a compound component that involves all the field stuff for each field type. As it stands now, `TextInput.Field` contains all styling information, so you couldn't pass `SelectOne` without messing up styling.

I was thinking a bit about that too, The only hesitation I have is for something like radio where you would also have to remember to make the correct modifications for the label. 

```tsx
<FormField>
  <FormField.Label as='legand'>Label</FormField.Label>
   <FormField.Field as={Radio} />
 </FormField>
 ```
 
 vs 
 
 ```tsx
<RadioInput>
  <RadioInput.Label>Label</RadioInput.Label>
   <RadioInput.Field />
 </RadioInput>
 ```

File: cypress/integration/TextInputLabs.spec.ts (hunk 5 above)
Comment (vibdev) [reply cypress/integration/TextInputLabs.spec.ts:23]: Agree, I just straight copied these tests from the non-labs component. Didn't want to mess with them too much until we had a solid agreement of what the spec of this component should be.

File: modules/labs-react/text-input/lib/TextInput.Field.tsx (hunk 6 above)
Comment (vibdev) [reply modules/labs-react/text-input/lib/TextInput.Field.tsx:1]: I can drop for consistency with the export :D

File: modules/labs-react/text-input/lib/TextInput.Field.tsx (hunk 7 above)
Comment (vibdev) [reply modules/labs-react/text-input/lib/TextInput.Field.tsx:19]: Ah thanks forgot about that bit, will update

File: modules/labs-react/text-input/lib/TextInput.Field.tsx (hunk 8 above)
Comment (vibdev) [reply modules/labs-react/text-input/lib/TextInput.Field.tsx:35]: hmmm do we need to omit BoxProps if they are used in a base style?

File: modules/labs-react/text-input/lib/TextInput.Label.tsx (hunk 11 above)
Comment (vibdev) [reply modules/labs-react/text-input/lib/TextInput.Label.tsx:78]: Is the correct behavior to hide the * from aria and just set the required attribute on the input field?

File: modules/labs-react/text-input/lib/TextInput.Field.tsx (hunk 8 above)
Comment (vibdev) [reply modules/labs-react/text-input/lib/TextInput.Field.tsx]: or actually should the style BoxProps get moved form styled down to TextInputField props

File: tsconfig.stories.json [hunk 12]
```diff
@@ -13,6 +13,7 @@
     "strictNullChecks": false,
     "noUnusedLocals": false,
     "types": [],
-    "incremental": false
+    "incremental": false,
+    "skipLibCheck": true
```
vibdev, line 17: This fixes the CI failures, is it ok to add?

modules/labs-react/text-input/lib/useTextInputModel.tsx
vibdev: Is this the right way to forward the state?
https://codesandbox.io/s/hopeful-kalam-x5cji?file=/tester/Tester.tsx:730-767

NicholasBoll: No. It would be more like:

```tsx
const useTextInputModel = ({hasError}) => {

  const state = {
    hasError,
  }

  const events = {}
  
  return { state, events }
}
```

modules/labs-react/text-input/lib/TextInput.Label.tsx
vibdev: Swapped to using css

modules/labs-react/text-input/lib/TextInput.Field.tsx
vibdev: went with css

modules/labs-react/text-input/lib/TextInput.Label.tsx
vibdev: Maybe if folks don't want it they can create their own label with behavior hook

tsconfig.stories.json
vibdev: This seems to be a conflict with the formik 2.x and the version of @types/react we are using, I tried moving to @types/react 16.9.12 but that introduced a bunch of conflicts with react-testing library, updating that caused issues with enzyme. So wondering if we should hold off for now and maybe revisit this later, perhaps after enzyme gets removed.

vibdev:
RE: Adding the FormField compound component that is can compose with, what do y'all think of me doing that in a follow up PR? This one is getting big already and wonder if it would be easier to reason if I made it alongside, say a text area component, so it is more obvious what can be factored out.

modules/labs-react/text-input/stories/examples/LabelPosition.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const LabelPosition = () => {
+  const [value, setValue] = React.useState('');
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setValue(event.target.value);
+  };
+
+  return (
+    <HStack spacing="l" alignItems="center">
```
vibdev, line 13: perhaps this could be a hook to use with `Stack`?
e.g.
`useLeftLabel = () => ({ flexDirection="row", spacing="l", alignItems="center" })`
`useTopLabel = () => ({ flexDirection="column", spacing="xxxs" alignItems="flex-start" })`

tsconfig.stories.json
NicholasBoll: Ugh. There is still a few tests written in Enzyme, `Menu` being the most complicated. The other files have only a few tests each. If we add this, I'd like an issue to get rid of it again.

NicholasBoll:
> RE: Adding the FormField compound component that is can compose with, what do y'all think of me doing that in a follow up PR? This one is getting big already and wonder if it would be easier to reason if I made it alongside, say a text area component, so it is more obvious what can be factored out.

This being a labs component (to be merged into main in a future version), that works for me. The current CI failure is something to do with IE11.

modules/labs-react/text-input/stories/examples/LabelPosition.tsx
NicholasBoll: Hmm. Or maybe a layout component? Like `HorizontalField` and `VerticalField`? If the worry is remembering what styles to set on every field.

vibdev: ¯\\_(ツ)_/¯ Eh, I don't feel strongly either way so would tend towards leaving up to the consumer.

vibdev:
@NicholasBoll Fixed IE11 not sure what the new failure is?

modules/labs-react/text-input/lib/TextInputLabel.tsx
```diff
@@ -0,0 +1,74 @@
...
+          {localModel.state.isRequired && (
+            <span
+              css={themeRTL(styles.asterisk, {color: theme.canvas.palette.error.main})}
+              aria-hidden="true"
+            >
+              *
+            </span>
+          )}
+        </HStack>
+      </Box>
```
NicholasBoll, line 71: I don't like the amount of elements here and lack of access and overrides. Can we get rid of `<Box>` entirely?

Also, the `isVisuallyHidden` seems to be such a one-off prop that could be used for any component via the `accessibleHide` utility style.

```suggestion
      <HStack
        spacing="xxxs"
        as="label"
        css={themeRTL(styles.label)}
        {...props}
      >
        <span>{children}</span>
        {localModel.state.isRequired && (
          <span
            css={themeRTL(styles.asterisk, {color: theme.canvas.palette.error.main})}
            aria-hidden="true"
          >
            *
          </span>
        )}
      </HStack>
```

The visually hidden example would then use the `accessibleHide` style manually - reinforcing its use.

modules/labs-react/text-input/lib/hooks/useTextInputModel.tsx
```diff
@@ -0,0 +1,45 @@
...
+  guards: {},
+  callbacks: {},
+});
+
+export type TextInputModelConfig = TextInputState &
+  Partial<ToModelConfig<TextInputState, TextInputEvents, typeof textInputEventMap>>;
+
+export const useTextInputModel = (config: TextInputModelConfig = {}): TextInputModel => {
+  const inputId = useUniqueId(config.inputId);
+  const hintId = useUniqueId(config.hintId);
```
NicholasBoll, line 30: Other components that have multiple ids use the same base id and derive new ones from there. `Tabs` for example has a single `id` and derives item ids from it. This decreases the amount of state necessary.

NicholasBoll:
> RE: Adding the FormField compound component that is can compose with, what do y'all think of me doing that in a follow up PR? This one is getting big already and wonder if it would be easier to reason if I made it alongside, say a text area component, so it is more obvious what can be factored out.

I've done it both ways. 2 similar components in 1 PR to get a good feel for how things work and compose. A second PR is an option, but I've found I often need to modify the first PR's code later and it can be annoying if feedback requires changes to both, so I tend to favor a larger PR of similar components to get a better idea of how things work so that future changes are less likely.

modules/labs-react/text-input/lib/hooks/useTextInputModel.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const textInputEventMap = createEventMap<TextInputEvents>()({
+  guards: {},
+  callbacks: {},
+});
+
+export type TextInputModelConfig = TextInputState &
+  Partial<ToModelConfig<TextInputState, TextInputEvents, typeof textInputEventMap>>;
+
+export const useTextInputModel = (config: TextInputModelConfig = {}): TextInputModel => {
```
NicholasBoll, line 28: Something to consider. All these models are field-type models. Will there be a specific model for each field type that is copy/pasted from this one, or is there a `FieldModel` instead of `*Model`? Popups work this way where all popup type components use a `PopupModel` instead of defining new model types for specific popups. `useModalModel` actually returns a `PopupModel` that has modal behaviors already baked in.

NicholasBoll (approved):

Issue #1204, opened 2021-08-10
Improve on Popup storybook example with Custom target


divyanshu023:
## 🐛 Bug Report

The storybook example for Popup with a Custom target does not entirely work. We see browser console errors

## To Reproduce
1. Create a Popup component with a custom target as described in storybook here: 
https://workday.github.io/canvas-kit/?path=/docs/components-popups-popup-react--custom-target

2. Notice the custom target component does not accept the ref which is passed to it by the Popup component. 

3. Open browser debugger window. There will be console errors asking you to use forward ref

## Expected Behavior
There should be no console errors and custom target should be able to accept a ref

## Actual Results

Custom target is not able to accept the ref

## Browser (if applicable)

n/a

## Link to repl or repo (highly encouraged)

https://workday.github.io/canvas-kit/?path=/docs/components-popups-popup-react--custom-target

### Error Output

Paste the error output here

```bash

```

or

Include a screenshot of a visual bug

PR #1209, opened 2021-08-10
docs(popup): Update custom target example to accept ref

divyanshu023:
JIRA: https://github.com/Workday/canvas-kit/issues/1204
## Checklist

- [x] `yarn test` passes
      applicable


modules/react/popup/stories/examples/CustomTarget.tsx
```diff
@@ -8,14 +8,18 @@ import {
...
 interface MyTargetProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   label: string;
 }
 
-const MyTarget = ({label, ...props}: MyTargetProps) => {
-  return <button {...props}>{label}</button>;
-};
+const MyTarget = React.forwardRef(({label, ...props}: MyTargetProps, ref) => {
+  const myRef = React.useRef(ref);
+	const elementRef = useForkRef(ref, myRef);
```
alanbsmith, line 19: Yessss! 👑

alanbsmith (approved):
Thanks, @divyanshu023!! Great catch and we appreciate the doc update. 👏🏻

modules/react/popup/stories/examples/CustomTarget.tsx
```diff
@@ -8,14 +8,18 @@ import {
...
+import {useForkRef} from '@workday/canvas-kit-react/common'
 
 interface MyTargetProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   label: string;
 }
 
-const MyTarget = ({label, ...props}: MyTargetProps) => {
-  return <button {...props}>{label}</button>;
-};
+const MyTarget = React.forwardRef(({label, ...props}: MyTargetProps, ref) => {
```
alanbsmith, line 17: Yesss! 👑

modules/react/popup/stories/examples/CustomTarget.tsx
NicholasBoll: I don't think we need to fork here. Using `React.forwardRef` will forward any ref assigned when our component is called. This means it might not be defined. If our component needs to use a `RefObject` (a `ref` with a `.current` value), _and_ we need to forward the ref, we must create our own and then fork it, passing the `elementRef` to the component in use.

Ref forking is also required anytime we need to merge 2 refs (a more accurate name might actually be `useMergeRef`, but the name was already solidified by material-ui). Popup hooks that deal with refs defined by the popup model also need to fork refs to make sure every ref instance points to the same value.

The following would be a reason to fork:

```tsx
const localRef = React.useRef<HTMLButtonElement>(null)
const elementRef = useForkRef(ref, localRef)

useMount(() => {
  // focus when mounted
  localRef.current.focus()
})

return <button ref={elementRef} />
```

This type of case is common and requires you to specify the element type `HTMLButtonElement`, even though it is already known. The following is exactly the same, but requires no Typescript overrides:

```tsx
const { localRef, elementRef } = useLocalRef(ref)
useMount(() => {
  // focus when mounted
  localRef.current.focus()
})

return <button ref={elementRef} />
```

My guess is this pattern was copied from the internal implementation of `Popup.Target`, which needs to merge the `model.state.targetRef` with a forwarded `ref`, which is not the case here.

modules/react/popup/stories/examples/CustomTarget.tsx
```diff
@@ -8,14 +8,18 @@ import {
...
+  const myRef = React.useRef(ref);
+	const elementRef = useForkRef(ref, myRef);
+
+  return <button {...props} ref={elementRef}>{label}</button>;
```
NicholasBoll, line 21: The ref does not need to be forked. See https://github.com/Workday/canvas-kit/pull/1209/files#r686991626

```suggestion
  return <button {...props} ref={ref}>{label}</button>;
```

NicholasBoll (approved):

modules/react/popup/stories/examples/CustomTarget.tsx
```diff
@@ -8,14 +8,15 @@ import {
   useInitialFocus,
   useReturnFocus,
 } from '@workday/canvas-kit-react/popup';
+import {useForkRef} from '@workday/canvas-kit-react/common'
```
NicholasBoll, line 11: Forgot to remove this import

```suggestion
```

modules/react/popup/stories/examples/CustomTarget.tsx
```diff
@@ -13,9 +13,9 @@ interface MyTargetProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   label: string;
 }
 
-const MyTarget = ({label, ...props}: MyTargetProps) => {
-  return <button {...props}>{label}</button>;
-};
+const MyTarget = React.forwardRef(({label, ...props}: MyTargetProps, ref) => {
+  return <button {...props} ref={ref}>{label}</button>;
+});
```
NicholasBoll, line 18: Type error dealing with the forwarded ref...

```suggestion
const MyTarget = React.forwardRef<HTMLButtonElement, MyTargetProps>(({label, ...props}, ref) => {
  return (
    <button {...props} ref={ref}>
      {label}
    </button>
  );
});
```

PR #1219, opened 2021-08-13
ci: Add Chromatic baseline to support branches

NicholasBoll:
This PR will enable automatic rebaselining in support branches. Without this change, every PR set to be merged to a support branch will have to re-accept the same changes in every PR after a visual change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,14 @@ jobs:
           branch: master
 
   - stage: support
+    script:
+      - >- # Chromatic relies on a built Storybook, so exit early if build-storybook fails
+        yarn build-storybook --quiet &&
+        yarn chromatic --quiet --auto-accept-changes --exit-once-uploaded --storybook-build-dir docs &&
+    after_failure:
+      - node utils/report-failure.js
+    env:
+      - CHROMATIC_APP_CODE="dlpro96xybh"
     deploy:
       - provider: script
         script: npm config set //registry.npmjs.org/:_authToken=$NPM_PUBLISH_TOKEN && node utils/publish-canary.js


.travis.yml
```diff
@@ -77,6 +77,14 @@ jobs:
...
 
   - stage: support
+    script:
+      - >- # Chromatic relies on a built Storybook, so exit early if build-storybook fails
+        yarn build-storybook --quiet &&
+        yarn chromatic --quiet --auto-accept-changes --exit-once-uploaded --storybook-build-dir docs &&
+    after_failure:
+      - node utils/report-failure.js
+    env:
+      - CHROMATIC_APP_CODE="dlpro96xybh"
```
alanbsmith, line 87: Just double-checking: I assume it's okay for this to be public.

NicholasBoll: Yes. It is already public here: https://github.com/Workday/canvas-kit/pull/1219/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R65

NicholasBoll: Making this code a secret would involve either switching to CircleCI, or a lot of extra work with Github Actions. I've talked to Chromatic and Cypress about the limitation on secrets in forked pull requests. Both providers are okay with a public key. There is no risk to Workday or our repository other than someone using the key to push junk results in our or incur monthly costs in overage which is already a risk if someone opens up junk PRs. The latter risk is mitigated by us approving new contributors which is better managed than a public key.

alanbsmith: Cool. Thanks for the context. 👍🏻

PR #1220, opened 2021-08-14
docs: Update inter-linking to match Canvas Site

NicholasBoll:
This PR adds Canvas Site URLs to Storybook MDX files and adds a `routes.js` file that maps the Canvas Site routes to Storybook stories. Also added is a Cypress test that makes sure all the Storybook stories are valid to prevent incorrect links.

This change will make it so that Storybook links and Canvas Site links are the same. We decided to put the burden of maintaining these connections on the Canvas Kit side.

After this PR, we'll need to maintain the `.storybook/routes.js` file when adding new components or changing URL routes on the Canvas Site. The CI job will make sure the Storybook IDs are always valid. Perhaps when the Canvas Site launches, we can have a Cypress job that also makes sure Canvas Site URLs are also valid.


modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -29,9 +29,10 @@ any questions about the update.
 
 ## Codemod
 
-We've introduced a [new codemod package](../../codemod) you can use to automatically update your
-code to work with a majority of the breaking changes in the migration from Canvas Kit v4 to v5.
-Simply run:
+We've introduced a
+[new codemod package](https://github.com/Workday/canvas-kit/tree/master/modules/codemod) you can use
```
NicholasBoll, line 33: Removing all relative links in any MDX file consumed by the Canvas Site.

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
```diff
@@ -823,8 +825,9 @@ interface MyButtonProps extends ButtonProps {
 
 ### Card
 
-Card is now a [compound component](./COMPOUND_COMPONENTS.mdx) composed of a `Card.Body` and an
-optional `Card.Heading`. This allows direct access to the heading and body elements.
+Card is now a [compound component](/getting-started/for-developers/resources/compound-components/)
```
NicholasBoll, line 828: This change will keep Storybook links working. This will break the link in Github, but I think that's okay since the document is not easy to find while looking at the file in Github.

.storybook/routes.js
```diff
@@ -0,0 +1,61 @@
...
+  '/components/inputs/text-input/': 'components-inputs-text-input-react--basic',
+  '/components/inputs/radio/': 'components-inputs-radio-react--basic',
+  '/components/inputs/switch/': 'components-inputs-switch-react--basic',
+  '/components/indicators/banner/': 'components-indicators-banner-react--basic',
+  '/components/indicators/loading-animation/':
+    'components-indicators-loading-animation-react--basic',
+  '/components/indicators/skeleton/': 'components-indicators-skeleton-react--basic',
+  '/components/indicators/status-indicator/': 'components-indicators-status-indicator-react--basic',
+  '/components/inputs/text-area/': 'components-inputs-textarea-react--basic',
+  '/components/navigation/breadcrumbs/': 'preview-breadcrumbs-react--default',
```
NicholasBoll, line 37:
```suggestion
  '/components/navigation/breadcrumbs/': 'preview-breadcrumbs-react--basic',
```

.storybook/routes.js
```diff
@@ -0,0 +1,61 @@
+// List of routes The key is the Canvas Site URL The value is the first Storybook example ID. This
+// prevents odd scrolling. This file is used by the `webpack-loader-redirect-mdx-to-github` loader
+// and will rewrite the URL to point to the correct URL for the story. The Storybook key will be
+// rewritten to something like `?path=/docs/${id}`
+const routes = {
+  'getting-started/introduction': 'welcome-getting-started--page',
```
jamesfan, line 6: This should include the beginning and trailing slashes as well, right?

```suggestion
  '/getting-started/introduction/': 'welcome-getting-started--page',
```

Also, just to clarify, we're including a mapping for every route we think could be linked to from an MDX exported by `@workday/canvas-kit-docs`? That's why `/getting-started/introduction/` is here (because we may want to link to it from, say, `CONTRIBUTING.mdx`), but `/content/overview/` isn't (because it's unlikely we'd want to link to it from anything in `@workday/canvas-kit-docs`).

In both cases, there's no `canvas-kit` equivalent: [`/getting-started/introduction/`](https://ghe.megaleo.com/pages/design/canvas-site/getting-started/introduction) and [`welcome-getting-started--page`](https://workday.github.io/canvas-kit/?path=/docs/welcome-getting-started--page) aren't the same, and [`/content/overview/`](https://ghe.megaleo.com/pages/design/canvas-site/content/overview) isn't remotely close to anything in `canvas-kit`. The difference is `/getting-started/introduction/` may need to be linked to from an MDX in `@workday/canvas-kit-docs` so we need to provide a mapping for it (even if it's an imperfect one).

Am I understanding that right?

NicholasBoll: Yes, I did forget the leading slash. I am trying to make a 1:1 mapping of links. It is possible that the getting started page is really a dev getting started page and should actually be linked to something like `/getting-started/for-developers/introduction/`.

There were many links that I say no equivalent for. It is possible that there's a special value type that allows us to link from Storybook straight to the Canvas Site since the content may only live there. Something like `canvas-site` or something that indicates no perfect matching and to redirect to the canvas site instead. This list will be subject to change over time and will be necessary to keep links up-to-date with Canvas Site content.

The main goal  is to keep interlinking mostly working within Storybook and Canvas Site without a lot of weird loops. For example, I've entirely been avoiding linking between documents to avoid the issue.

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
anicholls: This makes me a bit nervous. We have linked to github docs a lot, so any historical references (old slack posts, etc.) will have broken links after this. At the very least, we should swap these to fully qualified URLs once the site is live.

anicholls (approved):
While this approach definitely works, I'm hesitant to break the links in GitHub. Maybe it's fine if we swap them, to fully qualified once the site is launched. We could potentially flip this, where we link to storybook by default and only rewrite URLs that exist on the site, but I know we're trying to push people away from Storybook 😕 .

modules/docs/mdx/5.0-MIGRATION-GUIDE.mdx
NicholasBoll: We've broken URLs in the past when we moved these component documents from the root of the project to the `docs` package. It is possible to make links work for Github (interlinking), Storybook, and the Canvas Site, but complexity rises significantly. We have to drop relative links between Github docs at the least since the URLs won't exist on the Canvas Site, but then we run into a issue with active branches unless we always assume `master`.

I think it's time to drop support for Github doc rendering. We used Github's interlinking because we didn't have a good way to link between Storybook docs. The Canvas Site url scheme presents a better way of linking between documents and, with this change, works on both the Canvas Site and Storybook.

NicholasBoll:
> While this approach definitely works, I'm hesitant to break the links in GitHub. Maybe it's fine if we swap them, to fully qualified once the site is launched. We could potentially flip this, where we link to storybook by default and only rewrite URLs that exist on the site, but I know we're trying to push people away from Storybook 😕 .

We need to drop support for something. I don't think we should try to retain proper linking between Github, Storybook, and the Canvas Site. It is possible, but it would be extra effort to maintain. I'd prefer links to Storybook over Github if I had to choose since our Storybook docs are more curated than a doc rendering on Github. Honestly, I'd prefer links to the Canvas Site since it looks more professional, but most likely the Canvas Site will be slightly behind.

.storybook/routes.js
```diff
@@ -0,0 +1,61 @@
...
+  '/getting-started/for-developers/resources/api-pattern-guidelines/':
+    'welcome-dev-docs-api-pattern-guidelines--page',
+  '/getting-started/for-developers/resources/compound-components/':
+    'welcome-dev-docs-compound-components--page',
+  '/getting-started/for-developers/resources/creating-compound-components/':
+    'welcome-dev-docs-creating-compound-components--page',
+  '/getting-started/for-developers/resources/testing/': 'welcome-dev-docs-testing--page',
+  '/getting-started/for-developers/resources/4.0-migration-guide':
+    'welcome-migration-guides-v4-0--page',
+  '/getting-started/for-developers/resources/5.0-migration-guide':
```
NicholasBoll, line 57:
```suggestion
  '/getting-started/for-developers/resources/4.0-migration-guide/':
    'welcome-migration-guides-v4-0--page',
  '/getting-started/for-developers/resources/5.0-migration-guide/':
```

Issue #894, opened 2020-10-30
Remove `aria-label` requirement restrictions for IconButton


lychyi:
## 🐛 Bug Report

Currently for `IconButton` (**TODO** and I'm sure other components) we require `aria-label` as a prop/attribute but it really should require _either_ `aria-label` and `aria-labelledby`.

I would vouch to either 1) remove the restriction entirely and promote accessibility through documentation or 2) add a union type that specifies that we require one or the other.

NicholasBoll:
Closed by #1401

PR #1232, opened 2021-08-19
refactor: SidePanel [WIP]

alanbsmith:
## Checklist

- [ ] `yarn test` passes
      applicable

### DIFF
--- a/modules/labs-react/common/lib/Box.tsx
+++ b/modules/labs-react/common/lib/Box.tsx
@@ -22,7 +22,7 @@ export type BoxProps = BorderStyleProps &
     children?: React.ReactNode;
   };
 
-const omittedProps = ['display', 'color'];
+const omittedProps = ['display', 'color', 'height', 'overflow', 'width'];
 
 const shouldForwardProp = (prop: string) => {
   return isPropValid(prop) && !omittedProps.includes(prop);
--- a/modules/preview-react/side-panel/lib/NewSidePanel/SidePanelToggleButton.tsx
+++ b/modules/preview-react/side-panel/lib/NewSidePanel/SidePanelToggleButton.tsx
@@ -1,0 +1,70 @@
+/** @jsx jsx */
+import * as React from 'react';
+import {css, jsx} from '@emotion/core';
+import {space} from '@workday/canvas-kit-react/tokens';
+import {createComponent, useModelContext} from '@workday/canvas-kit-react/common';
+import {IconButton, IconButtonProps} from '@workday/canvas-kit-react/button';
+import {transformationImportIcon} from '@workday/canvas-system-icons-web';
+
+import {SidePanelModel, SidePanelModelContext, useSidePanelToggleButton} from './hooks';
+
+export interface SidePanelToggleButtonProps
+  extends Omit<IconButtonProps, 'aria-label'>,
+    React.ButtonHTMLAttributes<HTMLButtonElement> {
+  'aria-label'?: string;
+  /**
+   * Optionally pass a model directly to this component. Default is to implicitly use the same
+   * model as the container component which uses React context. Only use this for advanced use-cases
+   */
+  model?: SidePanelModel;
+}
+
+export const SidePanelToggleButton = createComponent(IconButton)({
+  displayName: 'SidePanel.ToggleButton',
+  Component: (
+    {
+      model,
+      icon = transformationImportIcon,
+      variant = 'plain',
+      ...elemProps
+    }: SidePanelToggleButtonProps,
+    ref,
+    Element
+  ) => {
+    const localModel = useModelContext(SidePanelModelContext, model);
+    const props = useSidePanelToggleButton(localModel, elemProps, ref);
+
+    // Note: Depending on the collapsed width, the button could "jump" to it's final position.
+    const buttonStyles = css({
+      position: 'absolute',
+      top: space.m,
+      right: !localModel.state.expanded
+        ? 0
+        : localModel.state.origin === 'start'
+        ? space.s
+        : undefined,
+      left: !localModel.state.expanded
+        ? 0
+        : localModel.state.origin === 'end'
+        ? space.s
+        : undefined,
+      margin: !localModel.state.expanded ? 'auto' : 0, // to override the -8px margin for IconButton.Plain
+      // transform:
+      //   localModel.state === 'collapsed' || localModel.state === 'collapsing'
+      //     ? `scaleX(${localModel.origin === 'left' ? '1' : '-1'})`
+      //     : `scaleX(${localModel.origin === 'left' ? '-1' : '1'})`,
+    });
+
+    return (
+      // @ts-ignore aria-label type error here. The user will decide to use aria-label or aria-labelledby
+      <Element
+        ref={ref}
+        css={buttonStyles}
+        type="button"
+        variant={variant}
+        icon={icon}
+        {...props}
+      />
+    );
+  },
+});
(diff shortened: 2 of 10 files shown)


modules/labs-react/common/lib/Box.tsx
```diff
@@ -22,7 +22,7 @@ export type BoxProps = BorderStyleProps &
     children?: React.ReactNode;
   };
 
-const omittedProps = ['display', 'color'];
+const omittedProps = ['display', 'color', 'height', 'overflow', 'width'];
```
alanbsmith, line 25: I saw these oddly being passed through with the `SidePanel` section using `Box`.

modules/preview-react/side-panel/lib/NewSidePanel/hooks/useSidePanel.ts
```diff
@@ -0,0 +1,52 @@
...
+    model.state.expanded === 'expanded'
+      ? `${motion.collapse} ease-out 200ms`
+      : `${motion.expand} ease-out 200ms`;
+
+  return {
+    ref,
+    css: {
+      // should Box have an animation prop?
+      animation: `${animation}`,
+    },
```
alanbsmith, line 48: This is an interesting situation. Typically, we've left styling out of behavior hooks, but this is a situation where it feels like it should live here.

modules/preview-react/side-panel/lib/NewSidePanel/hooks/useSidePanel.ts
```diff
@@ -0,0 +1,52 @@
...
+    onAnimationStart: handleAnimationStart,
+    onAnimationEnd: handleAnimationEnd,
```
alanbsmith, line 50: We could also have `width` and `maxWidth` styles live in this hook. But should they be in the CSS prop or should standalone? E.g.
 
 ```ts
     const width =
      localModel.state.expanded === 'expanded'
        ? localModel.state.expandedWidth
        : localModel.state.collapsedWidth;
 
 return {
   ref,
   // provided we add an `animation` prop to `Box`
   animation: '',
   width: width,
   maxWidth: width,
 }
 ```
 
 Or:
 
 ```ts
     const width =
      localModel.state.expanded === 'expanded'
        ? localModel.state.expandedWidth
        : localModel.state.collapsedWidth;
 
 return {
   ref,
   css: {
     animation: '',
     width: width,
     maxWidth: width,
   },
   onAnimationStart: handleAnimationStart,
   onAnimationEnd: handleAnimationEnd,
 }
 ```
 
The first implementation assumes a `Box` API, and merges the styles nicely. But is making that assumption a problem? You couldn't throw in a non-`Box` component without extracting those props into styles from the props and putting them into a style object.

The nice thing about the second solution is that it doesn't assume the element being used, but it will necessarily override the `width` and `maxWidth` Box props.

modules/preview-react/side-panel/lib/NewSidePanel/SidePanelToggleButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+import {space} from '@workday/canvas-kit-react/tokens';
+import {createComponent, useModelContext} from '@workday/canvas-kit-react/common';
+import {IconButton, IconButtonProps} from '@workday/canvas-kit-react/button';
+import {transformationImportIcon} from '@workday/canvas-system-icons-web';
+
+import {SidePanelModel, SidePanelModelContext, useSidePanelToggleButton} from './hooks';
+
+export interface SidePanelToggleButtonProps
+  extends Omit<IconButtonProps, 'aria-label'>,
+    React.ButtonHTMLAttributes<HTMLButtonElement> {
```
NicholasBoll, line 13: What's the reason for extending `React.ButtonHTMLAttributes` here? That should be handled automatically.

modules/preview-react/side-panel/lib/NewSidePanel/SidePanelToggleButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+  'aria-label'?: string;
```
NicholasBoll, line 14: I understand this is necessary as well as the `Omit` above because `IconButton` requires an `aria-label`. I think we have an issue that removes this requirement because there are plenty of examples where an `aria-label` is not required.

NicholasBoll: https://github.com/Workday/canvas-kit/issues/894

modules/preview-react/side-panel/lib/NewSidePanel/SidePanelToggleButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+        : undefined,
+      margin: !localModel.state.expanded ? 'auto' : 0, // to override the -8px margin for IconButton.Plain
+      // transform:
+      //   localModel.state === 'collapsed' || localModel.state === 'collapsing'
+      //     ? `scaleX(${localModel.origin === 'left' ? '1' : '-1'})`
+      //     : `scaleX(${localModel.origin === 'left' ? '-1' : '1'})`,
+    });
+
+    return (
+      // @ts-ignore aria-label type error here. The user will decide to use aria-label or aria-labelledby
```
NicholasBoll, line 59: #894

NicholasBoll: I don't like the `@ts-ignore`. Perhaps we should remove the restriction in `IconButton` now. It won't be a breaking change.

modules/preview-react/side-panel/lib/NewSidePanel/hooks/useSidePanelModel.ts
```diff
@@ -0,0 +1,86 @@
...
+export type SidePanelOrigin = 'start' | 'end';
+export type SidePanelTransitionStates = 'collapsed' | 'expanded' | 'collapsing' | 'expanding';
+
+export type SidePanelState = {
+  collapsedWidth: number | string;
+  expandedWidth: number | string;
+  expanded: SidePanelTransitionStates;
+  labelId: string;
+  origin: SidePanelOrigin;
+  panelId: string;
```
NicholasBoll, line 19: I've been using `id` and extrapolating an id from there. Tabs does this. If an application developer wishes to override these ids, they can do so on the subcomponent directly rather than in the model config.

modules/preview-react/side-panel/lib/NewSidePanel/SidePanelToggleButton.tsx
alanbsmith: Yeah, probably don't need to. I grabbed that from the current `SidePanel`.

modules/preview-react/side-panel/lib/NewSidePanel/SidePanelToggleButton.tsx
alanbsmith: Yeah, `aria-label` needs to be optional.

modules/preview-react/side-panel/lib/NewSidePanel/SidePanelToggleButton.tsx
alanbsmith: Yep. Agreed.

modules/preview-react/side-panel/lib/NewSidePanel/hooks/useSidePanelModel.ts
alanbsmith: I thought about that. I agree. applying an `id` to the element directly feels like the most natural way to override  it. The `panelId` name comes from the original API (it made more sense there).

alanbsmith:
I'm going to close this for now. If / when the work is ready and prioritized, we can reopen it or take bits that are helpful.

PR #1248, opened 2021-09-08
fix(tooltip): Fix tooltip placement overlap on keyboard focus

### LINKED ISSUE #1071: Tooltip overlaps keyboard focus border
## 🐛 Bug Report

Tooltip overlaps keyboard focus border

Steps to reproduce:

## To Reproduce

1. Open the storybook link (wd-components)
2. Navigate to popup widget --> Adapter canvas-kit tooltips.
3. Using keyboard, navigate to Tooltip widget.
4. Bring the focus on tooltip trigger element. Tooltips is displayed.
5. Notice the focus on the button.

## Expected Behavior

Tooltip should not overlap the button keyboard focus

## Actual Results

Tooltip overlaps the button keyboard focus

## Browser (if applicable)
Screen Shot 2021-05-14 at 1 15 40 PM

Safari-Mac, Windows - IE11, Chrome

Screen Reader: JAWS, NVDA, VO

### Error Output

Screen Shot 2021-05-14 at 1 16 27 PM

NicholasBoll:
Fixes #1071


alanbsmith (approved):
🐼

modules/react/tooltip/lib/TooltipContainer.tsx
```diff
@@ -68,6 +69,28 @@ export const TooltipContainer = styled('div')<TooltipContainerProps>(
...
+    },
+    '[data-popper-placement="top-end"] &, [data-popper-placement="bottom-end"] &': {
+      right: -spaceNumbers.xxxs,
+    },
+    '[data-popper-placement="left-start"] &, [data-popper-placement="right-start"] &': {
+      top: -spaceNumbers.xxxs,
+    },
+    '[data-popper-placement="left-end"] &, [data-popper-placement="right-end"] &': {
+      bottom: -spaceNumbers.xxxs,
+    },
```
NicholasBoll, line 93: I'm not 100% sold on these changes, but here's what they do:

**Before**
before

**After**
after

The subtle change is how the Tooltip aligns to the edge of the target. In the before, there's 4px spacing in the `before-start` to the left edge of the card. It is hard to see because of the border. This change removes that 4px spacing. The reason I did this was the original issue adds an additional 2px (focus ring thickness for consistency) which makes the gap more noticeable at 6px. These relative offsets fix the alignment issues. The focus ring still shifts the tooltips away from the edge, but it is much more difficult to tell. Here's what it looks like:

focus ring

You can see the focus ring adds 4px on it's own because of the double ring which is done to ensure visibility against either a dark or light background. I don't think we want to align to the edge of the focus ring since that is outside the bounds of the original button.

PR #1259, opened 2021-09-10
feat(modal): Add support for scrolling a modal

### LINKED ISSUE #790: Modal: Support scrolling entire modal / overlay for overflowing content.
## 🚀 Feature Proposal
For modals with a large amount of content that will not fit on the screen, it would be nice to have an option for the overlay itself to scroll (as opposed to always requiring the consumer to make the content of the modal itself scrollable).

An additional requirement of this feature would be that the background content should not be scrollable while the modal is open. This would also likely be useful even for modals that handle scrolling within their content.

## Motivation

Currently on PEX we have several use cases for modals where the content is user defined, and can be quite large. For example, on the Home page, we have modals for announcements that are user configured and can contain media and rich text. Likewise is true for Knowledge Base articles which can be shown in modals in many places and are almost always larger than will fit on screen at any resolution.

## Example

modal-overlay-scroll

NicholasBoll:
## Summary

Fixes #790

This PR changes the structure of Modals slightly by removing the extra `div` in the `Modal.Body` and `Popup.Body` and instead adds the type that `Card.Body` adds manually. This allows the `Modal.Body` to be more easily styled, but may be a breaking change. This change also allows overflow to be applied to Modals.

category


modules/react/popup/lib/PopupCard.tsx
```diff
@@ -87,7 +88,7 @@ export const PopupCard = createComponent('div')({
         padding={padding}
         {...props}
       >
-        <Card.Body>{children}</Card.Body>
+        {children}
```
NicholasBoll, line 91: This removes the extra wrapper `div` in Popups and instead manually brings in the `type.levels.subtext.large` typing to the `<StyledPopupCard>` styled component. This allows better style overrides.

modules/react/modal/stories/examples/BodyOverflow.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const BodyOverflow = () => {
+  const handleDelete = () => {
+    console.log('Deleted item');
+  };
+
+  return (
+    <Modal>
+      <Modal.Target as={DeleteButton}>Delete Item</Modal.Target>
+      <Modal.Overlay>
+        <Modal.Card as={VStack} spacing={0} maxHeight={`calc(100vh - ${space.xl} * 2)`}>
```
NicholasBoll, line 17: Question is should `Modal.Card` always be a columnar flex container? This would be a breaking change, but would allow for automatic overflow handling without a the large amount of work otherwise. For example, this example works, but isn't compliant with the different Modal card margins depending on screen size. That's probably better handled by the component.

breno-felipe (approved):

modules/react/popup/lib/PopupCard.tsx
NicholasBoll: Already done

modules/react/modal/stories/Modal.stories.mdx
```diff
@@ -81,6 +83,31 @@ requires a `label` prop.
 
 <ExampleCodeBlock code={CustomTarget} />
 
+### Body Content Overflow
+
+There are a few ways to handle body content overflowing. The easiest is to use `Modal.Card` as a
+vertical stack using `VStack` and setting the max height to the view height minus the desired margin
+around the card (default margin is `xl` or `40px`). The `VStack` is a flexbox container than
```
mannycarrera4, line 90: `is a flexbox container than` -> `is a flexbox container that`

modules/react/modal/stories/Modal.stories.mdx
```diff
@@ -81,6 +83,31 @@ requires a `label` prop.
...
+controls the height of the children. From there, you set up the content area to have an
+`overflow: scroll` style. Usually, you want the header and action buttons to stay put. In this
+example, the `Modal.Card` has an `as={VStack}` with no spacing, and a max height of
+`calc(100vh - ${space.xl} * 2)` which is the viewport height minus the default margin x 2 (top and
+bottom). The `Modal.Heading`, `Modal.Body`, and `HStack` are the 3 flexbox children that will try to
+equally fill the container up to the max height defined. The `overflow: scroll` on the `Modal.Body`
+is what restricts the content to an overflow container so that it scrolls. The effectively fixes the
```
mannycarrera4, line 97: `The effectively fixes` -> `This effectively fixes`

modules/react/modal/stories/Modal.stories.mdx
```diff
@@ -81,6 +83,31 @@ requires a `label` prop.
...
+position of the heading and action buttons.
+
```
mannycarrera4, line 99: This i an awesome description, i would just break it up into smaller paragraph or bullet points so it's a little easier to follow

modules/react/modal/stories/examples/BodyOverflow.tsx
mannycarrera4: Most content is in a column right? I think that makes sense even though it might be a breaking change

mannycarrera4: Especially if consumers can use or Hstack and styled properties to override things, it think it should be fine

modules/react/modal/stories/Modal.stories.mdx
NicholasBoll: Use the suggestion feature: 
image

It makes it easy to make suggestions that are easy to commit

modules/react/modal/stories/examples/BodyOverflow.tsx
NicholasBoll: I mean the Modal is a `VStack` by default. A modal is almost always a vertical stack of a header, body, and optional footer. Modal variants also only add to vertical stack elements: image, header, body, optional footer. It would probably break a lot of modals. The change would be from static block positioning (block elements naturally stack vertically) to flexbox vertical stacking. The problem might be how spacing is handled.

A naive approach might be built-in spacing with the option to override. This example adds no spacing since `Heading` has margins built in (would be removed). You can see `paddingTop` added to the `HStack` later. Basically building in spacing to `VStacks` would require an overhaul to built-in margin/padding of components. The transition from children control layout spacing to containers controlling layout spacing.

NicholasBoll: These examples show how different overflow can be achieved by direct manipulation of styling of the subcomponents. It would be nice for some styling to be built-in, but that might cause cascading breaking changes.

modules/react/modal/stories/Modal.stories.mdx
NicholasBoll: I made body overflow be the default, which removes the need for all this documentation.

modules/react/modal/stories/examples/BodyOverflow.tsx
NicholasBoll: I updated the `Popup.Card` to be a flex container and the complexity of this example is no longer required.

modules/labs-react/layout/lib/Stack.tsx
```diff
@@ -88,19 +72,19 @@ export const HStack = createComponent('div')({
...
-};
+}
 
 export const VStack = createComponent('div')({
   displayName: 'VStack',
-  Component: ({children, flexDirection = 'column', ...elemProps}: VStackProps, ref, Element) => {
+  Component: ({children, ...elemProps}: VStackProps, ref, Element) => {
     return (
-      <Stack as={Element} ref={ref} flexDirection={flexDirection} {...elemProps}>
+      <Stack as={Element} ref={ref} flexDirection="column" {...elemProps}>
```
NicholasBoll, line 87: Changes to this file are optional. The changes simplify `Stack` by removing unnecessary Typescript and decreasing prop name duplication.

mannycarrera4 (approved):

PR #1260, opened 2021-09-10
fix(tooltip): Update OverflowTooltip to support non-elements

NicholasBoll:
This change adds support for the following use case:

```tsx
<OverflowTooltip>
  <SomeCustomComponent />
</OverflowTooltip>
```

The problem was `react-innertext` was used to determine the text of the content inside the `OverflowTooltip`. `react-innertext` uses VDOM `children` to determine text. But in this case, `children` has no text, just a single child of type `SomeCustomComponent`. Therefore calculated `titleText` is `''` and the tooltip will never show. The fix is to use [DOM innerText](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText) instead which should work regardless of React JSX structure. This should be safe because the text is always calculated based on `onFocus` or `onMouseEnter` events; the same time that overflow is calculated.

The `Ellipsis` example was updated with this use case as well as a Cypress test ensuring it works properly.

### DIFF
--- a/modules/react/tooltip/lib/OverflowTooltip.tsx
+++ b/modules/react/tooltip/lib/OverflowTooltip.tsx
@@ -99,15 +99,17 @@ export const OverflowTooltip = ({
   children,
   ...elemProps
 }: OverflowTooltipProps) => {
-  const titleText = innerText(children);
-  const {targetProps, popperProps, tooltipProps} = useTooltip({type: 'label', titleText});
+  const [titleText, setTitleText] = React.useState('');
+  const {targetProps, popperProps, tooltipProps} = useTooltip({type: 'muted'});
 
   const onMouseEnter = (event: React.MouseEvent<HTMLElement>) => {
+    setTitleText(event.currentTarget.innerText);
     if (isOverflowed(event.currentTarget)) {
       targetProps.onMouseEnter(event);
     }
   };
   const onFocus = (event: React.FocusEvent<HTMLElement>) => {
+    setTitleText(event.currentTarget.innerText);
     if (isOverflowed(event.currentTarget)) {
       targetProps.onFocus(event);
     }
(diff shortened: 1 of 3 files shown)


modules/react/tooltip/lib/OverflowTooltip.tsx
```diff
@@ -99,15 +99,17 @@ export const OverflowTooltip = ({
   children,
   ...elemProps
 }: OverflowTooltipProps) => {
-  const titleText = innerText(children);
-  const {targetProps, popperProps, tooltipProps} = useTooltip({type: 'label', titleText});
+  const [titleText, setTitleText] = React.useState('');
+  const {targetProps, popperProps, tooltipProps} = useTooltip({type: 'muted'});
```
NicholasBoll, line 103: `muted` is the appropriate type in this case. This means a screen reader will not have any special affordances for overflow tooltips since overflowing is a visual concern and the calculated "name" of the element should already be known to assistive technology. Leaving as `label` in this case will actually remove the name from assistive technology because the `aria-label` will be calculated as `''` because `titleText` is an empty string until the user interacts with the element, either via `focus` or `mouseenter` events. `OverflowTooltip` is meant to be a visual aid when text is overflowed.

modules/react/tooltip/stories/examples/Ellipsis.tsx
```diff
@@ -28,6 +43,9 @@ export const Ellipsis = () => {
           Super Mega Ultra Long Content With Max Width
         </button>
       </OverflowTooltip>
+      <OverflowTooltip>
+        <CustomContent />
+      </OverflowTooltip>
```
NicholasBoll, line 48: This use-case previously failed with a tooltip of empty content.

alanbsmith (approved):
🐘

PR #1263, opened 2021-09-13
fix(tooltip): Fix OverflowTooltip with SVG in IE11

NicholasBoll:
IE11 doesn't define the `children` property in `Element`. It defines `children` as part of the `HTMLElement` interface. This means that `findEllipsisElement` will fail if `OverflowTooltip` contains any SVG element. An issue has been filed with MDN to update their docs. The code now does specific testing against an `HTMLElement` for `innerText` (not defined on `SVGElement` in any browser) as well as testing the presence of `.children` to avoid IE11 crashing on SVG elements.

More info here: https://github.com/mdn/content/issues/8890


alanbsmith (approved):
That is a wild edge case. Thanks for the fix!

PR #1266, opened 2021-09-14
fix: Update import of static states when creating labs component [skip ci]

mannycarrera4:
## Summary

I noticed when running the `createComponent` script for a labs component, it imported `StaticStates` from labs, so moved it so that when you run the script, the correct import is added.

category


utils/create-component/templates/react/stories_VisualTesting.js
```diff
@@ -2,8 +2,7 @@
 
 module.exports = (modulePath, storyPath, pascalCaseName, rootPath) => `import React from 'react';
 
-import {StaticStates} from '@workday/canvas-kit-preview-react/tokens';
-import {ComponentStatesTable} from '@workday/canvas-kit-labs-react/common';
+import {ComponentStatesTable, StaticStates} from '@workday/canvas-kit-labs-react/common';
```
mannycarrera4, line 5: should be from common, not labs common

alanbsmith (approved):
Thanks for the fix, @mannycarrera4 👍🏻

Issue #1244, opened 2021-09-02
Review and Update the MDX Doc for Tab


myvuuu:
## 💡 Suggestion

Review and update the MDX doc for Tab so that:

- [ ] The MDX lines up with the [standardized MDX template](https://github.com/Workday/canvas-kit/discussions/1131)
- [ ] Code snippets and examples are up-to-date and accurate

PR #1279, opened 2021-09-22
docs(tabs): Update Tabs MDX to adhere to standardized template [skip ci]

jamesfan:
## Summary

Fixes: #1244

Update the Tabs MDX to adhere to the [standardized MDX template](https://github.com/Workday/canvas-kit/discussions/1131).

category

---

## Checklist

- [x] Prepare Design Site PR with updated Storybook links


modules/react/tabs/stories/Tabs.stories.mdx
```diff
@@ -22,80 +33,67 @@ yarn add @workday/canvas-kit-react
...
+
+`Tabs` includes a container `Tabs` component and the following subcomponents: `Tabs.List`,
+`Tabs.Item` and `Tabs.Panel`.
 
-### Basic Usage
+<ExampleCodeBlock code={Basic} />
 
-Below are examples we expect will work for most cases. Please refer to the
-[Advanced Usage](#advanced-usage) section for details on supporting custom implementations.
+### Hoisted Model
```
jamesfan, line 43: I moved the `Hoisted Model` example right below the `Basic Example`. Previously, it was much further down in the Advanced Usage section (now gone) which was characterized with "we expect most people won't need to read this section," but it seems like a fairly common use case to me. Thoughts?

modules/react/tabs/stories/Tabs.stories.mdx
```diff
@@ -22,80 +33,67 @@ yarn add @workday/canvas-kit-react
...
-to have only a single tab panel using attribute overrides and callbacks. More info about attributes
-and callbacks are in property tables below for each subcomponent. In this example, we'll use a
-Hoisted model and the `activeTab` property of the state to show content of our `contents` object.
+In this example, we use a hoisted model and the `activeTab` property of the state to show content
+from the `contents` object.
 
 <ExampleCodeBlock code={SinglePanel} />
 
-#### Dynamic Tabs
+### Dynamic Tabs
```
jamesfan, line 96: I was a little confused by the `Dynamic Tabs` example. I thought I could move Add Tab above the `tabs.map` that creates the `Tabs.Items` and still have Add Tab show up at the end of the tabs list (given the `index={tabs.length}`), but that resulted in Add Tab showing up at the beginning of the list.

Also, removing use of the `index` property from the example code doesn't seem to affect it. Am I misunderstanding how `index` is supposed to work?

modules/react/tabs/stories/Tabs.stories.mdx
```diff
@@ -108,84 +106,149 @@ the end.
...
 
 <PropsTable of={Tabs.Panel} />
 
+## Hooks
+
+### useTabsModel
+
+#### Usage
+
+Creates a `TabsModel` using the provided configuration object.
```
jamesfan, line 192: Most of the changes in this PR are me shifting around and wordsmithing content, but the `Hooks` section is all new. Let me know if you think I'm missing or misrepresenting anything here.

modules/react/tabs/stories/Tabs.stories.mdx
NicholasBoll: Index is used to inform the model where in the list each tab should go. This example always adds a new tab at the end. If you add a tab at the middle, but didn't send an index, the rendering would be correct, but keyboard navigation will get messed up. What will happen is the tab will render at the middle of the list, but register with the model at the end of the list. So when you're focus is to the left of the new tab and you hit the right key, it will skip over the new tab. When you're focus is on the last visible tab and you hit the right key it will jump to the new tab in the middle of the list.

modules/react/tabs/stories/Tabs.stories.mdx
```diff
@@ -108,84 +106,151 @@ the end.
...
+Use `useTabsModel` if you wish to create a `TabsModel` rather than relying on the `Tabs` component
+to create and use its own model internally. Doing so allows you to access the model's state and
+events directly.
+
+```tsx
+const model = useTabsModel({
+  onActivate({data, prevState}) {
+    console.log('Activated Tab', data, prevState);
+  },
+});
```
NicholasBoll, line 203: This example is covered by model hoisting. The idea being that compound components follow a similar structure so that we don't need to repeat how these components work and rather link to a common section in the compound component docs.

modules/react/tabs/stories/Tabs.stories.mdx
```diff
@@ -108,84 +106,151 @@ the end.
...
 ```
 
-#### Component Props
+#### Props
+
+Undocumented props are spread to the underlying `div` element.
 
 <PropsTable of={Tabs.Panel} />
 
+## Hooks
```
NicholasBoll, line 186: The compound component docs try to be very specific in calling out [Behavior Hooks](https://workday.github.io/canvas-kit/?path=/docs/welcome-dev-docs-compound-components--page#behavior-hooks)

Tabs doesn't currently export these hooks yet, but we're planning on it. Similar to how Popups documents behavior hooks for its subcomponents.

modules/react/tabs/stories/Tabs.stories.mdx
```diff
@@ -108,84 +106,151 @@ the end.
...
+
+### useTabsModel
```
NicholasBoll, line 188: `useTabsModel` is technically a React hook, but more importantly it is the model of the `Tabs` compound component. Eventually I'd like a consistent way of documenting models automatically using source code vs manually documenting using MDX

modules/react/tabs/stories/Tabs.stories.mdx
```diff
@@ -108,84 +106,151 @@ the end.
...
 ```
 
-#### Component Props
+#### Props
 
-<PropsTable of={Tabs.Item} />
+Undocumented props are spread to the underlying `button` element.
 
----
+<PropsTable of={Tabs.Item} />
```
NicholasBoll, line 165: I noticed the use of both `PropsTable` and `ArgsTable`. I think we need to pick one.

modules/react/tabs/stories/Tabs.stories.mdx
alanbsmith: Yeah, I agree. That copy was from a while back. Seems right to move it up. 👍🏻

alanbsmith (approved):
Looks good! Thanks for the update, @jamesfan. 👍🏻

Issue #1290, opened 2021-09-28
Fix Broken MDX Examples for Menu


myvuuu:
## 🐛 Bug Report

The Context Menu and Control Button examples in the Menu MDX doc is broken and cannot be pulled into the Canvas site.

PR #1203, opened 2021-08-09
docs: Adds menu mdx docs

sahlhoff:
fixes #1164

PR #1291, opened 2021-09-29
docs(menu): Fix Menu examples to be self-contained and SSR-safe

jamesfan:
## Summary

Fixes: #1290

Updates the Menu examples to be self-contained and SSR-safe so they may be consumed properly from `@workday/canvas-kit-docs`.

category

---

## Checklist

- [ ] Prepare Design Site PR with updated Storybook links


modules/preview-react/menu/stories/Menu.stories.mdx
```diff
@@ -27,8 +26,15 @@ yarn add @workday/canvas-kit-preview-react
 
 ### Basic Example
 
+`Menu` is typically triggered by an action such as pressing a button. Here's an example of how you
+might implement that behavior using a [`Popup`](/components/popups/popup/).
+
 <ExampleCodeBlock code={Basic} />
```
jamesfan, line 33: Discussed this @alanbsmith a bit: I removed the current `Basic` example and made `ControlButton` the new `Basic` example instead. Seemed odd to have a Menu just floating on its own without a triggering element for our main example.

modules/preview-react/menu/stories/Menu.stories.mdx
```diff
@@ -27,8 +26,15 @@ yarn add @workday/canvas-kit-preview-react
...
 
+`Menu` will automatically assign focus to itself when it appears provided you set the `isOpen` prop
+correctly, so you do **not** need to use the `useInitialFocus` `Popup` hook. You **will**, however,
+need to use `useReturnFocus` to return focus back to the triggering button after closing the `Menu`.
```
jamesfan, line 36: I added this note about `useInitialFocus` not being necessary since Menu already [gives itself focus](https://github.com/Workday/canvas-kit/blob/master/modules/preview-react/menu/lib/Menu.tsx#L92) when it's opened. I tried not setting the `isOpen` prop and using `useInitialFocus` instead to set focus on the Menu, but that was causing the entire page to scroll to the top after clicking the button to activate the Menu.

Maybe an issue with how focus is being managed internally in Menu, not sure. I decided to continue relying on the Menu to focus itself.

modules/preview-react/menu/stories/examples/ContextMenu.tsx
```diff
@@ -1,49 +1,65 @@
...
 
   useAlwaysCloseOnOutsideClick(model);
   useCloseOnEscape(model);
 
   return (
-    <CanvasProvider theme={canvasTheme}>
-      <Popup model={model}>
-        <ContextMenuTarget style={{display: 'inline'}} tabIndex={0}>
-          Right click on this text (Context menu target)
-        </ContextMenuTarget>
```
jamesfan, line 31: I removed `CanvasProvider` from `ContextMenu` to simplify the code a bit. It wasn't adding much in this particular case, and we don't typically use it in our examples except for RTL examples (with the exception of SidePanel)... though maybe we should. It works well in the SidePanel docs and I know @alanbsmith's put thought into adding better support for RTL across all our examples.

modules/preview-react/package.json
```diff
@@ -60,7 +60,6 @@
     "uuid": "^3.3.3"
   },
   "devDependencies": {
-    "@material-ui/core": "^4.9.7",
```
jamesfan, line 63: We no longer need this after removing `ClickAwayListener` from our examples. Ran a `yarn remove @material-ui/core` from the `preview-react` folder.

modules/preview-react/menu/stories/examples/Basic.tsx
```diff
@@ -1,17 +1,74 @@
...
+  const isOpen = model.state.visibility !== 'hidden';
+
+  const handleButtonKeyDown = (event: React.KeyboardEvent) => {
+    let isShortcut = false;
+    if (event.key === `Spacebar` || event.key === ` ` || event.key === `Enter`) {
+      isShortcut = true;
+      if (isOpen) {
+        model.events.hide();
+      }
+      model.events.show();
```
willklein, line 30: Is it intended that when `isOpen` is true, it will hide then show? Or should it be an else clause wrapping this?

modules/preview-react/menu/stories/examples/Basic.tsx
```diff
@@ -1,17 +1,74 @@
...
+
+  useAlwaysCloseOnOutsideClick(model);
+  useCloseOnEscape(model);
+  useReturnFocus(model);
+
+  const isOpen = model.state.visibility !== 'hidden';
+
+  const handleButtonKeyDown = (event: React.KeyboardEvent) => {
+    let isShortcut = false;
+    if (event.key === `Spacebar` || event.key === ` ` || event.key === `Enter`) {
```
willklein, line 25: Note: I looked at making this a switch statement but hated it.

modules/preview-react/menu/stories/examples/Basic.tsx
```diff
@@ -1,17 +1,74 @@
...
+        model.events.hide();
+      }
+      model.events.show();
+    } else if (event.key === `ArrowDown`) {
+      isShortcut = true;
+      model.events.show();
+    } else if (event.key === `ArrowUp`) {
+      isShortcut = true;
+      model.events.show();
+    }
```
willklein, line 37:
```suggestion
    } else if (event.key === `ArrowDown` || event.key === `ArrowUp`) {
      isShortcut = true;
      model.events.show();
    }
```

modules/preview-react/menu/stories/examples/Basic.tsx
willklein: If it is intended, `models.events.show()` could go inside the `if (isShortcut)` block.

jamesfan: No, there should be an `else` wrapping the `model.events.show()`. Good catch 👍

willklein (approved):
I made some comments, nothing blocking. Super nice to drop that `@material-ui/core` dependency!

modules/preview-react/menu/stories/examples/CustomMenuItem.tsx
```diff
@@ -7,7 +7,7 @@ export const CustomMenuItem = () => {
     <Menu title="Menu Title">
       <MenuItem>First Item</MenuItem>
       <MenuItem>Second Item</MenuItem>
-      <li role="menuItem" id="customMenu-3" tabIndex={-1} style={{listStyle: 'none'}}>
```
jamesfan, line 10: I removed the `id` on the custom menu item since it's ignored (Menu overwrites the passed `id` with a `${menuId}-${index}` `id` when it clones the menu item [here](https://github.com/Workday/canvas-kit/blob/master/modules/preview-react/menu/lib/Menu.tsx#L145)).

Also, does that `title` need to be there for `Menu`? The old Menu stories didn't use `title` (see `stories.tsx` in the [Menu MDX PR](https://github.com/Workday/canvas-kit/pull/1203/files)), seems like `title` was added when creating the new examples in the MDX PR. I did notice `title` was used in [a code snippet](https://github.com/Workday/canvas-kit/pull/1203/files#diff-90044480b1e84345b67ce916bc763d6d48e617af129e933da7df24defd743458L34) in the old README, but that was the only place I saw it in the old Menu docs.

sahlhoff: I would say remove it if it's not recommended for a11y (it doesn't appear to be: https://www.w3.org/TR/wai-aria-1.1/#menu)

modules/preview-react/menu/stories/examples/CustomMenuItem.tsx
```diff
@@ -7,7 +7,7 @@ export const CustomMenuItem = () => {
     <Menu title="Menu Title">
```
alanbsmith, line 7:
```suggestion
    <Menu>
```

modules/preview-react/menu/stories/examples/CustomMenuItem.tsx
alanbsmith: Agreed. I don't think we're doing that on any of our menus.

modules/preview-react/package.json
alanbsmith: BYE MUI ✌🏻

alanbsmith (approved):
👑

Issue #1245, opened 2021-09-02
Review and Update the MDX Doc for Pagination


myvuuu:
## 💡 Suggestion

Review and update the MDX doc for Pagination so that:

- [ ]  The MDX lines up with the standardized MDX template
- [ ]  Code snippets and examples are up-to-date and accurate

PR #1295, opened 2021-10-05
docs(pagination): Update Pagination MDX to adhere to standardized template [skip ci]

jamesfan:
## Summary

Fixes: #1245

Update the Pagination MDX to adhere to the [standardized MDX template](https://github.com/Workday/canvas-kit/discussions/1131).

category

---

## Checklist

- [ ] Prepare Design Site PR with updated Storybook links


modules/react/pagination/lib/Pagination/GoTo/Label.tsx
```diff
@@ -7,6 +7,9 @@ import {PaginationModel} from '../types';
 
 export interface GoToLabelProps extends React.LabelHTMLAttributes<HTMLLabelElement> {
   model: PaginationModel;
+  /**
+   * Accepts child elements or a render prop.
+   */
   children?: (model: PaginationModel) => React.ReactNode | React.ReactNode;
```
jamesfan, line 13: I had to add a JSDoc comment for `children` in order for it to show up in `<ArgsTable of={Pagination.GoToLabel} />`. May have something to do with [this](https://github.com/storybookjs/storybook/issues/9921#issuecomment-655436418).

modules/react/pagination/lib/Pagination/types.ts
```diff
@@ -1,17 +1,56 @@
...
-  range: number[]; // array of page numbers
-  rangeSize: number; // size of the range
+  /**
+   * The page number for the current page
+   */
+  currentPage: number;
+  /**
+   * The page number for the first page
+   */
+  firstPage: number;
```
jamesfan, line 9: I added JSDoc comments for `PaginationState` and `PaginationEvents` based on the descriptions originally in the MDX file. This allowed me to use `<ArgsTable>` to show the prop tables for the model state and events using dummy components in `PropsTables.splitprops.tsx`.

modules/react/pagination/lib/Pagination/usePaginationModel.ts
```diff
@@ -5,10 +5,29 @@ import {getRangeMax, getRangeMin} from './common/utils/helpers';
...
+  /**
+   * The page number for the last page (it can also be used as a total page
+   * count)
+   */
   lastPage: number;
+  /**
+   * The page number for the first page
+   * @default 1
+   */
   firstPage?: number;
```
jamesfan, line 17: Maybe I'm missing something, but `firstPage` didn't work as I expected it to. For example, setting it to `2` still showed Page 1 in the range, although I couldn't interact with the Page 1 PageButton. How is it meant to work?

modules/react/pagination/stories/PropTables.splitprops.tsx
```diff
@@ -0,0 +1,47 @@
...
+} from '@workday/canvas-kit-react/pagination';
+
+// <ArgsTable of={Pagination} /> generates a very large props table given that
+// PaginationProps includes FlexProps. Use this dummy component instead to
+// limit the props shown.
+export const PaginationHoistedComponent = (_: {model: PaginationModel}) => <div />;
+
+// <ArgsTable of={Pagination.PageButton} /> generates a props table with
+// IconButton props. Use this dummy component instead to limit the props shown.
+export const PageButtonComponent = (_: {pageNumber: number}) => <div />;
```
jamesfan, line 15: I had to add a few dummy components to [`PropTables.splitprops.tsx`](https://github.com/Workday/canvas-kit/pull/1295/files#diff-4ff84213be099deb3d4d749878f1a4e74de5dadfc0887759a9830e5ff83b87a0) in order to keep their prop tables at a manageable size. Let me know if you can think of a better way to handle this.

modules/react/pagination/stories/examples/HoistedModel.tsx
```diff
@@ -16,30 +17,43 @@ export const HoistedModel = () => {
...
-    </Pagination>
+        </Pagination.AdditionalDetails>
+      </Pagination>
+      <SecondaryButton
+        onClick={() => {
+          model.events.goTo(7);
+        }}
+      >
+        Go to Page 7
+      </SecondaryButton>
```
jamesfan, line 56: I modified the Hoisted Model example to better demonstrate the value of hoisting the model. It's similar to the Hoisted Model [example](https://workday.github.io/canvas-kit/?path=/story/components-containers-tabs-react--hoisted-model) for Tabs.

modules/react/pagination/stories/pagination.stories.mdx
```diff
@@ -91,50 +35,45 @@ yarn add @workday/canvas-kit-react
...
 
 This example adds a form (`Pagination.GoToForm`) that allows you to skip to a specific page within
 the range.
 
 <ExampleCodeBlock code={GoToForm} />
 
----
-
-#### Additional Details
+### Additional Details
```
jamesfan, line 76: I'm guessing the `Additional Details` content should be included in the `Basic Example` section given that it adds a required `aria-live` region?

modules/react/pagination/stories/pagination.stories.mdx
```diff
@@ -212,172 +119,142 @@ provide a configuration as in the example below.
...
+If you're using the hoisted model pattern, follow this table instead:
 
-| name       | type              | required? | default | recommended                              |
-| ---------- | ----------------- | --------- | ------- | ---------------------------------------- |
-| aria-label | `string`          | ✅ `true` | n/a     | "Pagination" (and translated equivalent) |
-| model      | `PaginationModel` | ✅ `true` | n/a     | n/a                                      |
+<ArgsTable of={PaginationHoistedComponent} />
 
-This component also supports
-[all native HTMLElement props](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes).
```
jamesfan, line 261: I removed the MDN links throughout the MDX for consistency since we don't use them in other component docs, but I think they're a useful addition. May add them back in a later pass.

modules/react/pagination/stories/pagination.stories.mdx
```diff
@@ -212,172 +119,142 @@ provide a configuration as in the example below.
...
-| aria-label | `string` | ✅ `true` | n/a     | "First" (and translated equivalent) |
+#### Props
 
-This component also supports all `IconButton` props.
+Undocumented props are spread to the underlying `<button>` element. Note that you must set
+`aria-label` to meet accessibility standards. We recommend setting it to `First` or the translated
+equivalent.
 
----
+`Pagination.JumpToFirstButton` supports all `IconButton` props.
```
jamesfan, line 190: Rather than show an `<ArgsTable>` for `Pagination.JumpToFirstButton` and all of the other `IconButton` subcomponents (which would repeat the `IconButton` props several times throughout the doc), I just left it at this short sentence.

modules/react/pagination/lib/Pagination/usePaginationModel.ts
NicholasBoll: 😕 
Your observation matches implementation. In the `StepToPreviousButton` has an `isDisabled` to prevent going to a page less than the `firstPage` and the `goTo` event prevents going to a page lower than `firstPage`. The question might be why would a developer choose a "first page" other than 1

alanbsmith: Good find. There's a mismatch between what's here and what's in `buildPageRange` [here](https://github.com/Workday/canvas-kit/blob/master/modules/react/pagination/lib/Pagination/buildPageRange.ts#L16)

alanbsmith: > _The question might be why would a developer choose a "first page" other than 1_

That's fair. I thought about that a lot while building the component, but it felt like an artificial constraint. The underlying logic really doesn't care about "pages" (other than the bug you found). It's takes the range size, the current item, the first and last pages, and creates a list. Outside of that, it doesn't really care.

This is a range of -10 to 10

alanbsmith: Would you ever use this? Probably not. But the component is sort of agnostic to that.

modules/react/pagination/stories/pagination.stories.mdx
```diff
@@ -91,50 +35,45 @@ yarn add @workday/canvas-kit-react
...
-quickly without much explanation. Further detail on individual component props, behavior, and
-accessibility can be found in the [Components](#components) section.
+### Basic Example
 
----
+`Pagination` includes a container `Pagination` component and a number of subcomponents which can be
+composed in a variety of ways.
 
-### Basic Usage
+In this example, we set up a basic `Pagination` component with the default range of 5 pages, as well
```
alanbsmith, line 43: Optional nit: `s/5/five`

```suggestion
In this example, we set up a basic `Pagination` component with the default range of five pages, as well
```

modules/react/pagination/stories/pagination.stories.mdx
```diff
@@ -91,50 +35,45 @@ yarn add @workday/canvas-kit-react
...
+as step controls (`Pagination.StepToPreviousButton` and `Pagination.StepToNextButton`) that allow
+you to move to the next page or the previous page.
```
alanbsmith, line 45: Optional nit: keep the order of the control buttons.
```suggestion
you to move to the previous page or the next page.
```

modules/react/pagination/stories/pagination.stories.mdx
```diff
@@ -91,50 +35,45 @@ yarn add @workday/canvas-kit-react
...
 
-#### Step Controls
+### Hoisted Model
 
-In this example, the component uses step controls (`Pagination.StepToPreviousButton` and
-`Pagination.StepToNextButton`) that allow you to move to the next page and previous pages.
+By default, `Pagination` will create and use its own [model](#model) internally. Alternatively, you
+may configure your own model with `usePaginationModel` and pass it to `Pagination` via the `model`
+prop. This pattern is referred to as
+[hoisting the model](/getting-started/for-developers/resources/compound-components/#configuring-a-model)
```
alanbsmith, line 54: It's a little odd that the title for this section is `Configuring a Model` when we call out "hoisting the model" here. I'm not settled on a solution though.

modules/react/pagination/stories/pagination.stories.mdx
alanbsmith: Yes. That's correct. It should be in every example. 👍🏻

modules/react/pagination/stories/pagination.stories.mdx
alanbsmith: Yeah, I agree. I think it's a helpful resource, but I'd like it to be consistent too.

modules/react/pagination/stories/pagination.stories.mdx
```diff
@@ -212,172 +119,142 @@ provide a configuration as in the example below.
...
-| aria-label         | `string`                       | ✅ `true`  | n/a     | "Pagination" (and translated equivalent) |
-| lastPage           | `number`                       | ✅ `true`  | n/a     | n/a                                      |
-| initialCurrentPage | `number`                       | 🚫 `false` | 1       | n/a                                      |
-| rangeSize          | `number`                       | 🚫 `false` | 5       | n/a                                      |
-| firstPage          | `number`                       | 🚫 `false` | 1       | n/a                                      |
-| onPageChange       | `(pageNumber: number) => void` | 🚫 `false` | n/a     | n/a                                      |
+<ArgsTable of={PaginationModelConfigComponent} />
 
-And for the hoisted model pattern, follow this table:
+If you're using the hoisted model pattern, follow this table instead:
```
alanbsmith, line 149: Maybe we should link to the model section here? `PaginationModel` doesn't tell me much.

alanbsmith (approved):
Really nice work, @jamesfan. I left a few comments but nothing that should block this PR.

PR #1325, opened 2021-10-29
feat(tabs): Add overflow support to tabs

NicholasBoll:
## Summary

Add support for overflow tabs

category

### Release Note

### BREAKING CHANGES
Tabs API was updated to support a more generic selection model for all lists.
- `model.events.activate({tab})` -> `model.events.select({id})`
- `model.state.activeTab` -> `model.state.selectedKeys[0]`
- `useTabsModel({onActivate})` -> `useTabsModel({onSelect})`
- `useTabsModel({shouldActivate})` -> `useTabsModel({shouldSelect})`

---

## Checklist


cypress/integration/Tabs.spec.ts
```diff
@@ -33,8 +33,8 @@ describe('Tabs', () => {
         cy.findByRole('tab', {name: 'Second Tab'}).should('have.attr', 'aria-selected', 'false');
       });
 
-      it('should not have tabindex on the first tab', () => {
-        cy.findByRole('tab', {name: 'First Tab'}).should('not.have.attr', 'tabindex');
+      it('should not have tabindex=-1 on the first tab', () => {
+        cy.findByRole('tab', {name: 'First Tab'}).should('not.have.attr', 'tabindex', '-1');
```
NicholasBoll, line 37: **Before:**

Roving index was generic, but still assumed each item was a `button`. All `button` elements are focusable by default and no `tabIndex` was required. The specification assumed this and made sure not `tabindex` was defined to override the default focusability of the element.

**After:**

The roving focus behavior makes no assumption of the element being used. `tabindex` is always set regardless of inherit focusability status to ensure roving focus always works. So the button now looks like `button[tabindex=0]`. I could have asserted for `tabindex=0`, but it seemed the more generic way of asserting according to the spec was to not override with `tabindex=-1`. Thoughts?

modules/react/popup/lib/PopupCard.tsx
```diff
@@ -87,7 +88,7 @@ export const PopupCard = createComponent('div')({
         padding={padding}
         {...props}
       >
-        <Card.Body>{children}</Card.Body>
```
mannycarrera4, line 90: is there a reason this was deleted?

modules/react/tabs/lib/TabsItem.tsx
```diff
@@ -0,0 +1,217 @@
...
+import {TabsModel} from './useTabsModel';
+export interface TabsItemProps extends ExtractProps<typeof Box, never> {
+  /**
+   * Optionally pass index to tab item. This should be done if `Tabs.Item` components were created
+   * via a `Array::map` function. This index will ensure keyboard navigation works even if items are
+   * inserted out of order.
+   */
+  index?: number;
+  /**
+   * The label text of the Tab.
```
mannycarrera4, line 35: if it's children can it be anything? not just label text?

modules/react/tabs/lib/TabsItem.tsx
```diff
@@ -0,0 +1,217 @@
...
+   * Use `hasIcon={true}` when the Tab item contains an icon. This instructs the tab item to render
+   * children as a flex container to accommodate icons. You will have to use `Tabs.Item.Text` as a
+   * child element if this is set to true.
+   *
+   * @example
+   * <Tabs.Item hasIcon>
+   *   <Tabs.Item.Icon icon={someIcon} />
+   *   <Tabs.Item.Text>Tab Text</Tabs.Item.Text>
+   * </Tabs.Item>
+   */
```
mannycarrera4, line 81: I'm wondering if there's any way we can warn the consumer if they don't do this and pass an icon?

modules/react/tabs/lib/TabsOverflowButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const TabsOverflowButton = createComponent('button')({
+  displayName: 'Tabs.OverflowButton',
+  Component: ({model, children, ...elemProps}: OverflowButtonProps, ref, Element) => {
+    const localModel = useModelContext(TabsModelContext, model);
+
+    const props = useTabsOverflowButton(localModel, elemProps, ref);
+
+    return (
+      <StyledTabItem hasIcon={true} type="button" as={Element} {...props}>
```
mannycarrera4, line 39: do we want hadIcon true always set here?

NicholasBoll: Yes. The Overflow button has an icon. This can still be overridden.

modules/react/tabs/lib/TabsItem.tsx
NicholasBoll: That could definitely be a dev-only warning, but it might involve extra code and state tracking that is confusing and harder to maintain.

NicholasBoll: How about https://github.com/Workday/canvas-kit/pull/1325/files#diff-12fd6ab47eb13fddf64e87d63f64b95b38acd128ff0e4ed69e4ab0524e782bf4R185-R198

modules/react/tabs/lib/TabsItem.tsx
NicholasBoll: It can be anything. It is the contents of the tab. I can update to be more clear

modules/react/tabs/lib/TabsItem.tsx
```diff
@@ -0,0 +1,234 @@
...
+      props.ref = elementRef;
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      useMount(() => {
+        if (localRef.current.querySelector('svg') && !props.hasIcon) {
+          console.warn(
+            `A Tabs.Item with an icon should have the 'hasIcon' prop set to true. This ensures correct rendering`
+          );
+        }
+      });
+    }
```
mannycarrera4, line 198: <3 thank you!!!

modules/react/tabs/lib/TabsItem.tsx
mannycarrera4: awesome!

cypress/integration/Tabs.spec.ts
```diff
@@ -279,9 +315,9 @@ describe('Tabs', () => {
         cy.findByRole('tab', {name: 'Tab 1'}).click();
       });
 
-      context('then the Backspace/Delete key is pressed', () => {
+      context('then the Delete key is pressed', () => {
         beforeEach(() => {
-          cy.focused().type('{backspace}');
+          cy.focused().type('{del}');
```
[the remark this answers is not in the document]
mannycarrera4: interesting, TIL

alanbsmith (approved):

mannycarrera4 (approved):

Issue #1317, opened 2021-10-19
Status Indicator Refactoring for Truncated Text


myvuuu:
## 🚀 Feature Proposal

Work with A11y and Canvas designers to determine how to handle truncated text for status indicators. We have a couple of options:

1. - Increasing max width
2. - Using overflow tooltips for truncated status indicators (last resort...dundundun), not ideally for a11y standards
3. - Providing additional usage guidance around limiting the amount of characters in status indicators and redirecting users to pills when applicable

## Decisions

Dev / Design: 
- Will provide examples (Storybook) with overflow tooltips and give guidance around setting up an overflow tooltip for Status Indicators (make it focusable)
- We should keep a max width for Status Indicators. Users can override it if they absolutely need to, but this will ensure more consistency across WD UI
- Increase the max width value to 200px (this includes icon)

Doc:
- Provide more guidance around best practices for Status Indicator (short, concise text) 
- Status Indicators should never wrap
- Direct users to Pills when appropriate; Status Indicators are not actionable, but Pills are 
- Keep overflow tooltips, but should only be used when needed (text cannot be shorten down/condensed). Recommend that users consider shortening it or use plain text before they add overflow tooltips
- “Keep in mind that <indicators> increase the amount of visual noise, particularly when combined with other visual labelling elements, so use them in moderation.”

## Motivation

A11y wants us to move away from truncated text and if we need it, overtool tooltips should be used (current workaround for users, but the status indicator must be focused on, which can be confusing for users).

mannycarrera4:
Closing, verified in: https://github.com/Workday/canvas-kit/pull/1329

PR #1329, opened 2021-11-04
feat: Increase max width of Status Indicator, add prop and examples

anicholls:
## Summary

Fixes: #1317 

Status Indicators currently truncate when they reach their max width of `150px`. After receiving
requests to increase this, we have upped it by 25% to `200px`.

I've also added a prop (`maxWidth`) to configure this if you need to. I've added a new story to show an example of this, and with an `OverflowTooltip`

category

### BREAKING CHANGES
Status Indicators currently truncate when they reach their max width of `150px`. After receiving
requests to increase this, we have upped it by 25% to `200px`.

---

## Checklist

## Additional References

image


NicholasBoll (approved):

NicholasBoll:
Perhaps this should target v7 now? The migration guide is in conflict and points to v6 while the branch points to v7

Issue #1336, opened 2021-11-10
Add Version Support section to Readme


alanbsmith:
## Overview

We have told people informally what versions of CK we support and how we support them, but we don't actually have it written down anywhere I could find. This is something we should add for transparency and on-boarding.

PR #1337, opened 2021-11-10
docs: Add Version Support section to Readme

alanbsmith:
## Summary

Fixes: #1336

category

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,15 +10,17 @@ implement user experiences consistent with
 <a href="https://lerna.js.org">
   <img src="https://img.shields.io/badge/maintained%20with-lerna-cc00ff.svg" alt="Maintained with Lerna" />
 </a>
-[![Release](https://github.com/Workday/canvas-kit/actions/workflows/release.yml/badge.svg)](https://github.com/Workday/canvas-kit/actions/workflows/release.yml)
+<a href="https://github.com/Workday/canvas-kit/actions/workflows/release.yml">
+  <img alt="Release" src="https://github.com/Workday/canvas-kit/actions/workflows/release.yml/badge.svg">
+</a>
 <a href="./modules/docs/mdx/CONTRIBUTING.mdx">
   <img src="https://img.shields.io/badge/PRs-welcome-brightgreen.svg" alt="PRs welcome" />
 </a>
 
 ---
 
-**:white_check_mark: For a list of available react modules see
-[Canvas Kit Component Status](./modules/docs/mdx/COMPONENT_STATUS.mdx)**
+_For a list of available react modules see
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,15 +10,17 @@ implement user experiences consistent with
 <a href="https://lerna.js.org">
   <img src="https://img.shields.io/badge/maintained%20with-lerna-cc00ff.svg" alt="Maintained with Lerna" />
 </a>
-[![Release](https://github.com/Workday/canvas-kit/actions/workflows/release.yml/badge.svg)](https://github.com/Workday/canvas-kit/actions/workflows/release.yml)
+<a href="https://github.com/Workday/canvas-kit/actions/workflows/release.yml">
+  <img alt="Release" src="https://github.com/Workday/canvas-kit/actions/workflows/release.yml/badge.svg">
+</a>
```
alanbsmith, line 15: I noticed this badge was rendering fine in Storybook but not in our Readme. So I updated it to use HTML instead of MD.

README.md
```diff
@@ -10,15 +10,17 @@ implement user experiences consistent with
...
 <a href="./modules/docs/mdx/CONTRIBUTING.mdx">
   <img src="https://img.shields.io/badge/PRs-welcome-brightgreen.svg" alt="PRs welcome" />
 </a>
 
 ---
 
-**:white_check_mark: For a list of available react modules see
-[Canvas Kit Component Status](./modules/docs/mdx/COMPONENT_STATUS.mdx)**
+_For a list of available react modules see
```
alanbsmith, line 22: The ✅ wasn't rendering properly in MD, and I wasn't entirely sure if made sense, so I removed it.

README.md
```diff
@@ -112,6 +114,23 @@ Canvas Kit follows [semantic versioning](https://semver.org/) and is enforced au
...
+Actions.
+
+The current major version is also stable and receives new feature and patch updates. Patch releases are
+automatically deployed upon merge by GitHub Actions, and minor releases are manually deployed at the
+end of each sprint.
+
+The next major version is typically an unstable environment and has major breaking changes. You are
+welcome to pull this version down for local development and experimentation, but we generally
+recommend against using it in production until the first stable version has been released.
+
```
mannycarrera4, line 133: is it worth pointing out where this info can be found or where we can see the current version? Example: You can find out current `here` 

Also i know this has come up before but do we host a storybook with a previous version and a current version?

alanbsmith: Good callouts! I'd like to add another section about our repo branches and where to find things.

We don't have a stable URL reference for previous Storybook versions, but you can find the Chromatic artifacts.

I think I would like that information to live in our development docs in a dedicated section and / or an FAQ.

mannycarrera4 (approved):

PR #1350, opened 2021-11-22
fix(tabs): Fix focusability of the More button

NicholasBoll:
## Summary

The "More" button in tabs should be focusable so that keyboard users and screen reader users can access it.

category

### DIFF
--- a/cypress/integration/Tabs.spec.ts
+++ b/cypress/integration/Tabs.spec.ts
@@ -540,7 +540,7 @@ describe('Tabs', () => {
             cy.focused().tab();
           });
 
-          it('should focus on the tab panel', () => {
+          it('should focus on the "More" button', () => {
             cy.findByRole('button', {name: 'More'}).should('have.focus');
           });
         });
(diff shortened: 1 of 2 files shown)


cypress/integration/Tabs.spec.ts
```diff
@@ -510,6 +528,24 @@ describe('Tabs', () => {
...
+        });
+
+        context('when the Tab key is pressed', () => {
+          beforeEach(() => {
+            cy.focused().tab();
+          });
+
+          it('should focus on the tab panel', () => {
+            cy.findByRole('button', {name: 'More'}).should('have.focus');
+          });
```
jamesfan, line 545:
```suggestion
          it('should focus on the "More" button', () => {
            cy.findByRole('button', {name: 'More'}).should('have.focus');
          });
```

jamesfan (approved):
Copy-and-paste issue with a test, everything else looks good 👍

mannycarrera4 (approved):

PR #1352, opened 2021-11-23
fix(popup): Fix PopupCard styles to be more easily overridden

NicholasBoll:
## Summary

`PopupCard` defines `width`, `position`, and `maxWidth` in Emotion styles instead of style props making overriding styles inconsistent. This change moves all layout styles into style props.

category


modules/react/popup/lib/PopupCard.tsx
```diff
@@ -82,9 +71,10 @@ export const PopupCard = createComponent('div')({
       <StyledPopupCard
         as={Element}
         transformOrigin={transformOrigin}
-        depth={depth}
-        width={width}
-        padding={padding}
+        position="relative"
+        padding="l"
+        depth={2}
+        maxWidth={`calc(100vw - ${space.l})`}
```
mannycarrera4, line 77: ah i see, you moved them out of emotion and on to the element, and then anything they pass as props will override

mannycarrera4 (approved):

Issue #1200, opened 2021-08-03
Cannot set Ref prop to Banner component


youryss:
## 🐛 Bug Report

I'm trying to use the ref property of the Banner component but it seems the component doesn't have this api available.

## To Reproduce

try creating a ref and applying that ref to the banner and programmatically try using any html event method like focus()

https://codesandbox.io/s/frosty-cache-t584f?file=/src/CustomBanner.tsx

## Expected Behavior

I hope that when I apply ref to the component I have access to its DOM reference 

## Actual Results

ref does not seem to work and we have no DOM reference

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

## Link to repl or repo (highly encouraged)

https://codesandbox.io/s/frosty-cache-t584f?file=/src/CustomBanner.tsx

myvuuu:
Will be fixed as part of the v7 release!

PR #1359, opened 2021-11-29
refactor(banner): Update Banner to use the compound component pattern

vibdev:
## Summary

Fixes: #1200 

- [x] Update Banner to use the compound component pattern.
- [x]  Adds JSDoc comments Banner and its subcomponents DX
- [x] Adds Cypress Tests for the Banner component
- [x] Adds V7 codemod and tests
- [x] Updates storybook and provides additional examples

category

### BREAKING CHANGES
This change updates the banner component to use the compound component pattern. For more information, please see the V7 migration guide.

---

## Checklist


modules/react/banner/lib/Banner.tsx
```diff
@@ -1,127 +1,101 @@
...
-        {...props}
-      >
-        <BannerIcon icon={bannerIcon} color={iconColor} colorHover={iconColor} size={iconSize} />
-        <BannerLabel>{label}</BannerLabel>
-        <BannerViewAll variant={variant}>{actionText}</BannerViewAll>
-      </BannerWrapper>
+      <BannerModelContext.Provider value={value}>
+        <Flex
+          as={Element}
+          width={isSticky ? '222px' : '328px'}
```
mannycarrera4, line 80: is there a chance someone wants to change this or is this per design spec? just curious!

vibdev: These are from the og component so assume they are spec. That said, since it is a style prop it should be easy to override if needed.

modules/react/banner/lib/BannerAction.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <Box
+        display={localModel.state.isSticky ? 'none' : 'inline'}
+        as={Element}
+        css={[styles]}
+        {...props}
+      >
+        {children}
+      </Box>
+    );
+  },
```
mannycarrera4, line 41: Mostly curious here, i notice in the stories that the `onClick` is added to the top level component `Banner` is there a reason why this component isn't a `Tertiary` button or a button by default? Is the reasoning because the whole alert is clickable?

I figured if I were consuming this I would want to add the `onClick` the `Banner.Action` component, and the wrapping component just handles the model. Thoughts?

vibdev: > Is the reasoning because the whole alert is clickable?
Yes. I don't think it makes sense to have more than one click handler. Open to suggestion here.

mannycarrera4 (commented):
Hey @vibdev  thanks for taking charge and converting this to a compound component!
- The stories add so much more value than before so I really appreciate that!
- Also, the way you've broken into the sub components make a lot of sense to me!

I just had a couple questions but other than that, this is looking really good!

- `Banner.Action` i understand why you made this a span, but the component name makes it sound like I can add my own click event and that would be my natural instinct to do so, maybe we can rename this to `Banner.ActionText` and still keep the click event on the wrapping container?

- If the sub components will more than likely always be wrapped in a `Banner` component, do they each need the `model` as a prop? Or have you added the model prob for flexibility to the consumer?

- I know this was a conversion PR, but if the `Banner` Component is a button, does it need a different `aria-role` for accessibility reasons? Mostly just referencing: https://www.w3.org/TR/wai-aria-1.2/#banner

vibdev:
@mannycarrera4 Thanks for the review!
1) I think renaming to ActionText makes a lot of sense 
2) I think that is the convention for subcomponents, I think it is for flexibility maybe @alanbsmith can speak more to that
3) I don't think banner is the correct role it is more for 'site-oriented content, rather than page-specific content.' This component is more page level.

mannycarrera4 (approved):
@vibdev Thanks for working on this!!

Issue #1362, opened 2021-11-30
Create a PageHeader Example


alanbsmith:
## 💡 Suggestion

Create an example `PageHeader` component for users to reference to migrate from `DeprecatedPageHeader`.

## Motivation

Now that we've deprecated `PageHeader`, we should add an example to help users migrate from the deprecated component.

## Acceptance Criteria

- [x] Create a code example for `PageHeader` using existing canvas-kit components and patterns
- [ ] Add link to the example in the v6 migration guide

Issue #1363, opened 2021-11-30
Create a GlobalHeader Example


alanbsmith:
## 💡 Suggestion

Create an example `GlobalHeader` component for users to reference to migrate from `DeprecatedGlobalHeader`.

## Motivation

Now that we've deprecated the `GlobalHeader` in v6, we'd like to provide an example for users to use so that they can migrate from the deprecated component.

## Acceptance Criteria

- [x] Create a code example of `GlobalHeader` that looks like what's in Workday's UI today using existing canvas-kit components and patterns
- [ ] Add link to the example in the v6 migration guide

PR #1366, opened 2021-12-03
docs: Add headers examples to Storybook

RayRedGoose:
## Summary

Resolves: #1362, #1363  

Now that we've deprecated  `PageHeader` and `GlobalHeader` components in v6, we'd like to provide an example for users to use so that they can migrate from the deprecated component.

category

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -25,7 +25,8 @@ function storySort(a, b) {
     prefix('labs-', '4'),
     prefix('basic', 'aa'),
     prefix('default', 'ab'),
-    prefix('visual-testing', 'zz')
+    prefix('visual-testing', 'zz'),
+    prefix('examples', 'zzz')
   );
 
   const left = prefixFn(a[0]);
--- a/modules/docs/mdx/6.0-MIGRATION-GUIDE.mdx
+++ b/modules/docs/mdx/6.0-MIGRATION-GUIDE.mdx
@@ -220,6 +220,9 @@ const CustomGlobalHeader = props => {
 };
 ```
 
+You may continue to use this component exactly as you did in v5, but note that we plan to hard-deprecate it in Canvas Kit v7.
+If you would like to migrate away from this deprecated component now, you can reference [this example](https://workday.github.io/canvas-kit/?path=/story/examples-global-header-react--basic) instead.
+
 ### Page Header
 
 We are [soft-deprecating](#soft-deprecation) `PageHeader`. It has been renamed to
@@ -270,6 +273,9 @@ export const CustomPageHeader = (props: CustomPageHeaderProps) => {
   return <DeprecatedPageHeader {...props} />;
 };
 ```
+
+You may continue to use this component exactly as you did in v5, but note that we plan to hard-deprecate it in Canvas Kit v7.
+If you would like to migrate away from this deprecated component now, you can reference [this example](https://workday.github.io/canvas-kit/?path=/story/examples-page-header-react--basic) instead.
 
 ## Component Migrations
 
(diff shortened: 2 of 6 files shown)


.storybook/preview.js
```diff
@@ -25,7 +25,8 @@ function storySort(a, b) {
     prefix('labs-', '4'),
     prefix('basic', 'aa'),
     prefix('default', 'ab'),
-    prefix('visual-testing', 'zz')
+    prefix('visual-testing', 'zz'),
+    prefix('examples', 'zzz')
```
alanbsmith, line 29: This will put examples at the very bottom of our sidenav list in Storybook. I think that's fine for now. When we figure out how we want to present examples to people, we can find a more visible location for it.

mannycarrera4:
Hey @RayRedGoose this is looking awesome! A couple notes:

1. When we wrap the Workday Logo in a hyperlink it gives it a weird grey hover background:
Screen Shot 2021-12-03 at 2 16 39 PM
Maybe we change this or style is differently?


2. This is mostly a question, should we show the `PageHeader` icons wrapped in tooltips? I would build it as if it where one of our products and my guess is design would want a tooltip with those icon buttons? Thoughts? 

Code wise, everything looks great!

modules/react/_examples/stories/GlobalHeader.stories.mdx
```diff
@@ -0,0 +1,10 @@
+import {Meta, Story} from '@storybook/addon-docs/blocks';
+import {Basic} from './examples/GlobalHeader';
+
+<Meta title="Examples/GlobalHeader/React" />
+
+# Canvas Kit Examples
+
+## GlobalHeader
+
```
alanbsmith, line 9: Let's put some additional context here.

```suggestion
Developers building internal Workday applications will likely not need to create this component. However, if you're building components to be used outside of Workday, this is a helpful reference for building a global navigation header that looks like our internal `GlobalHeader`.
```

modules/react/_examples/stories/GlobalHeader.stories.mdx
```diff
@@ -0,0 +1,10 @@
+import {Meta, Story} from '@storybook/addon-docs/blocks';
+import {Basic} from './examples/GlobalHeader';
+
+<Meta title="Examples/GlobalHeader/React" />
+
+# Canvas Kit Examples
+
+## GlobalHeader
+Developers building internal Workday applications will likely not need to create this component. However, if you're building components to be used outside of Workday, this is a helpful reference for building a global navigation header that looks like our internal `GlobalHeader`.
+<ExampleCodeBlock code={Basic} />
```
alanbsmith, line 10: Oops. I missed a space here. Sorry!
```suggestion

<ExampleCodeBlock code={Basic} />
```

modules/docs/mdx/6.0-MIGRATION-GUIDE.mdx
```diff
@@ -271,6 +274,9 @@ export const CustomPageHeader = (props: CustomPageHeaderProps) => {
 };
 ```
 
+You may continue to use this component exactly as you did in v5, but note that we plan to hard-deprecate it in Canvas Kit v7.
+If you would like to migrate away from this deprecated component now, you can reference [this example](https://workday.github.io/canvas-kit/?path=/story/examples-page-header-react--basic) instead.
+
```
RayRedGoose, line 279: I've added message for users with links to Storybook deprecated conponents.

RayRedGoose:
> When we wrap the Workday Logo in a hyperlink it gives it a weird grey hover background

@mannycarrera4 I think the same. I used `HyperLink` CK component as it was used in DeprecatedGlobalHeader. But if it's not fit I can create a custom style for hyperlink, because I didn't find any good replacement for this in existing components.

alanbsmith (approved):
Looks great! Thanks, @RayRedGoose! 🎉

mannycarrera4 (approved):
Nice work!!

PR #1367, opened 2021-12-03
chore: Bump uuid to stable non deprecated version

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/1311

category

PR #1371, opened 2021-12-06
fix: Revert uuid upgrade to get releases working [skip release]

NicholasBoll:
## Summary

Fix broken releases introduced by #1367 that unintentionally removed `@types/node` from Typescript which is causing `process.env` to be undeclared. Also updating the `yarn typecheck` scripts to catch this type of error. The PR job does not run `yarn build`, but runs `yarn typecheck`. Before the whole project was typechecked as a single project which worked because the `css` module included storyfiles using `@storybook/react` which included `webpack` which included definitions files for `process` and thus this PR was able to be merged even though a `yarn build` would fail. `yarn typecheck` is now being run the same way as `yarn build` is. This should prevent this type of issue.

category


mannycarrera4 (approved):

yarn.lock
```diff
@@ -4325,10 +4310,12 @@
...
-"@types/uuid@^8.3.3":
-  version "8.3.3"
-  resolved "https://registry.yarnpkg.com/@types/uuid/-/uuid-8.3.3.tgz#c6a60686d953dbd1b1d45e66f4ecdbd5d471b4d0"
-  integrity sha512-0LbEEx1zxrYB3pgpd1M5lEhLcXjKJnYghvhTRgaBeUivLHMDM1TzF3IJ6hXU2+8uA4Xz+5BA63mtZo5DjVT8iA==
+"@types/uuid@^3.4.4":
+  version "3.4.4"
+  resolved "https://registry.yarnpkg.com/@types/uuid/-/uuid-3.4.4.tgz#7af69360fa65ef0decb41fd150bf4ca5c0cefdf5"
+  integrity sha512-tPIgT0GUmdJQNSHxp0X2jnpQfBSTfGxUMc/2CXBU2mnyTFVYVa2ojpoQ74w0U2yn2vw3jnC640+77lkFFpdVDw==
+  dependencies:
+    "@types/node" "*"
```
mannycarrera4, line 4318: this little bugger....

NicholasBoll: Yeah. `yarn` doesn't try to decrease dependencies to a minimal amount of dependencies. We use `yarn-deduplicate` for that

Issue #1370, opened 2021-12-06
Tabs overflow menu has unnecessary gutters


anicholls:
## 🐛 Bug Report

The tabs overflow menu has some unexpected spacing on the inside of the menu container. The items should extend right to the edge

![image](https://user-images.githubusercontent.com/908478/144925384-1caf0f2f-7929-4ea1-86f6-6f0bab44b71c.png)

PR #1378, opened 2021-12-07
fix(tabs): Remove gutter from tabs overflow menu

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/1370

Remove the `margin` on the menu. I also moved the border down one level mostly because I was seeing some visual inconsistency in chrome

category


mannycarrera4:
Screen Shot 2021-12-07 at 2 39 25 PM
Without the override of the border changes in this PR, this is what the dropdown looks like.

modules/react/tabs/lib/menu/MenuList.tsx
```diff
@@ -39,9 +39,10 @@ export const MenuList = createComponent('div')({
         borderRadius="m"
         padding="zero"
         marginY="zero"
-        marginX="xxs"
+        marginX="zero"
```
alanbsmith, line 42: This could probably just be`margin="zero"` instead of using `marginX` and `marginY`.

NicholasBoll:
I see a small problem in the screenshots. Since the margin was removed, the `MenuItem` elements sit on top of the border radius of the `Menu` element. Do you see that rendering issue in the corners, @mannycarrera4?

NicholasBoll:
It looks like the preview menu avoids this by adding margin to the top/bottom: 
image

NicholasBoll:
The spec has it too: 
image

mannycarrera4:
ah i also see the 1px margin, thanks for this! fixing!

NicholasBoll:
I'm not sure the 1px margin is intentional from the Figma, I think it is trying to avoid the border.

NicholasBoll (approved):

Issue #1382, opened 2021-12-08
Action Bar Refactoring


myvuuu:
## 🚀 Feature Proposal

Refactor Action Bar by:

- [ ] Create a new compound component version of `ActionBar`
- [ ] Update tests and documentation
- [ ] Update migration guide
- [ ] Create a codemod for the update (if feasible)
- [ ] ~Updating the Icon button to point to the new icon only primary/secondary/tertiary icon buttons (#1379)~
- [ ] ~Add overflow behavior to allow buttons to collapse down based on screen size~

## 🗒️ Notes

- We're cutting the Menu + Menu button work from the initial scope until we decide which Menu to use. See #1391 

- We'll update the component to use the new buttons when #1379 is complete

PR #1396, opened 2021-12-16
feat(action-bar): Refactor ActionBar component

RayRedGoose:
## Summary

- Convert to compound component.
- Remove `ChildrenContainer` and replace it styles to `ActionBarContainer`.
- Update documentation
- Add jest component test and visual tests for storybook.

Resolves: #1382

category

### BREAKING CHANGES
Change `fixed` prop from component to `position` to set container position (`fixed` position has been set as default).


modules/react/action-bar/lib/ActionBar.tsx
```diff
@@ -1,76 +1,55 @@
...
+type PropertyPosition =
+  | Globals
+  | '-webkit-sticky'
+  | 'absolute'
+  | 'fixed'
+  | 'relative'
+  | 'static'
+  | 'sticky';
+
+export type ActionBarProps = StyledType & {
```
NicholasBoll, line 15: Move the `StyledType` to line 23 instead. `createComponent` adds the `as` type. Having `StyledType` might interfere with the `as` provided by `createComponent`

NicholasBoll:
Overall I'm torn on this refactor. I didn't have comments on specific lines, so I'll review here:

With many conversions to compound components, we've been moving to using layout components which give us a lot of niceties: many style props come for free and defaults are easily overridden.

For example:

### Using styled
```tsx
export interface SomeStyledProps {
  position: ...
}

const SomeStyledComponent = styled('div')<SomeStypedProps & StyledType>({
  display: 'flex',
  ...
  position: 'fixed',
  left: 0,
  right: 0,
  top: 0,
  ...
  ({position}) => ({position})
})

const SomeComponent = createComponent({
  Component(props: SomeComponentProps, ref, Element) {
    return <SomeStyledComponent ref={ref} as={Element} {...props} />
  }
})
```

### Layout components

```tsx
interface SomeComponentProps extends FlexProps {}

const SomeComponent = createComponent({
  Component(props: SomeComponentProps, ref, Element) {
    return (
      <Flex
        ref={ref}
        as={Element}
        position="fixed"
        top={0}
        left={0}
        right={0}
        {...props}
      />
    )
  }
})
```

We don't export prop types anymore with `createComponent`. The reason is the component is polymorphic, meaning exporting props aren't as useful as it used to be. Instead we use `ExtractProps` from the common module:

```tsx
interface Props {
  foo: string
}

const Component = createComponent('div')({
  Component(props: Props) {
    return <div />
  }
})

type Props1 = Props // Props
type Props2 = ExtractProps<typeof Component> // Props & React.HTMLAttributes<HTMLDivElement>
type Props3 = ExtractProps<typeof Component, never> // Props
type Props4 = ExtractProps<typeof Component, 'aside'> // Props & React.HTMLAttributes<HTMLElement>
```

If someone wants to extend `Component` and use the same props and get the HTML attributes that come with the element, they can use `ExtractProps` to get the interface. If they use `as`, they can add the second argument:

```tsx
const MyComponent = (props: ExtractProps<typeof Component, 'aside'>) => {
  return <Component as="aside" {...props} />
}
```

Also, `fixed` used to default to `false`... Should we maintain that? If so, the codemod is incomplete. It should change `fixed={true}` to `position="fixed"`.

If `ActionBar` was actually a `HStack` element, the flex + gap would be handled and wouldn't need to be in the styled component.

modules/react/action-bar/stories/ActionBar.stories.mdx
```diff
@@ -36,6 +36,6 @@ At 768px, responsive styles will take effect:
 
 ## Props
 
-Undocumented props are spread to the underlying container `<div>` element.
+Undocumented props are spread to the underlying `HBlock` component.
```
NicholasBoll, line 39:
```suggestion
Undocumented props are spread to the underlying `HStack` component.
```

modules/react/action-bar/lib/ActionBar.tsx
```diff
@@ -1,76 +1,55 @@
...
+type PropertyPosition =
+  | Globals
+  | '-webkit-sticky'
+  | 'absolute'
+  | 'fixed'
+  | 'relative'
+  | 'static'
+  | 'sticky';
+
+interface ActionBarProps {
```
NicholasBoll, line 16: This interface isn't accurate since all `HStack` props are valid. The only thing this interface does is add the `default` value to `position`. Should we get rid of the interface? Or maybe change it instead?

```suggestion
interface ActionBarProps extends HStackProps {
```

modules/react/action-bar/lib/ActionBar.tsx
```diff
@@ -1,76 +1,55 @@
...
+    <ResponsiveHStack
+      ref={ref}
+      as={Element}
+      position={position}
+      spacing="s"
+      {...containerStyles}
+      {...elProps}
+    />
+  ),
+});
```
NicholasBoll, line 55: I'm not sure there's a need for `ActionBarProps`. If so, I think we could more consistent with default props vs overriding props. For example, `spacing="s"` is an override prop while `position={position}` is a default prop. Also, we use `elemProps` everywhere else instead of `elProps`. I don't have a problem with `elProps`, but is there a strong reason to go with `elProps` instead?

```suggestion
export const ActionBar = createComponent('div')({
  displayName: 'ActionBar',
  Component: (elemProps: ActionBarProps, ref, Element) => (
    <ResponsiveHStack
      ref={ref}
      as={Element}
      position="fixed"
      spacing="s"
      {...elemProps}
    />
  ),
});
```

NicholasBoll (approved):

modules/codemod/lib/v7/spec/compoundActionBar.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+      <ActionBar fixed="true" />
+    `;
+
+    const expected = stripIndent`
+      import {ActionBar} from '@workday/canvas-kit-react/action-bar'
+
+      <ActionBar />
+    `;
+
+    expectTransform(input, expected);
```
alanbsmith, line 21: Two questions:

- Will this support default imports? `import ActionBar from '@workday/canvas-kit-react/action-bar`
- Will this support imports from `@workday/canvas-kit-react`?

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -35,3 +36,16 @@ rollback more easily if necessary.**
...
+
+We've refactored our ActionBar components to simplify logic.
+
+PR: https://github.com/Workday/canvas-kit/pull/1396
+
+**Changes:**
+
+- Component has been converted to compound component.
+- `fixed` prop has been removed from component.
+- Now ActionBar uses `position` prop to set container position instead `fixed` props.
```
alanbsmith, line 50:
```suggestion
- Now ActionBar uses `position` prop to set container position instead `fixed` prop.
```

alanbsmith (dismissed):
Just one small copy nit. The rest looks great! Thanks!!

modules/react/action-bar/lib/ActionBar.tsx
```diff
@@ -1,76 +1,37 @@
...
+      ref={ref}
+      as={Element}
+      spacing="s"
+      depth={1}
+      background={commonColors.background}
+      borderTop={`solid 1px ${colors.soap400}`}
+      padding={`${space.s} ${space.xl} `}
+      position="fixed"
+      bottom={0}
+      left={0}
```
alanbsmith, line 32: Worth noting that the `left` style prop is not RTL-safe. It will always be `left`. Is that what we want? Or should we use `insetInlineStart` which is RTL-safe? Here's some documentation on position props for reference: https://workday.github.io/canvas-kit/?path=/docs/labs-common-box--position

RayRedGoose: I added alignment to the right side instead of setting full width for container. So now it should work correctly for both directions.

modules/codemod/lib/v7/spec/compoundActionBar.spec.ts
RayRedGoose: - No. There is no default import from action-bar files.
- Yes. It supports imports from main directory.

modules/react/action-bar/index.ts
```diff
@@ -1,5 +1 @@
-import ActionBar from './lib/ActionBar';
-
-export default ActionBar;
```
alanbsmith, line 3: WIll the codemod also update any default imports to named imports? E.g.

```ts
import ActionBar from '@workday/canvas-kit-react/action-bar';
```
becomes this:

```ts
import { ActionBar } from '@workday/canvas-kit-react/action-bar';
```

modules/react/action-bar/lib/ActionBar.tsx
alanbsmith: Cool. Good to know. Thanks!

RayRedGoose:
This PR will be on hold by tomorrow. @alanbsmith found a change that not covered by codemod but It's reasonable to have opportunity to change default import to named.

modules/react/action-bar/index.ts
RayRedGoose: Thanks for catching it! I've updated existing codemod to add a script that changes default import to named or renamed import from `action-bar` package. Now codemod should cover this problem.

NicholasBoll (approved):

NicholasBoll:
This is a visual regression test that isn't showing the buttons. Is this intentional? https://www.chromatic.com/test?appId=5d854c26ba934e0020f5e98a&id=61d8c63dd73a51003ae4758a

Issue #1393, opened 2021-12-14
Create a CookieBanner Example


alanbsmith:
## 💡 Suggestion

Create an example `CookieBanner` component for users to reference to migrate from `DeprecatedCookieBanner`.

## Motivation

Now that we've deprecated the `CookieBanner` in v6, we'd like to provide an example for users to use so that they can migrate from the deprecated component.

## Acceptance Criteria

- [ ] Create a code example of `CookieBanner` that looks like what's in Workday's UI today using existing canvas-kit components and patterns
- [ ] Add link to the example in the v6 migration guide

PR #1402, opened 2021-12-21
docs: Add CookieBanner example

RayRedGoose:
## Summary

Resolves: #1393 

docs

### DIFF
--- a/modules/react/_examples/stories/examples/CookieBanner.tsx
+++ b/modules/react/_examples/stories/examples/CookieBanner.tsx
@@ -1,0 +1,113 @@
+import * as React from 'react';
+
+import {createComponent, styled, StyledType} from '@workday/canvas-kit-react/common';
+import {colors, commonColors, depth, type, space} from '@workday/canvas-kit-react/tokens';
+import {Hyperlink, HyperlinkProps, PrimaryButton} from '@workday/canvas-kit-react/button';
+
+const CookieBannerItem = createComponent('div')({
+  displayName: 'CookieBanner.Item',
+  Component: ({isRow, ...elProps}: ItemProps, ref) => (
+    <BannerItem ref={ref} isRow={isRow} {...elProps} />
+  ),
+});
+
+const CookieBannerSettings = createComponent('a')({
+  displayName: 'CookieBanner.Settings',
+  Component: (props: CookieSettingsProps, ref, Element) => (
+    <CookieSettings ref={ref} as={Element} {...props} />
+  ),
+});
+
+const CookieBanner = createComponent('div')({
+  displayName: 'CookieBanner',
+  Component: (props: CookieBannerProps, ref, Element) => (
+    <Banner ref={ref} as={Element} {...props} />
+  ),
+  subComponents: {Item: CookieBannerItem, Settings: CookieBannerSettings},
+});
+
+export const Basic = () => {
+  const DefaultNotice = `We use cookies to ensure that we give you the best experience on our website. 
+    If you continue without changing your settings, we'll assume that you are willing to receive cookies.`;
+  const onClickSettings = e => e.preventDefault();
+
+  return (
+    <ExampleContainer>
+      <CookieBanner isClosed={false}>
+        <CookieBanner.Item>{DefaultNotice}</CookieBanner.Item>
+        <CookieBanner.Item isRow>
+          <CookieBanner.Settings onClick={onClickSettings}>Settings</CookieBanner.Settings>
+          <PrimaryButton>Continue</PrimaryButton>
+        </CookieBanner.Item>
+      </CookieBanner>
+    </ExampleContainer>
+  );
+};
+
+interface CookieBannerProps extends StyledType {
+  isClosed?: boolean;
+}
+
+interface ItemProps extends StyledType {
+  isRow?: boolean;
+}
+
+interface CookieSettingsProps extends HyperlinkProps, StyledType {}
+
+const ExampleContainer = styled('div')({
+  minHeight: 84,
+  margin: space.xs,
+  position: 'relative',
+});
+
+const Banner = styled('div')(
+  type.levels.subtext.medium,
+  {
+    backgroundColor: commonColors.background,
+    borderTop: `1px solid ${colors.soap400}`,
+    display: 'flex',
+    ...depth[1],
+    padding: space.m,
+    alignItems: 'center',
+    justifyContent: 'space-between',
+    position: 'absolute',
+    bottom: 0,
+    left: 0,
+    right: 0,
+    zIndex: 99,
+    transition: 'transform 0.2s ease-out',
+    '@media (max-width: 450px)': {
+      flexDirection: 'column',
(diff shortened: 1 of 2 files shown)


modules/react/_examples/stories/examples/CookieBanner.tsx
```diff
@@ -0,0 +1,113 @@
...
+  displayName: 'CookieBanner.Item',
+  Component: ({isRow, ...elProps}: ItemProps, ref) => (
+    <BannerItem ref={ref} isRow={isRow} {...elProps} />
+  ),
+});
+
+const CookieBannerSettings = createComponent('a')({
+  displayName: 'CookieBanner.Settings',
+  Component: (props: CookieSettingsProps, ref, Element) => (
+    <CookieSettings ref={ref} as={Element} {...props} />
```
mannycarrera4, line 17: Should we render a tertiary button here in stead?

modules/react/_examples/stories/examples/CookieBanner.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const DefaultNotice = `We use cookies to ensure that we give you the best experience on our website. 
+    If you continue without changing your settings, we'll assume that you are willing to receive cookies.`;
+  const onClickSettings = e => e.preventDefault();
+
+  return (
+    <ExampleContainer>
+      <CookieBanner isClosed={false}>
+        <CookieBanner.Item>{DefaultNotice}</CookieBanner.Item>
+        <CookieBanner.Item isRow>
+          <CookieBanner.Settings onClick={onClickSettings}>Settings</CookieBanner.Settings>
```
mannycarrera4, line 39: I think we can just render a `tertiary button here`

RayRedGoose: Thanks for catching this. I agree, it's better to put `TertiaryButton` here.

mannycarrera4 (approved):
nice work!!

modules/react/_examples/stories/examples/CookieBanner.tsx
```diff
@@ -0,0 +1,99 @@
...
+  );
+};
+
+interface CookieBannerProps extends StyledType {
+  isClosed?: boolean;
+}
+
+interface ItemProps extends StyledType {
+  isRow?: boolean;
+}
```
NicholasBoll, line 45: Little nitpicks. I don't think these 2 prop interfaces need to extend `StyledType`. `StyledType` adds the `as` prop to a styled component. These are used inside a `createComponent` which already adds the `as` prop correctly. Extending `StyledType` should be removed. This way, `Banner` doesn't have to use `Pick<CookieBannerProps, 'isClosed'>` since that is the only prop available.

```suggestion
interface CookieBannerProps {
  isClosed?: boolean;
}

interface ItemProps {
  isRow?: boolean;
}
```

modules/react/_examples/stories/examples/CookieBanner.tsx
```diff
@@ -0,0 +1,99 @@
...
+    zIndex: 99,
+    transition: 'transform 0.2s ease-out',
+    '@media (max-width: 450px)': {
+      flexDirection: 'column',
+      alignItems: 'stretch',
+      textAlign: 'center',
+      padding: `${space.s} 0`,
+    },
+  },
+  ({isClosed}: Pick<CookieBannerProps, 'isClosed'>) => isClosed && {transform: 'translateY(100%)'}
```
NicholasBoll, line 76: It is weird for `CookieBanner` to have props that `Banner` uses since `CookieBanner` is a more specific version of `Banner`. I think it should be renamed `BannerProps`

```suggestion
  ({isClosed}: BannerProps) => isClosed && {transform: 'translateY(100%)'}
```

modules/react/_examples/stories/examples/CookieBanner.tsx
```diff
@@ -0,0 +1,99 @@
...
+const CookieBannerItem = createComponent('div')({
+  displayName: 'CookieBanner.Item',
+  Component: ({isRow, ...elProps}: ItemProps, ref) => (
+    <BannerItem ref={ref} isRow={isRow} {...elProps} />
+  ),
+});
+
+const CookieBanner = createComponent('div')({
+  displayName: 'CookieBanner',
+  Component: (props: CookieBannerProps, ref, Element) => (
```
NicholasBoll, line 16:
```suggestion
  Component: (props: BannerProps, ref, Element) => (
```

modules/react/_examples/stories/examples/CookieBanner.tsx
```diff
@@ -0,0 +1,99 @@
...
+  isRow?: boolean;
+}
+
+const ExampleContainer = styled('div')({
+  minHeight: 84,
+  margin: space.xs,
+  position: 'relative',
+});
+
+const Banner = styled('div')(
```
NicholasBoll, line 53: I think you might need to add `StyledType` here instead:

```suggestion
const Banner = styled('div')<BannerProps & StyledType>(
```

modules/react/_examples/stories/examples/CookieBanner.tsx
```diff
@@ -0,0 +1,99 @@
...
+      flexDirection: 'column',
+      alignItems: 'stretch',
+      textAlign: 'center',
+      padding: `${space.s} 0`,
+    },
+  },
+  ({isClosed}: Pick<CookieBannerProps, 'isClosed'>) => isClosed && {transform: 'translateY(100%)'}
+);
+
+const BannerItem = styled('div')(
```
NicholasBoll, line 79: I think you might need to add `StyledType` here:

```suggestion
const BannerItem = styled('div')<ItemProps & StyledType>(
```

modules/react/_examples/stories/examples/CookieBanner.tsx
```diff
@@ -0,0 +1,99 @@
...
+    '@media (max-width: 450px)': {
+      '&:not(:first-of-type)': {
+        marginTop: space.s,
+        '> *': {
+          flex: 1,
+        },
+      },
+    },
+  },
+  ({isRow}: Pick<ItemProps, 'isRow'>) =>
```
NicholasBoll, line 92: If you commit the suggestion above, you can remove the type override here:

```suggestion
  ({isRow}) =>
```

modules/react/_examples/stories/examples/CookieBanner.tsx
NicholasBoll: If you commit the above, you can remove the type override:

```suggestion
  ({isClosed}) => isClosed && {transform: 'translateY(100%)'}
```

modules/react/_examples/stories/examples/CookieBanner.tsx
RayRedGoose: Thanks for review. Suggested types look better than what I had. I've added this changes to the last commit.

NicholasBoll (approved):

Issue #978, opened 2021-02-23
Popper's containerElement doesn't work correctly/support full screen API


NicholasBoll:
## 🐛 Bug Report

Since the introduction of the PopupStack, `containerElement` no longer works properly. The PopupStack works with only a single container (`document.body`) for coordinating popup stacking and z-indexes. `Popper` does render inside the `containerElement`, but all behaviors are based on `document.body`. This causes an issue with stack hooks like `useCloseOnOutsideClick` not detecting clicks _inside_ the popup correctly.

A valid use-case is when using the [Full Screen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API). If the popup isn't a child of the full screen element, it will not be shown while in full screen mode. Full screen might be used for a video player.

## To Reproduce

Set a `containerElement` in a `Popper` component and use `useCloseOnOutsideClick`. Inside clicks will not be detected correctly.

Link: https://codesandbox.io/s/color-picker-popup-custom-container-xvmmz
Steps:
- click color picker icon
- hold the mouse down on an item in the popup menu

## Expected Behavior

The menu should not close

## Actual Results

The menu closes. This prevents clicks from working properly in the menu.

## Additional Notes

This will require changes to the PopupStack to allow for new popup stacks per container element. Popups still have to behave correctly relative to other items in the stack.

divyanshu023:
More Reproduction steps:
If the _containerElement_ is not the root, the Popper is never able to reposition the contents properly. The contents of _Popper_ does render inside _containerElement_, but _Popper_ does not position them properly. While inspecting the dom, i found a phantom dom element attached to _document.body_ which _Popper_ was trying to position

https://codesandbox.io/s/nifty-forest-i2ys3?file=/src/index.tsx

Issue #891, opened 2020-10-22
Select component doesn't work in conjuction with the Fullscreen API


zinas:
## 🐛 Bug Report

We are using the Select component inside the Workday Media widget. When we make the video go fullscreen, the Select components no longer work: when you click on them, the list of options doesn't appear. The reason this happens is because the list of options is appended to the body, while fullscreen is requested on a different element. This means that the stacking order of the options is lower than anything inside the fullscreen element and thus they wont appear

## To Reproduce

The code I mentioned is not yet merged, so I can't link to it. But you can reproduce by having a very simple page and requesting fullscreen on any element

## Expected Behavior

List of choices of the Select component appears

## Actual Results

As described in the first paragraph

## Browser (if applicable)

It applies to every browser

## Link to repl or repo (highly encouraged)

Please provide a reference to a repo or branch to help us reproduce this bug.

### Error Output

Paste the error output here

```bash

```

or

Include a screenshot of a visual bug

zinas:
Potential solution would be to be able to define a containerElement for the component

PR #1403, opened 2021-12-21
fix(popup-stack): Add support for the fullscreen API

NicholasBoll:
## Summary

Add fullscreen support to all popups. New hooks were added if your popup should transfer or close when entering/exiting a fullscreen context.

Fixes: #978
Fixes: #891

category

### Release Note
Fullscreen support was added to all Popups. 3 new hooks were added to help support fullscreen in whatever way you see fit:
- `useTransferOnFullscreenEnter`: Use if your popup should remain open and be transfer into the fullscreen element
- `useTransferOnFullscreenExit`: Use if your popup should remain open and transfer out of the fullscreen element back to the body element
- `useCloseOnFullscreenExit`: Use if your popup should close when fullscreen is exited

---

This fix is added to v5 with the option for popup stack adapters to opt into additional support for fullscreen. By default, a popup adapter does not need to support fullscreen. If fullscreen is entered, the PopupStack will use its own popup stack methods and when fullscreen is exited, control is returned back to the adapter. When a fullscreen popup stack context is entered, a new `stack` is pushed onto `window.__popupStackOfStacks`. If the adapter choses to support multiple stacks, an adapter must be pushed into this stack. This is how `PopupStack` keeps track of adapters on each stack.

A few components were updated to handle fullscreen:
- Tooltip
  - Closes when fullscreen is exited
- Select
  - Transfers into/out of the fullscreen context. Select menus can stay open during a transfer in/out of fullscreen

No Cypress tests were added to verify Fullscreen functionality. Cypress isn't able to test entering/exiting of fullscreen due to [not using native events](https://github.com/cypress-io/cypress/issues/311).


modules/popup-stack/lib/PopupStack.ts
```diff
@@ -1,3 +1,5 @@
+import screenfull from 'screenfull';
```
NicholasBoll, line 1: [screenfull](https://github.com/sindresorhus/screenfull) was chosen because it is as lightweight as possible and already used by some teams.

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -106,6 +108,7 @@ function getChildPopups(item: PopupStackItem, items: PopupStackItem[]): PopupSta
 
 interface Stack {
   items: PopupStackItem[];
+  container?: () => HTMLElement;
```
NicholasBoll, line 111: Optional. Falls back to `document.body` if not defined.

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -183,13 +186,23 @@ const setToWindow = (path: string, value: any) => {
...
   zIndex: {min: 30, max: 50, getValue: getValue},
-  _adapter: {} as Partial<typeof PopupStack>,
+  _adapter: {},
 };
 setToWindow('workday.__popupStack', stack);
 
+const stacks: Stack[] = getFromWindow('workday.__popupStackOfStacks') || [stack];
+
+(stacks as any).description = 'Global stack of popup stacks from @workday/canvas-kit/popup-stack';
+setToWindow('workday.__popupStackOfStacks', stacks);
```
NicholasBoll, line 200: This is the new variable added to `window`. This can be observed to see what stack we're currently in. Adapters should still manipulate this object.

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -212,31 +226,37 @@ export const PopupStack = {
...
    */
   add(item: PopupStackItem): void {
+    const stack = getTopStack();
     if (stack._adapter?.add) {
       stack._adapter.add(item);
       return;
     }
     stack.items.push(item);
-    document.body.appendChild(item.element);
+    (stack.container?.() || document.body).appendChild(item.element);
```
NicholasBoll, line 235: All the `|| document.body` is backwards compatible code. If this version of `PopupStack` isn't the first loaded, the `container` would not be defined. The `PopupStack` tries to register a global shared `window.workday.__popupStack` variable. This code cannot be removed.

modules/popup-stack/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "Apache-2.0",
   "main": "dist/commonjs/index.js",
   "module": "dist/es6/index.js",
-  "sideEffects": false,
+  "sideEffects": true,
```
NicholasBoll, line 9: This should have always been `true`. Since we manipulate `window` outside of a function call, this is required to support tree shaking. Proper tree shaking would break previous versions of `PopupStack`. I'm surprised we haven't received any bug reports.

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -183,13 +186,23 @@ const setToWindow = (path: string, value: any) => {
 // defined on the page, we need to use that one. Never, ever, ever change this variable name on
 // window
 const stack: Stack = getFromWindow('workday.__popupStack') || {
-  description: 'Global popup stack from @workday/canvas-kit-popup-stack',
-  items: [] as PopupStackItem[],
+  description: 'Global popup stack from @workday/canvas-kit/popup-stack',
+  container: () => document.body,
+  items: [],
   zIndex: {min: 30, max: 50, getValue: getValue},
-  _adapter: {} as Partial<typeof PopupStack>,
```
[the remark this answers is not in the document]
NicholasBoll: `stack` is already declared as type `Stack` which includes typing for `_adapter`, so this casting is not necessary.

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -361,3 +466,24 @@ export function resetStack() {
...
+  screenfull.on('change', () => {
+    if (screenfull.isFullscreen) {
+      if (screenfull.element) {
+        element = screenfull.element as HTMLElement;
+        PopupStack.pushStackContext(element);
+      }
+    } else if (element) {
+      PopupStack.popStackContext(element);
+    }
+    console.log('change', screenfull.isFullscreen, stacks.length);
```
NicholasBoll, line 487:
```suggestion
```

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -361,3 +466,24 @@ export function resetStack() {
...
+  });
```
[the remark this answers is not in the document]
NicholasBoll: Something would still need to call this method. Setting it to `null`  would mean the caller would have to do a nullable check. A global state variable would be the only way to ensure this method does not get called more than once, but it does put a weird cross-version dependency.

The way it is done here doesn't require a cross-version dependency, but does require `pushStackContext` to ignore requests from the same element. This requirement isn't only for this call though as pushing a new stack context could be from any number of components that could be placed on the same element. I think multiple calls to push a new stack context on the same element should be silently ignored (the side-effect is the same whether this was the first call or the 10th) rather than an error or other attempts to make sure the application only pushes a stack context element only once.

The question about the location of this call is around where could it be called where lifecycle is controlled. There are a few problems:
- How do we handle only adding a full screen listener if we have popups?
- How do we remove a full screen listener if there are no more popups?

I thought about other places this could go:
- In `Popper.tsx` in a `useEffect` hook
  - The problem here is the listener is added many more times than once per `PopupStack` import instance
- In `Popper.tsx` outside the component (side-effect)
  - The problem here is tree shaking. Tree shaking will theoretically remove non-reachable code. In practice, this probably isn't a problem because abuse of tree-shaking is so common. Many tree-shaking bundlers also need a special "pure" comment to flag removal and a side effect list this would not be flagged as "pure" and therefore not removed. There are higher optimizer settings that could still remove this code as "dead code" since there is no caller. For this reason, `@workday/canvas-kit-popup-stack` has been changed to a side-effect dependency to help remove chances of side-effect removal
- Only a method that is truly side-effect free
  - This would be a breaking change as the `PopupStack` library already relies on side effects to create a globally shared stack. It would mean the application would need to opt in to full screen support and could not be released as a patch. The reason the `PopupStack` relies on side-effects to remove the coordination required by a global call. It isn't obvious in larger application who is responsible for making this call. In a single app like Workday, that can be a platform team, but what about independent application development where an app may or may not be running in a context of the Workday application? That complexity would be pushed onto many developers. We made a decision to hide this complexity and simply using any popups would mean the system worked and it was designed with the multi-app use-case in mind.

I haven't thought of a better place to add the listener than `PopupStack.ts`. The lifecycle is tied to the page. This can mean slow memory leaks as any additional dynamic imports could add listeners that never get cleaned up, but that would be a very slow leak. The only real change that could mitigate slow memory leaks would be an additional global variable that checks if the listener is added, but that runs into the same multi-version dependency issue where the listener could never be changed by later versions. I still think this is the best place for this listener to be added to add support for full-screen without pushing the complexity to add development teams.

modules/popup-stack/lib/PopupStack.ts
```diff
@@ -183,13 +186,23 @@ const setToWindow = (path: string, value: any) => {
...
-  description: 'Global popup stack from @workday/canvas-kit-popup-stack',
-  items: [] as PopupStackItem[],
+  description: 'Global popup stack from @workday/canvas-kit/popup-stack',
+  container: () => document.body,
+  items: [],
   zIndex: {min: 30, max: 50, getValue: getValue},
-  _adapter: {} as Partial<typeof PopupStack>,
+  _adapter: {},
 };
 setToWindow('workday.__popupStack', stack);
```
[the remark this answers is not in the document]
NicholasBoll: That's possible, but a single stack needs to be created as the base stack and that needs to be shared. It could be that the first import of `PopupStack` is a version before this change and this variable would still exist. It could be that this code is removed and an old version that uses this still adds the global variable. It is important that all instances of this file use the same reference.

It is possible to deprecate, but that would take a lot of thought and testing. I have no plans to remove it at the moment.

alanbsmith (approved):

Issue #1311, opened 2021-10-15
Bump uuid dependency


alanbsmith:
## 💡 Suggestion

Bump `uuid` dependency. The versions we're using now, `^3.3.3` and `^3.4.4` are deprecated.

## Motivation

Just the usual housekeeping. When we bump uuid from v3 to v4, we’ll need to [update these imports](https://github.com/Workday/canvas-kit/search?q=uuid%2Fv4) to look like this:

```ts
// old
import uuid from 'uuid/v4';

// new
import { v4 as uuidv4 } from 'uuid';
```

NicholasBoll:
I actually plan on getting rid of `uuid` altogether! But we can bump it.

NicholasBoll:
Closed by #1408

Issue #1106, opened 2021-06-11
UUID dep is out of date and throwing a warning on install


alanbsmith:
## 🐛 Bug Report

When I installed our packages, I got the following warning:

```sh
warning @workday/canvas-kit-react > uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
```

## To Reproduce

```sh
npx create-react-app test-app
cd test-app
yarn add @workday/canvas-kit-react
```

## Expected Behavior

No warnings

## Browser (if applicable)

n/a

## Link to repl or repo (highly encouraged)

n/a

NicholasBoll:
Closed by #1408

Issue #664, opened 2020-05-20
Generate stable IDs


matsprea:
## 🚀 Feature Proposal

I'd like to have stable generated IDs instead of completely different at each execution.

## Motivation

As a developer I want to control these IDs and be able to use snapshots in automation to check their values. At the moment I have to mock the IDs because the ones generated by `uuid` are changing at every execution.

lychyi:
@matsprea Thanks for the issue.

Can you add a bit more context or a code block? I'm curious if mocking `uuid` causing another downstream issue for you? In the past, that's how we recommend that people do this for stable snapshots.

Also, which components are you specifically talking about? Some components have a manual id override like `FormField`. Others are ids generated as part of the component itself and they're not overridable by design.

We're not closed to the idea entirely but some more context about the pain point would help us a ton!

matsprea:
My specific issue is with the `popup` component that is generating an id for the `aria-labelledby` to reference the tittle of the popup and every time we run the automation, the id changes.

`NotificationsPopup >  › NotificationsPopup > › snapshot with open set to true

    expect(received).toMatchSnapshot()

    Snapshot name: `NotificationsPopup >  NotificationsPopup > snapshot with open set to true 1`

    - Snapshot
    + Received

    @@ -536,11 +536,11 @@
      
      <div
        class="emotion-20"
      >
        <div
    -     aria-labelledby="0f7757d3-5286-459e-9c98-28de998ca51e"
    +     aria-labelledby="58ae28cd-88d6-4158-996a-038fa8be04c7"
          aria-modal="true"
          class="emotion-19"
          role="dialog"
          tabindex="-1"
        >
    @@ -580,11 +580,11 @@
          <div
            class="emotion-18"
          >
            <h3
              class="emotion-3"
    -         id="0f7757d3-5286-459e-9c98-28de998ca51e"
    +         id="58ae28cd-88d6-4158-996a-038fa8be04c7"
            >
              Mentions (1)
            </h3>
            <div
              class="emotion-17"

      52 | export const takeSnapshot = (component, state) => {
      53 |   const { container } = getRenderedResult(component, state);
    > 54 |   expect(getTarget(container)).toMatchSnapshot();
         |                                ^
      55 | };
      56 | 
      57 | export const takeTableElementSnapshot = (component, state) => {

      at toMatchSnapshot (src/__tests__/testUtils.jsx:54:32)
[...]

Issue #602, opened 2020-04-28
Add a ID utility function for generating valid IDREFs


NicholasBoll:
## 🚀 Feature Proposal

We need a utility function to create valid strings to be added as `id`s of DOM elements. Perhaps we should make a hook while we're at it.

We will also need to be able to generate stable ids as a result of this feature (#664)

## Motivation

We make use of `uuid` to generate unique ids for accessibility requirements (label/for, input/id, etc). But the output of a `uuid` is not a valid [IDREF](https://www.w3.org/TR/xmlschema11-2/#IDREF). The spec says the id needs to always start with a letter and `uuid` sometimes returns a string that starts with a number.

The following will throw an error in the browser:
```js
document.querySelector('#1234')
```

We have to compensate in tests like this:
```js
document.querySelector('[id="1234"]')
```

A hook would also be nice since we use `React.useState(() => uuid())` based on https://codesandbox.io/s/p2ndq

## Example

```tsx
import { generateId, useId } from '@workday/canvas-kit-react-core'

class MyComponent extends React.Component {
  this.inputId = generateId()

  // ...
}

const MyComponent = () => {
  const id = useId();

  return (
    <input id={id} /* ... */ />
  )
}
```

Edit:
There is a utility: https://github.com/Workday/canvas-kit/blob/master/modules/common/react/lib/utils/useUniqueId.ts

jpante:
See #664 for additional context.

NicholasBoll:
Closed by #1408

PR #1408, opened 2022-01-05
feat(common): Remove uuid and update unique id generation

NicholasBoll:
## Summary

Our current id generation has issues.

- It generates very long ids that add noise to the page
- It is possible to generate a blank id
- Id generation is slow
- Snapshot tests require mocking the `uuid` module

Fixes: #1311, #1106, #664, #602

category

### Release Note
**NOTE for jest snapshots**: This change removes the `uuid` package and instead will generate a one-time client seed and then create auto-incrementing ids. This change will not break UI or automated UI tests. It will break snapshot tests however. Previously, the only way to get stable ids for snapshot tests was to mock the `uuid` module. This was an implementation detail. To make snapshots work again, add the following to your jest setup file:

```ts
import {setUniqueSeed, resetUniqueIdCount} from '@workday/canvas-kit-react/common';

beforeEach(() => {
  setUniqueSeed('a'); // force set the seed
  resetUniqueIdCount(); // reset the unique id count
});
```

This will ensure each Jest snapshot has ids that look like `a0` and `a1` and will be the same every time the snapshot is run. Do not use these methods in production though - it may lead to inaccessible applications due to IDREF collisions.

---

## Additional References

Test of different ID generation methods https://codesandbox.io/s/tkod6

Remaining dependencies on older versions of uuid:
- [x] `request` - this is a deprecated lib that we use for scripts
- [ ] `node-sass` - We're using v4. Newest is v7. This PR removed `request` which is using `uuid@3`, but hasn't been released yet: https://github.com/sass/node-sass/pull/3193. Currently blocked. `sass-loader` that is compatible with `node-sass@7` requires `webpack@5` which Storybook doesn't currently support.
- [x] `jest-unit` - We're using v10. Current version is v13
- [ ] `lerna` - We're using v3. Current version is v4
- [ ] `react-docgen-typescript-loader` - We're on v3.3.0. Latest is v3.7.2
- [ ] `storybook/addon-docs` - Oh, Storybook...


cypress/integration/Combobox.spec.ts
```diff
@@ -37,7 +37,7 @@ describe('Combobox', () => {
     });
 
     it('should not show menu', () => {
-      cy.findByRole('listbox').should('not.be.visible');
+      cy.findByRole('listbox').should('not.exist');
```
NicholasBoll, line 40: All these were due to a breaking change in [Cypress v6](https://docs.cypress.io/guides/references/changelog#6-0-0).

This "fix" did find legitimate issues with our tests, but I feel like we've lost some semantics. I want to know if the element is no longer accessible to the user. Whether it is still in the DOM or just hidden is an implementation detail. Perhaps I'll make an `be.inaccessible` assertion that does this.

cypress/integration/Tooltip.spec.ts
```diff
@@ -155,14 +155,14 @@ describe('Tooltip', () => {
     });
 
     it('the span element should not have an aria-describedby attribute', () => {
-      cy.get('button').should('not.have.attr', 'aria-describedby');
+      cy.get('span').should('not.have.attr', 'aria-describedby');
```
NicholasBoll, line 158: This was caught by the change in Cypress v6. Before, `not.have.attr` would pass even if `cy.get('button')` never actually matched anything. This was a typo from long ago that Cypress was perfectly happy to incorrectly pass.

cypress/integration/Tooltip.spec.ts
```diff
@@ -155,14 +155,14 @@ describe('Tooltip', () => {
...
     });
 
     it('the span element should not have an aria-label attribute', () => {
-      cy.get('button').should('not.have.attr', 'aria-describedby');
+      cy.get('span').should('not.have.attr', 'aria-describedby');
     });
 
-    context('when the "Delete" button is hovered', () => {
+    context('when the "Some Text" text is hovered', () => {
```
NicholasBoll, line 165: This was a copy/paste error. The tooltip isn't from a delete button, but a span with some text.

cypress/support/commands.ts
```diff
@@ -19,24 +19,33 @@ Cypress.Commands.add('injectAxe', () => {
 });
 
 // Needed for https://github.com/Bkucera/cypress-plugin-tab/issues/46
-Cypress.Commands.overwrite('visit', (originalFn, url, options = {}) => {
-  if (typeof url === 'object') {
-    // eslint-disable-next-line no-param-reassign
-    url = options.url;
+Cypress.Commands.overwrite('visit', (originalFn, urlOrOptions, inputOptions = {}) => {
```
NicholasBoll, line 22: Changes in this file were from Cypress adding type defintions to `Cypress.Commands.add` and `Cypress.Commands.overwrite` in [Cypress v9](https://docs.cypress.io/guides/references/changelog#9-0-0). In most cases, this is an improvement. In the case of commands that use Typescript overrides, this causes problems. I've rearranged command overrides and added some `any` in cases where the Cypress types introduce problems that cannot be resolved.

cypress/support/cypress-plugin-tab.d.ts
```diff
@@ -1,5 +1,5 @@
 declare namespace Cypress {
-  interface Chainable {
-    tab(options?: Partial<{shift: Boolean}>): Chainable;
+  interface Chainable<Subject = any> {
```
NicholasBoll, line 2: This was the correct typing all along

cypress/support/index.ts
```diff
@@ -6,3 +6,11 @@ import 'cypress-keyboard-plugin';
...
 
 import './commands';
+
+const resizeObserverLoopErrRe = /^[^(ResizeObserver loop limit exceeded)]/;
+Cypress.on('uncaught:exception', err => {
+  /* returning false here prevents Cypress from failing the test */
+  if (resizeObserverLoopErrRe.test(err.message)) {
+    return false;
+  }
+});
```
NicholasBoll, line 16: The `ResizeObserver loop limit exceeded` error started happening after upgrading Cypress from `5.6.0` to `9.2.0`. I'm not sure which Cypress version introduced the error. Perhaps later we can find out.

jest/setupTests.ts
```diff
@@ -14,3 +16,8 @@ configure({adapter: new Adapter()});
 // add convenience variables to the global context
 (global as any).verifyComponent = verifyComponent;
 (global as any).jest = jest;
+
+beforeEach(() => {
+  setUniqueSeed('a');
+  resetUniqueIdCount();
+});
```
NicholasBoll, line 23: This isn't actually needed for our tests, but it is nice to have something to point to as an example for teams that need this setup.

modules/preview-react/index.ts
```diff
@@ -6,3 +6,5 @@ export * from './select';
 export * from './side-panel';
 export * from './text-area';
 export * from './text-input';
+
+/// <reference types="@types/node" />
```
NicholasBoll, line 10: This removes Typescript errors complaining that `process` doesn't exist. `uuid` added `@types/node` to the global namespace for us.

jest/setupTests.ts
mannycarrera4: I'm sure teams are gonna wanna use this, nice addition

mannycarrera4: Should something like this go into the testing docs for reference?

mannycarrera4 (commented):
Looks good to me just some minor comments/ questions

modules/react/common/lib/utils/useUniqueId.ts
```diff
@@ -1,8 +1,24 @@
-import uuid from 'uuid/v4';
-
 import {useConstant} from './useConstant';
 
-export const generateUniqueId = () => uuid().replace(/^[0-9\-]*/gi, '');
+// Create a unique seed per import to prevent collisions from other versions of `useUniqueId`
+let seed = Math.random()
+  .toString(36)
+  .slice(2)
+  .replace(/[0-9]*/, '')
```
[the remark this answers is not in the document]
NicholasBoll: I added it to the `uuid` generation. Base 36 doesn't include dashes.

jest/setupTests.ts
mannycarrera4: nvm you documented this pretty well in the readme

mannycarrera4 (approved):

Issue #1334, opened 2021-11-09
Upgrade to React 17


myvuuu:
## 💡 Suggestion

Canvas Kit should upgrade to React 17. 

Open Questions:

- Will UI Plat to upgrade to React 17 in parallel? No
- What work is required for DS to upgrade to React 17? Upgrade dependency, no breaking changes

## Motivation

Our users have been asking for this upgrade for a while.

PR #1409, opened 2022-01-07
chore: Upgrade to React 17 and refactor our css prop

mannycarrera4:
## Summary
- Upgrade React and React Dom to 17
- Remove use `css` prop and `/** @jsx jsx */` pragma in examples and internal components
- Refactor components to use styled components where applicable

Fixes: https://github.com/Workday/canvas-kit/issues/1334

category

### BREAKING CHANGES
Canvas Kit will require teams to be on version ^17.x or at least 16.14 for backwards compatibility and we will remove the jsx pragma and use of css props. This does not mean consumers can’t use the css prop, just that internally we will stop using it. This would require teams to upgrade other parts of their app  including  React, potentially Babel and any other side effects that come with it.

## Additional References

- In React 17, they introduced a [new version of JSX Transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#whats-a-jsx-transform).  Although this is technically not a breaking change, it does introduce a complexity for teams using Babel, Emotion, and the /** @jsx jsx*/ pragma

- Canvas Kit uses this pragma because we have instances where we use the css prop from emotion. When upgrading to React 17, we saw the following error: `pragma and pragmaFrag cannot be set when runtime is automatic.`

- This error comes from babel from version 7.9 and greater when [they introduced automatic runtime](https://babeljs.io/docs/en/babel-preset-react)

- As part of this effort, we've also removed any use of enzyme since it will be [deprecating itself](https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl) and using [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/)


jest/verifyComponent.tsx
```diff
@@ -83,10 +81,10 @@ export function verifyComponent(
       });
     });
 
-    it('should merge css props correctly', () => {
+    it('should merge style props correctly', () => {
```
mannycarrera4, line 84: Wonder if for test purposes we either ignore or delete these tests?

modules/css/button/stories.tsx
```diff
@@ -1,3 +1,4 @@
+/** @jsxRuntime classic */
```
mannycarrera4, line 1: ignore cause css file

modules/labs-react/common/lib/utils/other.ts
```diff
@@ -0,0 +1,20 @@
...
+export function other<P extends OtherStyleProps>(props: P) {
+  for (const key in props) {
+    if (key in otherProps) {
+      const value = props[key as keyof OtherStyleProps] as keyof OtherStyleProps;
+      const styleFn = otherProps[key as keyof OtherStyleProps];
+      return styleFn(value);
+    }
+  }
+  return {};
+}
```
mannycarrera4, line 20: I added this issue to implement more of these: https://github.com/Workday/canvas-kit/issues/1423

modules/labs-react/header/stories/stories.tsx
```diff
@@ -220,17 +203,17 @@ storiesOf('Labs/Header/React', module)
...
+      <Box backgroundColor={colors.soap100} padding="m">
         <DeprecatedHeader
           title="Kitchen Sink"
           themeColor={DeprecatedHeader.Theme.Blue}
           brandUrl="#"
           onMenuClick={handleMenuClickTest}
           leftSlot={<SearchWithAutoComplete searchTheme={SearchForm.Theme.Dark} />}
           isCollapsed={false}
-          css={[{zIndex: 4}]}
+          style={{zIndex: 4}}
```
mannycarrera4, line 216: move this out into styled component

modules/preview-react/breadcrumbs/lib/Breadcrumbs/Dropdown/Button.tsx
```diff
@@ -36,14 +35,12 @@ export const DropdownButton = createComponent(IconButton)({
...
     const hasPlainVariant = variant === 'plain';
-    // gives the plain icon variant a little extra space between the chevron separators
-    const customStyles = css(hasPlainVariant ? {margin: '0 1px'} : {});
 
     return (
       <Element
         ref={ref}
         type="button"
-        css={customStyles}
+        style={hasPlainVariant ? {margin: '0 1px'} : {}}
```
mannycarrera4, line 43: Any ideas how to tackle this? It's an icon button underneath, can I use Box here?

modules/labs-react/header/stories/stories.tsx
mannycarrera4: or do we case since it's deprecated header?

modules/react/button/stories/visual-testing/utils.tsx
```diff
@@ -1,3 +1,4 @@
+/** @jsxRuntime classic */
```
mannycarrera4, line 1: this is a testing file so leaving it in for now

modules/labs-react/header/stories/stories.tsx
```diff
@@ -22,22 +20,7 @@ import {
...
-  padding: `0 0 64px 0`,
-  background: `linear-gradient(${chroma(colors.blueberry500)
-    .alpha(0.8)
-    .css()}, ${chroma(colors.blueberry400)
-    .alpha(0.8)
-    .css()}), url(${bgImg})`,
-  backgroundPosition: `0 50%`,
-  backgroundSize: 'cover',
-};
+import {Box} from '../../common';
```
NicholasBoll, line 23:
```suggestion
import {Box} from '@workday/canvas-kit-labs-react/common';
```

NicholasBoll (approved):

PR #1411, opened 2022-01-10
chore: Pin colors to version 1.4.0

Parker-Ledoux:
## Summary

Pin `colors` to unaffected version.

category
---
The maintainer of faker.js and colors.js went rogue and introduced some breaking changes into the latest patch versions of those modules. 

The code affecting `colors` is this infinite loop that was introduced: https://github.com/Marak/colors.js/commit/5d2d242f656103ac38086d6b26433a09f1c38c75#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1R18

I'm not sure what the long term strategy will be for CKR (maybe moving to a different library for console colors like [chalk](https://www.npmjs.com/package/chalk)) but this just makes sure that the local dev environments and CI does not go bust in case it pulls in the latest `colors` version.

### DIFF
--- a/modules/codemod/package.json
+++ b/modules/codemod/package.json
@@ -29,7 +29,7 @@
   },
   "homepage": "https://github.com/Workday/canvas-kit#readme",
   "dependencies": {
-    "colors": "^1.4.0",
+    "colors": "1.4.0",
     "jscodeshift": "^0.11.0",
     "yargs": "^16.2.0"
   },
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "babel-plugin-emotion": "^10.0.29",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2",
     "caniuse-lite": "^1.0.30001154",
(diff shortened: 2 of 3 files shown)


NicholasBoll:
We currently have a bigger issue with the CI using `npx` for `http-server` which currently uses the affected version of `colors`. Also this:

```
=> Found "colors@1.4.0"
info Has been hoisted to "colors"
info Reasons this module exists
   - "workspace-aggregator-7740824d-d993-4c47-b25a-85cd5607f0ba" depends on it
   - Specified in "devDependencies"
   - Hoisted from "_project_#jscodeshift#colors"
   - Hoisted from "_project_#colors"
   - Hoisted from "_project_#@workday#canvas-kit-codemod#colors"
   - Hoisted from "_project_#cypress#cli-table3#colors"
```

NicholasBoll:
`http-server` was updated to move to `chaulk`, which will unbork our CI.

NicholasBoll (approved):

PR #1413, opened 2022-01-12
chore: Remove colors dependency/upgrade dependencies

NicholasBoll:
## Summary

Removes use of the `colors` dependency in favor of `chalk`. The only dependency on `colors` is `cli-table3` currently used by Cypress.

I also upgraded the following dependencies to move off `colors`:

| Dependency              | Previous | Current | Notes                                                                                                                                   |
|-------------------------|----------|---------|-----------------------------------------------------------------------------------------------------------------------------------------|
| storybook               | 6.2.9    | 6.3.13  | 6.4+ requires TS 4.1                                                                                                                    |
| jscodeshift             | 0.11.0   | 0.13.1  | Included tighter type definitions which required codechanges                                                                            |
| cypress                 | 9.2.0    | 9.2.1   |                                                                                                                                         |
| react-docgen-typescript | 1.22.0   | 2.2.2   | Required small change to a function call and replaced `react-docgen-typescript-loader` with `@storybook/react-docgen-typescript-plugin` |

category

---

## Additional References

```
yarn why v1.22.10
[1/4] Why do we have the module "colors"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "colors@1.4.0"
info Reasons this module exists
   - "_project_#cypress#cli-table3" depends on it
   - Hoisted from "_project_#cypress#cli-table3#colors"
info Disk size without dependencies: "116KB"
info Disk size with unique dependencies: "116KB"
info Disk size with transitive dependencies: "116KB"
info Number of shared dependencies: 0
```


.storybook/docgen-plugin.js
```diff
@@ -1,5 +1,7 @@
 const docGen = require('react-docgen-typescript');
-const docGenLoader = require('react-docgen-typescript-loader/dist/generateDocgenCodeBlock.js');
+const {
+  generateDocgenCodeBlock,
+} = require('@storybook/react-docgen-typescript-plugin/dist/generateDocgenCodeBlock.js');
```
NicholasBoll, line 4: `react-docgen-typescript-loader` doesn't seem to be maintained and does use the newest version of `react-docgen-typescript`. The project was forked for use in Storybook.

.storybook/docgen-plugin.js
```diff
@@ -106,15 +108,15 @@ const processModule = (module, tsProgram) => {
...
-      })
-      .substring(module.userRequest.length) +
+    generateDocgenCodeBlock({
+      filename: module.userRequest,
+      source: module.userRequest,
+      componentDocs,
+      docgenCollectionName: 'STORYBOOK_REACT_CLASSES',
+      setDisplayName: true,
+      savePropValueAsString: true,
+      typePropName: 'type',
```
NicholasBoll, line 118: This was the value that was added as an optional param with a default. Since we don't use the default export and use a private one instead, this default isn't being set (even if it was released as a minor...). This was needed even with `react-docgen-typescript-loader`. I decided to keep `@storybook/react-docgen-typescript-plugin` since it is being actively updated and probably handles future Storybookisms. I may be around long enough to regret this decision, however.

.storybook/mdx-code-block-rewrite.js
```diff
@@ -19,7 +19,7 @@ module.exports = function rewriteExampleCodeBlock(source) {
   if (!hasArgsTable) imports.push('ArgsTable');
   return (
     (imports.length && hasSpecialBlocks
-      ? `import {${imports.join(',')}} from '@storybook/addon-docs/blocks';\n\n`
+      ? `import {${imports.join(',')}} from '@storybook/addon-docs';\n\n`
```
NicholasBoll, line 22: All changes from `@storybook/addon-docs/blocks` to `@storybook/addon-docs` is because of a deprecation added to 6.3 that isn't transpiled to ES5 which fails in IE11. The "fix" is to change imports. https://github.com/storybookjs/storybook/issues/15315

modules/codemod/lib/v6/deprecateCookieBanner.ts
```diff
@@ -1,124 +1,15 @@
...
-    .find(j.MemberExpression, {object: {type: 'Identifier', name: 'CookieBanner'}})
-    .forEach(nodePath => {
-      const identifier = nodePath.value.object as Identifier;
-      identifier.name = 'DeprecatedCookieBanner';
-    });
+  return renameImports(api, root, '@workday/canvas-kit-react/cookie-banner', {
+    CookieBanner: 'DeprecatedCookieBanner',
+    CookieBannerProps: 'DeprecatedCookieBannerProps',
+  }).toSource();
+};
```
NicholasBoll, line 13: I ran into type errors in v6 codemod files because of the upgrade of `jscodeshift` from `0.11.0` to `0.13.1`. The Typescript types were tightened and all these `find` commands didn't return a `boolean`. As I was looking to fix them all, I noticed almost all the codemods renamed import specifiers. So I made a utility to do this.

modules/codemod/lib/v6/index.ts
```diff
@@ -21,5 +21,7 @@ export default function transformer(file: FileInfo, api: API, options: Options)
     renameCanvasDepthValue,
   ];
 
-  return fixes.reduce((source, fix) => fix({...file, source}, api, options), file.source);
-}
+  return fixes.reduce((source, fix) => fix({...file, source}, api, options) as string, file.source);
```
NicholasBoll, line 24: `as string` is needed because the `Transform` type enforces the return type which `reduce` is fumbling over because initial value is a string.

.storybook/docgen-plugin.js
willklein: This import was weird but I looked and saw that `generateDocgenCodeBlock` wasn't exported from the index. I'm not sure if the `.js` extension is needed but maybe because of how it wasn't exported publicly.

modules/codemod/index.js
```diff
@@ -2,32 +2,32 @@
...
 
 const {spawn} = require('child_process');
-require('colors');
+const chalk = require('chalk');
 
 const {_: commands, path} = require('yargs')
   .scriptName('canvas-kit-codemod')
-  .usage('$0 <transform> [path]'.brightBlue.bold)
-  .command('v5 [path]', 'Canvas Kit v4 > v5 migration transform'.gray, yargs => {
+  .usage(chalk.bold.blueBright('$0 <transform> [path]'))
```
willklein, line 9: I prefer chalk's syntax. 😃

modules/codemod/lib/v5/getImportRenameMap.ts
```diff
@@ -1,4 +1,4 @@
-import {Collection, ImportDeclaration, JSCodeshift, CallExpression} from 'jscodeshift';
```
willklein, line 1: Did the codemod's need to get touched in this PR?

willklein (commented):
I approve everything except the modules/codemod/* changes, which I'll need more time to go over.

willklein (approved):
I went through the code mod refactor too. :shipit:

.storybook/docgen-plugin.js
NicholasBoll: That's a good question. `generateDocgenCodeBlock` wasn't exported from `react-docgen-typescript-loader`, but I'll check `@storybook/react-docgen-typescript-plugin`

NicholasBoll: `generateDocgenCodeBlock` is not exported: https://github.com/hipstersmoothie/react-docgen-typescript-plugin/blob/master/src/index.ts

We make our own plugin for processing prop types because we do extra processing. The Storybook plugin does not allow injecting any options to the generator.

For example:
- we mark the `as` prop as optional because `react-docgen-typescript` thinks it is supposed to be required because of the overload.
- We rewrite `as` prop type to a list of valid elements instead of `React.Element<any>`
- We filter style props from `Box` and instead link to `Box`. If we didn't do this, the prop list would be huge and the same for almost all compound components

modules/codemod/index.js
NicholasBoll: Yes. I don't like when prototypes are modified. It is easy to make a mistake assuming the prototype is modified, but you didn't modify it. Simple changes can break in unpredictable ways.

modules/codemod/lib/v5/getImportRenameMap.ts
NicholasBoll: Sort of. `jscodeshift` was updated from `0.11.0` to `0.13.1` to remove the `colors` dependency. This included a change to the type definitions of `find` which now requires a `boolean` return. This effected almost all the v6 codemods. This forced me to review the logic in the codemod. Upon inspection I noticed 90% of the codemods were the same. I had 2 options:

1. Add `boolean` returns to all instances of `root.find`
2. Create a `renameImports` utility with the changes and update codemods to use it thus removing a lot of redundant code

I chose 2️⃣ because it helps to understand what the codemods do. It also helps when more need to be created.

Issue #1414, opened 2022-01-12
Popups scroll to the top of the page when using useInitialFocus


NicholasBoll:
## 🐛 Bug Report

The `useInitialFocus` hook calls `focus` before PopperJS positions a Popup. This will cause the browser to scroll to the top of the page, losing context. The `useInitialFocus` should wait.

## Link to repl or repo (highly encouraged)

[Popup Scroll Bug](https://codesandbox.io/s/popup-scroll-bug-77sqh)

Scroll down to the buttons and click the "Delete Item" button. The browser will scroll to the top of the page. The popup will not be visible.

NicholasBoll:
Fixed in v5.3.8. Will be fixed in v6 shortly

PR #1415, opened 2022-01-12
fix(popup): Fix body scrolling using useInitialFocus

NicholasBoll:
## Summary

Fixes: #1414 

category


alanbsmith (approved):
Line breaks go brrrrrr. 🥶

PR #1418, opened 2022-01-13
test: Remove enzyme

NicholasBoll:
## Summary

Enzyme is not being supported long term. The React 17 adapter is unofficial and there may not be an adapter for future versions of React.

category

---

## Additional Notes

I was able to remove extra setup code required by Enzyme, but not by React Testing Library


modules/labs-react/header/spec/GlobalHeader.spec.tsx
```diff
@@ -1,111 +1,26 @@
...
-window.resizeBy = (x: number, y: number) => {
-  // @ts-ignore
-  window.innerWidth = x;
-  // @ts-ignore
-  window.innerHeight = y;
-  window.dispatchEvent(new Event('resize'));
-};
-
-// @ts-ignore
-window.requestAnimationFrame = cbFn => cbFn();
```
NicholasBoll, line 41: Luckily none of this is necessary anymore.

alanbsmith: haha BYE! ✌🏻

modules/preview-react/menu/lib/MenuItem.tsx
```diff
@@ -274,7 +274,7 @@ class MenuItem extends React.Component<MenuItemProps> {
           id={id}
           role={role}
           onClick={this.handleClick}
-          aria-disabled={!!isDisabled}
+          aria-disabled={isDisabled ? true : undefined}
```
alanbsmith, line 277: 👍🏻

alanbsmith (approved):
Thanks for removing Enzyme!

.github/workflows/pull-request.yml
```diff
@@ -51,7 +53,7 @@ jobs:
       - uses: actions/cache/@v2
         with:
           path: .cache/cypress
-          key: ${{ runner.os }}-cypress-cache-version-${{ steps.cypress-version.outputs.version }}
+          key: ${{ runner.os }}-cypress-cache-version-${{ needs.install.outputs.cypress-version }}
```
NicholasBoll, line 56: When everything was under a single job, the steps context could be used to get outputs of a specific step. When we parallelized the workflow, we had to break out the workflow into multiple jobs with dependencies to make sure everything ran in the proper order. Github doesn't allow a step to access the output of another job. I never updated the calls for the `cypress-version` string in the jobs. Instead, you have to add `outputs` to a job. From there, a different job can access those job outputs if the needs is added (meaning depending on the job). Then you can access those outputs via `needs.{job_id}.outputs.{output_name}`. So I had to add job outputs and update downstream jobs to access this output.

So instead of Github failing with a "could not find" something error message, it simply returns blank. Somehow this now invalid cache key returns Cypress 5.6.0 instead of 9.2.0. My guess is the invalid cache key will return Cypress 5.6.0 or 9.2.0 on whether the `support` or `master` job was run last that warmed the cache. I happened to find this because I ran a PR off master soon after a job on support primed the cache with the 5.6.0 version. Originally I found it because I changed Cypress from 9.2.0 to 9.2.1 and thus the binary missing. I thought it was just my PR, but that was a coincidence

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idoutputs

modules/preview-react/menu/lib/MenuItem.tsx
NicholasBoll: We didn't catch this before, but this was actually a bug.

PR #1424, opened 2022-01-18
fix: Allow repositioning of PopperJS instances

NicholasBoll:
## Summary

Allow access to the `PopperJS` instance to allow resize positioning of a Popup.

category

## Additional References

Before this change, the following video would show the popup size changing, but not recentering.

PopperResize


alanbsmith (approved):
Looks good! Thanks!

PR #1430, opened 2022-01-19
feat(button): Add ExternalHyperlink and missing stories for Hyperlinks

vibdev:
## Summary

Adds stories for the Hyperlink component. Provides guidance for how to use the component.
Adds new ExternalHyperlink component. 

Components

---

## Checklist

### New visual test
Screen Shot 2022-02-16 at 1 44 44 PM

### Design Spec
Screen Shot 2022-02-15 at 4 01 59 PM


modules/react/button/stories/button/examples/Hyperlink.tsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+import {Hyperlink} from '@workday/canvas-kit-react/button';
+
+export const Link = () => <Hyperlink href="#">Hyperlink</Hyperlink>;
```
vibdev, line 5: Is it better to do it this way and have 'link' in the storybook nav or 
```
import {Hyperlink as Link} from '@workday/canvas-kit-react/button';

export const Hyperlink = () => <Link href="#">Hyperlink</Link>;
```
To have nav match the component name but have a more convoluted code to copy

modules/react/button/stories/button/examples/HyperlinkExternal.tsx
```diff
@@ -0,0 +1,21 @@
...
+    as={Hyperlink}
+    target="_blank"
+    href="http://workday.com"
+    rel="noreferrer"
+    display="inline-flex"
+    alignItems="center"
+    spacing="xxxs"
+  >
+    <span>External HyperLink</span>
+    <SystemIcon icon={extLinkIcon} aria-label="Opens link in new window" size={16} />
```
vibdev, line 19: Design question, should we should how to make the icon color match the text color
```
const [hover, setHover] = React.useState(false)
...

<HStack
    as={Hyperlink}
    ...
    onMouseOut={() => setHover(false)}
    onMouseOver={() => setHover(true)}
    onBlur={() => setHover(false)}
    onFocus={() => setHover(true)}
>
     ...
    <SystemIcon
        icon={extLinkIcon}
        ...
        color={hover ? colors.blueberry500 : colors.blueberry400}
        colorHover={colors.blueberry500}
    />
</HStack>
```

NicholasBoll: That's a good question. And assuming yes, what's the best way to do that? It could be done via CSS selectors or by state like in your example. I tend to not like changing state when DOM state changes when CSS is capable of that. Emotion handles this fairly well, but the CSS selectors can get a bit tricky. Your example causes inefficiencies because the React reconciler and Emotion both get involved: https://codesandbox.io/s/external-link-css-differences-ly3mk

A profile of a mouse over on the React state (your suggested version):
image

A profile of a mouse over on the pre-calculated CSS version:

image

You can see both screenshots have a `mouseout` and a `mouseover`. The first one has a larger `mouseover` with a `Recalculate Style` block within the `mouseover` event because Emotion has to inject a new stylesheet which can cause layout cache invalidation. It is hard to see from the screenshot, but the first one contains 2 style recalculations. The first being from the browser and the second being triggered from a change in Emotion. The second example does not require a style recalculation for the hover because no styles have change. Both contain the usual paint and composition.

NicholasBoll: Perhaps these nuances warrant an `ExternalHyperlink` component?

vibdev: ExternalHyperlink makes a lot of sense to me

NicholasBoll (approved):

modules/react/button/lib/HyperlinkExternal.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {HStack} from '@workday/canvas-kit-labs-react/layout';
+import {extLinkIcon} from '@workday/canvas-system-icons-web';
+import {SystemIcon, systemIconStyles} from '@workday/canvas-kit-react/icon';
+import {Hyperlink, HyperlinkProps} from './Hyperlink';
+
+export interface HyperlinkExternalProps extends HyperlinkProps {
+  /**
+   * Override the aria-label for the icon
+   * @default 'Opens link in new window'
+   */
```
NicholasBoll, line 18: We should describe the purpose of this prop. It is meant to inform the user that the link opens in a new window and thus should be translated.

```suggestion
  /**
   * Informs a screen reader user the link will open in a new window. It is read after the link text.
   * This value will need to be translated.
   * @default 'Opens link in new window'
   */
```

modules/react/button/stories/button/Button.stories.mdx
```diff
@@ -117,6 +122,38 @@ Undocumented props are spread to the underlying `<button>` element.
...
+
+Undocumented props are spread to the underlying `<a>` element.
+
+<ArgsTable of={Hyperlink} />
+
+### HyperlinkExternal
+
+If your link should open in a new tab or window use the `<HyperlinkExternal>` component. It will let
+users know about the behavior by adding an icon, `extLinkIcon`. You can override `aria-label` of the
+icon by using the `iconLabel` attribute.
```
NicholasBoll, line 147: We've been moving away from default translations. In this case, it might make sense. I think we should add a bit about translation though, describing the purpose of `iconLabel`. A callout will help with translation efforts. Canvas Kit doesn't take care of translations itself as many teams using Canvas Kit have different translation solutions.

NicholasBoll (approved):

modules/react/button/stories/button/Button.stories.mdx
```diff
@@ -15,6 +17,9 @@ import {SecondaryInverse} from './examples/SecondaryInverse';
 import {Tertiary} from './examples/Tertiary';
 import {TertiaryInverse} from './examples/TertiaryInverse';
 import {Delete} from './examples/Delete';
+import {Link} from './examples/Hyperlink';
+import {LinkInverse} from './examples/HyperlinkInverse';
+import {LinkExternal} from './examples/HyperlinkExternal';
 
 <Meta title="Components/Buttons/Button/React" />
```
alanbsmith, line 24: I wonder if it's finally time to move these stories out of the Button section. Seems so odd there.

modules/react/button/stories/button/Button.stories.mdx
```diff
@@ -15,6 +17,9 @@ import {SecondaryInverse} from './examples/SecondaryInverse';
 import {Tertiary} from './examples/Tertiary';
 import {TertiaryInverse} from './examples/TertiaryInverse';
 import {Delete} from './examples/Delete';
+import {Link} from './examples/Hyperlink';
+import {LinkInverse} from './examples/HyperlinkInverse';
+import {LinkExternal} from './examples/HyperlinkExternal';
```
alanbsmith, line 22: This is a small nit, and it really has to do with the overall organization of stories. Because these examples are called Link, Link Inverse, and Link External, those names show up in the left nav and searching for "hyperlink" removes them from the results, which is some nonsense outside of our control.

searching the storybook left nav for hyperlink and seeing the aberrant behavior described above

alanbsmith: But that kind of makes me think these examples should live elsewhere.

```
Components
|- Hyperlink
|  |- Basic
|  |- External
|  |- Inverse
```

vibdev: 100% agree

vibdev: @alanbsmith should we save reorganizing to a new component for v7?

modules/react/button/lib/ExternalHyperlink.tsx
```diff
@@ -0,0 +1,51 @@
...
+  display: 'inline-flex',
+  flexDirection: 'row',
+  alignItems: 'center',
+});
+
+const iconSize = '1em';
+
+const StyledSystemIcon = styled(SystemIcon)<StyledType>({
+  ...iconStyles,
+  width: `calc(${iconSize} - 2px)`,
```
vibdev, line 31: We need to reduce the spacing after the icon otherwise our svg adds too much space between words.

modules/react/icon/lib/Icon.tsx
```diff
@@ -18,6 +18,7 @@ export default class Icon extends React.Component<IconProps> {
 
     if (size) {
       iconStyles['& svg'] = {
+        ...((iconStyles['& svg'] as CSSObject) ?? {}),
```
vibdev, line 21: @NicholasBoll had to add this so Icon doesn't yeet any existing svg styles in the styles object

modules/react/button/stories/button/Hyperlink.stories.mdx
```diff
@@ -0,0 +1,74 @@
...
+## Usage
+
+### Hyperlink
+
+Hyperlinks should be used when you want to navigate away from the current page or to an anchor on
+the current page.
+
+<ExampleCodeBlock code={Link} />
+
+Hyperlinks also have an `inverse` variant. Use this variant when you need to place a Hyperlink on a
```
alanbsmith, line 34:
```suggestion
#### Inverse

Hyperlinks also have an `inverse` variant. Use this variant when you need to place a Hyperlink on a
```

modules/react/button/stories/button/Hyperlink.stories.mdx
```diff
@@ -0,0 +1,74 @@
...
+### ExternalHyperlink
+
+If your link should open in a new tab or window use the `<ExternalHyperlink>` component. It will let
+users know about the behavior by adding an icon, `extLinkIcon`. By default, the icon has a label of
+"Opens link in new window". You can override `aria-label` of the icon by using the `iconLabel`
+attribute. This override will be needed for translations.
+
+<ExampleCodeBlock code={ExternalLink} />
+
+ExternalHyperlinks also have an `inverse` variant. Use this variant when you need to place a
```
alanbsmith, line 54:
```suggestion
#### Inverse

`ExternalHyperlink`s also have an `inverse` variant. Use this variant when you need to place a
```

modules/react/button/stories/button/Hyperlink.stories.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+```sh
+yarn add @workday/canvas-kit-react
+```
+
+## Usage
+
+### Hyperlink
+
+Hyperlinks should be used when you want to navigate away from the current page or to an anchor on
```
alanbsmith, line 29:
```suggestion
`Hyperlink`s should be used when you want to navigate away from the current page or to an anchor on
```

modules/react/button/stories/button/Hyperlink.stories.mdx
```diff
@@ -0,0 +1,74 @@
...
+Undocumented props are spread to the underlying `<a>` element.
+
+<ArgsTable of={Hyperlink} />
+
+### ExternalHyperlink
+
+If your link should open in a new tab or window use the `<ExternalHyperlink>` component. It will let
+users know about the behavior by adding an icon, `extLinkIcon`. By default, the icon has a label of
+"Opens link in new window". You can override `aria-label` of the icon by using the `iconLabel`
+attribute. This override will be needed for translations.
```
alanbsmith, line 50:
```suggestion
prop. This override will be needed for translations.
```

modules/react/button/stories/button/Hyperlink.stories.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+<ExampleCodeBlock code={ExternalLink} />
+
+ExternalHyperlinks also have an `inverse` variant. Use this variant when you need to place a
+Hyperlink on a dark or colorful background such as `blueberry400`.
```
alanbsmith, line 55:
```suggestion
`ExternalHyperlink`s also have an `inverse` variant. Use this variant when you need to place the link on a dark or colorful background such as `blueberry400`.
```

modules/react/button/stories/button/Hyperlink.stories.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+#### Props
+
+Undocumented props are spread to the underlying `<a>` element.
+
+<ArgsTable of={ExternalHyperlink} />
+
+### Accessible Use of the `as` Prop
+
+Like many of our components, Buttons accept an `as` prop, which lets you change the underlying
```
alanbsmith, line 67:
```suggestion
Like many of our components, `Hyperlink` and `ExternalHyperlink` accept an `as` prop, which lets you change the underlying
```

modules/react/button/stories/button/Hyperlink.stories.mdx
```diff
@@ -0,0 +1,74 @@
...
+semantic element - usually an `<a>`, or a `<button>`. This should be done with caution to ensure the
+best accessibility. Generally, `<button>` elements should be used for actions, and `<a>` elements
+should be used for navigation.
```
alanbsmith, line 70: Good call out. Thanks for adding this. 👍🏻

Issue #1444, opened 2022-01-27
Preview FormField.Hint has incorrect error color


vibdev:
## 🐛 Bug Report

When there is an error on the model, the text color should equal the themed error color

## To Reproduce

[See storybook](https://workday.github.io/canvas-kit/?path=/story/preview-inputs-text-input-react--error)

## Expected Behavior

Should be red

## Actual Results

Emotion is applying the css color form the type instead of the error.

PR #1445, opened 2022-01-27
fix(preview): Preview FormField.Hint error text not the correct color

vibdev:
## Summary

Fixes: #1444 

Components

---

## Checklist

## Additional References

Screen Shot 2022-01-26 at 5 56 12 PM

### DIFF
--- a/modules/preview-react/form-field/lib/FormFieldHint.tsx
+++ b/modules/preview-react/form-field/lib/FormFieldHint.tsx
@@ -39,9 +39,8 @@ export const FormFieldHint = createComponent('p')({
         as={Element}
         css={{
           ...type.levels.subtext.medium,
+          color: localModel.state.hasError ? theme.canvas.palette.error.main : undefined,
         }}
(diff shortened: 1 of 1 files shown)


modules/preview-react/form-field/lib/FormFieldHint.tsx
```diff
@@ -39,9 +39,8 @@ export const FormFieldHint = createComponent('p')({
         as={Element}
         css={{
           ...type.levels.subtext.medium,
+          color: localModel.state.hasError ? theme.canvas.palette.error.main : undefined,
         }}
-        color={localModel.state.hasError ? theme.canvas.palette.error.main : undefined}
-        marginY={space.xxs}
```
vibdev, line 44: Margin isn't being used in storybook example (overridden by VStack). Don't really think we need a default margin since consumers can add whatever they need here.

mannycarrera4 (approved):

PR #1446, opened 2022-01-28
feat(BrandAssets): Add alt property to Workday logos.

renato-rossi:
## Summary

Implement [empty alt attributes to image tags](https://www.w3.org/WAI/tutorials/images/decorative/) on Workday logos to improve accessibility for screen reader users.

category

---


modules/react/common/lib/parts/_brand-assets.ts
```diff
@@ -1,11 +1,11 @@
 const CDN_URI = '//design.workday.com/images/';
 
-export const dubLogoBlue = `<img src="${CDN_URI}ck-dub-logo-blue.svg">`;
+export const dubLogoBlue = `<img src="${CDN_URI}ck-dub-logo-blue.svg" aria-label="">`;
```
NicholasBoll, line 3: I think an `alt` attribute should be present and not an `aria-label`. And it shouldn't be empty. It should say "Workday". If any other word (like "logo") is included, we may have to translate it.

```suggestion
export const dubLogoBlue = `<img src="${CDN_URI}ck-dub-logo-blue.svg" alt="Workday">`;
```

modules/react/common/lib/parts/_brand-assets.ts
```diff
@@ -1,11 +1,11 @@
...
 
-export const dubLogoWhite = `<img src="${CDN_URI}ck-dub-logo-white.svg">`;
+export const dubLogoWhite = `<img src="${CDN_URI}ck-dub-logo-white.svg" aria-label="">`;
```
NicholasBoll, line 5:
```suggestion
export const dubLogoWhite = `<img src="${CDN_URI}ck-dub-logo-white.svg" alt="Workday">`;
```

modules/react/common/lib/parts/_brand-assets.ts
```diff
@@ -1,11 +1,11 @@
...
 
-export const wdayLogoBlue = `<img src="${CDN_URI}ck-wday-logo-blue.svg">`;
+export const wdayLogoBlue = `<img src="${CDN_URI}ck-wday-logo-blue.svg" aria-label="">`;
```
NicholasBoll, line 7:
```suggestion
export const wdayLogoBlue = `<img src="${CDN_URI}ck-wday-logo-blue.svg" alt="Workday">`;
```

modules/react/common/lib/parts/_brand-assets.ts
```diff
@@ -1,11 +1,11 @@
...
 
-export const wdayLogoWhite = `<img src="${CDN_URI}ck-wday-logo-white.svg">`;
+export const wdayLogoWhite = `<img src="${CDN_URI}ck-wday-logo-white.svg" aria-label="">`;
```
NicholasBoll, line 9:
```suggestion
export const wdayLogoWhite = `<img src="${CDN_URI}ck-wday-logo-white.svg" alt="Workday">`;
```

modules/react/common/lib/parts/_brand-assets.ts
```diff
@@ -1,11 +1,11 @@
...
 
-export const miniWdayLogoBlue = `<img src="${CDN_URI}ck-mini-wday-logo-blue.svg">`;
+export const miniWdayLogoBlue = `<img src="${CDN_URI}ck-mini-wday-logo-blue.svg" aria-label="">`;
```
NicholasBoll, line 11:
```suggestion
export const miniWdayLogoBlue = `<img src="${CDN_URI}ck-mini-wday-logo-blue.svg" alt="Workday">`;
```

modules/react/common/lib/parts/_brand-assets.ts
NicholasBoll: Github didn't allow me to span a suggestion across all lines

NicholasBoll: The W3C article linked mentioned an empty alt attribute and not an empty aria-label attribute. I'm asking if the alt value should be empty or "Workday".

renato-rossi: @NicholasBoll You're correct, instead of an empty aria-label it should be an empty alt property. Empty alt tells the browser that the image is decorative, if we set alt to Workday, screen readers would read the alt value as an image description, which is what we are trying to avoid. I'll fix this now.

NicholasBoll (approved):

PR #1450, opened 2022-02-01
fix(popup-stack): Fix popup element removal when adapter is used

NicholasBoll:
## Summary

Fixes an issue where an adapted PopupStack was never removed from the DOM

category

### DIFF
--- a/modules/popup-stack/lib/PopupStack.ts
+++ b/modules/popup-stack/lib/PopupStack.ts
@@ -246,7 +246,10 @@ export const PopupStack = {
    */
   remove(element: HTMLElement): void {
     // Find the stack the popup belongs to.
-    const stack = find(stacks, stack => !!find(stack.items, item => item.element === element));
+    const stack = find(
+      stacks,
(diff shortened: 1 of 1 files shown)


modules/popup-stack/lib/PopupStack.ts
```diff
@@ -246,7 +246,10 @@ export const PopupStack = {
    */
   remove(element: HTMLElement): void {
     // Find the stack the popup belongs to.
-    const stack = find(stacks, stack => !!find(stack.items, item => item.element === element));
+    const stack = find(
+      stacks,
+      stack => !!find(PopupStack.getElements(stack), el => el === element)
+    );
```
NicholasBoll, line 252: It was a mistake to use `stack.items` since that only exists on non-adapted stacks. `PopupStack.getElements` should always return elements.

willklein (approved):
:shipit:

NicholasBoll:
I fixed a bug where the `automerge` label could merge too soon

Issue #1447, opened 2022-01-28
Promote components: Box, Flex, Stack


mannycarrera4:
## 🚀 Feature Proposal

After an assessment done in https://github.com/Workday/canvas-kit/issues/1395, we've decided that `Box, Flex, and Stack` are good components to promote. 

As part of this promotion make sure to:
- Update any documentation that references these components
- Use script to promote to main module

PR #1460, opened 2022-02-07
chore: Migrate Flex, Stack and Box to main under layout 

mannycarrera4:
## Summary
Promote Flex, Stack and Box to the main react package

Fixes: https://github.com/Workday/canvas-kit/issues/1447

category

### BREAKING CHANGES
Flex, Box and Stack will no longer be imported from `@workday/canvas-kit-labs-react`. They will now be imported from `@workday/canvas-kit-react/layout`


modules/labs-react/common/spec/SSR.spec.tsx
```diff
@@ -3,7 +3,7 @@
  */
 import * as React from 'react';
 import {renderToString} from 'react-dom/server';
-import {Box} from '../';
+import {Box} from '@workday/canvas-kit-react/layout';
```
NicholasBoll, line 6: This whole spec file should be moved to `/layout`. It only tests `Box`

modules/react/layout/README.md
```diff
@@ -1,3 +1,20 @@
+# Canvas Kit React Layout
+
+<a href="https://github.com/Workday/canvas-kit/tree/master/modules/labs-react/README.md">
+  <img src="https://img.shields.io/badge/LABS-beta-orange" alt="LABS: Beta" />
+</a>  This component is work in progress and currently in pre-release.
+
```
NicholasBoll, line 6: Seems like this should be removed? This isn't a labs package.
```suggestion
```

modules/react/layout/README.md
```diff
@@ -1,3 +1,20 @@
...
+a collection of layout components
```
NicholasBoll, line 7: Is this a complete sentence?

```suggestion
A collection of layout components.
```

modules/react/layout/README.md
```diff
@@ -1,3 +1,20 @@
...
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-react/layout
+```
+
+For more detailed information on this component, please refer to the
+[storybook documentation](https://workday.github.io/canvas-kit/?path=/docs/labs-layout)
```
NicholasBoll, line 16: Is this path changing? Will it be `/docs/layout` instead?

modules/react/layout/stories/Box.stories.mdx
```diff
@@ -1,5 +1,5 @@
 import {Meta} from '@storybook/addon-docs';
-import {Box} from '@workday/canvas-kit-labs-react/common';
+import {Box} from '../';
```
NicholasBoll, line 2: I'm unsure why some imports were changed to be absolute while others were changed to be relative. Seems arbitrary, but this file will be imported by the doc site, so there can be no relative paths to packages. Only examples can be relative.

```suggestion
import {Box} from '@workday/canvas-kit-labs-react/layout';
``

mannycarrera4: I have to clean this up, this was before i found the reason why the package path was giving me an error because i had circular dependencies

NicholasBoll: I updated the suggestion. I noticed the path was pointing to labs.

NicholasBoll (approved):

PR #1478, opened 2022-02-17
feat(layout): Add cx prop, pseudo props, and other style props to Box

alanbsmith:
## Summary

category

---

Things still to be done:

- [ ] Clean up code
- [ ] Improve / clean up the types
  - [ ] Probably update csstype to v3?
- [ ] Add JSDoc comments
- [ ] Fix and add tests
- [ ] Add storybook docs 
  - [ ] Move `modules/labs-react/common/stories/examples/PropTables.splitprops.tsx` into `react/common`

## Checklist


modules/react/layout/index.ts
```diff
@@ -12,5 +12,5 @@ export * from './lib/utils/position';
 export {SpaceStyleProps} from './lib/utils/space';
 export * from './lib/Flex';
 export * from './lib/Stack';
-export type { FlexStyleProps } from './lib/utils/flex';
-export type { StackSpacing, StackDirection, StackStyleProps } from './lib/utils/stack';
+export type {FlexStyleProps} from './lib/utils/flex';
+export type {StackSpacing, StackDirection, StackStyleProps} from './lib/utils/stack';
```
alanbsmith, line 16: looks like my linter went after this file.

modules/react/layout/lib/utils/cx.ts
```diff
@@ -0,0 +1,89 @@
...
+  FlexItemStyleProps &
+  LayoutStyleProps &
+  OtherStyleProps &
+  PositionStyleProps &
+  SpaceStyleProps &
+  PseudoStyleProps & {
+    [key: string]: string | number;
+  };
+
+export function cxFn<S extends CXStyleProps>(styles: S, isRTL: boolean) {
```
alanbsmith, line 25: This currently does not support nested style objects.

modules/react/layout/lib/Box.tsx
```diff
@@ -1,7 +1,10 @@
+/** @jsxRuntime classic */
+/** @jsx jsx */
+import {jsx} from '@emotion/react';
```
alanbsmith, line 3: I'm not sure this is the way we want to go about this or not. I'm open to suggestions here. We could also try to make this a style function similar to the other style props.

mannycarrera4:
i know in our yarn.lock file we have csstype3 but in our react/package.json we have it to 2.x, we can upgrade this since i think emotion uses csstype3

modules/react/layout/lib/Box.tsx
mannycarrera4: I think when we were talking about this this is what would work, have the pragma in one file where we use the css prop and that's it, im ok with this!

alanbsmith:
Going to close this until I have some real time to dedicate to it.

Issue #924, opened 2020-12-07
Allow Labs Select to accept a prop for a custom render method for the selected option


jordandgold:
## 🚀 Feature Proposal

As it stands, the Labs Select component has a prop to pass a custom render method for each option in the select. Currently, this only works to change the options in the dropdown, but not the selected option that shows when the component is closed. This issue a request to add functionality to allow a custom render method for the selected option just like the options in the dropdown.

The current render prop (`renderOption`) used to render options in the dropdown is demonstrated [here](https://workday.github.io/canvas-kit/?path=/story/labs-select-react-top-label--default-with-custom-options-data) in Storybook. This proposal would extend this custom render functionality to the Select Button so the display of the selected option in the Button (when the dropdown menu is closed) could be customized as well.

## Motivation

This would support the designs for a feature I am working on.

giulialubet:
I'd like to work on this!

willklein:
Released in [v7.0.0](https://github.com/Workday/canvas-kit/releases/tag/v7.0.0)

PR #1512, opened 2022-03-18
feat(select): Accept custom render method prop for selected option

giulialubet:
## Summary

Resolves: #924 

These changes add a new prop, `renderSelected`, to the select component to custom render the selected option. This is similar to the existing `renderOptions` prop that provides a custom render method for the select options.

category

---

## Checklist

## Additional References

Selected option with icon due to custom render method


modules/preview-react/select/lib/SelectBase.tsx
```diff
@@ -49,6 +49,10 @@ export interface NormalizedOption extends Option {
   label: string;
 }
 
+export interface RenderSelectedOptionFunction {
+  (option: NormalizedOption): React.ReactNode;
```
giulialubet, line 53: I opted to use `NormalizedOption` instead of `RenderableOption` because the selected property is redundant and I thought the focused property could be confusing since it'll apply to an option that isn't rendered in the menu list

modules/preview-react/select/lib/SelectBase.tsx
```diff
@@ -261,7 +281,11 @@ const SelectWrapper = styled('div')<Pick<SelectBaseProps, 'grow' | 'disabled'>>(
 );
 
 const defaultRenderOption: RenderOptionFunction = option => {
-  return <div>{option.label}</div>;
+  return <div>{defaultRenderSelectedOption(option)}</div>;
+};
+
+const defaultRenderSelectedOption: RenderSelectedOptionFunction = option => {
+  return option.label;
```
giulialubet, line 288: I kept the default render methods separate to keep the underlying DOM unchanged. I am open to combining them into one default render method, but I was not sure if this is recommended

modules/preview-react/select/lib/SelectBase.tsx
```diff
@@ -261,7 +281,11 @@ const SelectWrapper = styled('div')<Pick<SelectBaseProps, 'grow' | 'disabled'>>(
 );
 
 const defaultRenderOption: RenderOptionFunction = option => {
-  return <div>{option.label}</div>;
+  return <div>{defaultRenderSelectedOption(option)}</div>;
+};
+
+const defaultRenderSelectedOption: RenderSelectedOptionFunction = option => {
```
jamesfan, line 287: Could shorten `defaultRenderSelectedOption` to `defaultRenderSelected` as well.

modules/preview-react/select/stories/stories.tsx
```diff
@@ -139,6 +149,13 @@ storiesOf('Preview/Select/React/Top Label', module)
     <FormField label="Label" inputId="select-grow" grow={true}>
       {controlComponent(<Select name="contact" options={options} />)}
     </FormField>
+  ))
+  .add('Default with Custom Render Selected Method', () => (
+    <FormField label="Label" inputId="select-default-custom-selected">
+      {controlComponent(
+        <Select name="icon" options={customOptions} renderSelected={customRenderSelectedOption} />
+      )}
+    </FormField>
```
jamesfan, line 158: It seems likely that folks would use `renderSelected` and `renderOption` together. Instead of creating this new "Default with Custom Render Selected Method" story, what do you think about showing the `renderSelected` prop using the existing ["Default with Custom Options Data"](https://github.com/giulialubet/canvas-kit/blob/custom-render-selected/modules/preview-react/select/stories/stories.tsx#L109) story? The [`<Select>`](https://github.com/giulialubet/canvas-kit/blob/custom-render-selected/modules/preview-react/select/stories/stories.tsx#L112) in "Default with Custom Options Data" would receive the `renderSelected` prop in addition to the existing `renderOption` prop. The example would then show an icon for the selected option in addition to the icons already shown for the menu options.

The `customRenderSelectedOption` function would remain unchanged (except perhaps renaming it to `customRenderSelected` to be consistent with the rest of my feedback above).

modules/preview-react/select/stories/stories.tsx
```diff
@@ -216,4 +233,15 @@ storiesOf('Preview/Select/React/Left Label', module)
...
+  .add('Default with Custom Render Selected Method', () => (
+    <FormField
+      labelPosition={FormField.LabelPosition.Left}
+      label="Label"
+      inputId="select-default-custom-selected"
+    >
+      {controlComponent(
+        <Select name="icon" options={customOptions} renderSelected={customRenderSelectedOption} />
+      )}
+    </FormField>
```
jamesfan, line 246: Same feedback as above with the Top Label story (consolidate story into Left Label > "Default with Custom Options Data").

jamesfan (commented):
Thanks for the contribution @giulialubet! This pretty much looks good to go, just left a few nits below.

modules/preview-react/select/lib/SelectBase.tsx
jamesfan: Makes sense 👍

modules/preview-react/select/lib/SelectBase.tsx
jamesfan: I can't recall why I wrapped the option label in a `<div>` in `defaultRenderOption` 🤔, but it does make sense to preserve the existing DOM structure.

jamesfan (approved):

PR #1521, opened 2022-03-30
fix(avatar): Resetting imageLoaded when new url is provided

BillGeoghegan:
## Summary

Resolves: resets the imageLoaded to false when a new image url is provided within the Avatar component. 

Motivation: This would allow us to swap back to the default image when a url value is no longer passed. 

category

---

## Checklist


NicholasBoll (approved):

modules/react/avatar/lib/Avatar.tsx
```diff
@@ -139,6 +139,10 @@ const Avatar: AvatarOverload = React.forwardRef(
       }
     };
 
+    React.useEffect(() => {
```
NicholasBoll, line 142: This might cause rendering issues in the UI visual regression tests. I'll switch to a `React.useLayoutEffect` to make sure there isn't a frame with an incorrect image. This is technically able to change the `imageLoaded` to `false` if the image is loaded and available _before_ the `onLoad` of the image is called. I'm seeing this only in IE11, but it is a race condition.

```suggestion
    React.useLayoutEffect(() => {
```

Issue #1491, opened 2022-02-26
Update the Tab Component to Support WebView


myvuuu:
## 🚀 Feature Proposal

Update Tabs to support mobile behavior/design for WebView.

## Motivation

Allow teams to create WebView experiences using Canvas Kit to quickly test out the feasibility and need for generating fully native experiences.

## Requirements

- [x] The "More" button and web overflow behavior (opens a menu with other tab options when the user selects "More") should be removed and replaced with scrollable tabs when the full set of tabs cannot fit on a single screen 
- [x] When overflow tabs are needed, the screen reader should announce the first fully visible tab (not faded) within the current screen 
- [x] The tab text label is used to define the tab width

Overall Requirements

- [x] Tab names do not truncate, regardless of the length of the tab label
- [x] Tabs support the use of icons
- [x] Developers are able to opt-in/out of the mobile specific behavior
- [x] Modality is determined through theming within the Canvas Provider

myvuuu:
Working out the following:
- Whether or not we need to make all the tabs equal size (for fixed tab)
- Min-width for a tab

myvuuu:
<img width="537" alt="WebView Overflow Tabs" src="https://user-images.githubusercontent.com/84945599/157139461-fc36dd38-ad1a-408a-bdab-b4bd9cd552eb.png">
<img width="541" alt="WebView Fixed Tabs" src="https://user-images.githubusercontent.com/84945599/157139465-9480d9cd-dc67-4441-a877-02710912634a.png">

PR #1523, opened 2022-03-31
feat: Add useModalityType hook and add modality support to tabs

NicholasBoll:
## Summary

Fixes: #1491

Adds a `useModalityType` hook that will return the current modality returned by [PointerEvent.pointerType](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType)

Tabs is updated to use this modality detection to change the overflow behavior of tabs.

category

---

## Additional References

To test, go to the Overflow tabs example and use the device emulator in the browser's dev tools. Once you do that, your mouse will be a simulated touch device. Clicking the mouse will change the modality to "touch" and cause the tabs to render the tab list in a horizontal overflow container.

https://user-images.githubusercontent.com/338257/161164634-b43b2d04-97cd-4c2b-95f1-bd6efa4684d4.mov


modules/react/tabs/lib/TabsItem.tsx
```diff
@@ -215,7 +216,7 @@ export const TabsItem = createComponent('button')({
 
     return (
       <OverflowTooltip>
-        <StyledTabItem as={Element} {...props}>
+        <StyledTabItem as={Element} maxWidth={modality === 'touch' ? undefined : 280} {...props}>
```
mannycarrera4, line 219: should the maxWidth be configurable?

mannycarrera4 (approved):
just a minor comment, but this is awesome!

modules/react/tabs/lib/TabsItem.tsx
NicholasBoll: `maxWidth` can be overridden by passing `maxWidth` to the `Tabs.Item` component:

```tsx
<Tabs>
  <Tabs.List>
     <Tabs.Item maxWidth={300}>First Tab</Tabs.Item>
```

We don't have a system for defining properties based on screen size, component size, or modality. I would not worry about it without a system in place.

Issue #1404, opened 2022-01-03
Remove Popper's containerElement prop


alanbsmith:
## 💡 Suggestion

Remove Popper's [`containerElement` prop](https://github.com/Workday/canvas-kit/blob/fa0f4d9226ada17cbd9f293eb914313d94d966d8/modules/react/popup/lib/Popper.tsx#L31).

## Motivation

We no longer need this with Fullscreen API support introduced in #1403.

## Acceptance Criteria

- [ ] Remove the `containerElement` prop from `Popper`
- [ ] Update documentation and examples
- [ ] Update V7 migration guide

jaclynjessup:
Closing was marked done Apr 27 22 in sprint

PR #1524, opened 2022-04-01
fix(popup): Remove containerElement prop from Popper

RayRedGoose:
## Summary

Fixes: #1404  

category

### Release Note
The `containerElement` prop has been removed from Popper component because It's no longer needed as Fullscreen API support implemented.


modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -154,6 +155,11 @@ If your code contains any hacks to make a `Modal` overflow, these hacks should n
 `max-height` of the `Modal.Body` element using calculations. These should be removed. The
 `Popup.Card` now has a max height and the `Popup.Body` height is automatically calculated.
 
+## Popper Props Update
+
+We removed the `containerElement` prop from Popper component because it's no longer needed with
+Fullscreen API.
```
mannycarrera4, line 161: do we have a codemod for this?

RayRedGoose: I've checked usage of Popper, so it will be good to add codemod for making migration easier. Thanks for catching it, I'll add it

modules/codemod/lib/v7/removePropFromPopper.ts
```diff
@@ -0,0 +1,41 @@
...
+          item.name.type === 'JSXIdentifier' &&
+          ['containerElement'].includes(item.name.name)
+            ? false
+            : true
+        );
+      }
+    });
+
+  return root.toSource();
+}
```
mannycarrera4, line 41: thanks for adding this!

mannycarrera4 (approved):

PR #1539, opened 2022-04-12
fix(button): Updated stories to match spec

josh-bagwell:
## Summary

Fixes: Updated buttons with relatedActionsVerticalIcon to match design spec

category


mannycarrera4 (approved):
nice work!! 🎉

Issue #1496, opened 2022-03-08
Codemod wrongly migrates Skeleton.Header to Skeleton.DeprecatedHeader


jaslloyd:
## 🐛 Bug Report

Codemod wrongly migrates Skeleton.Header to Skeleton.DeprecatedHeader

## To Reproduce

1. Have some code you need to migrate that includes `Skeleton.Header`
2. Run codemod script `npx @workday/canvas-kit-codemod v6 src`
3. Skeleton.Header is now changed Skeleton.DeprecatedHeader

## Expected Behavior

To not change Skeleton.Header to Skeleton.DeprecatedHeader, Skeleton.Header is right component

## Actual Results

Skeleton.Header gets migrated to Skeleton.DeprecatedHeader

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

## Link to repl or repo (highly encouraged)

Please provide a reference to a repo or branch to help us reproduce this bug.

### Error Output

Paste the error output here

```bash

```

or

Include a screenshot of a visual bug

willklein:
We also have an issue renaming any instance of `Header` to `DeprecatedHeader` even if it's not a Canvas Kit import, and this is a common component name.

PR #1546, opened 2022-04-20
fix(codemod): Limit specifier renaming to Canvas Kit imports

willklein:
## Summary

Fixes: #1496 

The current code mods will rename components that aren't imported from Canvas Kit. They also rename sub-components like `Skeleton.Header`. This change addresses both of these issues.

This change implements the following:
- [x] adds test cases to guard against undesired component renaming
- [x] identifies Canvas Kit component imports to track which components should be renamed
- [x] only renames identified components that have been imported from Canvas Kit
- [x] only renames identified components where they are not aliased to a different name
- [x] amends the `renameImports` utility to allow for specialized code mods to pass in its identified component imports
- [x] removes the `renameImportSpecifiers` utility since it was only used by one file

category

### Release Note
Fixed code mods that improperly rename non-Canvas Kit components as well as Canvas Kit sub-components. This typically includes non-Canvas Kit `Header` components and Canvas Kit's `Skeleton.Header` being improperly renamed in the JSX, but solves for all future cases of Canvas Kit component names.

---

## Checklist

## Additional References

Past behavior:
```jsx
        import { Header as CanvasHeader } from '@workday/canvas-kit-labs-react/header';
        import { Header } from './localProjectHeader';
        import { Skeleton } from '@workday/canvas-kit-labs-react/skeleton';

        const SomeComponent = () => {
          return <>
            <Header />
            <CanvasHeader />
            
            <Skeleton>
              <Skeleton.Header />
            </Skeleton>
          </>;
        };
```
would get code shifted to be:

```jsx
        import { DeprecatedHeader as CanvasHeader } from '@workday/canvas-kit-labs-react/header';
        import { Header } from './localProjectHeader';
        import { Skeleton } from '@workday/canvas-kit-labs-react/skeleton';

        const SomeComponent = () => {
          return <>
            <DeprecatedHeader />
            <CanvasHeader />
            
            <Skeleton>
              <Skeleton.DeprecatedHeader />
            </Skeleton>
          </>;
        };
```
...where `Header` and `Skeleton.Header` are incorrectly renamed. This problem potentially extends to any of our component renames, but `Header` is the most prevalent one that teams are using from local or external sources. This PR fixes all potential cases by building a Record of import identifiers where the import name is used (not aliased to a local name) and only in these cases updating the `Identifier` or `MemberExpression` in the rest of the file's AST.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -13,7 +13,7 @@ module.exports = {
 };
 
 // Extra scopes supported outside the `modules` folder
-const scopes = ['labs', 'preview', 'fonts', 'popup-stack'];
+const scopes = ['labs', 'preview', 'fonts', 'popup-stack', 'codemod'];
 
 /**
  * @param {{ cwd: string}} context
--- a/modules/codemod/lib/v6/moveAndRenameSearchBar.ts
+++ b/modules/codemod/lib/v6/moveAndRenameSearchBar.ts
@@ -96,6 +96,7 @@ const transform: Transform = (file, api) => {
     if (searchBarSpecifiers.length) {
       searchBarSpecifiers.forEach(specifier => {
         const specifierName = specifier.imported.name;
+
         if (specifierName in renameMap) {
           // if it hasn't been aliased, track it for updating JSX
           if (!specifier.local || specifier.local.name === specifier.imported.name) {
--- a/modules/codemod/lib/v6/moveAndRenameSearchBar.ts
+++ b/modules/codemod/lib/v6/moveAndRenameSearchBar.ts
@@ -1,10 +1,10 @@
 import {ASTPath, Transform, ImportDeclaration, ImportSpecifier} from 'jscodeshift';
 import {
   filterImportSpecifiers,
-  renameImportSpecifiers,
   ImportSpecifierArray,
   renameImports,
   hasImportSpecifiers,
+  RenameMap,
 } from './utils';
 
 const mainPackage = '@workday/canvas-kit-labs-react';
@@ -17,7 +17,7 @@ const searchBarImportSpecifiers = [
   'SearchThemes',
 ];
 
-const renameMap = {
+const renameMap: RenameMap = {
   SearchBar: 'SearchForm',
   SearchBarProps: 'SearchFormProps',
   SearchBarState: 'SearchFormState',
@@ -42,6 +42,8 @@ const transform: Transform = (file, api) => {
     nodePath.insertBefore(j.importDeclaration(specifiers, j.stringLiteral(sourcePath)));
   }
 
+  const discoveredImportAliases: Record<string, boolean> = {};
+
   // Rename search-bar named imports from @workday/canvas-kit-labs-react
   // e.g. import { SearchBar, SearchBarProps } from '@workday/canvas-kit-labs-react';
   // becomes import { SearchForm, SearchFormProps } from '@workday/canvas-kit-labs-react';
@@ -49,8 +51,16 @@ const transform: Transform = (file, api) => {
     const importSpecifiers = nodePath.value.specifiers || [];
     // Filter search bar imports
     const searchBarImports = filterImportSpecifiers(importSpecifiers);
+
     // Rename search bar imports
-    renameImportSpecifiers(searchBarImports, renameMap);
+    searchBarImports.forEach(specifier => {
+      const specifierName = specifier.imported.name;
+      if (specifierName in renameMap) {
+        discoveredImportAliases[specifier.imported.name] = true;
+
+        specifier.imported.name = renameMap[specifierName];
+      }
+    });
   });
 
   // Handle imports from @workday/canvas-kit-react/header
@@ -81,10 +91,18 @@ const transform: Transform = (file, api) => {
 
     // If SearchBar import specifiers were found, insert a new import declaration above the original import declaration
     if (searchBarSpecifiers.length) {
-      const renamedSearchBarSpecifiers = renameImportSpecifiers(searchBarSpecifiers, renameMap);
+      searchBarSpecifiers.forEach(specifier => {
+        const specifierName = specifier.imported.name;
+        if (specifierName in renameMap) {
+          discoveredImportAliases[specifier.imported.name] = true;
+
+          specifier.imported.name = renameMap[specifierName];
+        }
+      });
+
       insertImportDeclarationBefore(
         nodePath,
-        renamedSearchBarSpecifiers,
+        searchBarSpecifiers,
         '@workday/canvas-kit-labs-react/search-form'
       );
     }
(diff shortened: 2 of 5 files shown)


commitlint.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
 };
 
 // Extra scopes supported outside the `modules` folder
-const scopes = ['labs', 'preview', 'fonts', 'popup-stack'];
+const scopes = ['labs', 'preview', 'fonts', 'popup-stack', 'codemod'];
```
willklein, line 16: So we can commit with a message like `fix(codemod): Message here`.

modules/codemod/lib/v6/moveAndRenameSearchBar.ts
```diff
@@ -42,15 +42,25 @@ const transform: Transform = (file, api) => {
     nodePath.insertBefore(j.importDeclaration(specifiers, j.stringLiteral(sourcePath)));
   }
 
+  const discoveredImportAliases: Record<string, boolean> = {};
```
willklein, line 45: I'm going to rename this. I want to add another object map to keep track of aliases we don't rename cases like this:
```
import { Header as CanvasHeader } from '@workday/canvas-kit-labs-react/header';

const SomeComponent = () => {
  return <>
    <Header />
    <CanvasHeader />
  </>;
};
 ```

File: modules/codemod/lib/v6/moveAndRenameSearchBar.ts [hunk 3]
```diff
@@ -42,15 +42,25 @@ const transform: Transform = (file, api) => {
...
   // Rename search-bar named imports from @workday/canvas-kit-labs-react
   // e.g. import { SearchBar, SearchBarProps } from '@workday/canvas-kit-labs-react';
   // becomes import { SearchForm, SearchFormProps } from '@workday/canvas-kit-labs-react';
   root.find(j.ImportDeclaration, {source: {value: mainPackage}}).forEach(nodePath => {
     const importSpecifiers = nodePath.value.specifiers || [];
     // Filter search bar imports
     const searchBarImports = filterImportSpecifiers(importSpecifiers);
+
     // Rename search bar imports
-    renameImportSpecifiers(searchBarImports, renameMap);
```
willklein, line 53: I dropped this utility which was used in two places, only in this file. This means a little bit of duplication. I didn't keep it as a local function because of the `discoveredImportAliases` map.

modules/codemod/lib/v6/spec/deprecateHeader.spec.ts
```diff
@@ -8,7 +8,42 @@ describe('Canvas Kit Deprecate Header Codemod', () => {
   context('when transforming header imports', () => {
     it('should ignore non-canvas-kit imports', () => {
       const input = `import Header, {GlobalHeader} from "@workday/some-other-library";`;
-      const expected = `import Header, {GlobalHeader} from "@workday/some-other-library";`;
+      const expected = `${input}`;
```
willklein, line 11: I started doing this for test cases where nothing changed. I hope it makes it more obvious that no change is expected. Otherwise you might scan the `input` and `expected` and not tell right away if there is a difference.

modules/codemod/lib/v6/utils/index.ts
```diff
@@ -179,7 +159,8 @@ export function renameImports(
   api: API,
   root: Collection<any>,
   packageName: string,
-  specifierMap: Record<string, string>
+  specifierMap: Record<string, string>,
+  discoveredImportAliases: Record<string, boolean> = {}
```
willklein, line 163: Adding this not only fixes the `Header` import problem, but means we address a lot of possible overlap with non-Canvas Kit components that share the same name.

modules/codemod/lib/v6/moveAndRenameSearchBar.ts
```diff
@@ -42,15 +42,28 @@ const transform: Transform = (file, api) => {
...
+    searchBarImports.forEach(specifier => {
+      const specifierName = specifier.imported.name;
+      if (specifierName in renameMap) {
+        // if it hasn't been aliased, track it for updating JSX
+        if (!specifier.local || specifier.local.name === specifier.imported.name) {
```
willklein, line 60: This is a new check I've added here and in the `renameImports` utility. It will guard against renaming non-Canvas Kit components that have the same name, by checking to see if it was aliased to a local name. If it has been renamed, we don't need to track it for updating identifiers and member expressions.

modules/codemod/lib/v6/spec/deprecateHeader.spec.ts
```diff
@@ -8,7 +8,99 @@ describe('Canvas Kit Deprecate Header Codemod', () => {
...
+
+      expectTransform(input, expected);
+    });
+
+    it('should ignore non-canvas-kit header identifiers', () => {
```
willklein, line 16: Please review the new test cases to see if I covered anything we can think of. `Header` is the most common component name that conflicts with non-Canvas Kit components. We could have more general spec files to test the utilities, but I hope this is sufficient for now.

modules/codemod/lib/v6/spec/deprecateHeader.spec.ts
```diff
@@ -8,7 +8,99 @@ describe('Canvas Kit Deprecate Header Codemod', () => {
...
+            <CanvasHeader />
+          </>;
+        };
+      `;
+      const expected = `
+        import { DeprecatedHeader as CanvasHeader } from '@workday/canvas-kit-labs-react/header';
+
+        const SomeComponent = () => {
+          return <>
+            <Header />
```
willklein, line 74: This used to get renamed even if it was not a Canvas Kit header.

modules/codemod/lib/v6/spec/moveAndRenameSearchBar.spec.ts
```diff
@@ -78,18 +78,73 @@ describe('Canvas Kit Move and Rename Search Bar Codemod', () => {
...
+        }
+      `;
+      const expected = `
+        import { SearchForm as CanvasSearchBar } from "@workday/canvas-kit-labs-react";
+        import { SearchBar } from "@other/search-bar";
+
+        const CustomSearch = () => {
+          return <>
+            <CanvasSearchBar />
+            <SearchBar />
```
willklein, line 101: This is a less likely case than Header, but it's possible we have a similarly named non-Canvas Kit SearchBar. I added support for this just in case.

modules/codemod/lib/v6/utils/index.ts
```diff
@@ -195,26 +176,45 @@ export function renameImports(
...
+            discoveredImportSpecifiers[specifier.imported.name] = true;
+          }
+
           specifier.imported.name = specifierMap[specifier.imported.name];
         }
       }
     });
   });
 
-  // Transforms default imports from package
```
willklein, line 208: This actually confused me and needed to move a few lines down.

NicholasBoll:
I removed the `Breaking Changes` section. Adding the section (even if a `N/A` is the only content) causes the PR lint to think there's breaking changes and informs you to target `prerelease/major` instead.

NicholasBoll (approved):

PR #1547, opened 2022-04-21
feat: Add collection system

### LINKED ISSUE #1318: Multi-dimensional and hierarchical list
## Summary

Update Lists to support hierarchy and multiple dimensions.

## Motivation

Lists support a single dimension list and navigating and selecting within a list. A hierarchical dropdown menu will need an added notion of hierarchy and components like a color picker will need the concept of more than a single dimension.

NicholasBoll:
## Summary

Fixes #1463 
Fixes #1318 

This PR adds a collection system to Canvas Kit! A collection system adds models and elemProps hooks to help create components like tables, menus, selects, etc. The collection system supports selection, navigation, and registration.

category

### Release Note
We've added a collection system to help our users create custom components like 

### BREAKING CHANGES
- Upgrade Typescript to 4.1 and drop support for Typescript <4
- `Popup.Body` is an overflow container. If your popup buttons are inside the `Popup.Body` element (our examples did this), you will need to move those buttons outside the `Popup.Body` element, otherwise the focus rings may be cut off. See migration guide for more information.
- Changed the signature of model callbacks and guards. This is handled by our codemod
  - Before:
    - `shouldUpdate({data, state})`
    - `onUpdate({data, prevState})`
  - After:
    - `shouldUpdate(data, state)`
    - `onUpdate(data, prevState)`

---

## Checklist


modules/preview-react/breadcrumbs/lib/Breadcrumbs/Dropdown/Button.tsx
```diff
@@ -1,15 +1,15 @@
-import React from 'react';
-import {IconButton, IconButtonProps} from '@workday/canvas-kit-react/button';
-import {relatedActionsIcon} from '@workday/canvas-system-icons-web';
-import {CanvasSystemIcon} from '@workday/design-assets-types';
-import {colors} from '@workday/canvas-kit-react/tokens';
-import {createComponent} from '@workday/canvas-kit-react/common';
+import React from "react";
```
willklein: Do we need to lock in some Prettier settings?

modules/docs/mdx/changelog.stories.mdx
```diff
@@ -3,4 +3,5 @@ import Changelog from '../../../CHANGELOG.md';
 import Markdown from '../../../utils/storybook/Markdown.tsx';
 
 <Meta title="Welcome/Changelog" />
+
```
NicholasBoll, line 6: The changelog wasn't working before... This is annoying. Maybe a prettier update can fix this

modules/codemod/lib/v7/index.ts
```diff
@@ -1,7 +1,25 @@
...
+    recategorizeIconButtons,
+    removePropFromPopper,
+    renameIconPosition,
+    renameLayoutImports,
+    updateSegmentedControl,
+  ];
+  return fixes.reduce((source, fix) => fix({...file, source}, api, options) as string, file.source);
+};
+
+export default transform;
```
NicholasBoll, line 25: I have no idea why none of these transforms were hooked up yet

NicholasBoll: https://www.youtube.com/watch?v=0GypdsJulKE

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -538,6 +539,242 @@ main `react` package.
 - `Flex` was promoted from `@workday/canvas-kit-labs/layout` to `@workday/canvas-kit-react/layout`
 - `Stack` was promoted from `@workday/canvas-kit-labs/layout` to `@workday/canvas-kit-react/layout`
 
+## Model Changes
```
willklein, line 542: If you're reviewing this, I recommend viewing the Markdown file here: https://github.com/NicholasBoll/canvas-kit/blob/feat/add-grid-support-temp/modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx#model-changes

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -538,6 +539,242 @@ main `react` package.
...
+
+### Guards and Callbacks
+
+We changed the signature of model event guards and callbacks. In v6, the parameters were in an
+object. This has a less than ideal developer experience as intellisense isn't engaged immediately
```
willklein, line 547: When you say "intellisense isn't engaged immediately" are we describing that IntelliSense slower with an object parameter?

I will personally complain that with the object, it was one extra level of destructuring to get what I wanted. I agree with making prevState the second parameter too. This makes sense.

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -538,6 +539,242 @@ main `react` package.
...
+
+### Model Implementation
+
+If you don't extend models, you can skip this section.
+
+In v6 we supported Typescript 3.8 which limited the way types could be inferred and defined. Canvas
+Kit v7 now requires Typescript 4.1 which introduced
+[Template Literal Types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html).
+This means that the guards and callbacks no longer need to be manually defined via an `eventMap`.
+Event maps were a stopgap and were manual and suffered from user-error in defining them. Not all
```
willklein, line 581: Suggestions are not an option for some reason...
`Event maps were a stopgap and were manual and suffered from user-error in defining them.`
could be:
`Event maps were a stopgap, were manual, and suffered from user-error in defining them.`

willklein: I had some thoughts about the language choices (`were manual`, `suffered from user-error`) but I think it's fine to let your personal tone shine through!

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -538,6 +539,242 @@ main `react` package.
...
+events had guards and callbacks there's no check other than manual review to ensure the existence of
```
willklein, line 582: `events had guards and callbacks`
could be:
`events had guards or callbacks`

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -538,6 +539,242 @@ main `react` package.
...
+an event callback or guard. Template Literal Types allow us to properly type guards and callbacks
+without event maps. The previous types forced `data` to be an object. We've since dropped that
+restriction. You can still only pass a single argument to events. If you need additional
```
willklein, line 585: > You can still only pass a single argument to events. If you need additional
information, use an object.

I am actually a little confused reading this. I wonder if some of this should move past the next example so the code changes are presented before reading this. This doesn't refer to code in the previous example, does it?

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -538,6 +539,242 @@ main `react` package.
...
+information, use an object.
+
+This change also allowed us to remove a lot of boilerplate associated with models without
+sacrificing type safety. This change doesn't effect most use-cases, but will effect those whole
```
willklein, line 589: small thing, I don't think `use-cases` needs a hyphen

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
NicholasBoll: Yes. It is an additional level. The parameter given to callbacks and guards was called `event`, and the properties were `data`, and `state` for guards  and `prevState` for callbacks. This often lead to odd paths for events like `show` and `hide` where the parameter for the DOM event was also called `event`: `event.data.event`.

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
NicholasBoll: "prone to errors" is the same idea without sounding like we're pointing fingers

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
NicholasBoll: All events can only take a single parameter because of the signature of guards and callbacks:

```ts
// event signature
type updateValue = (value: string) => void

// guard signature - note the first parameter always gets named `data`
type shouldUpdateValue = (data: string, state: State) => boolean

// callback signature
type onUpdateValue = (data: string, prevState: State) => void
```

The first parameter of the event signature becomes the first parameter of a guard and callback. If we allowed more than a single argument, we'd run into issues:

```ts
// event signature
type updateValue = (value: string, delay = false) => void

// guard - only takes first parameter
type shouldUpdateValue = (data: string, state: State) => boolean

// callback signature - only takes first parameter
type onUpdateValue = (data: string, prevState: State) => void
```

The previous example would have to look like this:
```ts
// event signature
type updateValue = (event: { value: string, delay: boolean }) => void

// guard - only takes first parameter
type shouldUpdateValue = (data: { value: string, delay: boolean }, state: State }) => boolean

// callback signature - only takes first parameter
type onUpdateValue = (data: { value: string, delay: boolean }, prevState: State) => void
```

modules/codemod/lib/v7/index.ts
willklein: I have a pretty good idea why. I'll think about how we can prevent it from slipping again.

modules/codemod/lib/v7/updateModelSignatures.ts
```diff
@@ -0,0 +1,112 @@
...
+                }
+              });
+            }
+          });
+        }
+      });
+    }
+  });
+
+  return root.toSource();
```
mannycarrera4, line 111: don't forget to add this to the `index` file :)

NicholasBoll: Done

modules/codemod/lib/v7/index.ts
NicholasBoll: One way is to update all codemods to import from this index file instead of their individual transform file. Then the spec won't work until everything is hooked up correctly.

I think we did it this way in v5 because we had cascading transforms that we wanted to isolate, but that might be a mistake. It makes it too easy to have all tests pass, but none of the codemods actually work.

modules/react/collection/stories/Collection.stories.mdx
```diff
@@ -0,0 +1,130 @@
...
+well-supported way to accomplish accessibility requirements for focusing items within a list. This
+example shows how to use `useListRovingFocus`. You can either use the tab key for focus on an item
+or click on an item and then use the up/down keys to navigation the list. By default, the list is
+set to wrap navigation using the `wrappingNavigationManager`. Only a single item in the list is a
+focus stop that "roves" as the up/down arrows are pressed.
+
+**Note:** This example doesn't meet accessibility requirements. The list will have to have some type
+of context. Like "navigation list" or "menu list".
+
+<!--ExampleCodeBlock code={RovingFocus} /-->
```
mannycarrera4, line 114: did you still want this commented out?

modules/react/collection/stories/examples/MultiSelection.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <ul>{useListRenderItems(model, props.children)}</ul>
+      <p>Cursor ID: {model.state.cursorId}</p>
+      <p>
+        Selected ID: {(model.state.selectedIds !== 'all' ? model.state.selectedIds : []).join(',')}
+      </p>
+    </ListModelContext.Provider>
+  );
+};
+
+const useItem = composeHooks(useListItemSelect, useListItemRovingFocus, useListItemRegister);
```
mannycarrera4, line 33: the order of these matters right? is it `useListItemRegister` -> `useListItemRovingFocus` -> `useListItemSelect` where you need to register your items first and your select is last?

modules/react/menu/lib/Menu.tsx
```diff
@@ -40,13 +33,9 @@ export interface MenuProps<T> extends MenuModelConfig<T> {
  *   </Menu.Popper>
  * </Menu>
  */
-export const Menu = createComponent()({
+export const Menu = createContainer()({
```
mannycarrera4, line 36: do we need to give it a default element here?

modules/react/menu/lib/MenuCard.tsx
```diff
@@ -0,0 +1,69 @@
...
+);
+
+export const useMenuCard = createElemPropsHook(useMenuModel)(() => {
+  return {};
+});
+
+export const MenuCard = createSubcomponent('div')({
+  displayName: 'Menu.Card',
+  modelHook: useMenuModel,
+  elemPropsHook: useMenuCard,
```
mannycarrera4, line 52: I'm a little confused, can you explain why we need elemPropsHook here? It looks like it's returning an empty object?

mannycarrera4 (approved):
Overall this is huge and super excited to get comfortable with using this, especially with color picker.

A couple notes/ questions?
- when you use `createContainer` do you need to give it a `elemPropsHook` i think I'm a little confused by this. I think it just provides the underlying element the necessary attributes for accessibility is that correct?

- Do we have anywhere written when to use `createComponent` vs `createContainer` vs `createHook` having some sort of guidance would be great, could be something to add after this gets merged

- Do you have to use `createSubComponent` if the parent was created with `createContainer` ? Is this to avoid adding the `model` prop to each sub component?

Just a few questions but awesome work on this!!

modules/react/collection/stories/Collection.stories.mdx
NicholasBoll: I didn't revisit this after introducing a `ListBox` which takes care of most of this. Now I'm a bit torn on how to document the collection system...

Most cases are covered by `ListBox`, but not all. Originally I was showcasing how all the low-level models and hooks could be used together to create higher-level components. Now I've introduced a few abstractions that cover the common cases I'm outlining and `ListBox` should be used when possible. So... Do I stick with low-level or use the highest abstraction possible?

modules/react/collection/stories/examples/DynamicItems.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+const items: Item[] = Array(1000)
+  .fill(true)
+  .map((_, index) => ({id: String(index + 1), text: `Item - ${index + 1}`}));
+
+export const DynamicItems = () => {
+  return (
+    <ListBox items={items} maxHeight={300}>
+      {(item: Item) => <ListBox.Item data-id={item.id}>{item.text}</ListBox.Item>}
```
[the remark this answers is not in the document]
NicholasBoll: We don't need to in this example. `ListBox.Item` already does what we need.

modules/react/menu/lib/Menu.tsx
NicholasBoll: `Menu` does not represent an element. All containers are responsible for creating a model context, but not all also represent an element. Tabs is similar - according to the specification, there's no wrapper container element needed for tabs to work. This means the container would not represent a semantic element. We've opted to decrease the amount of elements needed for components.

NicholasBoll:
> * when you use `createContainer` do you need to give it a `elemPropsHook` i think I'm a little confused by this. I think it just provides the underlying element the necessary attributes for accessibility is that correct?

`elemPropsHook` only makes sense if your container component represents an element. If there's no element, there's also no elemPropsHook.

> Do we have anywhere written when to use createComponent vs createContainer vs createHook having some sort of guidance would be great, could be something to add after this gets merged

Not yet

> Do you have to use createSubComponent if the parent was created with createContainer ? Is this to avoid adding the model prop to each sub component?

Not necessarily. I'll do a writeup about the transition.

modules/react/menu/lib/MenuCard.tsx
NicholasBoll: It is a placeholder in case `useMenuCard` needs to do something more in the future.

modules/react/collection/stories/examples/MultiSelection.tsx
NicholasBoll: This one is weird and I tried to change it when I created it.

It runs hooks from right to left, but merges props from left to right. So the rightmost runs first, while the leftmost props win.

For example, the lodash `compose` works the same way - `compose` runs functions from right to left.

modules/codemod/lib/v7/recategorizeIconButtons.ts
```diff
@@ -106,14 +107,21 @@ export default function transformer(file: FileInfo, api: API, options: Options)
...
   // const StyledIconButton = styled(CorrectButtonMapping)
   root.find(j.VariableDeclarator).forEach(nodePath => {
     if (
       nodePath.value.init?.type === 'CallExpression' &&
       nodePath.value.init.callee.type === 'CallExpression' &&
-      nodePath.value.init.callee.arguments[0].type === 'Identifier'
+      nodePath.value.init.callee.callee.type === 'Identifier' &&
+      nodePath.value.init.callee.callee.name === 'styled' &&
+      nodePath.value.init.callee.arguments[0].type === 'Identifier' &&
+      nodePath.value.init.callee.arguments[0].name === 'IconButton'
```
NicholasBoll, line 124: Before this change, the following would happen:

```diff
-const MyComponent = createComponent(Button)({
+const MyComponent = createComponent(TerinaryButton)({
```

modules/codemod/lib/v7/renameLayoutImports.ts
```diff
@@ -79,12 +79,14 @@ const transform: Transform = (file, api) => {
     });
   } else {
     // create new import
-    foundImport[0].insertBefore(
-      j.importDeclaration(
-        reactLayoutSpecifiers.map(mapToSpecifiers),
-        j.stringLiteral('@workday/canvas-kit-react/layout')
-      )
-    );
+    if (foundImport.length) {
```
NicholasBoll, line 82: The codemod assumed an import would always be found, but it wouldn't if the file under the codemod didn't import from `Layout` which caused the codemod to crash.

modules/react/collection/lib/react-virtual.ts
```diff
@@ -0,0 +1,71 @@
+// retype interfaces
+// delete when `react-virtual` accepts this PR: https://github.com/TanStack/react-virtual/pull/277
```
NicholasBoll, line 2: Can be removed when https://github.com/TanStack/react-virtual/pull/277 is merged

modules/react/collection/lib/useBaseListModel.tsx
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// interface Collection<T> {
+//   getKeys(): string[];
+//   getItem(id: string): Item<T> | null;
+//   at(index: number): Item<T> | null;
+//   size: number;
+// }
+
+// TODO: Build an interface that allows representation of different collections. For example an array and a dynamically loaded dataset
```
NicholasBoll, line 34: I'm still working on this, but we don't need to hold up the PR

modules/codemod/lib/v7/recategorizeIconButtons.ts
mannycarrera4: ah so you're adding more specificity?

modules/codemod/lib/v7/renameLayoutImports.ts
mannycarrera4: ahhhh I also have this PR update this codemod but we can merge this in first! https://github.com/Workday/canvas-kit/pull/1567

modules/codemod/lib/v7/recategorizeIconButtons.ts
NicholasBoll: Yes. It is making sure the function used is named `styled` and the argument is named `IconButton`. Perhaps it should use the importmap instead in case they rename `IconButton`

Issue #1540, opened 2022-04-13
Add JSXExpressionContainer support to V7 IconButton codemod


alanbsmith:
## 🐛 Bug Report

Currently the V7 IconButton codemod supports `Literal` values, but does not identify `JSXExpressionContainer` values.

## Expected Behavior

This code should be transformed:

```
// v6
<IconButton variant={"inverseFilled"} />

// v7
<SecondaryButton variant={"inverse"} />
// or this would be fine too
<SecondaryButton variant="inverse" />
```

## Actual Results

The code is not transformed.

PR #1558, opened 2022-04-27
fix(codemod): Add JSXExpressionContainer support to V7 IconButton codemod

josh-bagwell:
Worked on codemod updates with Raisa. Updated codemod for v7 support. 

Fixes: #1540 

category


modules/codemod/lib/v7/recategorizeIconButtons.ts
```diff
@@ -73,7 +74,19 @@ export default function transformer(file: FileInfo, api: API, options: Options)
 
       // Default IconButton variant is `circle`
       if (variantProp) {
-        const variantPropValue = ((variantProp as JSXAttribute).value as StringLiteral)?.value;
```
willklein, line 76: I know `variantProp as JSXAttribute` was present before, but with so many more usages I'm checking a way to remove the need to use `as` over and over. I'll let you know if I figure it out.

modules/codemod/lib/v7/recategorizeIconButtons.ts
```diff
@@ -73,7 +74,19 @@ export default function transformer(file: FileInfo, api: API, options: Options)
```
willklein, line 74: So it's not letting me suggest this properly, but we can type the above assignment to `variantProp` and simplify the code below. Line 72 to line 85 would be the following:
```
      const variantProp = attrs?.find(
        attr => attr.type === 'JSXAttribute' && attr.name.name === 'variant'
      ) as JSXAttribute | undefined;

      // Default IconButton variant is `circle`
      if (variantProp) {
        const valueBody = variantProp.value;
        if (valueBody?.type === 'JSXExpressionContainer') {
          variantProp.value = valueBody.expression as StringLiteral;
        }

        const variantPropValue = (variantProp.value as StringLiteral).value;

        (variantProp.value as StringLiteral).value = variantPropValue.replace(
```

Could you give that a try? I think we can do more to simplify the types but I want to start there.

josh-bagwell: Just replaced lines 72-85, and see no issues/errors. 

Spec file isn't showing any issues either

modules/codemod/lib/v7/recategorizeIconButtons.ts
```diff
@@ -69,11 +70,19 @@ export default function transformer(file: FileInfo, api: API, options: Options)
...
       if (variantProp) {
-        const variantPropValue = ((variantProp as JSXAttribute).value as StringLiteral)?.value;
+        const valueBody = variantProp.value;
+        if (valueBody?.type === 'JSXExpressionContainer') {
+          variantProp.value = valueBody.expression as StringLiteral;
+        }
+
+        const variantPropValue = (variantProp.value as StringLiteral).value;
+
+        (variantProp.value as StringLiteral).value = variantPropValue.replace('Filled', '');
```
mannycarrera4, line 84: Thank you for fixing this! @jamesfan this fixes the issues you mention of the `inverseFilled` incorrectly mapping

modules/codemod/lib/v7/recategorizeIconButtons.ts
```diff
@@ -7,6 +7,7 @@ import {
   JSXAttribute,
   ASTPath,
   JSXElement,
+  JSXExpressionContainer,
```
RayRedGoose, line 10:
```suggestion
```

This causes an issue in building

modules/codemod/lib/v7/recategorizeIconButtons.ts
```diff
@@ -69,11 +70,19 @@ export default function transformer(file: FileInfo, api: API, options: Options)
...
-      );
+      ) as JSXAttribute | undefined;
 
       // Default IconButton variant is `circle`
       if (variantProp) {
-        const variantPropValue = ((variantProp as JSXAttribute).value as StringLiteral)?.value;
+        const valueBody = variantProp.value;
+        if (valueBody?.type === 'JSXExpressionContainer') {
+          variantProp.value = valueBody.expression as StringLiteral;
+        }
```
willklein, line 80: What do you think about this?
```suggestion
        if (valueBody?.type === 'JSXExpressionContainer' && valueBody.expression.type === 'StringLiteral') {
          variantProp.value = valueBody.expression;
        }
```

RayRedGoose: For me, previous variant seems cleaner than adding additional validation. `valueBody.expression` is always a `StringLiteral`, so It's just typescript issue

modules/codemod/lib/v7/recategorizeIconButtons.ts
```diff
@@ -69,15 +70,35 @@ export default function transformer(file: FileInfo, api: API, options: Options)
...
+            nodePath.value.openingElement.attributes?.splice(attrs?.indexOf(variantProp)!, 1);
+          }
+        } else {
+          console.warn(
+            chalk.bold.red('** CANNOT UPDATE EXPRESSION FOR ICON BUTTON VARIANT. **') +
+              '\nYou will need to manually update this file to fix this warning.' +
+              '\nFile Name: ' +
+              chalk.blue(file.path) +
+              '\nConsult Upgrade Guide Section for v7: "Removal of Icon Button"'
+          );
```
mannycarrera4, line 101: i like this a lot! Should we  make this a pattern

mannycarrera4 (approved):
this is a big improvement!

modules/codemod/lib/v7/recategorizeIconButtons.ts
willklein: FYI, this turned out not to be the case. When I see `as SomeType` that's a warning sign that it warrants investigation. Sometimes it really is fine, and there are definitely a couple AST types where I've seen this come up.

In this case, the expression really can be an expression, like a ConditionalExpression (ternary), or LogicalExpression (something && something).

PR #1567, opened 2022-05-04
fix: Update codemods to handle different imports

mannycarrera4:
## Summary

While updating canvas site, we found a few cases in which the codemod wasn't handling specific imports. This change updates the codemod to handle more use cases.

category


mannycarrera4:
@NicholasBoll  found some issues with our codemods, going to keep this open

modules/codemod/lib/v7/utils/getImportRenameMap.ts
```diff
@@ -12,7 +12,9 @@ export function getImportRenameMap(j: JSCodeshift, root: Collection<any>, packag
     const value = node.source.value;
     if (
       typeof value === 'string' &&
-      (value.startsWith('@workday/canvas-kit-react') || value === packageName)
+      (value === '@workday/canvas-kit-react' ||
+        value.startsWith('@workday/canvas-kit-react/') ||
+        value === packageName)
```
NicholasBoll, line 17: This change prevents old imports from matching. For example `@workday/canvas-kit-react-core`

NicholasBoll (approved):

PR #1568, opened 2022-05-04
fix: Fix imports in examples

jamesfan:
## Summary

Updates examples in Banner and Tabs to import from packages rather than a path (canvas-site requires package imports when consuming examples from canvas-kit).

category


alanbsmith (approved):
🚢

NicholasBoll (approved):

Issue #1571, opened 2022-05-05
Fix build-mdx handling of Flex splitprops import


jamesfan:
## 🐛 Bug Report

We have a [build-mdx](https://github.com/Workday/canvas-kit/blob/prerelease/major/modules/docs/utils/build-mdx.js) script in our docs module which processes our MDX files so they can be consumed by the canvas-site. In most cases, it will convert examples from named imports to default imports since the canvas-site requires examples to be default exports. For example...

```
import {Basic} from './examples/Basic';
```

is converted to...

```
import Basic from './examples/Basic';
```

The exports in the example files are [converted to default exports](https://github.com/Workday/canvas-kit/blob/prerelease/major/modules/docs/utils/build-mdx.js#L83) by build-mdx as well.

Occasionally, we'll use something called a splitprops file ([example](https://github.com/Workday/canvas-kit/blob/prerelease/major/modules/react/pagination/stories/PropTables.splitprops.tsx)) which includes dummy components we use to display prop tables for things that aren't actually components such as a model's config, state, and event objects.

Unlike examples, splitprops files are [NOT converted to default exports](https://github.com/Workday/canvas-kit/blob/prerelease/major/modules/docs/utils/build-mdx.js#L86). Therefore, the corresponding splitprops imports in the MDX files should NOT be converted from named imports to default imports. This [line](https://github.com/Workday/canvas-kit/blob/prerelease/major/modules/docs/utils/build-mdx.js#L31) in build-mdx which converts the named imports to default imports happens to avoid catching imports from splitprops files because the regex doesn't catch blocks of multiple named imports, multiline imports, or imports that aren't from the examples folder. One of these is always true for splitprops imports ([example](https://github.com/Workday/canvas-kit/blame/prerelease/major/modules/react/pagination/stories/pagination.stories.mdx#L11)).

EXCEPT for [this splitprops import](https://github.com/Workday/canvas-kit/blame/prerelease/major/modules/react/layout/stories/Flex.stories.mdx#L8) in the Flex MDX. This import is a single named import on a single line from the examples folder, so it's converted to...

```
import FlexStyle from './examples/PropTables.splitprops.tsx';
```

...which breaks when consuming the Flex MDX docs on canvas-site since `FlexStyle` remains a named export in PropTables.splitprops.tsx.

PR #1575, opened 2022-05-10
fix: Fix build mdx script to better handle splitprops files

jamesfan:
## Summary

Fixes: #1571 

Fixes an issue in the build-mdx script where it was incorrectly rewriting a named import in `Flex.stories.mdx` to a default import.

`import {FlexStyle} from './examples/PropTables.splitprops.tsx'` was being rewritten to `import FlexStyle from './examples/PropTables.splitprops.tsx'` when instead it should NOT have been rewritten at all. Imports from splitprops files need to remain named imports.

This PR updates the regex we use to rewrite the imports to properly exclude all imports from `splitprops` files. The code change itself is very small (adds a `(?!.*splitprops)` [negative lookahead](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Assertions#other_assertions) to the regex); most of the PR is better documentation around how the regex for rewriting the imports works.

I performed a `yarn build` from `modules/docs` before and after the fix and diff-ed the resulting `dist` folders. The folders were identical except for desired change to the `FlexStyle` import.

category

### DIFF
--- a/modules/docs/utils/build-mdx.js
+++ b/modules/docs/utils/build-mdx.js
@@ -27,8 +27,54 @@ const sanitizeMdxFile = (inFile, outFile) => {
       .replace(/import {.*} from '@storybook\/addon-docs';/g, '')
       .replace(/<Meta.* \/>\n/g, '')
       .replace(/^\s+|\s+$/g, '')
-      // Convert named example import to default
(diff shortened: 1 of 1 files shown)


alanbsmith (approved):
👍🏻

PR #1581, opened 2022-05-13
docs: Update Migration guide with tabs changes

NicholasBoll:
## Summary

Add Migration notes for `Tabs`. Fix codemod to include

category


modules/codemod/lib/v7/spec/updateTabs.spec.ts
```diff
@@ -5,7 +5,7 @@ import {stripIndent} from 'common-tags';
 const expectTransform = expectTransformFactory(transform);
 
 describe('updateModelSignatures', () => {
-  it('should replace the "name" attribute in Tabs.Items with "data-id"', () => {
+  it('should replace the "name" attribute in Tabs.Items and Tabs.Panel with "data-id"', () => {
```
jamesfan, line 8: `Tabs.Panel` test with `name` and `data-id` is a separate test below.
```suggestion
  it('should replace the "name" attribute in Tabs.Items with "data-id"', () => {
```

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -418,6 +419,20 @@ Two props where added to the `SidePanel.ToggleButton`:
...
+`Tabs.Menu.Item`. The attribute changed from `name` to `data-id`. We did this to not interfere with
+the HTML `name` attribute.
+
+### Breaking Change
+
+🤖 The codemod will handle all these changes automatically:
+
+- `<Tabs.Item name=` becomes `<Tabs.Item data-id=`
+- `<Tabs.Panel name=` becomes `<Tabs.Panel data-id=`
+- `<Tabs.Menu.Item name=` becomes `<Tabs.Menu.Item data-id=`
```
jamesfan, line 434: Is it strange that the overflow menu works correctly on the Canvas Site even though I didn't change `name` to `data-id` for `Tabs.Menu.Item`? I noticed the OverflowTabs example is still using [`name` for `Tabs.Menu.Item`](https://github.com/Workday/canvas-kit/blob/prerelease/major/modules/react/tabs/stories/examples/OverflowTabs.tsx#L36) as well.

NicholasBoll: That could be because I allowed the `useRegisterListItem` hook to accept either `name` or `data-id`. I've since removed support for `name`. The updated codemod should help with this.

jamesfan: Got it, I'll update Canvas Site to use `data-id` for `Tabs.Menu.Item`. Should we update the OverflowTabs example in this PR as well then?

jamesfan (approved):

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
NicholasBoll: Yes! Good catch!

PR #1584, opened 2022-05-14
docs: Clean up component docs

jamesfan:
## Summary

Cleans up the Flex and Stack MDXes to follow our [documentation template](https://github.com/Workday/canvas-kit/discussions/1131) more closely. Both documents should now flow more smoothly, particularly when displayed within the context of the Canvas Site.

I had to diverge a bit from the template given that the Flex and Stack MDXes are both imported into the same Layout page on the Canvas Site. For instance, the Flex and Stack MDXes cannot share the same level 2 heading names (e.g., "Props") since this results in those headings being assigned the same `id` slug which creates problems with our AnchorNav jump links.

Aside from removing the non-standard Table of Contents sections, I mostly moved content around and renamed headings -- very little was added or removed to the MDX docs.

Additionally, I made a few corrections to the Button docs regarding `iconPosition` changing from `left` and `right` to `start` and `end` in v7.

category

### DIFF
--- a/modules/react/button/lib/TertiaryButton.tsx
+++ b/modules/react/button/lib/TertiaryButton.tsx
@@ -27,9 +27,9 @@ export interface TertiaryButtonProps extends Themeable {
    */
   size?: ButtonSizes;
   /**
-   * Button icon positions can either be `left` or `right`.
-   * If no value is provided, it defaults to `left`.
-   * @default 'left'
+   * Button icon positions can either be `start` or `end`.
+   * If no value is provided, it defaults to `start`.
+   * @default 'start'
    */
   iconPosition?: IconPositions;
   /**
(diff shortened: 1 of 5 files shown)


modules/react/button/lib/TertiaryButton.tsx
```diff
@@ -27,9 +27,9 @@ export interface TertiaryButtonProps extends Themeable {
    */
   size?: ButtonSizes;
   /**
-   * Button icon positions can either be `left` or `right`.
-   * If no value is provided, it defaults to `left`.
-   * @default 'left'
+   * Button icon positions can either be `start` or `end`.
+   * If no value is provided, it defaults to `start`.
+   * @default 'start'
```
alanbsmith, line 32: good catch – thank you.

alanbsmith (approved):
Looks great! Thanks!

PR #1587, opened 2022-05-17
chore(disclosure): Remove event data wrapper in disclosure events

NicholasBoll:
## Summary

Remove the object wrapping around the `show` and `hide` events for the disclosure family of models: `useDisclosureModel`, `usePopupModel`, `useModalModel`, and `useDisclosureModel`. The `show` and `hide` events used an object due to limitations of Typescript 3.8 in our v6 models, but the object wrapper is more to use. For example, it is common to attach the `show` or `hide` as part of a DOM event.

```tsx
// v6
<button
  onClick={e => model.events.show({ event })}
/>

// v7
<button
  onClick={model.events.show}
/>
```

Now the `event` from the `onClick` can be forwarded directly to the `show` event. This also means the `shouldShow`, `onShow`, `shouldHide`, and `onHide` model config signatures have changes. Another change removed the object wrapper around `data`, `state`, and `prevState`. This change goes further and removes the data object wrapping around the `event`.

```tsx
// v6
const model = usePopupModel({
  shouldShow({ data: { event }, state }) {
    console.log(event, state)
  },
  onShow({ data: { event }, prevState }) {
    console.log(event, prevState)
  }
})

// v7
const model = usePopupModel({
  shouldShow(event, state) {
    console.log(event, state)
  },
  onShow(event, prevState) {
    console.log(event, prevState)
  }
})
```

category

### Release Note

### BREAKING CHANGES

Removed the data wrapper around the `event` in `show` and `hide` events. The arguments are not commonly used, however.

```tsx
// v6
const model = usePopupModel({
  onShow({ data: { event }, prevState }) {
    console.log(event);
  }
})

// v7
const model = usePopupModel({
  onShow(event, prevState) {
    console.log(event);
  }
})
```

🤖 The codemod will update all inline guards and callbacks like in this example. If a guard or
callback was defined outside the model config block, it will not be covered by the codemod.

---


mannycarrera4 (commented):
is this going to affect all consumers who use popup or modal as well?

NicholasBoll:
> is this going to affect all consumers who use popup or modal as well?

Yes. I'm waiting on the migration guide update to add to the migration guide. Very few people use the parameters of the `show`, `hide` events or in the `shouldShow`, `shouldHide`, `onShow`, and `onHide` guards and callbacks, so impact is minimal.

mannycarrera4:
ok cool, that was my only comment, everything else looks good

mannycarrera4 (approved):

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -789,6 +789,65 @@ anyway since the buttons will no longer scroll with overflowed `Body` content.
...
+
+// v7
+<button onClick={model.events.show} />
+```
+
+Removing the object wrapper around the `event` allows us to directly pass the `show` or `hide` event
+to the `onClick` handler which is much more convenient. We couldn't find many uses of the `event` in
+the wild, so impact will be minimum. Most usage of `show` or `hide` events are called without
+`event`, so this change will not impact most people. There is no codemod for this change, because
+the usage is much more difficult to detect since most people pass a callback that doesn't take
```
jamesfan, line 819:
```suggestion
`event`, so this change will not impact most people. There is **no** codemod for this change, because
the usage is very difficult to detect since most people pass a callback that doesn't take
```

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -789,6 +789,65 @@ anyway since the buttons will no longer scroll with overflowed `Body` content.
...
+parameters.
+
+Guards and Callback signatures have also changed to remove the object wrapper around the `event`.
```
jamesfan, line 822:
```suggestion
Guards and callback signatures have also changed to remove the object wrapper around the `event`.
```

modules/docs/mdx/7.0-MIGRATION-GUIDE.mdx
```diff
@@ -789,6 +789,65 @@ anyway since the buttons will no longer scroll with overflowed `Body` content.
...
+  shouldShow(event, state) {
+    console.log(event, state);
+  },
+  onShow(event, prevState) {
+    console.log(event, prevState);
+  },
+});
+```
+
+🤖 The codemod will update all inline guards and callbacks like in this example. If a guard or
```
jamesfan, line 848:
```suggestion
🤖 The codemod will update all inline guards and callbacks defined like they are in this example. If a guard or
```

Issue #1589, opened 2022-05-17
TableRow component does not support refs


alanbsmith:
## 🐛 Bug Report

The `TableRow` component does not export a `ref`.

## Expected Behavior

All CK components should allow a `ref`

## Actual Results

No ref is available.

PR #1590, opened 2022-05-17
fix(table): Refactor TableRow to use createComponent

alanbsmith:
## Summary

Fixes: #1589

category

---

## Checklist

### DIFF
--- a/modules/react/table/lib/TableRow.tsx
+++ b/modules/react/table/lib/TableRow.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import styled, {CSSObject} from '@emotion/styled';
 import {rgba} from 'polished';
 import {colors, space, spaceNumbers, statusColors} from '@workday/canvas-kit-react/tokens';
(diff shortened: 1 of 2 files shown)


mannycarrera4 (approved):

modules/react/table/lib/TableRow.tsx
```diff
@@ -192,16 +192,22 @@ const Row = styled('tr')<TableRowProps>(
...
-export default class TableRow extends React.Component<TableRowProps> {
-  public static State = TableRowState;
-
-  public render() {
-    const {state, header, children, ...elemProps} = this.props;
-
+const TableRow = createComponent('tr')({
+  displayName: 'TableRow',
+  Component: (
+    {state, header, children, ...elemProps}: TableRowProps & StyledType,
```
NicholasBoll, line 198: This is a bit weird. The external interface already supports the `as` prop, so `StyledType` shouldn't be needed here. It would be needed on the `styled` function if it wasn't a string.

PR #1592, opened 2022-05-18
fix(action-bar): Make hidden buttons non-interactive

RayRedGoose:
## Summary

PR fixes issue with focus order for ActionBar. It was still select hidden buttons before focusing overflow menu button.

category


alanbsmith (approved):
👍🏻

Issue #1566, opened 2022-05-04
Broken Link in Code Style Guide


sarah-duncan:
## 🐛 Bug Report

There is a link in https://github.com/Workday/canvas-kit/blob/master/modules/docs/mdx/CONTRIBUTING.mdx#code-style-guide
called [API & Pattern Guidelines](https://github.com/Workday/canvas-kit/blob/master/getting-started/for-developers/resources/api-pattern-guidelines). that results in 404 error and presumably should point to https://github.com/Workday/canvas-kit/blob/master/modules/docs/mdx/API_PATTERN_GUIDELINES.mdx instead

## To Reproduce

Go to https://github.com/Workday/canvas-kit/blob/master/modules/docs/mdx/CONTRIBUTING.mdx#code-style-guide and click the link at API & Pattern Guidelines.
## Expected Behavior

The page should point to the API & Pattern Guidelines https://github.com/Workday/canvas-kit/blob/master/modules/docs/mdx/API_PATTERN_GUIDELINES.mdx instead
## Actual Results

A clear and concise description of what actually happens
Page not found error (404)
## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

## Link to repl or repo (highly encouraged)

Please provide a reference to a repo or branch to help us reproduce this bug.

### Error Output

Paste the error output here
Broken link: https://github.com/Workday/canvas-kit/blob/master/getting-started/for-developers/resources/api-pattern-guidelines

myvuuu:
A router (routes.js) should be used to link to https://canvas.workdaydesign.com/getting-started/for-developers/resources/api-pattern-guidelines

willklein:
Thanks for taking the time to log this issue, @sarah-duncan!

I just created #1602 to track a broader effort needed to improve our links. We can't implement the expected behavior because that would break both our [Storybook](https://workday.github.io/canvas-kit/) and the [Canvas site](https://canvas.workday.com/).

We should make those links functional though, even when accessing the source files on github.com, as this issue suggests. We need to do some prerequisite work to improve our redirects, routing, and also missing content, and then can point all of these links to URLs on the Canvas site.

While this issue is valid, I'm going to close it in favor of #1602 to better describe the effort we need to track. Thank you again!

PR #1601, opened 2022-05-22
docs: Updated CONTRIBUTING.mdx for #1566

ANaphade:
## Summary

Updated file CONTRIBUTING.mdx for fixing broken link issue #1566 

## Release Category

- docs

### DIFF
--- a/modules/docs/mdx/CONTRIBUTING.mdx
+++ b/modules/docs/mdx/CONTRIBUTING.mdx
@@ -210,7 +210,7 @@ This will start the unit tests.
 ### Code Style Guide
 
 Refer to the
-[API & Pattern Guidelines](/getting-started/for-developers/resources/api-pattern-guidelines/).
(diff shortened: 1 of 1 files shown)


willklein:
I just commented on #1566 and opened #1602. This change will eventually be desired but at the moment we have a redirect issue where this will not work in certain scenarios. I'm going to close this, but make a note on #1602 that we could later merge this once we're ready.

Thank you for taking the time to open a pull request though. This drew more attention to #1566 and sparked a discussion within the team on how we'd like to improve how we link our documentation. We would indeed like to link to the Canvas site as your PR implements, but it won't work due to a redirect issue in certain environments. The link actually works within both our Storybook and the Canvas site, so we'd rather leave it as-is for now. I also tracked in #1602 the desire to point people to our Canvas site URL for the CONTRIBUTING.mdx content so visitors have a better experience with this.

PR #1607, opened 2022-05-24
fix(codemod): Support multiple imports from canvas-kit

NicholasBoll:
## Summary

Fix a conflict where `TertiaryButton` and `IconButton` were part of 2 different imports

category


mannycarrera4 (approved):
Ah thanks for working on this!

Issue #1527, opened 2022-04-05
React 18 Upgrade


myvuuu:

RayRedGoose:
[More info about React 18 upgrade](https://github.com/Workday/canvas-kit/discussions/1545).

PR #1636, opened 2022-06-14
chore: Upgrade to React 18

RayRedGoose:
## Summary
- Upgrade `react` and `react-dom` to v18
- Upgrade `@types/react` and `@types/react-dom` to v18
- Update component types to remove deprecated types and fix issues with children in types 
- Update testing library to support React 18

Resolved: #1527 

dependencies

### BREAKING CHANGES
Canvas Kit will require teams to be on version ^18.x or at least ^17.0 for backwards compatibility. To use have all React 18 features would require teams to upgrade root API, otherwise they will be ignored.

## Checklist


modules/react/collection/lib/ListBox.tsx
```diff
@@ -53,6 +53,7 @@ export const ListBox = createContainer('ul')({
   },
 })<ListBoxProps>(({height, maxHeight, ...elemProps}, Element, model) => {
   // TODO figure out what style props should go to which `Box`
+  const renderedItems = useListRenderItems(model, elemProps.children) as React.ReactNode;
```
mannycarrera4, line 56: i wonder if we need to change the return type of `useListRenderItems` so that we don't have to do this every time we use this hook

modules/react/layout/lib/utils/types.ts
```diff
@@ -1,19 +1,70 @@
...
+  | 'hidden'
+  | 'inset'
+  | 'none'
+  | 'outset'
+  | 'ridge'
+  | 'solid';
+
+type ContentDistribution = 'space-around' | 'space-between' | 'space-evenly' | 'stretch';
+type ContentPosition = 'center' | 'end' | 'flex-end' | 'flex-start' | 'start';
+
```
mannycarrera4, line 67: why did these types change?

modules/react/package.json
```diff
@@ -54,7 +54,6 @@
     "@workday/canvas-system-icons-web": "^3.0.0",
     "@workday/design-assets-types": "^0.2.4",
     "chroma-js": "^2.1.0",
-    "csstype": "^2.6.17",
```
mannycarrera4, line 57: oh... this is probably why?

modules/react/popup/lib/hooks/useTransferOnFullscreenEnter.ts
```diff
@@ -16,7 +16,7 @@ import {usePopupModel} from './usePopupModel';
  */
 export const useTransferOnFullscreenEnter = createElemPropsHook(usePopupModel)(model => {
   const handler = React.useCallback(
-    event => {
+    (event: any) => {
```
mannycarrera4, line 19: should these be mouse events instead of any?

modules/react/layout/lib/utils/types.ts
RayRedGoose: `csstype` causes the type check failure. Alan and I tried to figure out why, but unsuccessfully. Probably it requires to be updated to a version with React 18 support, but we can't do it because of emotion. So I just replace `csstype` imports by this.

modules/react/popup/lib/hooks/useTransferOnFullscreenEnter.ts
RayRedGoose: I need to test it.

modules/react/modal/lib/ModalOverlay.tsx
```diff
@@ -14,7 +14,7 @@ import styled from '@emotion/styled';
 import {Box, BoxProps} from '@workday/canvas-kit-react/layout';
 import {useModalModel} from './hooks';
 
-export interface ModalOverlayProps extends BoxProps {
+export interface ModalOverlayProps extends Omit<BoxProps, 'children'> {
```
mannycarrera4, line 17: Do we have to omit children? Can't we just remove `children` down below and allow BoxProps to pass that children down?

modules/react/tabs/lib/TabsList.tsx
```diff
@@ -17,7 +17,7 @@ import {
 import {useTabsModel} from './useTabsModel';
 
 // Use `Partial` here to make `spacing` optional
-export interface TabListProps<T = unknown> extends Partial<ExtractProps<typeof Stack, never>> {
+export interface TabListProps extends Omit<Partial<ExtractProps<typeof Stack, never>>, 'children'> {
```
mannycarrera4, line 20: I guess here we're giving children a different type which makes sense to omit! Nice!

modules/react/tabs/lib/TabsPanels.tsx
```diff
@@ -5,11 +5,11 @@ import {useListRenderItems} from '@workday/canvas-kit-react/collection';
 
 import {useTabsModel} from './useTabsModel';
 
-export interface TabsPanelsProps<T = unknown> {
+export interface TabsPanelsProps {
   /**
    *
    */
-  children: ((item: T) => React.ReactNode) | React.ReactNode;
+  children: ((item: any) => React.ReactNode) | React.ReactNode;
```
mannycarrera4, line 12: I think Nicholas was trying to avoid having type `any` here because `T` would give you a generic of the item, but not it just says any. I would sync up with him on this one

RayRedGoose: This one should be reviewed, because it cause type mismatch for all menu components. It's possible to set generic type for `List` items, but not for `Menu.Popper` part of components.

modules/react/modal/lib/ModalOverlay.tsx
RayRedGoose: yes, I guess it's possible

modules/labs-react/combobox/lib/Combobox.tsx
```diff
@@ -135,7 +135,7 @@ const optionIdPart = `option`;
 export const getOptionId = (baseId?: string, index?: number) =>
   `${baseId}-${optionIdPart}-${index}`;
 
-export const getTextFromElement = (children?: React.ReactNode) => {
+export const getTextFromElement = (children?: React.ReactNode | {}) => {
```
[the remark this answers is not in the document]
RayRedGoose: it was a temporary solution for fixing type issue with `child === {}` from line 141. But it will be removed because of fully removal of `child === {}`. This check doesn't have sense as it always false (even if child is empty object: `{} === {} // return false`.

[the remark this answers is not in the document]
mannycarrera4: gotcha thanks for explaining that!

modules/react/tabs/lib/TabsList.tsx
```diff
@@ -27,7 +28,7 @@ export interface TabListProps extends Omit<Partial<ExtractProps<typeof Stack, ne
    *   {(item) => <Tabs.Item>{item.text}</Tabs.Item>}
    * </Tabs.List>
    */
-  children: ((item: any) => React.ReactNode) | React.ReactNode;
+  children: ((item: T) => React.ReactNode) | React.ReactNode;
```
mannycarrera4, line 31: nice this is looking much better!

RayRedGoose: `any` still be as a default of `T` here instead of `unknown`, because `unknown` causes issue with incompatibility of types of provided children function and this children in components based on `Menu`. This issue start displayed because of some changes made in `@types/react` v17, but as we didn't update it with react 17 it shows now.

modules/react/popup/lib/hooks/useTransferOnFullscreenEnter.ts
RayRedGoose: Neither React events or `screenfull` event don't work correctly. I guess it was a reason why there wasn't assigned type to event before, but it didn't cause type failing. Now not assigned type fails checks, so `any` causes the same skip type checking for `event` argument as it was before.

jest/verifyComponent.tsx
```diff
@@ -18,11 +18,11 @@ import {render, fireEvent} from '@testing-library/react';
  * })
  */
 export function verifyComponent(
-  Component: React.ComponentType<any>,
+  Component: React.ComponentType<React.PropsWithChildren<any>>,
```
NicholasBoll, line 21: I'm curious about this change. It doesn't look like it does anything.

Here's the type for `React.PropsWithChildren<T>`:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ea4d4e50e11788ef0a8094d7de564ad2f01dfedd/types/react/index.d.ts#L799
```tsx
type PropsWithChildren<P = unknown> = P & { children?: ReactNode | undefined };
```

Passing `any` to it will result in:
```tsx
any & { children?: ReactNode | undefined }
```

which will result in `any`.

Was this a find/replace or did this solve something?

modules/css/button/stories.tsx
```diff
@@ -30,7 +30,7 @@ const spaceButton = css({
   },
 });
 
-class IconDemo extends Component {
+class IconDemo extends Component<React.PropsWithChildren<any>> {
```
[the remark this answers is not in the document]
NicholasBoll: My best guess with this change is required because of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210

`children` is no longer implicit with many React components. I don't think the `any` is what we want though - it means ANY props are valid.

modules/css/icon/stories.tsx
```diff
@@ -16,7 +16,7 @@ const containerStyle = css({
   },
 });
 
-class IconDemo extends Component {
+class IconDemo extends Component<React.PropsWithChildren<any>> {
```
NicholasBoll, line 19:
```suggestion
class IconDemo extends Component<React.PropsWithChildren> {
```

The default generic for `React.PropsWithChildren` are `unknown`. `unknown` ANDed with any other type is the type.

For example:
```tsx
type Foo = unknown & { children: any } // {children: any}
```

https://www.typescriptlang.org/play?#code/LAKALgngDgpgBAMQPZLgXjgVwHYGttIDu2cAZHAN4DGAFgJYA2AJgE4zYBccAhthAL5wA9EMq1Grdl14CgA

modules/labs-react/combobox/stories/stories.tsx
```diff
@@ -38,10 +38,12 @@ const groupOfResults = (
...
-export const Autocomplete: FC<Omit<ComboboxProps, 'children'> & {
-  group?: boolean;
-  showDisabledItems?: boolean;
-}> = ({showClearButton, group, showDisabledItems = false, ...props}) => {
+export const Autocomplete: FC<React.PropsWithChildren<
+  Omit<ComboboxProps, 'children'> & {
+    group?: boolean;
+    showDisabledItems?: boolean;
+  }
+>> = ({showClearButton, group, showDisabledItems = false, ...props}) => {
```
NicholasBoll, line 46: This is kind of messy... In CK7, `children` was added to the props interface automatically (in React 18 it was removed... so we went a different direction). BUT, the CK version of `children` allows you to override `children`. If your prop interface includes `children`, yours will be uses instead of the default `children: React.ReactNode`.

This is more of an FYI and not related to this specific change since `Autocomplete` doesn't use our createComponent utility functions. This type of messy typing is because `Combobox` restricts `children` where most of our compound components do not do this.

modules/preview-react/breadcrumbs/lib/Breadcrumbs/List/index.tsx
```diff
@@ -6,23 +6,21 @@ export type BreadcrumbsListProps = React.HTMLAttributes<HTMLUListElement>;
 
 export const BreadcrumbsList = createComponent('ul')({
   displayName: 'BreadcrumbsList',
-  Component: ({children, ...elemProps}: BreadcrumbsListProps, ref) => {
+  Component: (elemProps: BreadcrumbsListProps, ref, Element) => {
     return (
       <Flex
+        as={Element}
```
NicholasBoll, line 12: Kind of a sneaky change to a real polymorphic component.

modules/react/action-bar/lib/ActionBarList.tsx
```diff
@@ -30,7 +30,7 @@ export interface ActionBarListProps<T = unknown>
    *   {(item) => <ActionBar.Item key={item.id} name={item.name}>{item.text}</ActionBar.Item>}
    * </ActionBar.List>
    */
-  children: ((item: T) => React.ReactNode) | React.ReactNode;
+  children: ((item: T, index: number) => React.ReactNode) | React.ReactNode;
```
NicholasBoll, line 33: Good catch

NicholasBoll: I wonder if this shouldn't  use `ListBoxProps` instead.

modules/react/collection/lib/useListRenderItem.tsx
```diff
@@ -7,14 +7,14 @@ import {useBaseListModel} from './useBaseListModel';
...
-) {
+): React.ReactNode {
   const items =
     typeof children === 'function'
       ? model.state.isVirtualized
         ? model.state.UNSTABLE_virtual.virtualItems.map(virtual => {
             const item = model.state.items[virtual.index];
             const child = children(item.value, virtual.index);
-            return React.cloneElement(child, {
+            return React.cloneElement(child as React.ReactElement, {
```
NicholasBoll, line 17: I'm curious about this change. Casting like this could cause issues where we're making an assumption about what `child` could be and ignoring when `child` is not a `React.ReactElement`.

modules/react/layout/lib/Flex.tsx
```diff
@@ -5,7 +5,7 @@ import {Box, BoxProps} from './Box';
 
 import {flex, FlexStyleProps} from './utils/flex';
 
-export type FlexProps = BoxProps & FlexStyleProps;
+export type FlexProps = Omit<BoxProps, 'display'> & FlexStyleProps;
```
NicholasBoll, line 8: I don't see  `display` as part of the React upgrade. Why is the `Omit` needed now? Was the previous code not properly restricting the `display` prop?

modules/react/modal/lib/ModalOverlay.tsx
```diff
@@ -86,7 +84,8 @@ const OpenModalOverlay = createSubcomponent('div')({
   displayName: 'Modal.OpenOverlay',
   modelHook: useModalModel,
   elemPropsHook: useModalOverlay,
-})<ModalOverlayProps>((elemProps: ModalOverlayProps, Element, model) => {
+  /* @ts-ignore */
```
NicholasBoll, line 87: The `@ts-ignore` is a red flag. It should be removed or there should be a comment why it is needed. The only change I can see is around `children`. What are the errors?

modules/react/popup/spec/Popper.spec.tsx
```diff
@@ -84,15 +84,15 @@ describe('Popper', () => {
 
     // force PopperJS to run
     // eslint-disable-next-line compat/compat
-    await act(() => new Promise(requestAnimationFrame));
+    await act(() => new Promise<any>(requestAnimationFrame));
```
NicholasBoll, line 87: Why is this `any` needed? It should extract the type from `requestAnimationFrame`, but we don't actually care about the return type.

modules/react/tabs/lib/TabsList.tsx
```diff
@@ -17,7 +17,8 @@ import {
 import {useTabsModel} from './useTabsModel';
 
 // Use `Partial` here to make `spacing` optional
-export interface TabListProps<T = unknown> extends Partial<ExtractProps<typeof Stack, never>> {
+export interface TabListProps<T = any>
+  extends Omit<Partial<ExtractProps<typeof Stack, never>>, 'children'> {
```
NicholasBoll, line 21: I wonder if `ExtractProps` should remove children... It is easier to add `children` than to take it away. I don't like using `Omit` as the type becomes much messier. All the `Extract*` utility types avoid using it and instead return clean interfaces

jest/verifyComponent.tsx
RayRedGoose: This was automatic replacement by [types-react-codemod](https://github.com/eps1lon/types-react-codemod`. We can keep it as `any`.

modules/react/popup/spec/Popper.spec.tsx
NicholasBoll: ReactDOM test utils changed the type of the promise that could be returned to `act` and forced it to be `void` even though the promise return type is ignored. `requestAnimationFrame` takes a callback that expects a `number` and they 2 come into conflict.

The alternative is:
```ts
await act(() => new Promise(cb => requestAnimationFrame(() => cb())));
```

But that seems messier

willklein:
Note, I'm removing this text from the PR description as a correction.

> ### BREAKING CHANGES
> Canvas Kit will require teams to be on version ^18.x or at least ^17.0 for backwards compatibility. To use have all React 18 features would require teams to upgrade root API, otherwise they will be ignored.

PR #1638, opened 2022-06-16
fix: Update remaining Emotion 11 imports

willklein:
## Summary

This cleanups our `verifyComponent`'s import for Emotion (it might be using Emotion 10 incorrectly) as well as the example imports for using `@workday/canvas-kit-react-fonts`.

category

### Release Note
Updated documentation for `@workday/canvas-kit-react-fonts` to correctly import utilities from Emotion 11 rather than Emotion 10.

---

## Checklist

### DIFF
--- a/jest/verifyComponent.tsx
+++ b/jest/verifyComponent.tsx
@@ -1,6 +1,6 @@
 /** @jsxRuntime classic */
 /** @jsx jsx */
-import {jsx} from '@emotion/core';
+import {jsx} from '@emotion/react';
 import React from 'react';
 import {render, fireEvent} from '@testing-library/react';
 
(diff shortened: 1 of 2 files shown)


jest/verifyComponent.tsx
```diff
@@ -1,6 +1,6 @@
 /** @jsxRuntime classic */
 /** @jsx jsx */
-import {jsx} from '@emotion/core';
```
willklein, line 3: I believe this may have been working because Emotion 10 still comes in because of Storybook. Tests seem to pass all the same either way, but let's tidy this up.

mannycarrera4 (approved):
ahhh thank you!!

Issue #1641, opened 2022-06-17
npmjs.com repository links aren't correct


willklein:
## 🐛 Bug Report

I was checking published Canvas Kit packages on [npmjs.com](https://www.npmjs.com/package/@workday/canvas-kit-react) and noticed the repository URL (http://github.com/Workday/canvas-kit/tree/master/modules/canvas-kit-react) leads to a 404 page.

The repository link on NPM:
<img width="438" alt="Screen Shot 2022-06-17 at 12 34 31 AM" src="https://user-images.githubusercontent.com/1075861/174239599-2cffda6f-4907-43a9-9cfa-551cd48c3f14.png">

Here is the `package.json` repository field for `@workday/canvas-kit-react`:
```json
  "repository": {
    "type": "git",
    "url": "http://github.com/Workday/canvas-kit/tree/master/modules/canvas-kit-react"
  },
```

I looked around and saw that we have misconfigured these fields for multiple packages. These were probably correct at one time. In some cases we configured them correctly, but all of them can be improved to include the `directory` field so users can be brought directly to the right URL for each module on GitHub.

The corrected `repository` field for `@workday/canvas-kit-react` would be:
```json
  "repository": {
    "type": "git",
    "url": "https://github.com/workday/canvas-kit.git",
    "directory": "modules/react"
  },
```

NPM docs: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#repository

## To Reproduce

npmjs.com repository link:
1. Go to the NPM page for `@workday/canvas-kit-react`: https://www.npmjs.com/package/@workday/canvas-kit-react
2. Click on the repository link on the right

`npm repo` CLI output
1. In a shell, run `npm repo @workday/canvas-kit-react`

## Expected Behavior

npmjs.com repository link:
- the link should be https://github.com/Workday/canvas-kit

`npm repo` CLI output
- the browser should open to https://github.com/Workday/canvas-kit/tree/master/modules/react

## Actual Results

npmjs.com repository link:
- the link is https://github.com/Workday/canvas-kit/tree/master/modules/canvas-kit-react which is not found

`npm repo` CLI output
- the browser opens to https://github.com/Workday/canvas-kit/tree/master

PR #1642, opened 2022-06-17
chore: Update repository fields for NPM

willklein:
## Summary

Fixes: #1641 

I like links to work. This will fix the repository links on NPM to work correctly for these packages:
https://www.npmjs.com/package/@workday/canvas-kit-codemod
https://www.npmjs.com/package/@workday/canvas-kit-css-fonts
https://www.npmjs.com/package/@workday/canvas-kit-css
https://www.npmjs.com/package/@workday/canvas-kit-docs
https://www.npmjs.com/package/@workday/canvas-kit-labs-react
https://www.npmjs.com/package/@workday/canvas-kit-popup-stack
https://www.npmjs.com/package/@workday/canvas-kit-preview-react
https://www.npmjs.com/package/@workday/canvas-kit-react-fonts
https://www.npmjs.com/package/@workday/canvas-kit-react

Consumers will also be able to run `npm repo <package-name>` and open the correct page. For example: `npm repo @workday/canvas-kit-react` will now open to https://github.com/Workday/canvas-kit/tree/master/modules/react

I did not bother with all the CSS sub-packages, like `@workday/canvas-kit-css-banner`. There would be 22 more `package.json` updates. I can do this if someone things it's valuable, but I think we can leave that alone.

category

---

## Checklist

## Additional References
https://docs.npmjs.com/cli/v6/configuring-npm/package-json#repository
https://docs.npmjs.com/cli/v8/commands/npm-repo

https://github.com/facebook/react/blob/main/packages/react/package.json#L32
https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/package.json#L74


modules/codemod/package.json
```diff
@@ -11,7 +11,8 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/Workday/canvas-kit.git"
+    "url": "https://github.com/workday/canvas-kit.git",
```
NicholasBoll, line 14: Why not https://github.com/workday/canvas-kit? https://github.com/workday/canvas-kit.git will redirect to https://github.com/workday/canvas-kit, so why not save the user a redirection?

willklein: Going by npmjs.com docs and examples. I think this field can get picked up for other purposes (VCS tools in editors maybe) so they recommend the `.git` extension, but not the `git+` prefix. I honestly don't have a specific example of how this matters though.

willklein: From *Additional References* on the PR description:
https://docs.npmjs.com/cli/v6/configuring-npm/package-json#repository
https://docs.npmjs.com/cli/v8/commands/npm-repo

https://github.com/facebook/react/blob/main/packages/react/package.json#L32
https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/package.json#L74

NicholasBoll: I didn't see any documentation referencing the `.git` extension in the URL, but I do see their examples using it. I know redirects can delay the page from rendering on higher latency connections which is why I like to avoid them.

NicholasBoll (approved):

modules/codemod/package.json
willklein: I agree this is not properly explained. They just say:
> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification. It should not be a url to an html project page that you put in your browser. It's for computers.

React uses .git, Chakra UI does not. I confirmed on npmjs.com that it will NOT include the .git extension for the repository URL. It also ignores the directory, and goes straight to the main repo page on GitHub. I had trouble confirming whether it leaves off `.git` when using `npm repo` but I did observe that it generates a URL when a directory field is specified, so I highly suspect it will also leave off .git. I just couldn't absolutely confirm it leaves off .git, but it will generate a link like: `https://github.com/chakra-ui/chakra-ui/tree/master/packages/breadcrumb` which redirects to `https://github.com/chakra-ui/chakra-ui/tree/main/packages/breadcrumb`. That leads me to suspect `.git is probably left off altogether when it generates this URL.

If we want a human-oriented URL then we can add a homepage URL to each package. I left off this improvement for now.

TL;DR: I observed no redirects akin to https://github.com/workday/canvas-kit.git to https://github.com/workday/canvas-kit in testing, and I would feel safer leaving `.git` there for possible VCS (version control systems) that might rely on it to work seamlessly. I don't feel too strongly though, so if you still insist on taking off `.git` I do not object and I'm happy to make that change.

Issue #1643, opened 2022-06-18
SidePanel's Right Origin example has content overlapping with the SidePanel


willklein:
## 🐛 Bug Report

The Preview `SidePanel` has a Right Origin example that sits on top of the non-side panel content. It also expands the side panel beyond the containing `<div>`. We can update this to give better guidance to the user. Using `position: absolute` / `right: 0` for styling the `SidePanel` is problematic.

Here's the current issue:
<img width="534" alt="Screen Shot 2022-06-18 at 2 16 32 PM" src="https://user-images.githubusercontent.com/1075861/174455833-15ec2365-a72f-4480-ae77-3334a4f83a7b.png">

Here's the `div` it should be rendering within, but doesn't:
<img width="520" alt="Screen Shot 2022-06-18 at 4 59 58 PM" src="https://user-images.githubusercontent.com/1075861/174459633-1c2d624f-e38c-4658-98f6-ac65b56743bf.png">

## To Reproduce

Look at the current story: https://workday.github.io/canvas-kit/?path=/story/preview-side-panel-react--right-origin

## Expected Behavior

Content should not be overlapping with the `SidePanel` when it's on the right.

## Actual Results

Content currently overlaps.

PR #1644, opened 2022-06-18
docs: Update SidePanel right origin story [skip release]

willklein:
## Summary

Fixes: #1643 

I tested this in IE11 too; here's the link to the deployed story: https://5d854c26ba934e0020f5e98a-mqybeegohw.chromatic.com/?path=/story/preview-side-panel-react--right-origin

Here's a CodeSandbox to play with too: https://codesandbox.io/s/sidepanel-right-origin-f7peik?file=/src/RightOrigin.tsx

Before:

Screen Shot 2022-06-18 at 4 57 47 PM

After:
Screen Shot 2022-06-18 at 5 06 33 PM

category

---

## Checklist

## Additional References

Used [margin-left: auto](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-right#values) instead of absolute positioning.

### DIFF
--- a/modules/preview-react/side-panel/stories/examples/RightOrigin.tsx
+++ b/modules/preview-react/side-panel/stories/examples/RightOrigin.tsx
@@ -18,8 +18,7 @@ const StyledHeader = styled('h3')({
 });
 
 const StyledSidePanel = styled(SidePanel)({
-  position: 'absolute',
-  right: 0,
+  marginLeft: 'auto',
 });
 
 const RightPanel = () => {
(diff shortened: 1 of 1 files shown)


mannycarrera4 (approved):

NicholasBoll (approved):

NicholasBoll:
FYI, `[skip release]` will not create a release for the doc site. If a docs change has a page on the doc site, there's no way to update the doc site without a release. I'm not too concerned as we don't update the doc site with every release and there's bound to be other changes that will trigger a release later.

Issue #1650, opened 2022-06-28
Upgrade Storybook to version with React 18 support


RayRedGoose:
## 🚀 Feature Proposal

Upgrade Storybook to version with React 18 support, now it's v6.5+.

## Motivation

Upgrading Storybook to the React 18 supporting version allows to visualize the new functionality in storybook and test it works correctly in our components. Without upgrading storybook runs Canvas Kit components using the legacy root API, so most new features would be ignored.

RayRedGoose:
Unblock by replacing story reload from `cypress-storybook` by loading additional story before tested one with extended wait time.

PR #1655, opened 2022-06-29
chore: Bump Storybook to 6.5 version with react 18 support

RayRedGoose:
## Summary

Resolves: #1650

Upgrading Storybook to the React 18 supporting version allows to visualize the new functionality in storybook and test it works correctly in our components. Without upgrading storybook runs Canvas Kit components using the legacy root API, so most new features would be ignored.

dependencies


RayRedGoose:
Blocked by Cypress right now. It should be reviewed how to make cypress working with new Storybook version and React 18 root API

RayRedGoose:
PR has been unblocked by adding additional story load before tested with extended wait time instead of deprecated `unmountComponentAtNode` in React 18 causing story reload from `cypress-storybook` package.

cypress/helpers/stories.ts
```diff
@@ -8,7 +8,10 @@
  * h.stories.load('Button', 'Primary')
  */
 export function load(categorization: string, story: string) {
-  return cy.loadStory(categorization, story);
+  cy.loadStory('Components/Buttons/Button/React', 'Primary');
+  // cy.wait(100);
```
jaclynjessup, line 12: remove?

cypress/integration/Button.spec.ts
```diff
@@ -31,7 +31,7 @@ describe('Button', () => {
     });
 
     it('should render the correct text', () => {
-      cy.get('button')
+      cy.findAllByRole('button', {name: 'Click Me'})
```
jaclynjessup, line 34: Why are we using `findAllByRole` here when this story only renders 1 button? Could we use `findByRole` instead?

RayRedGoose: Thank for catch! This replacement was copied from previous one, it should be changed.

.storybook/preview-head.html
```diff
@@ -1,6 +1,8 @@
 <script>
   document.title = 'Canvas Kit Storybook';
 </script>
+<!-- Polyfill for IE11 WeakSet -->
+<script crossorigin="anonymous" src="https://polyfill.io/v3/polyfill.min.js?features=WeakMap%2CWeakSet%2CSet"></script>
```
mannycarrera4, line 5: do we still need this since we're not supporting IE11 in v8?

mannycarrera4:
This looks good to me! Only thing is that yarn.lock file, i saw some changes to babel i wonder if that's coming from storybook dependencies

.storybook/preview-head.html
RayRedGoose: yes, it's needed until we finally deprecate IE.  It's not possible to set browser list only for specific branch but we still needs to run IE tests for v7.

RayRedGoose:
It's from emotion 11 update. As Storybook 6.5 finally uses emotion 11., we completed emotion 11 package renaming, one of this is emotion babel plugin.

.storybook/preview-head.html
mannycarrera4: Gotcha! But this is going into `major` for some reason i thought we weren't support IE11 in `major`

mannycarrera4 (approved):

.storybook/preview-head.html
RayRedGoose: I forgot to tell that it's Chromatic tests. we should run visual tests on all browsers, because on v7 we still support it. In chromatic there is no way to exclude IE only for this branch.

mannycarrera4: ahhh! yep yep ok thanks!

mannycarrera4: Maybe once we release this we can add a ticket to remove it since it won't be necessary

NicholasBoll (approved):

Issue #1660, opened 2022-06-30
Skeleton loading example still renders skeleton post-loading


willklein:
## 🐛 Bug Report

The Skeleton Simulation story still renders the skeleton after content loads:

![Screen Shot 2022-06-30 at 2 40 07 PM](https://user-images.githubusercontent.com/1075861/176774944-bc928d32-ae35-416c-b8ee-08aaee38abc2.png)

## To Reproduce

- wait 3 seconds for the loading to finish and you'll see the issue

## Expected Behavior

The skeleton should not render anymore, just the loaded content.

## Actual Results

The skeleton still renders when it should not

PR #1676, opened 2022-07-12
fix(skeleton): Move simulation code block within loading condition

anishatulai:
## Summary

Fixes: #1660

Skeleton Simulation on Storybook was always present even after the data completed loading. To fix, moved the simulation code block within a `loading` condition to ensure it only appears while in the loading state.

category

---

## Additional References

Screen Shot 2022-07-12 at 9 21 47 AM


willklein (approved):
:shipit:

Issue #1670, opened 2022-07-08
TertiaryButton icon-only variant doesn't have theme colors


RayRedGoose:
## 🐛 Bug Report

TertiaryButton doesn't apply theme colors to icon if it's icon only variant. 

![2022-07-07 18 21 20](https://user-images.githubusercontent.com/48605821/177892261-e277b324-c5ff-4432-8177-8fc9dc5f8d04.gif)

Because `getTertiaryButtonColors` function has this checks for existing children, it uses default colors for icon-only button. Primary and Secondary buttons takes colors from theme for icon-only button.

```
default: {
  icon: children ? themePrimary.main : colors.blackPepper400,
  label: themePrimary.main,
},
hover: {
  background: colors.soap200,
  icon: children ? themePrimary.dark : colors.blackPepper500,
  label: themePrimary.dark,
},
...

PR #1677, opened 2022-07-12
fix(button): Make icon-only tertiary button themeable

RayRedGoose:
## Summary

Fixes: #1670

Now only the icon-only tertiary button can't be modified by adding theme and uses default colors. PR contains fix allowing to make this buttons themeable.

category

### Release Note
We added `isThemeable` prop to `TertiaryButton` to allow applying theme colors for icon-only variant. Because default icon button doesn't use default theme colors, it makes icon-only `TertiaryButton` not themeable.


mannycarrera4:
Do we also want this for Primary and Secondary? We should probably add a visual test for this as well! thanks for catching this!

RayRedGoose:
@mannycarrera4 Primary and Secondary already can be themable for all variants.

modules/react/button/lib/TertiaryButton.tsx
```diff
@@ -99,27 +108,27 @@ const getTertiaryButtonColors = (
   } else {
     return {
       default: {
-        icon: children ? themePrimary.main : colors.blackPepper400,
+        icon: hasThemeStyles ? themePrimary.main : colors.blackPepper400,
```
mannycarrera4, line 111: if they add this prop with our default theme won't it change the icon color?

mannycarrera4: This is kind of tricky because for tertiary buttons we change the icon color completely and there isn't a 1:1 color match in our default theme

Issue #1490, opened 2022-02-26
Build Mobile Modal Variant for WebView


myvuuu:
## 🚀 Feature Proposal

Add a Modal variant for WebView based on the mobile design for Alerts (aka Modal in CK).

## Motivation

Allow teams to create WebView experiences using Canvas Kit to quickly test out the feasibility and need for generating fully native experiences.

## Requirements
- iOS styling is used for the WebView variant
- The Modal appears centered on the page, on top of the page content

myvuuu:
<img width="474" alt="WebView Alert:Modal" src="https://user-images.githubusercontent.com/84945599/157139333-9230fbcc-b92f-4ac0-8711-f798876bc651.png">

PR #1700, opened 2022-07-25
feat: Add responsive styles to Modal

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/1490

category

### Release Note
Add responsive styles for Modal. What this change includes:
- On smaller screen sizes, the padding reduces from `32px` to `24px`
- Border radius on the Modal card changes from `8px` to `24px`
- `useCloseOnOverlayClick` disables click events on `touch` devices.
- On smaller screen sizes, the modal with animated from the `bottom center` and be aligned towards the bottom of the viewport


modules/react/common/lib/utils/useModalityType.ts
```diff
@@ -11,24 +11,24 @@ type Modality = 'mouse' | 'touch' | 'pen';
...
-//   'mouse') as Modality;
+let localStorageValue = ((typeof localStorage !== 'undefined'
+  ? localStorage.getItem('modality')
+  : '') ||
+  (typeof document !== 'undefined'
+    ? document.documentElement.clientWidth < 768
+      ? 'touch'
+      : ''
+    : '') ||
+  'mouse') as Modality;
```
mannycarrera4, line 22: This was all commented out so just uncommented it.

modules/react/modal/lib/ModalCard.tsx
```diff
@@ -1,18 +1,31 @@
...
 export interface ModalCardProps extends ExtractProps<typeof Popup.Card, never> {}
 
+const ResponsiveModalCard = styled(Popup.Card)<ModalCardProps & StyledType>(({theme}) => ({
+  margin: space.xl,
+  padding: space.l,
+  [theme.canvas.breakpoints.down('s')]: {
+    margin: `0 0 ${space.xl}`,
+    padding: space.m,
+  },
+}));
```
mannycarrera4, line 23: The two biggest things that change are padding and margin. The padding decreases on smaller screen sizes and since the modal should transform from the bottom, I update the margin to give the modal some space.

modules/react/modal/lib/ModalOverlay.tsx
```diff
@@ -44,22 +44,24 @@ const Container = styled(Box)<StyledType>({
...
   // IE11 fix for setting min-height in a flex container
   ':before': {
     display: 'block',
     content: "''",
     height: '100vh',
   },
-});
+  [theme.canvas.breakpoints.down('s')]: {
+    alignItems: 'end',
+  },
```
mannycarrera4, line 63: On smaller screen sizes, we align the modal to the bottom instead of the center.

modules/react/modal/lib/hooks/useCloseOnOverlayClick.ts
```diff
@@ -30,14 +31,15 @@ export const useCloseOnOverlayClick = createElemPropsHook(usePopupModel)(model =
 
       if (
         dialog &&
+        modality !== 'touch' &&
```
mannycarrera4, line 34: We use modality in this case because on touch devices, Modals should not be dismissed if you click on the over overlay. This is a behavior change which is why modality makes sense in this case.

modules/react/popup/lib/PopupCard.tsx
```diff
@@ -65,11 +67,15 @@ export const PopupCard = createSubcomponent('div')({
...
+  const theme = useTheme();
   return (
     <StyledPopupCard
       as={As}
-      transformOrigin={transformOrigin}
+      transformOrigin={
+        theme.canvas.breakpoints.down('s')
+          ? {horizontal: 'center', vertical: 'bottom'}
+          : transformOrigin
+      }
```
mannycarrera4, line 78: Change the origin of the transform on smaller screen sizes.

modules/react/popup/stories/examples/Basic.tsx
```diff
@@ -27,7 +27,7 @@ export const Basic = () => {
     <Popup model={model}>
       <Popup.Target as={DeleteButton}>Delete Item</Popup.Target>
       <Popup.Popper placement="bottom">
-        <Popup.Card width={400} padding="s">
+        <Popup.Card width={400} padding="s" zIndex={9999}>
```
mannycarrera4, line 30: remove, i was testing something

mannycarrera4:
https://user-images.githubusercontent.com/7966550/182956331-c1c186e7-ae22-4d72-9d9e-2314ab4ceda9.mp4

modules/react/modal/stories/stories_VisualTesting.tsx
```diff
@@ -57,11 +61,13 @@ const TestModal = () => {
 export const ModalSmallWidth = withSnapshotsEnabled(() => <TestModal />);
 
 export const ModalRTL = withSnapshotsEnabled(() => {
+  const {direction, ...rest} = defaultCanvasTheme;
```
mannycarrera4, line 64: Curious of anyone has thoughts on this approach. The styling of the modal depends on `theme`. If someone only provides a partial theme to override direction, the rest of the canvas theme isn't provide and this causes the component to not render because it doesn't know what breakpoints is. I had to import defaultCanvasTheme and just change direction and then still spread the rest of the theme into the provider. CC @NicholasBoll @alanbsmith

modules/react/common/lib/utils/useModalityType.ts
RayRedGoose: This logic looks huge and there is a lot of repeats elements. Can we switch it to something like this:

```
const localStorageValue =
  ((localStorage && localStorage.getItem('modality')) as Modality) ||
  (document && document.documentElement.clientWidth < 768 && 'touch') ||
  'mouse';

```
So it validates if localStorage has some values (it covers all false possibilities) to try to get 'modality' value (and this is only value type that should be defined instead of the whole block)
if it's not it validates document and document client width. Otherwise it uses 'mouse'.

modules/react/modal/lib/ModalOverlay.tsx
```diff
@@ -44,22 +44,24 @@ const Container = styled(Box)<StyledType>({
 // positioning, but seems to when using flexbox centering. This messes up Popper calculations inside
 // the Modal. The centering container forces a "center" pixel calculation by making sure the width
 // is always an even number
-const CenteringContainer = styled('div')({
+const ResponsiveCenteringContainer = styled('div')(({theme}) => ({
```
RayRedGoose, line 47: What do you think about rename to just `ResponsiveCenteringContainer`? It seems that Responsive is more important here for meaning.

RayRedGoose (approved):
Overall It looks good for me. I have some comments, but it's not blockers.

modules/react/modal/stories/stories_VisualTesting.tsx
RayRedGoose: Have you tried: `const theme = useTheme({canvas: {direction: ContentDirection.RTL}});` and pass it into `CanvasProvider`? I guess it should create a the same theme but only with changed direction.

modules/react/modal/lib/ModalOverlay.tsx
mannycarrera4: oo can you clarify? I think the change shows that I did rename it to `ResponsiveCenteringContainer` ? Or are you saying to not add the responsive piece?

modules/react/modal/stories/stories_VisualTesting.tsx
mannycarrera4: YUS! thank you!!

modules/react/common/lib/utils/useModalityType.ts
mannycarrera4: works for me, nice clean up!

modules/react/modal/lib/ModalOverlay.tsx
RayRedGoose: I meant to rename to `ResponsiveContainer`.

Issue #1656, opened 2022-06-29
Adjust fade effect in Tabs to work with non-white backgrounds


jamesfan:
## 🐛 Bug Report

The Tabs component displays a white gradient after the last tab in the `Tabs.List` as a result of the `linear-gradient` added [here](https://github.com/Workday/canvas-kit/pull/1523/files#diff-5e3729c5f87ee54b9d212a45937aac7b46d5314877e6b5b1e8182b7d55a5901dR52) to render a fade-out effect for overflow tabs when viewed on a mobile touch device ([spec screenshot](https://github.com/Workday/canvas-kit/issues/1491#issuecomment-1061268486)).

This gradient is imperceptible if the Tabs are displayed on a white background, but it creates a visual issue if they're displayed on non-white background:

<img width="816" alt="tabs-white-gradient" src="https://user-images.githubusercontent.com/146020/176557294-06aaa33b-7a57-4e4a-8061-8dc51ac2ab67.png">

## To Reproduce

Render the tabs component on any non-white background.

PR #1705, opened 2022-07-29
fix(tabs): Use mask image to display fade effect on tabs

anishatulai:
## Summary

Fixes: #1656 

Replaced background css prop with mask-image to achieve a fade effect on tabs regardless of the background color applied.

category

## Additional References
Screen Shot 2022-07-29 at 11 47 08 AM
Screen Shot 2022-07-29 at 11 48 15 AM


modules/react/tabs/lib/TabsList.tsx
```diff
@@ -44,12 +44,12 @@ export const useTabsList = composeHooks(
 );
 
 const StyledStack = styled(Stack)<StyledType>({
+  maskImage: `linear-gradient(to right, white 95%, transparent)`,
   '::after': {
     content: '""',
     position: 'sticky',
     height: 52,
     minWidth: 30,
-    background: `linear-gradient(to right,rgba(255,255,255,0),white);`,
```
jamesfan, line 52: Thanks for taking this on @anishatulai! After reviewing your PR, this has turned out to be a more complicated issue than I anticipated 🤔 

I noticed `mask-image` doesn't work when applied to the `::after` pseudo-element: the `-webkit-mask-image/mask-image` style is applied, but it's not producing the desired visual fade effect. I'm guessing this is why you chose to apply `maskImage` to the StyledStack instead.

Unfortunately, this produces an undesired fade effect for users aren't using a touch device. This is a screenshot of the [Basic Tabs story](https://5d854c26ba934e0020f5e98a-aoijkqciui.chromatic.com/?path=/story/components-containers-tabs-react--basic) in the current preview build:

undesired-fade

Note how the thin gray bottom border of the tab list fades out at the very right (it's very subtle, but it's there). This is what's triggering the regression in the Chromatic test.

---

We could potentially address this by only applying `maskImage` to the StyledStack if the `modality` is set to `touch`. The subtle fade out on the right edge of the tab list border would still be present on touch devices (it wasn't before), but that may be acceptable. Here are before and after screenshots from my phone:

**Before**

phone-before

**After** (only applying `maskImage` if the `modality` is set to `touch`)

phone-after

---

Finally, we would also need to flip the direction of the `maskImage` gradient for RTL. Note the incorrect placement of the fade effect in the [RTL story](https://5d854c26ba934e0020f5e98a-aoijkqciui.chromatic.com/?path=/story/components-containers-tabs-react--right-to-left) in the current preview build (this is also causing a visual regression in the Chromatic test):

fade-rtl

modules/react/tabs/lib/TabsList.tsx
```diff
@@ -44,12 +44,12 @@ export const useTabsList = composeHooks(
 );
 
 const StyledStack = styled(Stack)<StyledType>({
+  maskImage: `linear-gradient(to right, white 95%, transparent)`,
   '::after': {
     content: '""',
     position: 'sticky',
     height: 52,
     minWidth: 30,
```
jamesfan, line 52: I suppose we could remove the `::after` pseudo-element entirely since it's sole purpose was to enable the fade effect (can I get a confirmation on this @NicholasBoll?) and we're now applying the effect to the StyledStack itself.

modules/react/tabs/lib/TabsList.tsx
anishatulai: I actually first tried adjusting the linear gradient background colors to match whatever is set on the element but that didn't look right. The mask-image, as you have seen, only works on the StyledStack itself. I can look into it more to see if there is any way to avoid the gray line fading as well. And I completely forgot about the RTL, good catch! Thanks for the feedback @jamesfan ! I will make the adjustments.

modules/react/tabs/lib/TabsList.tsx
```diff
@@ -44,12 +44,12 @@ export const useTabsList = composeHooks(
 );
 
 const StyledStack = styled(Stack)<StyledType>({
+  maskImage: `linear-gradient(to right, white 95%, transparent)`,
```
vibdev, line 47: Not sure if this is meant for the next major, but [mask-image doesn't have IE11 support](https://caniuse.com/?search=mask-image).

NicholasBoll: This is okay since this should only apply to touch modality. It is possible a touch device supports IE11, but we're okay with the mask image not working in this case.

modules/react/tabs/lib/TabsList.tsx
NicholasBoll: `::after` is used to prevent the use of non-semantic elements and to not fade the border. For example:

```html
<div role="tablist"> // stack with overflow and a border
  <div> // non-semantic element for the fade effect
```

We avoid these extra styling elements because you cannot get rid of them via the Component API of compound components and it serves to complicate "where to extra attributes/refs point to" and you no longer have access to all elements for easy styling. You're forced to use the Emotion styling API for nested styles - which have been an unofficial and undocumented API. Not a deal breaker, but we're trying to actively reduce elements that aren't tied to the semantic structure if possible. One current exception is `ListBox` requires multiple elements to render correctly which is unavoidable.

We may want to make this more complicated to support a fade effect only when there is something to be scrolled. This should reduce the likelihood that the last tab is active and faded which looks a bit silly.

modules/react/tabs/lib/TabsList.tsx
NicholasBoll: The fade doesn't bother me, but it may help if the fade is only apparent when the scrolling overflow is active (mobile) and when there is room to scroll. If you implement the latter, the former isn't necessary.

anishatulai: Updated to only show fade effect on touch devices. Fixed RTL support.

modules/react/tabs/lib/TabsList.tsx
anishatulai: Removed ::after styling now that fade effect is on the StyledStack itself.

modules/react/tabs/lib/TabsList.tsx
```diff
@@ -43,33 +44,29 @@ export const useTabsList = composeHooks(
...
-    height: 52,
-    minWidth: 30,
-    background: `linear-gradient(to right,rgba(255,255,255,0),white);`,
-    zIndex: 1,
-    right: 0,
-    top: 0,
-    pointerEvents: 'none',
-  },
-});
+const StyledStack = styled(Stack)<StyledType>(({maskImage}) => ({
```
jamesfan, line 47: If we do stick with this approach of applying `maskImage` to `StyledStack`, I would limit the `maskImage` typing to the `StyledStack` in TabsList rather than applying it to `StyledType` in the components.ts file (which isn't specific to Tabs and is used all over the place):

```suggestion
const StyledStack = styled(Stack)<StyledType & {maskImage?: string}>(({maskImage}) => ({
```

That said, to @NicholasBoll's point in his [earlier comment](https://github.com/Workday/canvas-kit/pull/1705#discussion_r937905533):

> We may want to make this more complicated to support a fade effect only when there is something to be scrolled. This should reduce the likelihood that the last tab is active and faded which looks a bit silly.

I do wonder if this is worth looking into given the odd visual effect Nicholas is referring to of the last active tab being faded:

last-active-tab-fade

jamesfan: Hi @anishatulai! Checking in to see if there's been any updates to this PR based on my last comments above. Do you have any leads on how to address the fadeout effect on the last tab?

jamesfan: No worries, this has been a trickier issue than anticipated. Thanks for the update!

modules/react/tabs/lib/TabsList.tsx
```diff
@@ -43,33 +44,30 @@ export const useTabsList = composeHooks(
...
   const modality = useModalityType();
+  const isRTL = useIsRTL();
   return (
     <StyledStack
       as={Element}
       position="relative"
       borderBottom={`1px solid ${commonColors.divider}`}
-      paddingX={modality === 'touch' ? 'zero' : 'm'}
+      paddingLeft={modality === 'touch' ? 'zero' : 'm'}
+      paddingRight="m"
```
jamesfan, line 64: We added this `paddingRight` to give the fade out gradient space to occupy so it doesn't show up on top of the last tab (see screenshot [here](https://github.com/Workday/canvas-kit/pull/1705#discussion_r938342058)). However, it doesn't seem to show up for me when I view the Basic Tabs story on my phone:

IMG_2223

Strangely, it does show up on desktop Chrome in device mode:

Screen Shot 2022-09-21 at 9 52 22 AM

jamesfan: This appears to be fixed in iOS 16. Sharing a video from @jaclynjessup which compares iOS 15 vs iOS 16 using the Xcode Simulator:

https://user-images.githubusercontent.com/146020/191618273-43034ce5-992e-4595-8c51-7b49558e2332.mov

Note the `paddingRight1` shows up in iOS 16 but not in iOS 15.

jamesfan:
@anishatulai Thanks for making the adjustment with the touch event handlers!

It looks like the fade only shows while touching and dragging on the device (i.e., there's no fade on the right side when the [Basic](https://5d854c26ba934e0020f5e98a-uqifarrwdf.chromatic.com/?path=/story/components-containers-tabs-react--basic) story is rendered _before_ you've touched the screen), and even then it only shows on one side at a time: if I drag my finger to the right, the fade only shows on the left edge of the tabs list; if I drag to the left, the fade only shows on the right side. Here's a video where I touch `First Tab` and then drag horizontally back and forth to illustrate what I'm referring to:

https://user-images.githubusercontent.com/146020/192372705-cb5488e9-37f8-4b0c-9452-bf33081101fe.MP4

The fade should show on whatever side has more content to scroll to (on both sides if you're scrolling through the middle of the tabs list) and it should be visible from the start (even before you've touched the screen).

I'm also realizing another challenge with using touch events to trigger the fade. It wouldn't make sense for the full fade effect to show on the left side as soon as _any_ portion of the first tab scrolls out of view. Otherwise, you end up with a jarring effect where the full fade immediately pops into view even though just the left sliver of the first tab is out of view:

tabs-too-much-fade

It seems like the fade effect itself needs to change based on the scroll/touch position, not just whether or not the fade is visible -- this sounds complicated. The padding-based solution is simpler in this regard: we just apply padding to the left and right of the tabs list, fill the space with the fade gradients, and the tabs slide under the fades as you scroll left or right. I wonder if we should be sticking to that approach.

cc: @NicholasBoll @jaclynjessup

jamesfan:
Putting this PR on hold until we release Canvas Kit v8 and to allow time for the mobile Tabs design spec to be updated. I'm leaving a few notes so we can resume where we left off when we're ready to pick this back up.

We have two potential approaches here:

1. Use a touch or scroll event handler to track when (and how) to display the fade effect. We'll need to be mindful of how the fade looks as we approach the left and right edges of the tabs list. I assume we don't want the entire fade to pop into existence at once (visually jarring) and instead present a gradual fade effect as the left and right edges of the tabs list fade out of view.
2. Use the padding on the left and right edges of the tabs list to give the fade gradients on each side room to occupy; tabs will slide underneath the fade gradients as you scroll through the tabs list. This only works if the width of the fade effect is **less** than the width of the padding.

anishatulai:
> Putting this PR on hold until we release Canvas Kit v8 and to allow time for the mobile Tabs design spec to be updated. I'm leaving a few notes so we can resume where we left off when we're ready to pick this back up.
> 
> We've identified two potential approaches to applying the fade effect so far:
> 
> 1. Use a touch or scroll event handler to track when (and how) to display the fade effect. We'll need to be mindful of how the fade looks as we approach the left and right edges of the tabs list. I assume we don't want the entire fade to pop into existence at once (visually jarring) and instead present a gradual fade effect as the left and right edges of the tabs list fade out of view.
> 2. Use the padding on the left and right edges of the tabs list to give the fade gradients on each side room to occupy; tabs will slide underneath the fade gradients as you scroll through the tabs list. This only works if the width of the fade effect is **less** than the width of the padding. This approach is simpler, but presents an issue in iOS 15 where the right-side padding does not render

To add to James' notes, if we do want to take the padding approach, there is currently an optional `spacing` prop on the StyledTabItem component to allow devs to override the default `marginLeft` prop. This prop may impact the fade effect. We may need either eliminate the optional prop or re-think what to get around variable marginLeft values.

mannycarrera4 (approved):

Issue #1437, opened 2022-01-21
Add a `useCloseOnScroll` hook that closes a popup when the target is no longer in the viewport


NicholasBoll:
## 🚀 Feature Proposal

Add a `useCloseOnScroll` hook that closes a popup when the target is no longer in the viewport. This would work well with the `useReturnFocus` hook.

## Motivation

The `useReturnFocus` hook will return focus back to the original target when the popup is closed (regardless of the reason). If the popup closes when the user scrolls away from the target, the browser will scroll back to the target element which is jarring for users.

## Example

A dropdown menu typically closes when a user moves the mouse away and this is often based on an intent timer. If a dropdown menu is opened and the user scrolls away, the `mouseout` event will fire and the browser will force scroll back to the original target element.

Here is a link to an example and a video of what happens.

https://codesandbox.io/s/dropdown-scroll-closes-popup-8i8j7

![MenuCloseScroll](https://user-images.githubusercontent.com/338257/150565363-2951d520-ba59-4475-b4c9-b894bc7fd7f7.gif)

NicholasBoll:
A Story demonstrating the functionality: https://5d854c26ba934e0020f5e98a-zuhyrbizeg.chromatic.com/?path=/story/testing-react-popups-popup--close-on-target-hidden-test

PR #1709, opened 2022-07-29
fix: Update `useReturnFocus` to change focus if target is visible

### LINKED ISSUE #1436: useReturnFocus can scroll page when clicking outside
## 🐛 Bug Report

A popup using `useCloseOnOutsideClick` and `useReturnFocus` will scroll to the `Popup.Target` element when scrolling away from the popup while it is open and clicking outside the popup.

## To Reproduce

Open the linked code sandbox and follow the instructions.

## Expected Behavior

The browser should not change the scroll position.

## Actual Results

A clear and concise description of what actually happens

https://codesandbox.io/s/popup-click-outside-focus-c6dhy

### Error Output

PopupClickScroll

NicholasBoll:
## Summary

Fixes #1436
Also fixes issue in #1437, but doesn't implement `useCloseOnScroll`

`useReturnFocus` will now check if the target is rendered onscreen and if the user clicked on a focusable element. It will not return focus to the target in these cases.

category

---

## Checklist

## For the Reviewer

- [x] PR summary describes the change (Fixes/Resovles linked correctly)

Extra heuristics were added to `useReturnFocus` when focus is meant to be returned. If the user clicks outside a popup, the element that is clicked on is checked to see if it is a focusable element. If it is a focusable element, `useReturnFocus` will _not_ return focus. This fixes the issue where focus is "stolen". Not returning focus when the target element is off screen fixes the issue where scrolling "snaps back" to the target.

## Testing Manually

Go to the new "Return Focus Test" and follow the instructions. Also automated tests verify functionality.


modules/react/popup/lib/hooks/useReturnFocus.tsx
```diff
@@ -14,13 +27,25 @@ import {usePopupModel} from './usePopupModel';
 export const useReturnFocus = createElemPropsHook(usePopupModel)(model => {
   const visible = model.state.visibility !== 'hidden';
 
-  React.useLayoutEffect(() => {
+  React.useEffect(() => {
```
NicholasBoll, line 30: `useEffect` must be used for `document.activeElement` to be the correct element. `useLayoutEffect` happens before the browser shifts focus.

The downside is there may be a frame rendered by the browser where the target element doesn't have focus. The upside is we can add extra heuristics to determine if the user clicked on a focusable element (we should not change focus in this case).

modules/react/common/lib/utils/elements.ts
```diff
@@ -2,7 +2,7 @@
  * Is an element focusable? This function performs various tests to see if the element in question
  * can receive focus. Should skip disabled elements as they are not focusable.
  */
-export const isFocusable = (element: HTMLElement) => {
+export const isFocusable = (element: Element) => {
```
NicholasBoll, line 5: Turns out I can make this type more permissive which helps for some callers that use `Element` instead of `HTMLElement`

modules/react/common/lib/utils/elements.ts
```diff
@@ -35,7 +35,11 @@ export const getFirstFocusableElement = (container: HTMLElement): HTMLElement |
 
   for (let i = 0; i < elements.length; i++) {
     const element = elements.item(i);
-    if (element && isFocusable(element as HTMLElement)) {
+    if (
+      element &&
+      isFocusable(element as HTMLElement) &&
```
mannycarrera4, line 40: Do you still need to cast it as HTMLElement?

mannycarrera4:
Only thing I'm not sure of the intended behavior. If I open the popup and try to tab, and I tab out of the close button, what is the expected behavior?

modules/react/common/lib/utils/elements.ts
NicholasBoll: No, we do not. Nice catch

modules/react/popup/lib/hooks/useReturnFocus.tsx
```diff
@@ -30,19 +57,62 @@ export const useReturnFocus = createElemPropsHook(usePopupModel)(model => {
...
+  // next focusable element. This means that the `useFocusRedirect` will change focus back to the
+  // target, then the tab key is processed by the browser to focus on the next item after the target
+  // without us having to use a focus manager to guess what the next focusable element is.
+  //
+  // But because we're using `useLayoutEffect` instead of `useEffect`, we cannot use
+  // `document.activeElement` (because the focus hasn't changed yet because the browser hasn't
+  // triggered a default action yet) to determine if a user clicked on something that can accept
+  // focus itself. If we don't short-circuit return focus, the user will intend to click on an
+  // `input` element only to get focus changed back. This is annoying to users and unintuitive. Instead,
+  // we'll use `mousedown` and `mouseup` to determine if we're closing because of a mouse click.
```
mannycarrera4, line 84: Do we have to account for `blur` events or does mouse down and up cover our cases?

mannycarrera4 (approved):
Looks good to me! Had a question on the math we're doing to calculate element height vs scroll parent and some reasoning behind it!

Issue #1718, opened 2022-08-10
Create component generator uses old patterns


vibdev:
## 🐛 Bug Report

Using the `yarn run create-component` uses the old createComponent and is missing mdx file for stories.

PR #1719, opened 2022-08-10
fix: Update create-component scripts

vibdev:
## Summary

Fixes: #1718 

Infrastructure

### Release Note
Updates the create create-component script to generate components using the new 7.x functions. It also updates the generated stories to use mdx.


modules/preview-react/index.ts
```diff
@@ -7,5 +7,3 @@ export * from './select';
 export * from './side-panel';
 export * from './text-area';
 export * from './text-input';
-
-/// <reference types="@types/node" />
```
vibdev, line 11: Not sure why this was here, but it was breaking being able to create a preview component. I didn't see anything like this in other index files, so removing seems safe?

utils/create-component/writeModuleFiles.js
```diff
@@ -14,7 +15,7 @@ module.exports = (files, modulePath) => {
     const file = files[key];
     const filePath = path.join(modulePath, file.path);
 
-    console.log('Creating ' + `.${filePath.replace(cwd, '')}`.cyan);
```
vibdev, line 17: This is what was causing the undefined printing bug. Sad typescript noises.

modules/preview-react/index.ts
alanbsmith: Yeah, no idea why that's there.

alanbsmith: Okay, I read up on [triple-slash directives](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html). And based on this, I think it's safe to remove it.

> Triple-slash directives are only valid at the top of their containing file. A triple-slash directive can only be preceded by single or multi-line comments, including other triple-slash directives. If they are encountered following a statement or a declaration they are treated as regular single-line comments, and hold no special meaning.

That said, if all our types blow up, this will be the line we need to add back.

utils/create-component/createReactComponent.js
```diff
@@ -22,12 +27,11 @@ const cwd = process.cwd();
 module.exports = (modulePath, name, description, prerelease, category) => {
   const moduleName = `@workday/canvas-kit-${prerelease && prerelease + '-'}react/${name}`;
 
-  console.log('\nCreating ' + chalk.blue.underline(`${moduleName}\n`));
+  consoleMessage('\nCreating', `${moduleName}\n`);
```
alanbsmith, line 30: Q: Is changing blue underline to cyan intentional?

utils/create-component/createReactComponent.js
```diff
@@ -22,12 +27,11 @@ const cwd = process.cwd();
...
 
   mkdirp.sync(modulePath);
 
   const prereleaseTitle = prerelease && prerelease.charAt(0).toUpperCase() + prerelease.slice(1);
-  const camelCaseName = getCamelCaseName(name);
   const pascalCaseName = getPascalCaseName(name);
```
alanbsmith, line 35: I'm chuckling that PascalCase is written in camelCase. 🤓

utils/create-component/templates/react/component.js
```diff
@@ -1,37 +1,33 @@
...
+import {${pascalCaseName}Target} from './${pascalCaseName}Target';
+import {${pascalCaseName}Content} from './${pascalCaseName}Content';
 
-export const ${pascalCaseName}ModelContext = React.createContext<${pascalCaseName}Model>({} as any);
-
-export interface ${pascalCaseName}Props extends ${pascalCaseName}ModelConfig {
-  model?: ${pascalCaseName}Model;
+export interface ${pascalCaseName}Props extends ExtractProps<typeof Box, never> {
+  /**
+   * Children of the ${titleCaseName}. Should contain a \`<${pascalCaseName}.Target>\`, a \`<${pascalCaseName}.Content>\`
```
alanbsmith, line 13: Q: I thought comment blocks were ignored? 🤔

alanbsmith (dismissed):
Thanks for this, @vibdev! I left a couple questions for you, but nothing that would hold this up.

modules/preview-react/index.ts
vibdev: yoloooooo

utils/create-component/templates/react/component.js
vibdev: Screen Shot 2022-08-10 at 5 49 12 PM

 Looks like comments generate correctly

utils/create-component/createReactComponent.js
vibdev: I just thought it looked nicer. The blue was kinda harsh :D 
Screen Shot 2022-08-10 at 5 50 22 PM

utils/create-component/templates/react/stories.mdx.js
```diff
@@ -0,0 +1,35 @@
...
+## Installation
+
+\`\`\`sh
+yarn add @workday/canvas-kit-${prerelease && prerelease + '-'}react
+\`\`\`
+
+
+## Usage
+
+### Basic
```
jamesfan, line 24: Thank you for updating this script to generate a starter MDX! Small nit to keep things consistent with our [standard MDX structure](https://github.com/Workday/canvas-kit/discussions/1131):

```suggestion
### Basic Example
```

Would love to build on this in a future PR to create a more complete starter document.

utils/create-component/createReactComponent.js
vibdev: Screen Shot 2022-08-10 at 5 51 37 PM

utils/create-component/templates/react/stories.mdx.js
vibdev: Sure thing. Totally into the idea of building on this. IMO it is one of the best ways to learn how to create really good components. Sometimes it is hard to figure out which component you should model after.

vibdev: Would love to add cypress to the mix too

modules/preview-react/index.ts
jamesfan: I checked the blame and @NicholasBoll added an explanation for why this line exists: https://github.com/Workday/canvas-kit/pull/1408/files#r779218020

vibdev: interesting, thanks for looking that up.  I wasn't getting any type errors running storybook, but could have missed something.

NicholasBoll: If there are no type errors, I'm glad to have it removed.

NicholasBoll (approved):
I've verified the triple slash directive is no longer needed.

Issue #1704, opened 2022-07-28
`get{First|Last}FocusableElement` doesn't account for tabindex=-1


NicholasBoll:
## 🐛 Bug Report

`useFocusRedirect` is using `getFirstFocusableElement` and `getLastFocusableElement` which use `isFocusableElement` which considers `button[tabindex=-1]` as a focusable element. It should ignore `tabindex=-1` which takes elements out of the focus order.

The problem is in here:

https://github.com/Workday/canvas-kit/blob/74de1cca8aec50b5299784480551a3b49d5c3e30/modules/react/common/lib/utils/elements.ts#L15-L22

The logic considers all `button` elements to be focusable, but doesn't take `tabindex=-1` into consideration. Those lines could be changed to the following to fix it:

```ts
  const validNativelyFocusable =
    ['button', 'details', 'embed', 'iframe', 'select', 'textarea'].includes(nodeName) &&
    element.getAttribute('tabindex') !== '-1';
```

## To Reproduce

1. Go to https://workday.github.io/canvas-kit/?path=/story/components-popups-menu-react--basic
2. Click "Open Menu"
3. Hit the tab key

## Expected Behavior

The menu should close. It will close only if the cursor is on the last menu item (you can use the down arrow until the last menu item is focused, then hit tab)

## Link to repl or repo (highly encouraged)

[Please provide a reference to a repo or branch to help us reproduce this bug.](https://workday.github.io/canvas-kit/?path=/story/components-popups-menu-react--basic)

chaiwattsw:
Can I take this issue?

willklein:
@chaiwattsw Thanks for asking! We have someone already taking a look. Thanks.

PR #1723, opened 2022-08-10
fix: Ignore tabIndex to take elements out of the focus order

smopur:
## Summary

Fixes: #1704  

category

### Release Note
Fixes tabbing out of a a control where elements in the control are excluded from the tab order


mannycarrera4 (approved):
Thanks for contributing this!!

Issue #1682, opened 2022-07-14
Buttons placed within `Popup.Body` are getting their focus rings cut off


derek-h-kim:
## To Reproduce

CK@7.0.14
Place any CK button (Primary/Secondary/Tertiary) within a `Popup.Body`. Focus the button.

## Expected Behavior

The focus ring should not get cut off by the Popup's padding.

## Actual Results

The focus ring is cut off on the sides that are too close to the Popup's borders. Fix should also work for buttons that cannot be moved outside of `Popup.Body`

## Browser (if applicable)

Edge and Chrome on Mac

## Link to repl or repo (highly encouraged)

https://workday.github.io/canvas-kit/?path=/docs/components-popups-dialog-react--basic
https://workday.github.io/canvas-kit/?path=/docs/components-popups-toast-react--error#with-an-action-link

### Error Output

![image](https://user-images.githubusercontent.com/104469540/179035933-4c4d9224-1089-4564-a979-d23833893bc6.png)
![image](https://user-images.githubusercontent.com/104469540/179035999-fd67c0ff-50de-42a2-b7a7-2fc4e4216e86.png)

mannycarrera4:
Not sure yet but based on this issue, it's not encouraged to have an element break the stacking context: https://github.com/w3c/csswg-drafts/issues/3417

Still need to dig into this

derek-h-kim:
FYI @mannycarrera4 , there are use cases where buttons might not want to go at a footer (i.e. 2nd screenshot). Some popups will have buttons that are placed on the left or right borders of the Popup which wouldn't make sense to move to a `Modal.Footer` or equivalent

PR #1621, opened 2022-06-01
docs: Update Modal examples

emroller16:
## Summary

Fixes: #1620

Currently, our Modal Documentation page on Canvas is lacking documentation and just shows examples from Storybook. Since all the examples we have feature a "Delete" action, the page could appear to have errors, as referenced in our support channel.

I'm proposing that we could improve the examples to show more neutral dialogs, launched by a primary button, with neutral text such as "Open Basic Modal." We could include a dedicated "Delete Item" example as well, if we wanted to show that example specifically.

category

---

## Checklist

- [ ] code has been documented
- [ ] design approved final implementation
- [ ] a11y approved final implementation
- [ ] code adheres to the [API & Pattern guidelines](https://workday.github.io/canvas-kit/?path=/story/welcome-dev-docs-api-pattern-guidelines--page)

PR #1727, opened 2022-08-11
fix: Prevent focus ring from getting cut off in Popup.Body

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/1682

category

### Release Note

- `Popup.Body` now has `8px` of padding. This addresses an issue introduced in v7 where focus rings were being clipped within the Body due to its built-in overflow; this padding allows the focus ring on any focusable element contained within the Body to be fully visible. The default space between the edge of the Card and the contents of the Body is still `32px`.
- `Popup.Heading` now has `8px` of padding to keep it visually aligned with the updated `Popup.Body`. It also now has a `marginBottom` of `8px`, bringing the default vertical space between the Heading and the Body to `24px` (previously `16px`) in order to meet the component spec.
- The padding in `Popup.Card` has been reduced by `8px` from `32px` to `24px` to compensate for the padding added to `Popup.Body` and `Popup.Heading` and to preserve the component's existing spacing. If you've customized the padding for any `Popup`, `Modal`, or `Dialog` subcomponents, you may need to adjust your padding values to get the same results after upgrading.
- `Popup.Card`'s default element is now a `Flex` element (previously a `Stack`) and thus the `spacing` prop has been removed. It made more sense to use `Flex` given the presence of the Close Icon, which didn't make sense as a `Stack` item.
- Given the changes above, if you're creating a footer for your `Popup`, `Modal`, or `Dialog` (using an `HStack`, for instance) you'll need to update it with the following props/styles to align it with the updated Heading and Body: `padding="xxs" (8px)` and `marginTop="xxs" (8px)`.

---

Screen Shot 2022-09-08 at 19 49 50


modules/react/modal/stories/examples/CustomFocus.tsx
```diff
@@ -12,31 +12,31 @@ export const CustomFocus = () => {
...
       <Modal.Overlay>
         <Modal.Card>
           <Modal.CloseIcon aria-label="Close" />
-          <Modal.Heading>Delete Item</Modal.Heading>
-          <Modal.Body>
+          <Modal.Heading>Modal Heading</Modal.Heading>
+          <Modal.Body paddingBottom="zero">
             <Box as="p" marginTop={0} marginBottom="m">
               Enter name to confirm deletion
             </Box>
```
RayRedGoose, line 29: I'm curious why did you change heading and button texts? New heading text is not connected to content it has. Maybe text should be abstract text as you had for heading and buttons, such it has in CustomTarget story.

mannycarrera4: Yeah, i was just trying to go off of this https://chakra-ui.com/docs/components/modal/usage#usage which is very generic

mannycarrera4: I can add some lorem ipsum!

RayRedGoose: Also, it seems like @emroller16 did something close to this: https://github.com/Workday/canvas-kit/pull/1621/files

mannycarrera4: Yeah, I followed up with her, she's on PTO, if she wants to commit that change, I'll go ahead and keep the stories as they where and let her make the change!

RayRedGoose: Awesome! 👍

josh-bagwell:
Manny and I discussed alternative approaches. One option was to move the buttons outside the `Modal.Body` component and that fixes the issue. However, it doesn't address the issue of links/buttons losing focus ring in `Modal.Body`. It looks as if the `overflow-y` is causing the focus ring issue.

mannycarrera4:
### do not merged until `7.3` is released

modules/react/modal/lib/ModalBody.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+import {createSubcomponent, ExtractProps} from '@workday/canvas-kit-react/common';
+import {Card} from '@workday/canvas-kit-react/card';
+import {Popup, usePopupModel} from '@workday/canvas-kit-react/popup';
+
+export const ModalBody = createSubcomponent('div')({
+  displayName: 'Modal.Body',
+  modelHook: usePopupModel,
+})<ExtractProps<typeof Card.Body>>(elemProps => {
+  return <Popup.Body overflowY="auto" padding={'l'} paddingTop="zero" {...elemProps} />;
```
jamesfan, line 11: I noticed the vertical spacing of a "default" Modal looks a bit off:

default-modal

Note the large vertical space between the Modal Body and the Modal Footer (`48px` vs `24px` in the spec). Even the vertical space between the Modal Heading and Modal Body is smaller than it should be (`16px` vs `24px` in the spec) -- this issue existed even before this PR.

I'm using the basic version of the Modal to render this example with almost no props applied to any of the subcomponents:

```
<Modal.Card>
  <Modal.CloseIcon aria-label="Close" />
  <Modal.Heading>Test</Modal.Heading>
  <Modal.Body>
    This is a test.
  </Modal.Body>
  <Modal.Footer>
    <Modal.CloseButton as={DeleteButton}>
      Delete
    </Modal.CloseButton>
    <Modal.CloseButton>Cancel</Modal.CloseButton>
  </Modal.Footer>
</Modal.Card>
```

What do you think about adjusting the default padding of the Modal subcomponents to match the `24px` of vertical spacing between the Header, Body, and Footer from the spec? I see there's a bottom padding on ModalBody Popup.Body, but that the top padding has been set to `zero`. Seems like we should add top padding to prevent content at the top of the Body from being cut off as well? We could add `paddingY="m"` to the ModalBody and remove all other vertical padding/margin from the subcomponents.

modules/react/modal/stories/stories_testing.tsx
```diff
@@ -101,17 +101,17 @@ export const WithRadioButtons = () => {
...
-                name="contact"
-                data-testid="radiogroup"
-                value={value}
-                onChange={value => setValue(String(value))}
-              >
-                <Radio id="1" value="email" label="E-mail" />
-                <Radio id="2" value="phone" label="Phone" />
-              </RadioGroup>
-            </VStack>
+            {/* <VStack spacing="s"> */}
```
jamesfan, line 104: This commented out `VStack` should be removed (or added back in).

modules/react/popup/lib/PopupCard.tsx
```diff
@@ -64,7 +64,7 @@ export const PopupCard = createSubcomponent('div')({
   }, [model.state.placement]);
 
   // As is a Stack that will render an element of `Element`
-  const As = useConstant(() => Stack.as(Element));
+  const As = useConstant(() => Flex.as(Element));
```
jamesfan, line 67: I noticed your comment about this in the PR summary:

> We've changed the base element on Popup.Card to use Flex instead of Stack to prevent breaking changes. If you were using spacing on that component, you'll have to use FlexStyleProps instead.

Can you clarify why this change was made? Which breaking changes are you referring to?

modules/react/modal/lib/ModalBody.tsx
mannycarrera4: That's perfect, thanks for that suggestion James!

modules/react/popup/lib/PopupCard.tsx
jamesfan: Just noticed this bit from the [7.4.0 Announcement](https://canvas.workday.com/whats-new/announcements#canvas-kit-740-release):

> As part of the Modal focus ring bug fix, each section (Heading, Body, and Footer) was separated to have their own padding and margins. Because of this change, the Stack in Modal.Card removed the margin top for the first element (Heading) in the Modal. As a result, the underlying Stack element for Modal.Card has also been replaced with the Flex element to add the 24px margin top back to the Modal Heading. Teams using the spacing prop on Modal.Card will need to remove it. If vertical alignment is required, teams should implement a VStack element.

mannycarrera4: Yeah i would love some word smithing for this. What happened was because i moved margin to the heading element, `Stack` has a selector that was overriding this margin and setting it to `marginTop: 0` so I had to change this to use flex stack so that it would avoid this.

jamesfan:
I chatted with @mannycarrera4 about this, and IMO it makes sense to apply these fixes to `Popup` instead of `Modal`. As it currently stands, this PR will fix the cut-off focus rings in `Modal`, but not in `Popup` or anything else which uses `Popup` (such as `Toast`). In order to address the original issue (#1682 which itself references the `Popup` and `Toast` components rather than `Modal`), we should fix the issue once at the source in `Popup` which will then propagate the fix to all other Popups (`Modal`, `Toast`, etc).

modules/react/popup/lib/PopupCard.tsx
```diff
@@ -71,12 +70,11 @@ export const PopupCard = createSubcomponent('div')({
       as={As}
       transformOrigin={transformOrigin}
       position="relative"
-      padding="l"
       depth={5}
       maxWidth={`calc(100vw - ${space.l})`}
-      spacing={0}
```
mannycarrera4, line 77: We can keep the underlying element a `Stack` however, if we keep this spacing="0" this will remove marginTop on the first element in the stack. I suggest we either remove this prop or change the underlying element to be `Flex`

modules/react/popup/lib/PopupCard.tsx
```diff
@@ -71,12 +70,11 @@ export const PopupCard = createSubcomponent('div')({
...
       flexDirection="column"
       minHeight={0}
+      padding="zero"
```
mannycarrera4, line 77: Move the padding from the `Card` to `Popup.Body` and margin to `Popup.Heading`

modules/react/popup/stories/examples/Basic.tsx
```diff
@@ -27,11 +27,13 @@ export const Basic = () => {
     <Popup model={model}>
       <Popup.Target as={DeleteButton}>Delete Item</Popup.Target>
       <Popup.Popper placement="bottom">
-        <Popup.Card width={400} padding="s">
```
mannycarrera4, line 30: I removed this padding, to keep the same basic  styles at Modal

modules/react/modal/stories/examples/Basic.tsx
```diff
@@ -17,7 +17,7 @@ export const Basic = () => {
           <Modal.CloseIcon aria-label="Close" />
           <Modal.Heading>Delete Item</Modal.Heading>
           <Modal.Body>
-            <Box as="p" marginTop={0} marginBottom="m">
+            <Box as="p" marginTop="zero" marginBottom="m">
```
mannycarrera4, line 20: consistency

modules/react/popup/lib/PopupBody.tsx
```diff
@@ -8,5 +8,7 @@ export const PopupBody = createSubcomponent('div')({
   displayName: 'Popup.Body',
   modelHook: usePopupModel,
 })<ExtractProps<typeof Card.Body>>(elemProps => {
-  return <Card.Body overflowY="auto" {...elemProps} />;
+  return (
+    <Card.Body overflowY="auto" paddingX={'l'} paddingTop="s" paddingBottom="l" {...elemProps} />
```
mannycarrera4, line 12: This is the main change, I removed the padding from `Popup.Card` and added it to the Body so that the focus ring is visible.

jamesfan: Hmm, I'm seeing issues with our approach of adding vertical padding to the Body. Go to the Modal Body Overflow story and adjust the frame for the story to trigger overflow on the Body. Scroll down the Body and note how the content scrolls out of view at the top of the Body right below the Heading:

overflow-scrolling

modules/react/modal/stories/examples/BodyOverflow.tsx
```diff
@@ -43,7 +42,7 @@ export const BodyOverflow = () => {
             <p>Are you sure you want to delete the item?</p>
             <p>Are you sure you want to delete the item?</p>
           </Modal.Body>
-          <HStack spacing="s" paddingTop="s">
+          <HStack spacing="s" paddingTop="s" paddingX="l" paddingBottom="l">
```
jamesfan, line 45: Now that we've removed `Modal.Footer`, it feels odd to me to require consumers to know they need `paddingX="l"` and `paddingBottom="l"` to get their `HStack` Footer to render with complementary padding to match `Modal.Body` and `Modal.Heading`. Previously, this was padding was baked into `Modal.Footer`.

mannycarrera4: if we make this a Minor release, it makes sense to add this back in. But I agree the consumer would have to know what padding to add

modules/react/modal/lib/ModalBody.tsx
jamesfan: We're going with Alan's approach instead (see top-level PR comment).

modules/react/popup/lib/PopupCard.tsx
jamesfan: We're changing PopupCard to use `Flex` instead of `Stack`. `Stack` doesn't work conceptually given the presence of the Popup.CloseIcon within the container (CloseIcon doesn't make sense as a `Stack` child).

modules/react/popup/lib/PopupBody.tsx
jamesfan: We've decided a small reduction in the vertical spacing between the Heading and Body as you scroll down the Body is acceptable (up to 8px reduction given the 8px of padding in Popup.Body, see Alan's approach in the top-level PR comment).

modules/react/modal/stories/examples/BodyOverflow.tsx
jamesfan: For now, the consumer will need to add their own footer (using an `HStack`, for instance) with the appropriate paddings and margins to align with the Heading and Body subcomponents (see screenshot in top-level PR comment). We'll update Popup, Modal, and Dialog examples to include example footers with the needed padding and margin.

modules/react/modal/stories/examples/BodyOverflow.tsx
```diff
@@ -18,8 +17,10 @@ export const BodyOverflow = () => {
           <Modal.CloseIcon aria-label="Close" />
           <Modal.Heading>Delete Item</Modal.Heading>
           <Modal.Body>
+            <PrimaryButton>FOCUSSSSSSSSSS</PrimaryButton>
```
jamesfan, line 20: Remove the test `PrimaryButton`s here (there are a few of them) and the `PrimaryButton` import at the top.

modules/react/popup/lib/PopupCard.tsx
```diff
@@ -64,19 +65,18 @@ export const PopupCard = createSubcomponent('div')({
   }, [model.state.placement]);
 
   // As is a Stack that will render an element of `Element`
```
jamesfan, line 67:
```suggestion
  // As is a Flex that will render an element of `Element`
```

jamesfan (approved):

Issue #1720, opened 2022-08-10
Link to non-interactive examples don't work


willklein:
## 🐛 Bug Report

[See the specifications table for Tooltip](https://workday.github.io/canvas-kit/?path=/docs/components-popups-tooltip-react--default-story#specifications). The links for "NonInteractive" examples don't work. The link text should read "Non-Interactive" and the URL should have a hyphen between non and interactive. Currently it is:
`[NonInteractive](https://workday.github.io/canvas-kit/?path=/story/testing-react-popups-tooltip--noninteractive)`

and it should be:

`[Non-Interactive](https://workday.github.io/canvas-kit/?path=/story/testing-react-popups-tooltip--non-interactive)`

## Expected Behavior

This page should work from the link.

<img width="1268" alt="Screen Shot 2022-08-10 at 11 10 42 AM" src="https://user-images.githubusercontent.com/1075861/183974814-ff533156-17fb-4669-a7a3-29ec1b969b41.png">

## Actual Results

Instead this is what is loading.

<img width="1271" alt="Screen Shot 2022-08-10 at 11 10 32 AM" src="https://user-images.githubusercontent.com/1075861/183974861-6e2643a0-f575-4270-a1d7-5ab3500da5b3.png">

PR #1731, opened 2022-08-15
docs: Change link URL to non-interactive examples

stefano-puzzuoli:
## Summary

Fixes: #1720 

The links for "NonInteractive" examples didn't work.  Before this change the link read:
https://workday.github.io/canvas-kit/?path=/story/testing-react-popups-tooltip--noninteractive 

but now with the fix, we have the correct link:

https://workday.github.io/canvas-kit/?path=/story/testing-react-popups-tooltip--non-interactive 

category

### DIFF
--- a/cypress/integration/Tooltip.spec.ts
+++ b/cypress/integration/Tooltip.spec.ts
@@ -279,9 +279,9 @@ describe('Tooltip', () => {
     );
   });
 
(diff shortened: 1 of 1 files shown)


alanbsmith (approved):
Looks good! Thanks for the fix, @stefano-puzzuoli! 🏆

PR #1735, opened 2022-08-18
fix: Remove alt text from Expandable Avatar and make element a div

mannycarrera4:
## Summary

category
---


willklein:
Changes implemented in #1739.

PR #1736, opened 2022-08-19
fix: Update accessibility on avatar in expandable component

mannycarrera4:
## Summary

category


modules/labs-react/expandable/lib/ExpandableAvatar.tsx
```diff
@@ -19,9 +15,12 @@ const StyledAvatar = styled(Avatar)<StyledType>({
...
+// When the component is created, it needs to be a button element to match AvatarProps.
+// Once Avatar becomes a `createComponent` we can default the element type to a `div`
+// and the types should be properly extracted
 export const ExpandableAvatar = createComponent('button')({
   displayName: 'Expandable.Avatar',
-  Component: ({...elemProps}: ExpandableAvatarProps, ref, Element) => {
-    return <StyledAvatar altText={undefined} as={Element} ref={ref} size={32} {...elemProps} />;
+  Component: ({altText, ...elemProps}: ExpandableAvatarProps, ref) => {
+    return <StyledAvatar altText={undefined} as={'div'} ref={ref} size={32} {...elemProps} />;
   },
```
RayRedGoose, line 25: Interesting thing: Avatar component isn't built by our `createComponent` util function, so this component has some limitation on `as` prop. For example, with other components we can easily modify element in styled function using `as` method:

```
const StyledActionBarText = styled(Box.as('span'))<StyledType>({
  textDecoration: 'underline',
})
```

`Avatar` component doesn't have this possibility 🤔 Also, our create component functions have a lot of typescript validation under the hood, and they are not exists in Avatar component because of it.

Should we review this moment and refactor `Avatar` component (an other components, if they exist) to use util functions?

mannycarrera4: Good catch Raisa! Yeah the fact that Avatar is not a createComponent is hindering us here :( We can refactor this now, I don't think it should be a breaking change but it would greatly help us. Let me write up an issue to track this and maybe we can refactor it!

mannycarrera4: Here's the issue: https://github.com/Workday/canvas-kit/issues/1738

RayRedGoose: yeah, Avatar is a main component, so we should be careful with making changes in it. 
it shouldn't be any change of functionality for users, just inner refactoring so we are probably good to do it.

willklein:
Changes implemented in #1739.

PR #1739, opened 2022-08-22
fix: Update avatar on expandable component

mannycarrera4:
## Summary

category

### DIFF
--- a/modules/labs-react/expandable/lib/ExpandableAvatar.tsx
+++ b/modules/labs-react/expandable/lib/ExpandableAvatar.tsx
@@ -21,6 +21,6 @@ const StyledAvatar = styled(Avatar)<StyledType>({
 export const ExpandableAvatar = createComponent('button')({
   displayName: 'Expandable.Avatar',
   Component: ({altText, ...elemProps}: ExpandableAvatarProps, ref) => {
-    return <StyledAvatar altText={undefined} as={'div'} ref={ref} size={32} {...elemProps} />;
+    return <StyledAvatar altText={undefined} as="div" ref={ref} size={32} {...elemProps} />;
   },
 });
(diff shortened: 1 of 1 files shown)


modules/labs-react/expandable/lib/ExpandableAvatar.tsx
```diff
@@ -19,9 +15,12 @@ const StyledAvatar = styled(Avatar)<StyledType>({
   flexShrink: 0,
 });
 
+// When the component is created, it needs to be a button element to match AvatarProps.
+// Once Avatar becomes a `createComponent` we can default the element type to a `div`
+// and the types should be properly extracted
```
jamesfan, line 20: I appreciate the explanation on this.

modules/labs-react/expandable/lib/ExpandableAvatar.tsx
```diff
@@ -19,9 +15,12 @@ const StyledAvatar = styled(Avatar)<StyledType>({
...
 export const ExpandableAvatar = createComponent('button')({
   displayName: 'Expandable.Avatar',
-  Component: ({...elemProps}: ExpandableAvatarProps, ref, Element) => {
-    return <StyledAvatar altText={undefined} as={Element} ref={ref} size={32} {...elemProps} />;
+  Component: ({altText, ...elemProps}: ExpandableAvatarProps, ref) => {
+    return <StyledAvatar altText={undefined} as={'div'} ref={ref} size={32} {...elemProps} />;
```
jamesfan, line 24: Nit: to save us a couple unnecessary characters.
```suggestion
    return <StyledAvatar altText={undefined} as="div" ref={ref} size={32} {...elemProps} />;
```

jamesfan (approved):
Left a small code style nit below. Looks good otherwise 👍

Issue #1743, opened 2022-08-26
Main Menu.Item.Icon changes color on hover on active state


alanbsmith:
## 🐛 Bug Report

When a `Menu.Item` is active, and you hover over the icon, the icon color changes from white to its default hover color.

## To Reproduce

* Open a menu
* Hover over an icon on an active menu item
* See the color change

## Expected Behavior

The icon color should remain white on hover.

## Actual Results

The icon color changes from white to its default hover color.

## Link to repl or repo (highly encouraged)

[Sandbox Example](https://codesandbox.io/s/menu-item-icon-issue-v8-ntuomy?file=/src/App.tsx)

### Error Output

![a gif showing the icon behavior](http://g.recordit.co/9NNvLD5BpN.gif)

PR #1744, opened 2022-08-26
fix(menu): Fix issue with color change on menu item icon hover

RayRedGoose:
## Summary

Fixes: #1743 

This PR fixes bug with remaining hover behavior in `Menu.Item.Icon`.

category

### DIFF
--- a/modules/react/menu/lib/MenuItem.tsx
+++ b/modules/react/menu/lib/MenuItem.tsx
@@ -73,6 +73,7 @@ const StyledItem = styled(Box.as('button'))<StyledType>(
         outline: 'none',
         backgroundColor: theme.canvas.palette.primary.main,
         color: typeColors.inverse,
+        fill: iconColors.inverse,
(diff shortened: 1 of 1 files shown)


mannycarrera4:
Looks like this is also existing on master already, but can you play with different icons
Screen Shot 2022-08-26 at 11 03 03 AM
!

RayRedGoose:
Oh, this defined behavior. This icon has accent color so it shouldn't be changed

RayRedGoose:
The latest commit fixes the next problems with icon color:

- icons with accent part has blue color of accent parts (now the whole icon has the same color for every state)
- icon color changes happening on hover of disabled button (now the whole icon has disabled color even in hover or focus)

alanbsmith (approved):
Confirmed on the latest SB build that this works. Thanks for the quick fix, @RayRedGoose!

svagi (approved):
Thanks for the fix! 👍

Issue #1761, opened 2022-09-01
Add a custom lint rule to catch imports from Canvas Kit imports from base exports


alanbsmith:
## 💡 Suggestion

Add a custom lint rule to catch Canvas Kit imports statements from the root export. For example:

```ts
// instead of this
import { Card, PrimaryButton } from "@workday/canvas-kit-react";

// encourage this
import { Card } from "@workday/canvas-kit-react/card";
import { PrimaryButton } from "@workday/canvas-kit-react/button";
```

## Motivation

Inconsistent imports have can create [obscure errors in the build process](https://github.com/Workday/canvas-kit/runs/8122454770?check_suite_focus=true#step:18:110) when building type definitions. We can prevent this by catching these issues in our PR checks to make sure our imports don't create problems for us later.

PR #1762, opened 2022-09-01
chore: Add custom lint rule to encourage CK slash imports

alanbsmith:
## Summary

Fixes: #1761

category

---

## For the Reviewer

This PR:

- Adds a custom lint rule for check CK import statements for slash imports
  - This rule only applies to `canvas-kit-react`, `canvas-kit-preview-react`, and `canvas-kit-labs-react`
  - `canvas-kit-popup-stack`, `canvas-kit-codemod`, `canvas-kit-docs`, and `canvas-kit-css-*` are excluded
- Adds tests to verify the rule works
- Updates CK import statements that are not using slash imports

## Areas for Feedback?

I would like to know whether the error messages make sense, or if they should be reworded a bit.

`utils/custom-lint-rules/use-ck-slash-imports.js`

## Testing Manually

The only way to test this manually would be:
-  pull the branch down
- update a CK import to use the base package
- run `yarn lint` and see the test fail

But the test I added is probably sufficient: `utils/custom-lint-rules/use-ck-slash-imports.spec.ts`. If it passes the lint rule works as expected.

## Thank You Gif (optional)

a happy pack of dachshunds running through a field wearing hotdog costumes


jaclynjessup (approved):

modules/labs-react/combobox/stories/stories.tsx
```diff
@@ -1,7 +1,11 @@
 import React, {useState, ReactNode, ReactElement, FC, ChangeEvent} from 'react';
 import {storiesOf} from '@storybook/react';
 
-import {Combobox, ComboboxProps, ComboBoxMenuItemGroup} from '@workday/canvas-kit-labs-react';
```
mannycarrera4, line 4: how did this not blow up before... lol

utils/custom-lint-rules/use-ck-slash-imports.js
```diff
@@ -0,0 +1,52 @@
...
+            return specifier.local.name;
+          }
+          if (specifier.type === 'ImportSpecifier') {
+            return specifier.imported.name;
+          }
+        });
+        const importSourceRegExp = /^@workday\/canvas-kit-(labs-|preview-)*react*$/;
+        if (importSourceRegExp.test(value)) {
+          const importsInEnglish = formatter.format(specifierNames);
+          const multipleSpecifierMessage = `Import ${importsInEnglish} from their Canvas Kit subdirectories instead of ${value}`;
```
mannycarrera4, line 42: I'm channeling my inner Nicholas. I feel like the error message should also give you a way to solve this. 
`Import PrimaryButton from their Canvas Kit subdirectories. Example Import {PrimaryButton} from "@workday/canvas-kit-react" should be import {PrimaryButton} from "@workday/canvas-kit-react/button""`

I know this is super verbose but this leave no doubt as to what you have to do to fix the error?

modules/labs-react/combobox/stories/stories.tsx
alanbsmith: The difference is that imports in stories don't matter. These don't have to change, but I made the lint rule catch them anyway.

utils/custom-lint-rules/use-ck-slash-imports.js
alanbsmith: Yeah, I thought about that too. We would need to have an up-to-date map of where every import comes from. The closest thing we have is the V5 codemod, but that is now out of date.

alanbsmith: If there was a way to know what subdirectory to pull from, I'd happily add that.

mannycarrera4: yeah that's fair, I was just thinking of an example text and not one that represents the import you're fixing which was hacky. Either way, big win

mannycarrera4 (approved):

Issue #1694, opened 2022-07-21
Create an initial Maintaining doc


alanbsmith:
## 💡 Suggestion

We have a contributing doc but no maintaining doc. Let's fix that. 😄

## Motivation

There's a lot of implicit knowledge in our heads about how to maintain Canvas Kit. But that knowledge isn't accessible or easily transferable. Let's make a doc so anyone on the team can access information and feel confident in their ability to maintain this library.

Some initial topics:

- [ ] Branches and versioning
- [ ] Create a maintaining doc
- [ ] CI / CD processes / GH Actions

PR #1764, opened 2022-09-02
docs: Add initial maintaining doc

alanbsmith:
## Summary

Fixes: #1694

category

---

## For the Reviewer

This PR adds an initial maintaining doc with information about:

- branches and versions
- GitHub Actions
- Releases

`modules/docs/mdx/MAINTAINING.mdx`

- [ ] Code
- [x] Documentation
- [ ] Testing
- [ ] Codemods

I really want to know if this makes sense to other maintainers. Feel free to edit and ask questions! I also have this living in our MDX docs, but it could live elsewhere. Let me know if you'd like to go somewhere else!

## Testing Manually

I would review the GH diff and toggle the "display source code" and "display rich diff" button.

## Thank You Gif (optional)

Thanks for taking a look!

a cat with a bowtie and glasses intently watching a laptop screen


modules/docs/mdx/MAINTAINING.mdx
```diff
@@ -0,0 +1,250 @@
...
+maintain Canvas Kit with confidence. If you see some information that's unclear, incomplete, or
+incorrect, please update this doc to help your future self and others. Thanks for maintaining Canvas
+Kit!
+
+- [Branches](#branches)
+- [GitHub Actions](#github-actions)
+  - [Forward Merge Workflow](#forward-merge-workflow)
+  - [Release Minor Workflow](#release-minor-workflow)
+  - [Release Workflow](#release-workflow)
+- [Releases](#releases)
```
alanbsmith, line 17: One question I have: Should these be separate docs? Should we have a standalone "Releasing" doc?

jaclynjessup: I think that having it live here is beneficial. Since releasing would only fall to this audience. Unless you think that the release doc is too large and will just create more noise than good within this doc

willklein: Let me think about this... probably good here, but I will say contributors and consumers can be interested in how code gets shipped at the right level of detail for that audience. For example, if code is merged, when does it ship?

alanbsmith: Yeah, I think we could have an abbreviated section for consumers on our release cadence.

alanbsmith: Let's see how long this doc gets and then parse it out. I'm okay with it living here for now.

jaclynjessup (approved):

Issue #1756, opened 2022-08-31
`SegmentedControlButton` doesn't have possibility to use size prop correctly


RayRedGoose:
## 🐛 Bug Report

`size` prop causes ts issue because it's not in `SegmentedControlButtonProps` interface.
Using `size` prop doesn't change icon size.

## To Reproduce

- SegmentedControl with SegmentedControlButton inside
- Add `size` prop to `SegmentedControlButton`
- See TS warning
- See that button size changed, icon remains 24px in width&height

## Expected Behavior

`SegmentedControlButton` shouldn't show ts issue for `size` prop and icon size should be changed based on button size.

## Actual Results

[Reproduced in Sandbox](https://codesandbox.io/s/v7-segmented-control-forked-kfhmmm?file=/src/App.tsx)

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

## Link to repl or repo (highly encouraged)

[Sandbox](https://codesandbox.io/s/v7-segmented-control-forked-kfhmmm?file=/src/App.tsx)

PR #1766, opened 2022-09-06
fix: Enable to use size prop in SegmentedControl.Button component

RayRedGoose:
## Summary

Fixes: #1756  

`SegmentedControl.Button` can have `size` prop to modify button and and icon size with a `small`, `medium` or `large` value.

category


modules/react/segmented-control/lib/SegmentedControlButton.tsx
```diff
@@ -79,13 +94,17 @@ const StyledButton = styled(BaseButton)<StyledType>(
...
       },
     },
+  }),
+  ({size}: ButtonContainerProps) => ({
+    ...getSizeStyles(size),
   })
 );
 
-export interface SegmentedControlButtonProps {
+export interface SegmentedControlButtonProps extends ButtonContainerProps {
```
mannycarrera4, line 103: yassssss

modules/react/segmented-control/lib/SegmentedControlButton.tsx
```diff
@@ -35,12 +40,22 @@ const getIconButtonColors = (toggled?: boolean): ButtonColors => {
   };
 };
 
-const StyledButton = styled(BaseButton)<StyledType>(
+const getSizeStyles = (size?: ButtonSizes) => {
```
mannycarrera4, line 43: don't we also have `extraSmall` ?

mannycarrera4 (approved):
thank you for fixing this!

modules/react/segmented-control/lib/SegmentedControlButton.tsx
RayRedGoose: no, SegmentedControl has only 3 size

Issue #1631, opened 2022-06-08
Deprecate IE11


myvuuu:
## 💡 Suggestion

IE11 support will be dropping in September 2022 and will need to be deprecated from Canvas Kit by:

- [x] Removing IE specific code
- [x] Update documentation to indicate that IE11 is no longer support

mannycarrera4:
I think we can just uncheck IE from chromatic
![Screen Shot 2022-08-18 at 8 54 57 AM](https://user-images.githubusercontent.com/7966550/185427024-147d5930-2969-458d-977d-83a94b632e6f.png)

mannycarrera4:
Update supported browser list https://github.com/Workday/canvas-kit#supported-browsers

mannycarrera4:
Look into if we still need whatinput and canvas provider

PR #1785, opened 2022-09-19
chore: Remove IE11 specific code

mannycarrera4:
## Summary

Fixes: #1631

category

### BREAKING CHANGES
We've removed any IE11 specific code including any polyfills. This means your application will no longer run in IE11 and fail to bootstrap with an error and a white screen. Do not upgrade to this version if your application needs to support IE11.


modules/react/tooltip/lib/OverflowTooltip.tsx
```diff
@@ -41,7 +41,8 @@ export const findEllipsisElement = (element: Element): Element | null => {
   if (style.textOverflow === 'ellipsis' || Number(style.webkitLineClamp) > 0) {
     return element;
   } else if (element.children) {
-    // `children` is not defined for SVGElement in IE11
+    // `children` is not defined for SVGElement in IE11 Do we still need this?
+
```
mannycarrera4, line 45: So I tried to remove this but that broke, any ideas if this has to be rewritten or kept?

NicholasBoll: I don't fully remember the context, but I think it was the difference of the if statement:

```ts
} else {
```

I think that gave an error in IE because `element.children` is not defined for SVG elements in IE, so this `else` became a type guard instead. The type definitions say `children` is always available on all `Element` interfaces (`SVGElement` extends `Element`). The `} else if (element.children) {` simply guarded against IE failing when an SVG element was encountered because SVG elements in IE don't properly extend `Element`s

NicholasBoll:
Does "not supporting IE11" mean intentionally breaking IE11 or no longer testing and fixing new issues related to IE11?

modules/react/tooltip/lib/OverflowTooltip.tsx
```diff
@@ -40,8 +40,7 @@ export const findEllipsisElement = (element: Element): Element | null => {
   const style = getComputedStyle(element);
   if (style.textOverflow === 'ellipsis' || Number(style.webkitLineClamp) > 0) {
     return element;
-  } else if (element.children) {
```
mannycarrera4, line 43: @NicholasBoll let me know if this change makes sense

modules/react/form-field/lib/Label.tsx
```diff
@@ -72,8 +72,6 @@ const RequiredAstrisk = styled('abbr')({
 });
 
 // Used inside the fieldset component instead of a label for accessible radio groups.
-// Extra `width: 100%` style is to help IE11 wrap text properly in a top-positioned
-// FormField's legend element.
 const LegendComponent = styled('legend')<LabelProps>(...labelStyles, ({labelPosition}) => {
   if (labelPosition === Label.Position.Top) {
```
willklein, line 76: Can this go away?

mannycarrera4: When this was removed, chromatic caught a regression in chrome, so I left it in

mannycarrera4:
We still use `element-closest` in our `InputProvider` should we drop some of that code?

NicholasBoll (approved):

Issue #1802, opened 2022-09-23
Buttons shouldn't use `style` to style anything because it will always override styles from consumers


mannycarrera4:
## 🐛 Bug Report

Because we use `style` to style border radius, consumers can't overwrite this.

In the [sandbox](https://codesandbox.io/s/bold-lehmann-wok2mv?file=/src/App.tsx) the border radius doesn't get overwritten. 

We should instead use `borderRadius` to update the border radius.

PR #1803, opened 2022-09-23
fix: Remove use of style to style border radius on TertiaryButton

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/1802

category

### Release Note
We were using `style` to apply border radius which prevent users from overwriting styles. Using `borderRadius` allows for customization.


jaclynjessup (approved):

RayRedGoose (approved):
I just have question about story fails, but otherwise look good!

Issue #1805, opened 2022-09-26
Tooltip - Tooltip doesn't reappear when the popup is closed while using useReturnFocus hook


clarehuang:
## 🐛 Bug Report

Noticed a use case with using the popup and tooltip components. When users close the popup and the focus is returned back to the popup trigger (ex. buttons), the tooltip doesn't reappear when the popup is closed. The tooltip gets re-opened only when the next-time focus happens. 

In addition, the hook, `useReturnFocus`, has been implemented in this case. From what I learned, the `onFocus` of the target isn't fired by the browser because this focus was part of another event or an unrelated target. 

## To Reproduce

- [Visit Codesandbox:](https://codesandbox.io/s/nameless-thunder-kuzsqi?file=/src/App.tsx:320-328)
- Open the Popup
- Close the Popup

## Expected Behavior

The tooltip on the popup target should show a tooltip

PR #1806, opened 2022-09-26
fix(popup): Fix return focus timing for tooltips and focus events

NicholasBoll:
## Summary

Fixes: #1805

There is a bug where if a popup is closed and `useReturnFocus` is used for the popup and the popup target has a `Tooltip`, the tooltip is not shown when the popup is closed. This is due to the synchronous nature of `useReturnFocus`. `useReturnFocus` uses a synchronous focus change to handle `useFocusRedirect` correctly. Additional event state tracking was added to detect when we want the browser to change focus on the popup target or when we want the browser to trigger a focus event on the popup target.

category

---

A new testing story was created with an icon button with a tooltip to demonstrate the issue. A codesandbox is also linked in #1805. A new Cypress specification is added to verify the behavior works as expected. Existing Cypress specifications verify behavior still works with `useFocusRedirect`.

Check implementation of [modules/react/popup/lib/hooks/useReturnFocus.tsx](https://github.com/Workday/canvas-kit/compare/master...NicholasBoll:canvas-kit:fix/return-focus-event?expand=1#diff-52b085c43a1ae970dadb57e46047918977c605b1cd8bff88583a552e7db7d9f1)

## Testing Manually

An automated test covers the fix. A new testing story was added under Testing/React/Popups/Popup/TooltipReturnFocus. The Chromatic URL to testing is https://5d854c26ba934e0020f5e98a-zpqkltkmsr.chromatic.com/?path=/story/testing-react-popups-popup--tooltip-return-focus


modules/react/popup/lib/hooks/useReturnFocus.tsx
```diff
@@ -14,16 +14,50 @@ import {usePopupModel} from './usePopupModel';
...
   React.useLayoutEffect(() => {
     if (!visible) {
       return;
     }
     // capture the element here. The refs will be null by the time the cleanup function is called
-    const element = (model.state.returnFocusRef || model.state.targetRef).current;
+    const element = (model.state.returnFocusRef || model.state.targetRef).current as HTMLElement;
+
+    document.addEventListener('keydown', onKeyDown, true);
+    document.addEventListener('keyup', onKeyDown, true);
```
NicholasBoll, line 41:
```suggestion
    document.addEventListener('keyup', onKeyUp, true);
```

modules/react/popup/lib/hooks/useReturnFocus.tsx
```diff
@@ -14,16 +14,50 @@ import {usePopupModel} from './usePopupModel';
...
+
     return () => {
-      changeFocus(element);
+      document.removeEventListener('keydown', onKeyDown, true);
+      document.removeEventListener('keyup', onKeyDown, true);
```
NicholasBoll, line 45:
```suggestion
      document.removeEventListener('keyup', onKeyUp, true);
```

mannycarrera4 (approved):
verified in storybook

jaclynjessup (approved):

Issue #1618, opened 2022-05-31
Tooltip scrolls with page


bvasisht:
## 🐛 Bug Report

After a tooltip appears on a page, the tooltip continues to appear even after the page is scrolled up or down. 

## To Reproduce

Add Tooltip to any element and try to hover over that element and the tooltip should appear. Then try to scroll.

## Expected Behavior

The tooltip should close when target is scrolled off the screen.

## Actual Results

The tooltip continues to show up and does even stick to the element. 

## Browser (if applicable)

Chrome

### Error Output

Paste the error output here

```bash

```

or

mannycarrera4:
Been doing some investigating and it looks like we need to update our tooltip to hide when `data-popper-reference-hidden` on the element. Here's [more info ](https://popper.js.org/docs/v2/modifiers/hide/#:~:text=The%20hide%20modifier%20lets%20you,is%20in%20a%20different%20context.)on popperjs's website

As far as where this change should go, I'm guessing https://github.com/Workday/canvas-kit/blob/master/modules/react/tooltip/lib/TooltipContainer.tsx might be a good place?

PR #1836, opened 2022-10-10
feat: Add `useCloseOnTargetHidden` popup hook

stefano-puzzuoli:
## Summary
Add `useCloseOnTargetHidden` hook to close popups when their target element is no longer visible. Updated tooltips to use this hook.

Fixes: #1618
Fixes: #1437

## Release Category
Components

### Release Note
If you need a popup to close when the target element is no longer visible (useful for tooltips and dropdown menus), there's now a 'useCloseOnTargetHidden` popup hook. This will close the popup when less than 50% of the target element is visible. This hook uses an [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).

---

## Checklist

## For the Reviewer

- [x] PR summary describes the change (Fixes/Resovles linked correctly)

## Testing Manually

Add Tooltip to any element and try to hover over that element and the tooltip should appear. Then try to scroll. The tooltip should now close when target is scrolled off the screen.

Screen recordings of the behavior before and after the bug fix can be found here: 

Before:

https://user-images.githubusercontent.com/7966550/195160321-cd2f1964-9ece-46ae-9fa6-e179ba36895b.mov

After:

https://user-images.githubusercontent.com/7966550/195160371-a928842b-e9fe-4deb-ad60-a23a3fefe022.mov


modules/react/tooltip/lib/TooltipContainer.tsx
```diff
@@ -78,6 +78,12 @@ export const TooltipContainer = styled('div')<TooltipContainerProps>(
       },
     },
 
+    // Hide tooltip when element is scrolled off the screen
```
mannycarrera4, line 81:
```suggestion
    // Hide tooltip when the reference element is either clipped or fully hidden
```

mannycarrera4:
Putting this on hold for now! Even though the tooltip isn't visually shown, events are still possible so we need a way to actually close the tooltip as well other than just visually hiding it!

NicholasBoll:
I looked to see how this could be done. I see 2 options:
1. Add "target visibility" into the model, add a "target visibility" callback to "Popper.tsx" and edit the PopperJS modifier to call this callback, call this callback in `PopupPopper.tsx` to update the model. Add a `useCloseOnTargetHidden` that calls the hide event when the model's state changes to target visibility = hidden
2. Add a `useCloseOnTargetHidden` hook that uses an [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to call the `hide` event when popup's target is no longer visible.

I _really_ dislike option 1. It feels like implementation bleeds everywhere.

Option 2 can only go into v8 because `InterSectionObserver` is not supported in IE11. It can be an isolated hook that subscribes to `model.state.visibility` to add an intersection observer when the popup is visible. It would have to figure out the "scroll parent" of the target element and create an intersection observer from it. It would then need to observe the target element and fire a `hide` event on the popup model when the target is no longer visible. The `useReturnFocus` checks that at least 50% of the target is visible when it calculates if focus should be returned, so `0.5` might be a good threshold for the `IntersectionObserver`.

NicholasBoll:
Something like this seems to work:

```tsx
function getScrollParent(element: HTMLElement): HTMLElement {
  if (element === document.body || !element.parentElement) {
    return element;
  }
  const styles = getComputedStyle(element);

  if (
    styles.overflowY === 'scroll' ||
    styles.overflowY === 'auto' ||
    styles.overflowX === 'scroll' ||
    styles.overflowX === 'auto'
  ) {
    return element;
  }

  return getScrollParent(element.parentElement);
}

export const useCloseOnTargetHidden = createElemPropsHook(usePopupModel)(model => {
  const visible = model.state.visibility !== 'hidden';

  const callback: IntersectionObserverCallback = React.useCallback(
    entries => {
      if (entries[0].intersectionRatio < 0.5) {
        model.events.hide();
      }
    },
    [model.events]
  );

  React.useLayoutEffect(() => {
    if (!visible || !model.state.targetRef.current) {
      return;
    }

    const target = model.state.targetRef.current;
    const scrollParent = getScrollParent(target);

    const observer = new IntersectionObserver(callback, {
      root: scrollParent,
      threshold: 0.5,
    });

    observer.observe(model.state.targetRef.current);
    return () => {
      observer.disconnect();
    };
  }, [callback, model.state.targetRef, visible]);

  return {};
});
```

NicholasBoll:
Codesandbox with above example: https://codesandbox.io/s/usecloseontargethidden-5xjlbx?file=/src/App.tsx

stefano-puzzuoli:
Hi @NicholasBoll - Thank you so much for the suggestion and for the help. Please let me know if any other improvements or changes are needed. Thanks again!

mannycarrera4:
Because this is using `IntersectionObserver` it might have to wait until after v8 since we're still supporting IE11 in v7

NicholasBoll:
I'll target against v8. I'll add documentation for this new hook. I'll also add Cypress tests to verify the behavior. I encourage you to check them out once added.

Since we're updating the baseline, you'll have to pull the latest from this branch, then run `yarn install` again.

stefano-puzzuoli:
That sounds good, I'll keep an eye out for when v8 is released. Thanks!

NicholasBoll:
I added tests: https://github.com/Workday/canvas-kit/pull/1836/commits/155c43dfcae8ba949b606f25dcab1f4e771fd9a1

NicholasBoll (approved):

PR #1846, opened 2022-10-12
ci: Fix popup tests and ci bugs [skip release]

NicholasBoll:
## Summary

Fixed multiple ci failures:
- Popup tests in React 18 due to different `useLayoutEffect` timing
- forward-merge Action not erroring on Chromatic changes
- pull-request Action not failing properly
- `prerelease/minor` canary builds failing to publish

## Release Category
Infrastructure

---

`useReturnFocus`

### DIFF
--- a/.github/workflows/forward-merge.yml
+++ b/.github/workflows/forward-merge.yml
@@ -130,7 +130,6 @@ jobs:
           appCode: dlpro96xybh
           storybookBuildDir: docs
           exitOnceUploaded: false
-          exitZeroOnChanges: true
 
       - name: Start Server
         run: npx http-server docs -p 9001 & npx wait-on http://localhost:9001
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -185,7 +185,11 @@ jobs:
 
   verify:
     runs-on: ubuntu-latest
+    if: always()
     needs: ['check', 'integration-test', 'visual-test']
     steps:
-      - name: Done
-        run: echo "Done"
+      - name: Failure Test
+        if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') # Test the result of all dependencies
+        run: exit 1
+      - name: Success
+        run: exit 0
--- a/modules/react/popup/lib/hooks/useReturnFocus.tsx
+++ b/modules/react/popup/lib/hooks/useReturnFocus.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 import {changeFocus, createElemPropsHook, isFocusable} from '@workday/canvas-kit-react/common';
 
 import {usePopupModel} from './usePopupModel';
+import {PopupStack} from '@workday/canvas-kit-popup-stack';
 
 function getScrollParent(element: HTMLElement): HTMLElement {
   if (element === document.body || !element.parentElement) {
@@ -53,19 +54,18 @@ export const useReturnFocus = createElemPropsHook(usePopupModel)(model => {
     }
   }, []);
 
-  const onKeyUp = React.useCallback((event: KeyboardEvent) => {
-    requiresFocusChangeRef.current = false;
-  }, []);
-
   // track mousedown events to determine if the mouse target is a focusable element. If it is, we
   // should not return focus
-  const onMouseDown = React.useCallback((event: MouseEvent) => {
-    elementRef.current = event.target as Element;
-  }, []);
-
-  const onMouseUp = React.useCallback(() => {
-    elementRef.current = null;
-  }, []);
+  const onMouseDown = React.useCallback(
+    (event: MouseEvent) => {
+      if (model.state.stackRef.current && event.target instanceof HTMLElement) {
+        if (!PopupStack.contains(model.state.stackRef.current, event.target)) {
+          elementRef.current = event.target;
+        }
+      }
+    },
+    [model.state.stackRef]
+  );
 
   // We use `useLayoutEffect` because the callback will be called _before_ the browser changes
   // focus. This allows the browser to change focus as normal. For example, if the popup closes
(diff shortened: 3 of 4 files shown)


.github/workflows/forward-merge.yml
```diff
@@ -130,7 +130,6 @@ jobs:
           appCode: dlpro96xybh
           storybookBuildDir: docs
           exitOnceUploaded: false
-          exitZeroOnChanges: true
```
NicholasBoll, line 133: This told Chromatic to not fail CI if any visual changes were detected. This is not intentional. This causes unreviewed changes in the ancestor builds of pull requests which is confusing inside a pull request.

modules/react/popup/lib/hooks/useReturnFocus.tsx
```diff
@@ -53,19 +54,18 @@ export const useReturnFocus = createElemPropsHook(usePopupModel)(model => {
...
-
   // track mousedown events to determine if the mouse target is a focusable element. If it is, we
   // should not return focus
-  const onMouseDown = React.useCallback((event: MouseEvent) => {
-    elementRef.current = event.target as Element;
-  }, []);
-
-  const onMouseUp = React.useCallback(() => {
-    elementRef.current = null;
-  }, []);
```
RayRedGoose, line 68: Why has `onMouseUp` and `onKeyUp` been removed? Do they cause any issues?

.github/workflows/pull-request.yml
```diff
@@ -185,7 +185,11 @@ jobs:
...
+    if: always()
     needs: ['check', 'integration-test', 'visual-test']
     steps:
-      - name: Done
-        run: echo "Done"
+      - name: Failure Test
+        if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') # Test the result of all dependencies
+        run: exit 1
+      - name: Success
+        run: exit 0
```
NicholasBoll, line 195: The `always` forces the `verify` job to run no matter if dependent jobs have failed. This is required because Github Actions skip jobs when a failure happens. Also Github status checks consider a skip to be a success.

The `verify` job is a required check for all PRs, so it should always be run. The `if: contains ...` with the `exit 1` ensures a proper failure if any dependent job has a result of `failure` or `cancelled`.

modules/react/popup/lib/hooks/useReturnFocus.tsx
```diff
@@ -108,12 +105,14 @@ export const useReturnFocus = createElemPropsHook(usePopupModel)(model => {
...
-        (elementRef.current && getFocusableElement(elementRef.current)) ||
-        elementRect.top + elementRect.height / 2 < scrollParentRect.top ||
-        elementRect.bottom - elementRect.height / 2 > scrollParentRect.bottom ||
-        elementRect.left + elementRect.width / 2 < scrollParentRect.left ||
-        elementRect.right - elementRect.width / 2 > scrollParentRect.right
+        (elementRef.current && getFocusableElement(elementRef.current)) || // did the user click on a focusable element?
+        elementRect.top + elementRect.height / 2 < scrollParentRect.top || // is the top half of the target element visible?
+        elementRect.bottom - elementRect.height / 2 > scrollParentRect.bottom || // is the bottom half of the target element visible?
+        elementRect.left + elementRect.width / 2 < scrollParentRect.left || // is the left half of the target element visible?
+        elementRect.right - elementRect.width / 2 > scrollParentRect.right // is the right half of the target element visible?
```
RayRedGoose, line 112: Thanks for adding this comments!

utils/publish-canary.mjs
```diff
@@ -101,7 +101,7 @@ exec('git diff --name-only HEAD HEAD^')
         bump = 'premajor';
       } else {
         // we'll use `next` for pre minors
-        preid = 'next';
+        preid = `${process.env.GITHUB_RUN_NUMBER || 0}-next`;
```
NicholasBoll, line 104: Here's the problem. The previous code made release tags like `v7.4.0-next.1`. The `.1` at the end is how many commits exist since the last tag. This doesn't work for us, because we forward merge commits. This means if there's a commit to master that creates a new release tag, the number starts over. The canary job then fails to publish the tag because a `v7.4.0-next.0` already exists. This means for a new `prerelease/minor` to create a canary release, we'll have to have enough commits on the `prerelease/minor` branch to exceed the previous number.

This change instead uses the Github run number to generate a unique release version each time.

modules/react/popup/lib/hooks/useReturnFocus.tsx
```diff
@@ -53,19 +54,18 @@ export const useReturnFocus = createElemPropsHook(usePopupModel)(model => {
...
-
-  const onMouseUp = React.useCallback(() => {
-    elementRef.current = null;
-  }, []);
+  const onMouseDown = React.useCallback(
+    (event: MouseEvent) => {
+      if (model.state.stackRef.current && event.target instanceof HTMLElement) {
+        if (!PopupStack.contains(model.state.stackRef.current, event.target)) {
+          elementRef.current = event.target;
+        }
```
NicholasBoll, line 64: The top `if` statement makes Typescript happy. The second `if` statement makes sure we're only counting `mousedown` events that are outside the popup when tracking the next element the user is clicking on for focus element calculations.

Before we were using `mouseup` to null out the `elementRef`. This worked for elements inside the popup because the popup didn't close and the mouseup was fired. When clicking outside the popup, focus changes as part of the `mousedown` event and the `mouseup` event is unsubscribed. The `elementRef` was nulled out by the `useLayoutEffect` cleanup function.

The old logic worked fine with Cypress and React 17 because the `useLayoutEffect` callback was called synchronously BEFORE the `mousedown` event released the event loop. React 18 calls the `useLayoutEffect` callback asynchronously. Cypress calls a `mousedown`, `mouseup`, and `click` within the same frame of the event loop which is impossible for a user to do. That means the `mouseup` event was clearing the `elementRef` in Cypress and React 18.

This new logic doesn't rely on the `mouseup` quirk that fails Cypress and React 18 and instead does popup contains detection manually and clears the `elementRef` consistently when the `useLayoutEffect` callback fires.

modules/react/popup/lib/hooks/useReturnFocus.tsx
NicholasBoll: They are not needed. Focus changes during `onMouseDown` and `onKeyDown` events.

jamesfan (approved):

PR #1861, opened 2022-10-21
fix: Popup close buttons no longer interfere with forms

NicholasBoll:
## Summary

Update the default `type` of popup `CloseIcon` and `CloseButton` components to `button` instead of `submit`. Also, we added a form modal example.

## Release Category
Components

### BREAKING CHANGES

We've updated all popup `CloseIcon` and `CloseButton` to default to `type=button`. We did this so
these close buttons did not submit the forms unintentionally. Without this change, you have to
manually add `type="button"` to these components if you wrap popup contents in a `form` element.
While this is a very unlikely that a `CloseButton` was intentionally used as an implicit submit
button, it is still a breaking change. This change effects the following components:

- `Popup.CloseIcon`
- `Dialog.CloseIcon`
- `Modal.CloseIcon`
- `Popup.CloseButton`
- `Dialog.CloseButton`
- `Modal.CloseButton`

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods

## Testing Manually

The "Form Modal" example should console.log the form contents when the submit button is clicked or when the user hits the "enter" key in either the "First Name" or "Last Name" fields.

image


modules/react/modal/stories/examples/FormModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+            <FormField label="First Name">
+              <TextInput name="first" />
+            </FormField>
+            <FormField label="Last Name">
+              <TextInput name="last" />
+            </FormField>
+          </Modal.Body>
+          <HStack spacing="s" padding="xxs" marginTop="xxs">
+            <Modal.CloseButton>Cancel</Modal.CloseButton>
+            <PrimaryButton type="submit">Submit</PrimaryButton>
```
mannycarrera4, line 44: Do you have to add submit since buttons default to submit, or is this more of a clear way to indicate hey this is submit and this button will trigger the `onSubmit` ?

modules/docs/mdx/8.0-UPGRADE-GUIDE.mdx
```diff
@@ -701,3 +702,20 @@ will have the following visual differences compared to the standard `Modal`:
...
+
+---
+
+### Popups
+
+We've updated all popup `CloseIcon` and `CloseButton` to default to `type=button`. We did this so
+these close buttons did not submit the forms unintentionally. Without this change, you have to
+manually add `type="button"` to these components if you wrap popup contents in a `form` element.
+While this is a very unlikely that a `CloseButton` was intentionally used as an implicit submit
+button, it is still a breaking change. This change effects the following components:
```
mannycarrera4, line 714: Any MDN docs to reference here that might help other developers?

modules/react/modal/stories/examples/FormModal.tsx
NicholasBoll: It is a clear indication that this is a submit button. All examples I could find explicitly use `type=submit` if a `form` element is used. This is the recommended way to do submit buttons. We'll change all buttons to `type=button` in the future, so this is future-proof code.

modules/docs/mdx/8.0-UPGRADE-GUIDE.mdx
NicholasBoll: Default `type` for buttons: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type

React has a very small amount of documentation here: https://reactjs.org/docs/forms.html

There isn't much documentation on how forms work.

modules/react/modal/stories/examples/FormModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+    // if it looks good, submit to the server and close the modal
+    model.events.hide();
+  };
+
+  return (
+    <Modal model={model}>
+      <Modal.Target icon={plusIcon}>Create New User</Modal.Target>
+      <Modal.Overlay>
+        <Modal.Card as="form" onSubmit={onSubmit}>
+          <Modal.CloseIcon aria-label="Close" />
```
willklein, line 32: Do you think `"Close"` still makes sense as the label as opposed to `"Cancel"`?

modules/docs/mdx/8.0-UPGRADE-GUIDE.mdx
NicholasBoll: At least the React documentation shows:
```jsx
<form onSubmit={handleSubmit}>
  <input .../>
  <button type="submit">Submit</button>
</form>
```

Most of the code I've seen in the wild does this instead:
```jsx
const handleKeyDown = e => {
  // handles the "enter" key submitting the form, but it isn't perfect... The enter key doesn't always submit
  // the form. A select, for example. The enter key when choosing a select option should close the select menu,
  // but not submit the form. The enter key should only submit the form if the select menu is closed AND the
  // `Select` has focus
  if (e.key === "Enter") {
    handleSubmit()
  }
}

<div onKeyDown={handleKeyDown}>
  <input ... />
  <button onClick={handleSubmit}>Submit</button>
  <button onClick={handleCancel}>Cancel</button>
</div>

File: modules/react/modal/stories/examples/FormModal.tsx (hunk 3 above)
Comment (NicholasBoll) [reply modules/react/modal/stories/examples/FormModal.tsx:32]: This is currently unspecified. Some form modals may show a "Are you sure you want to close the modal. You will lose unsaved work." due to unsaved work, especially if the form is longer. It depends on the use-case. Some form modals may omit the close icon altogether, and only allow an explicit "Cancel" button.

There are so many examples we could make around form modals. A `shouldHide` model guard could be added to prevent closing if the form isn't submitted (triggering an additional "are you sure model").

File: modules/react/modal/stories/examples/FormModal.tsx (hunk 1 above)
Comment (mannycarrera4) [reply modules/react/modal/stories/examples/FormModal.tsx:44]: ok cool that's what i thought!

File: modules/docs/mdx/8.0-UPGRADE-GUIDE.mdx (hunk 2 above)
Comment (mannycarrera4) [reply modules/docs/mdx/8.0-UPGRADE-GUIDE.mdx:714]: Should this be part of our documentation as in the links or od you think this is sufficient enough?

File: modules/docs/mdx/8.0-UPGRADE-GUIDE.mdx [hunk 4]
```diff
@@ -701,3 +702,20 @@ will have the following visual differences compared to the standard `Modal`:
 
 Additionally, users interacting with `Modal` on touch will not be able to exit out of the modal by
 tapping on the overlay.
+
+---
+
+### Popups
+
+We've updated all popup `CloseIcon` and `CloseButton` to default to `type=button`. We did this so
+these close buttons did not submit the forms unintentionally. Without this change, you have to
```
mannycarrera4, line 711:
```suggestion
these close buttons did not submit when a form element is present. Without this change, you have to
```

mannycarrera4: Feel free to word this any way, but I'm trying to say, when there is a form element in the modal or popup, the close button will not submit, not all instances of popup or modal have forms

mannycarrera4 (approved):
🚢 it

PR #1898, opened 2022-11-03
fix: Fail gracefully if localStorage is blocked in useModalityType

NicholasBoll:
## Summary

There could be many reasons localStorage is blocked (content security policy, iframe cross origin, etc). `useModalityType` uses localStorage to avoid possible flashing of UI for users. But a blank screen due to an access denied error is worse, so we'll wrap `localStorage` access in a `try/catch` block to gracefully fallback to guessing when localStorage is not available.

## Release Category
Components

---

## For the Reviewer

### DIFF
--- a/modules/react/common/lib/utils/useModalityType.ts
+++ b/modules/react/common/lib/utils/useModalityType.ts
@@ -2,6 +2,22 @@ import React from 'react';
 
 type Modality = 'mouse' | 'touch' | 'pen';
 
+function safeLocalStorageGet(key: string): string | null {
+  try {
+    return localStorage.get(key);
+  } catch (_) {
+    return '';
+  }
+}
+
+function safeLocalStorageSet(key: string, value: string): void {
+  try {
+    localStorage.set(key, value);
+  } catch (_) {
+    // Do nothing
+  }
+}
+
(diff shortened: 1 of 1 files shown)


modules/react/common/lib/utils/useModalityType.ts
```diff
@@ -2,6 +2,22 @@ import React from 'react';
 
 type Modality = 'mouse' | 'touch' | 'pen';
 
+function safeLocalStorageGet(key: string): string | null {
+  try {
+    return localStorage.get(key);
+  } catch (_) {
+    return '';
```
willklein, line 9: Not sure which is better, but what do you think about `return null;` here?

NicholasBoll: `null` would be better, but the effect is the same. `''` and `null` are both falsey

willklein (approved):
:shipit:

modules/react/common/lib/utils/useModalityType.ts
willklein: Ah, for the usage. I get you. Thanks for changing it. Just in case we refactor it later to be used more.

PR #1902, opened 2022-11-07
fix: Security access error on `typeof localStorage`

NicholasBoll:
## Summary

Chrome throws an access error if cookies aren't allowed even when using `typeof`. We'll move the `typeof` check that's required by SSR into a `try...catch` block to ensure there are no errors in Chrome.

## Release Category
Components


modules/react/common/lib/utils/useModalityType.ts
```diff
@@ -27,9 +27,7 @@ function safeLocalStorageSet(key: string, value: string): void {
 //   a. if localStorage isn't primed, use clientWidth of the browsers
 //   b. if < 768, default to 'touch'
 //   c. else default to 'mouse'
-let localStorageValue = ((typeof localStorage !== 'undefined'
-  ? safeLocalStorageGet('modality')
-  : '') ||
+let localStorageValue = (safeLocalStorageGet('modality') ||
```
mannycarrera4, line 30: where does `safeLocalStorageGet` come from?

NicholasBoll: Expand the file to see it

mannycarrera4 (approved):

Issue #1905, opened 2022-11-08
Codesandbox render of v8 Text component failed


RayRedGoose:
## 🐛 Bug Report

Codesandbox shows error that Text component is undefined when imported

## Codesandbox

[Codesandbox](https://codesandbox.io/s/wonderful-antonelli-7opzvp?file=/src/App.tsx)

## Screenshot
![Error](https://user-images.githubusercontent.com/23533895/200664214-9db6772e-d0f9-42de-9277-6f476a2a3e3a.png)

PR #1906, opened 2022-11-08
fix(text): Update text component imports

RayRedGoose:
## Summary

Fixes: #1905  

Changed text package import to relative and removed `Text.as` from `LabelText` and `Breadcrumbs.Item` to fix CSB render.

## Release Category
Components

### DIFF
--- a/modules/react/breadcrumbs/lib/BreadcrumbsCurrentItem.tsx
+++ b/modules/react/breadcrumbs/lib/BreadcrumbsCurrentItem.tsx
@@ -1,12 +1,9 @@
 import React from 'react';
 import {
-  styled,
   createElemPropsHook,
   useLocalRef,
   composeHooks,
   createSubcomponent,
-  StyledType,
-  ellipsisStyles,
   useForkRef,
 } from '@workday/canvas-kit-react/common';
 import {OverflowTooltip, OverflowTooltipProps} from '@workday/canvas-kit-react/tooltip';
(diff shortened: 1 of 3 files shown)


modules/react/breadcrumbs/lib/BreadcrumbsCurrentItem.tsx
```diff
@@ -21,7 +21,7 @@ export interface BreadcrumbsCurrentItemProps extends TextProps {
   tooltipProps?: OverflowTooltipProps;
 }
 
-const ListItem = styled(Text.as('li'))<BreadcrumbsCurrentItemProps & StyledType>(
```
NicholasBoll, line 24: Did you mean to remove the `.as('li')` here?

modules/react/text/lib/LabelText.tsx
```diff
@@ -1,21 +1,19 @@
...
-import {Text, TextProps} from '@workday/canvas-kit-react/text';
+import {Text, TextProps} from './Text';
 import {inputColors} from '@workday/canvas-kit-react/tokens';
 
 export interface TypeLabelProps extends TextProps {
   cursor?: Property.Cursor;
   disabled?: boolean;
 }
 
-const StyledLabel = styled(Text.as('label'))<StyledType & TypeLabelProps>(
```
NicholasBoll, line 12: Why are we removing the `.as('label')` here?

mannycarrera4: i think this was just clean up, since `createComponent` and `StyledType` handles this for us.

mannycarrera4 (approved):
If we want to keep the `.as` I'm ok with that as well, not blocking

alanbsmith (approved):
🦁

NicholasBoll (approved):

Issue #1916, opened 2022-11-16
React 16.14 isn't working with Canvas Kit 8


willklein:
## 🐛 Bug Report

If you actually try to install Canvas Kit 8 with React 16.14, which should be supported, you will get an npm error:
```bash
> npm install @workday/canvas-kit-react

npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: react-16-canvas-kit-8@1.0.0
npm ERR! Found: react@16.14.0
npm ERR! node_modules/react
npm ERR!   react@"^16.14.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^17.0 || ^18.0" from @workday/canvas-kit-react@8.1.1
npm ERR! node_modules/@workday/canvas-kit-react
npm ERR!   @workday/canvas-kit-react@"*" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR!
npm ERR! See /Users/will.klein/.npm/eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/will.klein/.npm/_logs/2022-11-16T23_30_05_960Z-debug-0.log
```

## To Reproduce

Create an empty directory and then inside there run:
```
npm init
npm install react@16 react-dom@16
npm install @workday/canvas-kit-react
```

You'll see the error mentioned above.

PR #1917, opened 2022-11-16
fix: Allow installing Canvas Kit with React 16.14

willklein:
## Summary

Fixes: #1916 

React 16.14 isn't working with Canvas Kit 8, because our `peerDependencies` specifies `"^17.0 || ^18.0"`. By specifying `">=16.14"` we are including React 17.x and 18.x.

I confirmed this version range strategy is supported at least as far back as [v6](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#dependencies); the current version is v9.

## Release Category
Dependencies

---

## Checklist

## For the Reviewer

NPM reference: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#dependencies
NPM SemVer calculator: https://semver.npmjs.com/

## Areas for Feedback?

- [ ] Verify the new approach to specifying versions.
- [ ] Documentation

## Screenshot

This is the error if you install Canvas Kit 8 with React 16.14.0:
Screen Shot 2022-11-16 at 4 52 46 PM

## Thank You Gif

that's better


modules/labs-react/package.json
```diff
@@ -41,7 +41,7 @@
     "workday"
   ],
   "peerDependencies": {
-    "react": "^17.0 || ^18.0"
+    "react": ">=16.14"
```
willklein, line 44: If we want to throw an npm error on React 19.x (whenever that comes out), we'd make this:
```
"react": "16.14 - 18"
```

I think it would be better to be permissive in case it works fine, rather than restrictive in case it is not.

mannycarrera4:
Do we also need to update the `docs` package?

willklein:
> Do we also need to update the `docs` package?

I don't believe so. We don't specify `peerDependencies` in its `package.json`. We are also the only supported consumer of this package, correct? It seems unnecessary to specify this.

mannycarrera4 (approved):

Issue #1990, opened 2023-01-16
useReturnFocus overrides custom focus moves


christine-m:
## 🐛 Bug Report

useReturnFocus always returns focus, even if focus has moved away from the modal via a custom focus movement in a handler.

## To Reproduce

See [Codesandbox](https://codesandbox.io/s/focus-moves-rjrs6t?file=/src/App.js).

With CK version 7.3.4, focus moves to the text input on close. With version 7.4.6, focus always moves back to the button.

## Expected Behavior

In codesandbox above, focus should move to the text input and remain on the text input.

## Actual Results

In codesandbox above, focus always moves to the button.

## Link to repl or repo (highly encouraged)

See [Codesandbox](https://codesandbox.io/s/focus-moves-rjrs6t?file=/src/App.js).

PR #1991, opened 2023-01-17
fix: Allow overriding of return focus when closing a popup

christine-m:
## Summary

Fixes: #1990 

Allows components to override which element focus is returned to when closing a popup. When not overridden, the default behaviour continues.

## Release Category
Components

---

## Checklist

## For the Reviewer

modules/react/popup/lib/hooks/useReturnFocus.tsx

## Testing Manually

The Testing/React/Popups/Popup/ReturnFocusTest story has been updated to include an example. Click on the link in the popup and see that focus is moved to the text input. When clicking anywhere else, existing focus move behaviour is preserved.

### DIFF
--- a/modules/popup-stack/lib/PopupStack.ts
+++ b/modules/popup-stack/lib/PopupStack.ts
@@ -344,10 +344,7 @@ export const PopupStack = {
    * opt-in to toggling. Otherwise there is no way to opt-out of toggle behavior (because the target
    * is not inside `element`).
    */
-  contains(element: HTMLElement | null, eventTarget: Element | null): boolean {
-    if (!element || !eventTarget) {
-      return false;
-    }
+  contains(element: HTMLElement, eventTarget: HTMLElement): boolean {
     const stack = getTopStack();
     if (stack._adapter?.contains) {
       return stack._adapter.contains(element, eventTarget);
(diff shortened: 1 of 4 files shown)


modules/popup-stack/lib/PopupStack.ts
```diff
@@ -344,7 +344,10 @@ export const PopupStack = {
    * opt-in to toggling. Otherwise there is no way to opt-out of toggle behavior (because the target
    * is not inside `element`).
    */
-  contains(element: HTMLElement, eventTarget: HTMLElement): boolean {
+  contains(element: HTMLElement | null, eventTarget: Element | null): boolean {
+    if (!element || !eventTarget) {
+      return false;
+    }
```
NicholasBoll, line 350: I don't think we should change the PopupStack. The PopupStack is meant to be global and shared. This means that your application is using a newer version of the PopupStack, but another application is using an older version of the PopupStack and has registered it first. In this scenario, this behavior will break.

modules/react/popup/lib/hooks/useReturnFocus.tsx
```diff
@@ -116,7 +116,11 @@ export const useReturnFocus = createElemPropsHook(usePopupModel)(model => {
         return;
       }
 
-      if (requiresFocusChangeRef.current) {
+      if (
+        (document.activeElement !== document.body &&
+          !PopupStack.contains(model.state.stackRef.current, document.activeElement)) ||
+        requiresFocusChangeRef.current
+      ) {
```
NicholasBoll, line 123: Update this logic to contain what is necessary without having to change `PopupStack`

christine-m: Updated!

NicholasBoll (approved):

Issue #1950, opened 2022-12-07
Modify the Canvas QA process to include testing across different screen sizes and modality


myvuuu:
## 💡 Suggestion

A clear and concise description of what the suggestion is.

## Motivation

What are the key benefits if this suggest was included to Canvas-Kit?

## Example

Please provide some examples of this suggestion in practice.

PR #2028, opened 2023-02-08
test: Modify visual testing

RayRedGoose:
## Summary

Resolves: #1950  

Improving our testing coverage by adding cypress and visual tests to cover different behavior and view because of modality, viewport sizes and container widths.

The following changes have been made:

- **ActionBar**
- [x] add visual tests for different container widths
- [x] add all cypress tests
- **Breadcrumbs**
- [x] add cypress tests for different container widths
- **Tabs**
- [x] add visual tests for overflow variant
- [x] add visual test for touch devices
- [x] add cypress tests for touch modality
- **Modal**
- [x] add visual tests for mobile screen
- [x] add cypress tests for touch modality

## Release Category
Infrastructure

---

## Checklist

## For the Reviewer

cypress folder or ActionBar, Breadcrumbs, Tabs, Modal packages


NicholasBoll:
I don't see any code that forces modality to touch for visual regression purposes.

modules/react/tabs/stories/stories_testing.tsx
```diff
@@ -66,18 +73,58 @@ export const TabStates = withSnapshotsEnabled(() => (
   </>
 ));
 
-export const Bidirectionality = withSnapshotsEnabled(() => {
+export const BasicTabs = withSnapshotsEnabled(() => {
+  return <Basic />;
+});
+
+BasicTabs.parameters.chromatic.viewports = [480, 1200];
```
NicholasBoll, line 80: I don't understand the purpose of changing the viewport here. This is the basic example which does not overflow, so the 480px width version simply gets clipped in the screenshot. The `TabStates` story already captures what the `Basic` example does. What does the `BasicTabs` story add for visual regression testing? We pay per Story captured, so each story should add something useful.

cypress/integration/Breadcrumbs.spec.ts
```diff
@@ -133,13 +139,105 @@ describe('Breadcrumbs', () => {
         getDropdownButton().should('have.attr', 'aria-controls', 'menu');
       });
 
-      context('when the dropdown button is clicked', () => {
+      context('when action list container is only 440px wide', () => {
```
josh-bagwell, line 142: Should this be `480px` ?

cypress/integration/Modal.spec.ts
```diff
@@ -156,6 +156,17 @@ describe('Modal', () => {
             cy.findByRole('dialog', {name: 'MIT License'}).should('not.exist');
           });
         });
+
+        context(`when clicking outside the modal on mobile view`, () => {
+          beforeEach(() => {
+            cy.viewport(380, 720);
```
josh-bagwell, line 162: Is this supposed to be `480, 720` ?

RayRedGoose: It does not really matter, I just need to simulate any phone screen. I changed it to specific phone screen to confuse less why I used this numbers.

josh-bagwell:
Per Nicholas idea for the Chromatic issue, the text inside the `Modal` for the visual testing should leave context for why Chromatic's version of Chrome places the `Modal` in the wrong position. When Chromatic updates the Chrome instance, the position will change which will be a visual diff and we will have to understand why it changed. 

For example, "This text should appear on the bottom of the screen. Chromatic uses a version of Chrome that makes it appear on the top and is a known issue.".

RayRedGoose:
@josh-bagwell @NicholasBoll 

> Per Nicholas idea for the Chromatic issue, the text inside the `Modal` for the visual testing should leave context for why Chromatic's version of Chrome places the `Modal` in the wrong position. When Chromatic updates the Chrome instance, the position will change which will be a visual diff and we will have to understand why it changed.
> 
> For example, "This text should appear on the bottom of the screen. Chromatic uses a version of Chrome that makes it appear on the top and is a known issue.".

I can't set a specific text to only mobile snapshots because both for mobile and regular view snapshots are made from one `Modal Small Width` story. I can change the text inside this story and it will be the same for all snapshots. 

Something like showed below: 
Screen Shot 2023-02-22 at 1 45 46 PM

Screen Shot 2023-02-22 at 1 46 33 PM

If we are okay with this changes, I can push them.

NicholasBoll:
> @josh-bagwell @NicholasBoll
> 
> > Per Nicholas idea for the Chromatic issue, the text inside the `Modal` for the visual testing should leave context for why Chromatic's version of Chrome places the `Modal` in the wrong position. When Chromatic updates the Chrome instance, the position will change which will be a visual diff and we will have to understand why it changed.
> > For example, "This text should appear on the bottom of the screen. Chromatic uses a version of Chrome that makes it appear on the top and is a known issue.".
> 
> I can't set a specific text to only mobile snapshots because both for mobile and regular view snapshots are made from one `Modal Small Width` story. I can change the text inside this story and it will be the same for all snapshots.
> 
> Something like showed below. If we are okay with this changes, I can push them.
> 
> Screen Shot 2023-02-22 at 1 45 46 PM
> 
> Screen Shot 2023-02-22 at 1 46 33 PM

@RayRedGoose I think this is okay. We don't recommend using our `Testing` stories for documentation purposes. We also don't hide them (which is technically possible), because sometimes they are useful to reference.

josh-bagwell (approved):

Issue #2084, opened 2023-03-14
Duplicate props in Modal Storybook documentation


sheelah:
## 🐛 Bug Report
Noticed that the `Modal` component has some props that appear two times in the props documentation in Storybook.

Example:  https://workday.github.io/canvas-kit/?path=/docs/components-popups-modal--basic

## To Reproduce

Visit https://workday.github.io/canvas-kit/?path=/docs/components-popups-modal--basic and check the props table. You'll see that `show`, `hide`, and `updatePlacement` are all listed in the same table twice.

## Expected Behavior

I'd expect each prop to appear once.

## Actual Results

Those props are listed more than once.

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

## Link to repl or repo (highly encouraged)

https://workday.github.io/canvas-kit/?path=/docs/components-popups-modal--basic

### Error Output

Paste the error output here

```bash

```

or

Include a screenshot of a visual bug

PR #2085, opened 2023-03-14
docs: Filter out model props in container components

NicholasBoll:
## Summary

Fixes: #2084

## Release Category
Documentation

---

### DIFF
--- a/modules/docs/docgen/plugins/enhancedComponentParser.ts
+++ b/modules/docs/docgen/plugins/enhancedComponentParser.ts
@@ -471,10 +471,12 @@ export const enhancedComponentParser = createParserPlugin<SupportedValues>((node
       const type = node.initializer.typeArguments
         ? parser.checker.getTypeAtLocation(node.initializer.typeArguments[0])
         : undefined;
-      const props = (type ? getComponentProps(parser, signature, type, baseElement) : []).filter(
-        // Filter out `model` and `elemPropsHook` that might come if we extend an interface
-        p => !['model', 'elemPropsHook'].includes(p.name)
-      );
+      const props = (type ? getComponentProps(parser, signature, type, baseElement) : [])
+        .filter(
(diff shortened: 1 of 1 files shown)


modules/docs/docgen/plugins/enhancedComponentParser.ts
```diff
@@ -471,10 +471,12 @@ export const enhancedComponentParser = createParserPlugin<SupportedValues>((node
...
         ? parser.checker.getTypeAtLocation(node.initializer.typeArguments[0])
         : undefined;
-      const props = (type ? getComponentProps(parser, signature, type, baseElement) : []).filter(
-        // Filter out `model` and `elemPropsHook` that might come if we extend an interface
-        p => !['model', 'elemPropsHook'].includes(p.name)
-      );
+      const props = (type ? getComponentProps(parser, signature, type, baseElement) : [])
+        .filter(
+          // Filter out `model` and `elemPropsHook` that might come if we extend an interface
+          p => !['model', 'elemPropsHook'].includes(p.name)
```
mannycarrera4, line 477: When we filter out model, i assume that account for any state and events?

mannycarrera4:
## Before
Screen Shot 2023-03-15 at 10 15 41 AM


## After 
Screen Shot 2023-03-15 at 10 14 25 AM

Looks much cleaner and then you can open `Model` which shows state and events! I like

mannycarrera4 (approved):

modules/docs/docgen/plugins/enhancedComponentParser.ts
NicholasBoll: The `model` is added as a external prop by an enhanced component. When we use `ExtractProps` of an enhanced component, the `model` and `elemPropsHook` are added. We filter those out because we add more specific ones because of the `modelHook`

PR #2096, opened 2023-03-20
fix(collection): Correctly use custom getId

### LINKED ISSUE #2095: getId is ignored by useCursorListModel because it imports defaultGetId and uses that instead
## 🐛 Bug Report

`getId` is ignored by `useCursorListModel` because it imports `defaultGetId` and uses that instead:

## To Reproduce

[Override t](https://fe3ut0.csb.app/)
TAB into the list
Use ARROWUP and ARROWDOWN to move focus

## Expected Behavior

Focus changes between list items

## Actual Results

No focus change

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

Please provide a reference to a repo or branch to help us reproduce this bug.

### Error Output

Paste the error output here

```bash

```

or

NicholasBoll:
## Summary

Fix the custom list model to correctly use custom `getId` config. The BaseListModel creates `items` that are wrapped with extra metadata. `useListItemRovingFocus` will now use the wrapped item metadata correctly and `useCursorListModel` will set the correct cursorId

Fixes #2095 

## Release Category
Components

### Release Note
An example was incorrectly using the wrong id. Note that `state.items` in a dynamic collection wraps your provided items so there is a `.id` property on `state.items`. `model.getId` is no longer needed and will be removed in the future to avoid confusion.

---

## Checklist

## For the Reviewer

### DIFF
--- a/modules/react/tabs/stories/examples/DynamicTabs.tsx
+++ b/modules/react/tabs/stories/examples/DynamicTabs.tsx
@@ -32,9 +32,9 @@ export const DynamicTabs = () => {
    * @param id The id of the item that will be removed
    */
   const removeItem = <T extends unknown>(id: string, model: ReturnType<typeof useTabsModel>) => {
-    const index = model.state.items.findIndex(item => model.getId(item) === model.state.cursorId);
+    const index = model.state.items.findIndex(item => item.id === model.state.cursorId);
     const nextIndex = index === model.state.items.length - 1 ? index - 1 : index + 1;
-    const nextId = model.getId(model.state.items[nextIndex]);
+    const nextId = model.state.items[nextIndex].id;
     if (model.state.selectedIds[0] === id) {
       // We're removing the currently selected item. Select next item
       model.events.select({id: nextId});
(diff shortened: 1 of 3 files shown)


mannycarrera4:
Does this affect anywhere we use `config.getId` like segmented control?

modules/react/tabs/stories/examples/DynamicTabs.tsx
```diff
@@ -32,9 +32,9 @@ export const DynamicTabs = () => {
    * @param id The id of the item that will be removed
    */
   const removeItem = <T extends unknown>(id: string, model: ReturnType<typeof useTabsModel>) => {
-    const index = model.state.items.findIndex(item => model.getId(item) === model.state.cursorId);
+    const index = model.state.items.findIndex(item => item.id === model.state.cursorId);
     const nextIndex = index === model.state.items.length - 1 ? index - 1 : index + 1;
-    const nextId = model.getId(model.state.items[nextIndex]);
+    const nextId = model.state.items[nextIndex].id;
```
mannycarrera4, line 37: >  Optional function to return an id of an item. If not provided, the default function will return
>      * the `id` property from the object of each item.

So the default behavior is that each item has an id property, so no need to call `getId`

mannycarrera4 (approved):

Issue #2058, opened 2023-03-01
Table Component Refactor


jaclynjessup:
## 🚀 Feature Proposal
The current [table component](https://workday.github.io/canvas-kit/?path=%2Fstory%2Fcomponents-containers-table--basic) requires updates for current needs. The outcome will be a styled Table component within Preview.

## AC
- [ ] Table Prototype during Design Sprint
- [ ] Polish Table compound component once Design specs are complete
- [ ] Testing
- [x] CK Site Documentation
- [ ] Update Canvas site docs
- [x] Upgrade Guide
- [ ] Review our quality checks (design, a11y, documentation, engineering, developer experience)

PR #2118, opened 2023-03-28
feat: Tables Component

josh-bagwell:
## Summary

New `Tables` component.

Fixes: #2058  

## Release Category
Components

### Release Note
We've built a new `Table` compound component that will have more flexibility and has access to lower level components like [Grid](https://canvas.workdaydesign.com/components/layout/grid) and [Flex](https://canvas.workdaydesign.com/components/layout/flex).  

---

## Checklist

## For the Reviewer

`/modules/preview-react/table/lib/Table.tsx`

- [ ] Code
- [ ] Documentation


modules/preview-react/table/index.ts
```diff
@@ -0,0 +1,9 @@
+export * from './lib/Table';
+export * from './lib/TableCaption';
+export * from './lib/TableDivider';
+export * from './lib/TableBody';
+export * from './lib/TableCell';
+export * from './lib/TableFooter';
+export * from './lib/TableHeader';
+export * from './lib/TableHead';
+export * from './lib/TableRow';
```
mannycarrera4, line 9: Is there anything else exported that might be useful on the files that aren't Table? Table itself attaches all the other sub components and I'm not sure if you can use `TableFooter` without the parent `Table` component. Look at Modal's index.ts file for example or Pills index.ts

mannycarrera4: If you tried to use `TableRow` without table it would give you an error right?

modules/preview-react/table/lib/TableBody.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <Table.Row>
+          <Table.Header>Table Header</Table.Header>
+          <Table.Cell>Table Cell</Table.Cell>
+        </Table.Row>
+      </Table.Body>
+    </Table>
+  );
+}
+```
+ */
```
mannycarrera4, line 25: I'm still learning here, but if we have the same documentation at the `Tabel` level do we still need it here?

modules/preview-react/table/lib/TableCell.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import {Grid, GridProps} from '@workday/canvas-kit-react/layout';
+import {createComponent} from '@workday/canvas-kit-react/common';
+
+export interface TableCellProps extends GridProps {
+  headers?: string;
```
mannycarrera4, line 6: does this need `colspan` and `rowspan` ? Referencing: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td

modules/preview-react/table/index.ts
josh-bagwell: Yeah definitely good point. I will update this because it will give an error

modules/preview-react/table/lib/TableBody.tsx
josh-bagwell: I don't think we do. It's redundant and I will update it.

modules/preview-react/table/lib/TableCell.tsx
josh-bagwell: So colspan and rowspan are attributes on the `td` and `th` however, since `Table.Row` is built on `Grid`, that is how those will be made to span rows or columns since colspan and rowspan are just visual attributes.

modules/preview-react/table/lib/TableDivider.tsx
```diff
@@ -0,0 +1,23 @@
...
+import {createComponent} from '@workday/canvas-kit-react/common';
+
+export const TableDivider = createComponent('hr')({
+  displayName: 'Table.Divider',
+  Component({...elemProps}: BoxProps, ref, Element) {
+    return (
+      <Box
+        as={Element}
+        ref={ref}
+        border={0}
```
mannycarrera4, line 13: Does the box by default add a border? If you just have borderBottom do you still need to set this to 0?

mannycarrera4:
### Table Basic
Screen Shot 2023-04-04 at 2 28 40 PM
Should there be a borderTop here above `TableHeader` ?
### Table With Heading
Screen Shot 2023-04-04 at 2 28 46 PM
This has a border top, they both look different. Does this match the design spec?

modules/preview-react/table/stories/Tables.stories.mdx
jamesfan, line 1: Nit: this file should be renamed to `Table.stories.mdx` (we typically use the singular form in our component `.stories.mdx` files).

modules/preview-react/table/stories/Tables.stories.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+## Usage
+
+Each piece of the compound component renders an HTML element that is part of a table element:
+
+- `Table`: Renders [table](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table)
+- `Table.Caption`: Renders
+  [caption](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption)
+- `Table.Head`: Renders [thead](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/thead)
+- `Table.Body`: Renders [tbody](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tbody)
```
jamesfan, line 30: The `SymbolDoc` component from Nicholas' recent docgen work will handle a lot of this documentation in your Usage section automatically (i.e., which HTML elements each of the Table components renders and even which layout component serves as the foundation for each of those Table components).

You'll need to make the following changes to your stories MDX:
```
// Top of Table.stories.mdx
import {SymbolDoc} from '@workday/canvas-kit-docs';

...

// Bottom of Table.stories.mdx
## Component API

<SymbolDoc name="Table" fileName="/preview-react/" />
```

It's customary to include the `## Component API` section after the `## Usage` section. See any of our other stories.mdx files for an example (such as [Modal.stories.mdx](https://github.com/Workday/canvas-kit/blame/master/modules/react/modal/stories/Modal.stories.mdx)).

modules/preview-react/table/lib/TableDivider.tsx
josh-bagwell: Yeah it does. So for the `hr` element, it will add a border all the way around the element which makes it look like a box

modules/preview-react/table/stories/Tables.stories.mdx
josh-bagwell: Updated

RayRedGoose:
How do you feel about making examples look more as a real tables instead of using repeating placeholders?

jamesfan:
> How do you feel about making examples look more as a real tables instead of using repeating placeholders?

+1 to this. I find examples to be more helpful when they model something you might see in actual use. It provides a more compelling argument to use the component; plus, it often helps us out as developers (sometimes you don't notice the issues with a component until you apply it to a real-world problem).

modules/preview-react/table/stories/Table.stories.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+## Installation
+
+```sh
+yarn add @workday/canvas-kit-preview-react
+```
+
+## Usage
+
+### Basic Example with Heading
```
jamesfan, line 25: Is there a recommendation from a11y on using a heading vs a caption? From the [W3C](https://www.w3.org/WAI/tutorials/tables/caption-summary/):

> A caption functions like a heading for a table.

Would you ever use both a heading and a caption? I ask because it would be ideal to pick a single "Basic" example (for consistency with our other component docs) and then have the other option be the next example (either "With Heading" or "With Caption").

modules/preview-react/table/lib/TableDivider.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import {BoxProps, Box} from '@workday/canvas-kit-react/layout';
+import {colors} from '@workday/canvas-kit-react/tokens';
+import {createComponent} from '@workday/canvas-kit-react/common';
+
+export const TableDivider = createComponent('hr')({
```
jamesfan, line 6: Interesting, I wasn't aware you could have an `<hr />` directly inside a `<tbody>`. Doesn't seem like this is valid HTML based on the [MDN page for tbody](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tbody):

image

modules/preview-react/table/stories/Table.stories.mdx
```diff
@@ -0,0 +1,52 @@
...
+purposes.
+
+<ExampleCodeBlock code={Basic} />
+
+### Fixed Column and Column Dividers
+
+In this example, users may want to have columns divided and have a fixed left column for their
+headers so users can scroll horizontally on bigger data sets.
+
+<ExampleCodeBlock code={FixedColumn} />
```
jamesfan, line 48: I feel like column dividers warrants its own example separate from fixed columns (though it does make sense to use them together).

modules/preview-react/table/lib/TableDivider.tsx
jamesfan: @mannycarrera4 gave a little context on why you might have implemented the row dividers this way (to avoid having to write extra CSS to style the first and last rows differently). Let me know if you have any additional info on why an `<hr />` was used to implement the border between rows.

Trying to think about how I would implement the design from the spec (with the row borders not extending the full width of the container) and the only thing I'm coming up with is applying a padding to the `<tbody>` and styling the first and last rows differently to compensate.

modules/preview-react/table/stories/Table.stories.mdx
josh-bagwell: Yeah definitely. 

I don't think there is a recommendation. Captions help with accessibility but, since the table itself will be accessible since it's a semantic table element, I don't think captions are absolutely necessary. Heading is probably the way most users will go

modules/preview-react/table/lib/TableDivider.tsx
josh-bagwell: Yeah, I think the dividers might need to be removed altogether. 

The padding on `<tbody/>` doesn't work well. The issue is because when users want to implement a sticky left column, then the scrolling happens under that column and doesn't look great. 

So maybe, I will have to remove that divider altogether.

modules/preview-react/table/stories/Table.stories.mdx
josh-bagwell: Column dividers are just borders on the cells and headers. I wonder if it does need an example on it's own?

modules/preview-react/table/lib/Table.tsx
```diff
@@ -0,0 +1,279 @@
...
+import {TableRow} from './TableRow';
+import {TableHeader} from './TableHeader';
+import {TableCell} from './TableCell';
+import {TableFooter} from './TableFooter';
+import {TableCaption} from './TableCaption';
+
+/**
+ * `Table` is a simple styled compound component that renders a [table](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table) element. It is used to present information in a two-dimensional table comprised of rows and columns of cells containing data.
+ *
+ * @example
```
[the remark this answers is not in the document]
josh-bagwell: Confirmation of this, Nicholas did mention that this should be removed and not used

josh-bagwell:
Doc Parser Update will need to be done
Also checking v9 upgrade guide

modules/preview-react/table/lib/TableCaption.tsx
```diff
@@ -0,0 +1,23 @@
...
+  displayName: 'Table.Caption',
+  Component: ({children, ...elemProps}: FlexProps, ref, Element) => {
+    return (
+      <Flex
+        as={Element}
+        ref={ref}
+        borderBottom={`1px solid ${colors.soap500}`}
+        paddingY="xxs"
+        paddingX="s"
+        textAlign="left"
```
mannycarrera4, line 16: I don't think you need textAlign here

modules/preview-react/table/lib/TableBody.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const StyledBody = styled(Grid.as('tbody'))<StyledType>({
+  'tr:last-of-type td': {
+    borderBottom: 'none',
+  },
+  'tr:last-of-type th': {
+    borderBottom: 'none',
+  },
+  'td:last-of-type': {
+    borderRight: 'none',
```
mannycarrera4, line 13: do you need this borderRight? I'm trying to test it and I don't see it impacting anything?

mannycarrera4: If you're adding just border bottom and the table has border all around, I don't think you need to get border right to 0?

modules/preview-react/table/lib/TableCell.tsx
```diff
@@ -0,0 +1,28 @@
...
+        as={Element}
+        ref={ref}
+        alignItems="center"
+        backgroundColor="frenchVanilla100"
+        borderBottom={`1px solid ${colors.soap400}`}
+        justifyContent="left"
+        minHeight="48px"
+        paddingY="xxs"
+        paddingX="s"
+        textAlign="left"
```
mannycarrera4, line 20: same thing here, if you have justifycontent left, do you need text align left as well?

modules/preview-react/table/lib/TableHead.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import {GridProps, Grid} from '@workday/canvas-kit-react/layout';
+import {createComponent, styled, StyledType} from '@workday/canvas-kit-react/common';
+
+const StyledHead = styled(Grid.as('thead'))<StyledType>({
+  'th:last-of-type': {
+    borderRight: 'none',
```
mannycarrera4, line 7: do you need borderRight none here?

modules/preview-react/table/lib/TableHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+        as={Element}
+        ref={ref}
+        alignItems="center"
+        backgroundColor="frenchVanilla100"
+        borderBottom={`1px solid ${colors.soap400}`}
+        justifyContent="left"
+        minHeight="48px"
+        paddingY="xxs"
+        paddingX="s"
+        textAlign="left"
```
mannycarrera4, line 20: same thing, I don't think you need textAlign left

modules/preview-react/table/stories/Table.stories.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+<ExampleCodeBlock code={BasicWithHeading} />
+
+### Example with Container
+
+In this example, users may want to have some padding around the `table` and have inset dividers for
+their rows. This showcases how a user can achieve this with creating a styled component to wrap the
+`table`.
+
+<ExampleCodeBlock code={TableWithContainer} />
```
mannycarrera4, line 41: You removed this component no? I don't think we need this anymore

modules/preview-react/table/stories/Table.stories.mdx
```diff
@@ -0,0 +1,62 @@
...
+In this example, we built a simple table that includes a `caption` without a Heading. A `caption` is
+not required but it is good for
+[accessibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table#accessibility_concerns)
+purposes.
+
+<ExampleCodeBlock code={Basic} />
+
+### Fixed Column with TableContainer
+
+In this example, users may want to have some padding surrounding their table and have a fixed
```
mannycarrera4, line 54: You can remove `in this example`. It's already under usage.

modules/preview-react/table/stories/examples/TableWithContainer.tsx
```diff
@@ -0,0 +1,48 @@
...
+            <Table.Row>
+              <Table.Cell>Basil</Table.Cell>
+              <Table.Cell>10 Leaves</Table.Cell>
+            </Table.Row>
+          </Table.Body>
+        </Table>
+      </Container>
+    </>
+  );
+};
```
mannycarrera4, line 48: Do we still need this story? I think this just showed styling around the table.

modules/preview-react/table/stories/Table.stories.mdx
josh-bagwell: This is an example I built with a styled component to show users what it looks like with a padded container so it has inset dividers.

modules/preview-react/table/lib/TableCaption.tsx
josh-bagwell: Correct. 

I updated this

modules/preview-react/table/lib/TableBody.tsx
josh-bagwell: So what this is doing is, it's removing the `border-right` for the last cells of a row. 

So if a user was to apply a `border-right` so have vertical dividers between cells, then this would remove the border from the right of the last cell so there isn't a double border.

josh-bagwell: This will also be updated to include left and right borders

modules/preview-react/table/lib/TableCell.tsx
josh-bagwell: Correct. This has been updated.

modules/preview-react/table/lib/TableHead.tsx
josh-bagwell: So like above, this would be removing the last cell's `border-right`. This will be updated for left and right just in case.

modules/preview-react/table/lib/TableHeader.tsx
josh-bagwell: Correct. Updated.

modules/preview-react/table/stories/Table.stories.mdx
josh-bagwell: Updated

modules/preview-react/table/stories/examples/TableWithContainer.tsx
josh-bagwell: This was requested by design so I created something to showcase the inset dividers for the table.

RayRedGoose (approved):
Love this real data for tables! Thank you for adding it.

Issue #2126, opened 2023-03-31
Stack Codemod Typos


josh-bagwell:
## 🐛 Bug Report

Typos in Stack Codemod.

PR #2127, opened 2023-03-31
fix: Stack Codemod Fix

josh-bagwell:
## Summary
Some typos/copy paste issues were caught. This is a PR to fix them. 

Fixes: #2126  

## Release Category
Codemods
---

### DIFF
--- a/modules/codemod/lib/v9/spec/deprecateStack.spec.ts
+++ b/modules/codemod/lib/v9/spec/deprecateStack.spec.ts
@@ -88,7 +88,7 @@ describe('Canvas Kit Deprecate Stack Codemod', () => {
       expectTransform(input, expected);
     });
 
-    it('should properly transform mutliple imports from @workday/canvas-kit-react/layout into one', () => {
+    it('should properly transform multiple imports from @workday/canvas-kit-react/layout into one', () => {
       const input = `
         import {Grid,
           Stack,
@@ -203,7 +203,7 @@ describe('Canvas Kit Deprecate Stack Codemod', () => {
       expectTransform(input, expected);
     });
 
-    it('should properly transform VStack JSX identifiers and add flexDirection attr', () => {
+    it('should properly transform VStack JSX identifiers and add flexDirection prop', () => {
       const input = `
       import {VStack} from '@workday/canvas-kit-react/layout';
 
@@ -260,7 +260,7 @@ describe('Canvas Kit Deprecate Stack Codemod', () => {
       expectTransform(input, expected);
     });
 
-    it('should properly transform StackProps type reference identifiers', () => {
+    it('should properly transform spacing prop only on Stack JSX Elements', () => {
       const input = `
       import { Stack } from '@workday/canvas-kit-react/layout';
       import { CustomCard } from '../components/CustomCard';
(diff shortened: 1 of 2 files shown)


mannycarrera4 (approved):

modules/codemod/lib/v9/spec/deprecateStack.spec.ts
```diff
@@ -317,7 +317,7 @@ describe('Canvas Kit Deprecate Stack Codemod', () => {
       expectTransform(input, expected);
     });
 
-    it('should properly transform the spacing prop on specific components that extend Stack', () => {
+    it('should properly transform the spacing prop on the Breadcrumbs component that extends Stack', () => {
```
jamesfan, line 320: Maybe I'm missing something, but I'm confused about this test. As far as I can tell, there are no Breadcrumbs subcomponents that extend `Stack` in v8. I see in the [v8 Breadcrumbs docs](https://workday.github.io/canvas-kit/?path=/docs/components-navigation-breadcrumbs--basic) that `Breadcrumbs.List` and `Breadcrumbs.Item` extend `Flex`, but I don't see any subcomponents that extend `Stack`.

If this is true, then the codemod shouldn't be touching canvas-kit Breadcrumbs at all (and this test doesn't need to exist).

modules/codemod/lib/v9/spec/deprecateStack.spec.ts
```diff
@@ -336,7 +336,7 @@ describe('Canvas Kit Deprecate Stack Codemod', () => {
       expectTransform(input, expected);
     });
 
-    it('should properly transform the spacing prop on specific components that extend Stack', () => {
+    it('should properly transform the spacing prop on the Breadcrumbs component that extends Stack and transform Stack component', () => {
```
jamesfan, line 339: Is this test supposed to ensure the codemod does _not_ modify Breadcrumbs that do _not_ come from canvas-kit? Though, if you look at my earlier comment, I'm not sure the codemod should be modifying Breadcrumbs from canvas-kit either.

modules/codemod/lib/v9/spec/deprecateStack.spec.ts
```diff
@@ -361,7 +361,7 @@ describe('Canvas Kit Deprecate Stack Codemod', () => {
       expectTransform(input, expected);
     });
 
-    it('should properly transform the spacing prop on specific components that extend Stack', () => {
+    it('should properly transform the spacing prop on the ActionBar component that extends Stack', () => {
```
jamesfan, line 364: This test makes more sense to me since `ActionBar.List` _does_ extend `HStack`. However, the `ActionBar` container component doesn't extend `Stack` so it shouldn't have a `spacing` in the test input (or a `gap` in the test output).

jamesfan (changes requested):
Hey @josh-bagwell! Left a few comments about the tests related to Breadcrumbs which don't make sense to me (see below). I double-checked my feedback, but maybe I'm missing something -- let me know.

modules/codemod/lib/v9/spec/deprecateStack.spec.ts
```diff
@@ -203,7 +203,7 @@ describe('Canvas Kit Deprecate Stack Codemod', () => {
       expectTransform(input, expected);
     });
 
-    it('should properly transform VStack JSX identifiers and add flexDirection attr', () => {
+    it('should properly transform VStack JSX identifiers and add flexDirection attribute', () => {
```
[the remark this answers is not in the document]
josh-bagwell: Yes it is. Thanks for catching this! 

I think I was referring to the codemod portion with AST

modules/codemod/lib/v9/spec/deprecateStack.spec.ts
josh-bagwell: I have removed breadcrumbs from the array for checking those imports

modules/codemod/lib/v9/spec/deprecateStack.spec.ts
josh-bagwell: Updated

jamesfan (approved):
Looks good 👍

Issue #1919, opened 2022-11-17
Popup Card maxHeight breaks calc when multiple values are provided


alanbsmith:
## 🐛 Bug Report

When multiple values are provided to the Popup Card's margin prop, it breaks [the `calc` function](https://github.com/Workday/canvas-kit/blob/master/modules/react/popup/lib/PopupCard.tsx#L87-L89) used in the `maxHeight` property and invalidates the style.

E.g.
```tsx
// valid use of the margin prop
<Popup.Card margin="32px 24px">

// but it produces this which is not valid.
maxHeight: calc(100vh - 32px 24px * 2)
```

## To Reproduce

```tsx
<Popup.Card margin="32px 24px">
```

## Expected Behavior

You should be able to add a valid style to the margin shorthand prop and `maxHeight` not invalidated.

## Actual Results

`maxHeight` is invalidated.

### Reproduction

[CodeSandbox](https://codesandbox.io/s/v7-overflow-modal-ye2ze7?file=/src/OverflowModal.tsx)

thunguyen19:
@alanbsmith When there are more than 2 values for `margin`, which value should we respect? `top`, `right`, `bottom`, `left`?

PR #2130, opened 2023-04-01
fix: PopupCard maxHeight calc function

thunguyen19:
## Summary

Fixes: #1919 

Fixed the bug in PopUpCard where multiple margin value broke `maxHeight` calculation.

## Release Category
Components

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods

### DIFF
--- a/modules/react/popup/lib/PopupCard.tsx
+++ b/modules/react/popup/lib/PopupCard.tsx
@@ -71,9 +71,18 @@ export const PopupCard = createSubcomponent('div')({
     return getTransformFromPlacement(model.state.placement || 'bottom');
   }, [model.state.placement]);
 
+  const getMargin = (margin?: string | number) => {
+    if (margin) {
+      const result = space[margin as CanvasSpaceKeys];
+      // Get the top margin from "25px 50px" or "10px 15px 20px" format.
+      const marginTop = typeof margin === 'string' ? margin.split(' ')[0] : margin;
+      return result ? result : marginTop;
+    }
+    return space.xl;
+  };
+
   // As is a Flex that will render an element of `Element`
   const As = useConstant(() => Flex.as(Element));
-
   return (
     <StyledPopupCard
       as={As}
@@ -84,9 +93,7 @@ export const PopupCard = createSubcomponent('div')({
       flexDirection="column"
       minHeight={0}
       padding="m"
-      maxHeight={`calc(100vh - ${
-        elemProps.margin ? space[elemProps.margin as CanvasSpaceKeys] || elemProps.margin : space.xl
-      } * 2)`}
+      maxHeight={`calc(100vh - ${getMargin(elemProps.margin)} * 2)`}
       overflowY="auto"
       {...type.levels.subtext.large}
       {...elemProps}
--- a/modules/react/popup/lib/PopupCard.tsx
+++ b/modules/react/popup/lib/PopupCard.tsx
@@ -70,7 +70,7 @@ function getSpace(value?: string | number) {
   }
 }
 
-function getMaxHeight(margin: PopupCardProps['margin']) {
+function getMaxHeight(margin?: string | number) {
   // set the default margin offset to space.xl
   let marginOffset: string | number = space.xl;
 
--- a/modules/react/popup/lib/PopupCard.tsx
+++ b/modules/react/popup/lib/PopupCard.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import {keyframes} from '@emotion/react';
 
 import {Card} from '@workday/canvas-kit-react/card';
-import {space, type, CanvasSpaceKeys} from '@workday/canvas-kit-react/tokens';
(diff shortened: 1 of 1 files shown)


modules/react/popup/lib/PopupCard.tsx
```diff
@@ -71,9 +71,18 @@ export const PopupCard = createSubcomponent('div')({
...
+  const getMargin = (margin?: string | number) => {
+    if (margin) {
+      const result = space[margin as CanvasSpaceKeys];
+      // Get the top margin from "25px 50px" or "10px 15px 20px" format.
+      const marginTop = typeof margin === 'string' ? margin.split(' ')[0] : margin;
+      return result ? result : marginTop;
+    }
+    return space.xl;
+  };
+
```
thunguyen19, line 83: Retrieve the `top` margin when multiple values are provided. I don't know what is the best way to achieve this unless we create a new prop just for the `maxHeight` calculation and users would have to explicitly pass that in.
Also for this function, I assumed that we wanted the `top` value. Let me know if this needs to be changed otherwise.

alanbsmith: I appreciate you picking this up, @thunguyen19. 😄 I think this is generally the right idea, but I have some suggestions.

- Let's pull this function outside of the `PopupCard` – it doesn't need to live inside the component render cycle.
– Since this function is only used by `maxHeight`, let's pull all the calculations into one spot
– I think `marginTop` is a good rough proxy, but let's make this a little more robust to account for `marginBottom` was well.

alanbsmith: I wrote this somewhat quickly, so it might not work perfectly, but have a look at this:

```ts
function getSpace(value?: string | number) {
  if (value && value in space) {
    return space[value as keyof typeof space]
  } else {
    return value;
  }
};

// This does not account for explicit
function getMaxHeight(margin: PopupCardProps['margin']) {
  // set the default margin offset to space.xl
  let marginOffset: string | number = space.xl;

  if (margin) {
    // parse the margin prop
    if (typeof margin === 'string') {
      const marginValues = margin.split(' ');
      const marginTop = getSpace(marginValues[0]);
      // If provided, use the specific margin-bottom in the shorthand, otherwise use the margin-top value
      const marginBottom = getSpace(marginValues[2]) || marginTop;

      marginOffset = `calc(${marginTop} + ${marginBottom})`;
    } else {
      // if margin is a number, double it to get the offset
      marginOffset = margin * 2;
    }
  }
  // return the offset value
  return `calc(100vh - ${marginOffset})`;
}
```

modules/react/popup/lib/PopupCard.tsx
```diff
@@ -62,6 +62,36 @@ const StyledPopupCard = styled(Card)<
...
 
+function getSpace(value?: string | number) {
+  if (value && value in space) {
+    return space[value as keyof typeof space];
+  } else {
+    return value;
+  }
+}
+
+// This does not account for explicit
```
alanbsmith, line 73:
```suggestion
// This does not account for explicit 
```
```suggestion
```

alanbsmith (approved):
Thank you so much, @thunguyen19!

thunguyen19:
Tested with these examples:
```
<Popup.Card margin="20px">
maxHeight=calc(100vh - (20px + 20px))

<Popup.Card margin="20px 30px 25px">
maxHeight=calc(100vh - (20px + 25px));

<Popup.Card margin="25%">
maxHeight=calc(100vh - (25% + 25%))

<Popup.Card margin={20}>
maxHeight=calc(100vh - 40px)
```

modules/react/popup/lib/PopupCard.tsx
```diff
@@ -62,6 +62,35 @@ const StyledPopupCard = styled(Card)<
...
+    if (typeof margin === 'string') {
+      const marginValues = margin.split(' ');
+      const marginTop = getSpace(marginValues[0]);
+      // If provided, use the specific margin-bottom in the shorthand, otherwise use the margin-top value
+      const marginBottom = getSpace(marginValues[2]) || marginTop;
+
+      marginOffset = `(${marginTop} + ${marginBottom})`;
+    } else {
+      // if margin is a number, double it to get the offset
+      marginOffset = `${margin * 2}px`;
```
thunguyen19, line 88: Getting yarn typecheck error because `margin` is not a number type. 
```
modules/react/popup/lib/PopupCard.tsx:88:7 - error TS2362: The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.

88       marginOffset = `${margin * 2}px`;
                           ~~~~~~
Found 1 error.
```
Can we do `Numeric(margin)` here or update the type of margin to string | numeric? Any suggestion?

alanbsmith: Yeah, TS is getting confused. I think it's easier to update the type of margin at the top.

PR #2168, opened 2023-04-26
fix: Only pass correct model to model components when using 'as'

NicholasBoll:
## Summary

Sometimes we need to compose model-based components using the `as`. If this happens, the `model` prop is in conflict. This change updates both Typescript and the runtime to understand this case and not pass an incompatible model to a component that cannot handle it.

## Release Category
Components

---

## Checklist

## For the Reviewer

Start with the tests.

- [x] Code
- [x] Testing


modules/react/common/spec/components.spec.tsx
```diff
@@ -481,3 +511,101 @@ describe('composeHooks', () => {
...
+
+      expect(result.current).toEqual(model);
+    });
+
+    it('should return the model if the `as` is a regular component', () => {
+      const Component = () => null;
+      const {result} = renderHook(() => useModelContext(context, model, Component));
+
+      expect(result.current).toEqual(model);
+    });
```
josh-bagwell, line 563: Duplicate
```suggestion
```

modules/react/common/spec/components.spec.tsx
```diff
@@ -481,3 +511,101 @@ describe('composeHooks', () => {
...
+
```
josh-bagwell, line 602: Duplicate
```suggestion
```

modules/react/common/spec/components.spec.tsx
```diff
@@ -481,3 +511,101 @@ describe('composeHooks', () => {
...
```
NicholasBoll, line 564:
```suggestion
```

josh-bagwell (approved):

PR #2182, opened 2023-04-30
feat(text-input): Add InputGroup component

NicholasBoll:
## Summary

Add an `InputGroup` component to handle icons in a `TextInput` component.

There is already an `InputIconContainer` component that is not documented, but is limited to only right-aligned icons and doesn't handle RTL.

## Release Category
Components

### Release Note
`InputGroup` will replace `InputIconContainer`. `InputIconContainer` does not handle bidirectionality or icons at the start of an input. `InputIconContainer` will be deprecated and later removed in future versions.

Before:
```tsx
<InputIconContainer icon={<SystemIcon icon={exclamationCircleIcon} />} />
```

After
```tsx
<InputGroup>
  <InputGroup.Input />
  <InputGroup.InnerEnd>
    <SystemIcon icon={exclamationCircleIcon} />
  </InputGroup.InnerEnd>
</InputGroup>
```

---

## Checklist

## For the Reviewer

- [x] Code
- [x] Documentation
- [x] Testing

image


modules/react/common/lib/utils/dispatchInputEvent.ts
```diff
@@ -0,0 +1,16 @@
...
+    const nativeInputValue = Object.getOwnPropertyDescriptor(Object.getPrototypeOf(input), 'value');
+    if (nativeInputValue && nativeInputValue.set) {
+      nativeInputValue.set.call(input, value);
+    }
+
+    const event = new Event('input', {bubbles: true});
+
+    input.dispatchEvent(event);
+  }
+}
```
NicholasBoll, line 16: This seemed like a useful function for sending change info to an input. It could go in the `text-input` package, but it is useful for any input-like element.

modules/react/form-field/stories/visual-testing/stories_TextInput.tsx
```diff
@@ -76,3 +80,44 @@ TextInputThemedStates.parameters = {
     theme: customColorTheme,
   },
 };
+
+export const InputGroupStates = () => (
```
NicholasBoll, line 84: RTL example:
 https://5d854c26ba934e0020f5e98a-wgbwrnfmtx.chromatic.com/?path=/story/testing-inputs-text-input--input-group-states
 
 
image

modules/react/text-input/spec/InputGroup.spec.tsx
```diff
@@ -0,0 +1,28 @@
...
+    const ssrRender = () =>
+      renderToString(
+        <InputGroup>
+          <InputGroup.Start pointerEvents="none">
+            <SystemIcon icon={searchIcon} size="small" />
+          </InputGroup.Start>
+          <InputGroup.Input />
+          <InputGroup.End>
+            <TertiaryButton role="presentation" icon={xSmallIcon} size="small" tabIndex={-1} />
+          </InputGroup.End>
```
alanbsmith, line 22: Seems like adding a `tabIndex={-1}` is going to be an important caveat for teams.

alanbsmith (changes requested):
A few nits, but overall, this looks good. Thanks, Nicholas.

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -0,0 +1,137 @@
...
+export const InputGroup = createComponent('div')({
+  displayName: 'InputGroup',
+  Component({children, ...elemProps}: ExtractProps<typeof Flex>, ref, Element) {
+    const isRTL = useIsRTL();
+    let paddingInlineStart: number;
+    let paddingInlineEnd: number;
+
+    React.Children.forEach(children, child => {
+      if (React.isValidElement<any>(child) && child.type === InputGroupStart) {
+        paddingInlineStart = (paddingInlineStart || 0) + (child.props.width || 40);
```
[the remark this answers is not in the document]
NicholasBoll: Do multiple layers of `calc` work?

modules/react/text-input/spec/InputGroup.spec.tsx
NicholasBoll: I was thinking about introducing a `InputGroup.ClearButton` that has this built in, including clearing the input when clicked.

We don't have specification for it yet. The Preview combobox has this button as focusable. I've also seen this be a non-focusable button because there's already keyboard functionality to clear an input. In my opinion, it should not be focusable via the tab keys. Our accessibility SMEs tend to agree on less focus stops of non-important elements. It makes the interface less predictable.

modules/react/text-input/stories/examples/Icons.tsx
```diff
@@ -0,0 +1,51 @@
...
+      </InputGroup.Start>
+      <InputGroup.Input ref={inputRef} onChange={handleChange} {...props} />
+      <InputGroup.End>
+        <TertiaryButton
+          role="presentation"
+          icon={xSmallIcon}
+          size="small"
+          tabIndex={-1}
+          transition="opacity 300ms ease"
+          style={{opacity: value ? 1 : 0}}
```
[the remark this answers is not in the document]
NicholasBoll: I don't think we should encourage a style prop for styles that change during the lifecycle of the component. Changing the `opacity` will trigger a style recalculation that the browser cannot predictably cache the result.

Here's a performance profile of using the `style` attribute:
image


Here's a performance profile of using the `opacity` style attribute (uses Emotion):
image

The notable difference is the "Recalculate Style". The  `style` attribute has a time of 0.18ms while the `opacity` style prop has a time of 2.24ms. The latter timing will depend on the amount of elements and style selectors on the page. This example only has a single element + Storybook styling and already the difference is significant. I saw an example of the Workday Sheets product took 80ms of style recalc.

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -0,0 +1,137 @@
...
+});
+
+export const InputGroupInput = createComponent(TextInput)({
+  Component(elemProps: ExtractProps<typeof Flex>, ref, Element) {
+    return <Flex ref={ref} as={Element} width="100%" {...elemProps} />;
+  },
+});
+
+/**
+ * An `InputGroup` is an Input with optional start and end elements. The start and end elements are
```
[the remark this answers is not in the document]
NicholasBoll: I think it should be `TextInput`, but the wording could be better.

modules/react/text-input/lib/InputGroup.tsx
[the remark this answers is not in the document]
NicholasBoll: `paddingInlineStart` is a number. I'll try out these changes locally to make sure everything works.

> Should we set the default for the width prop to space.xl here? Would that save you a lot of || logic?

No, because `child.props` is what's being passed to the child, not the default props of the child, which is why the `|| 40` is repeated here.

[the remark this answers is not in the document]
NicholasBoll: Multiple layers of `calc` work: `calc(calc(0px + 40px) + 40px);`

NicholasBoll:
I've updated to use `calc` based on @alanbsmith's suggestions. I've simplified the logic and added an extra visual test to ensure multiple inline elements are allowed.

image

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -0,0 +1,137 @@
+import React from 'react';
+
+import {createComponent, ExtractProps, useIsRTL} from '@workday/canvas-kit-react/common';
+import {Flex} from '@workday/canvas-kit-react/layout';
+
```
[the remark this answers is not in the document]
NicholasBoll: It is added at the top now.

modules/react/form-field/stories/visual-testing/stories_TextInput.tsx
NicholasBoll: Updated:

image

modules/react/text-input/lib/InputIconContainer.tsx
```diff
@@ -19,6 +19,7 @@ const IconContainer = styled('div')({
   right: space.xxs,
 });
 
+/** @deprecated Use `InputGroup` instead. */
```
alanbsmith, line 22: Remind me to put something about this in the release guide, so My doesn't have a surprise. 😄

modules/react/text-input/spec/InputGroup.spec.tsx
alanbsmith: I think this is fine for now. Just seems like an area we could improve in the future.

modules/react/text-input/stories/examples/Icons.tsx
```diff
@@ -0,0 +1,143 @@
...
+      </InputGroup.Start>
+      <InputGroup.Input ref={inputRef} onChange={handleChange} {...props} />
+      <InputGroup.End>
+        <TertiaryButton
+          role="presentation"
+          icon={xSmallIcon}
+          size="small"
+          tabIndex={-1}
+          transition="opacity 300ms ease"
+          opacity={value ? 1 : 1}
```
alanbsmith, line 72: Did you decide not to use the `style` attr here? And I think this should be:
```tsx
opacity={value ? 1 : 0}
```

alanbsmith (approved):
Appreciate the updates. 👍🏼

modules/react/text-input/lib/InputIconContainer.tsx
NicholasBoll: I'll deprecate in 9.1 instead and not change v9 comms

modules/react/text-input/stories/examples/Icons.tsx
NicholasBoll: This was my performance test story. I didn't mean to commit it. I tried a style prop, a style attribute, and a data attribute to test timing and demonstrate elements effected.

The browser does optimize over time. For example, the data attribute has 2 elements effected the first time, but caches the result so that the second time only 1 element is effected. The style prop is 14 effected the first time and 5 the second time. As the page changes, these optimizations are flushed, so it is still better to avoid Emotion for state-based style changes.

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -0,0 +1,155 @@
...
+export const InputGroup = createComponent('div')({
+  displayName: 'InputGroup',
+  Component({children, ...elemProps}: ExtractProps<typeof Flex>, ref, Element) {
+    const isRTL = useIsRTL();
+    let paddingInlineStart: string;
+    let paddingInlineEnd: string;
+    const offsetsStart: string[] = [];
+    const offsetsEnd: string[] = [];
+
+    React.Children.forEach(children, child => {
```
mannycarrera4, line 82: This is offsetting the padding based on if there's an element at the start or end so that the input has the correct padding/size instead of it being overlapped by the position absolute elements.

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    return (
+      <Flex ref={ref} as={Element} position="relative" {...elemProps}>
+        {mappedChildren}
+      </Flex>
+    );
+  },
+  subComponents: {
+    /**
+     * A component to show at the start of the input. The input's padding will be adjusted to not
```
mannycarrera4, line 138: not sure if we want to mention the word "inside" here. Cause start can be outside.

modules/react/text-input/stories/TextInput.stories.mdx
```diff
@@ -81,6 +82,22 @@ Labels for required fields are suffixed by a red asterisk.
 
 <ExampleCodeBlock code={Required} />
 
+### Icons
+
+`IconGroup` is available to add icons to the `TextInput`. Internally, a container `div` element is
```
mannycarrera4, line 87: Did you mean `InputGroup` ?

modules/react/text-input/stories/examples/Icons.tsx
```diff
@@ -0,0 +1,143 @@
...
+      <InputGroup.Input ref={inputRef} onChange={handleChange} {...props} />
+      <InputGroup.End pointerEvents="none">
+        <SystemIcon icon={mailIcon} size="small" />
+      </InputGroup.End>
+      <InputGroup.End>
+        <TertiaryButton
+          role="presentation"
+          icon={xSmallIcon}
+          size="small"
+          tabIndex={-1}
```
mannycarrera4, line 36: There's some context here as to why we don't want this to be focusable right?

modules/react/text-input/stories/TextInput.stories.mdx
NicholasBoll: No. An `InputGroup` adds icons "inside" `TextInput` component

mannycarrera4: where is icongroup as a component?

modules/react/text-input/stories/examples/Icons.tsx
NicholasBoll: I added a comment

modules/react/text-input/lib/InputGroup.tsx
NicholasBoll: I added a comment

modules/react/text-input/lib/InputGroup.tsx
NicholasBoll: I updated

modules/react/text-input/stories/TextInput.stories.mdx
NicholasBoll: Oh. I thought you meant `TextInput` should be `InputGroup`. You can use the suggestion feature to help narrow it down.
```suggestion
`InputGroup` is available to add icons to the `TextInput`. Internally, a container `div` element is
```

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -0,0 +1,167 @@
...
+  }
+  if (values.length === 1) {
+    return values[0];
+  }
+  return `calc(${values.map(toPx).join(' + ')})`;
+};
+
+/**
+ * An `InputGroup` is a container around a {@link TextInput} with optional start and end elements.
+ * The start and end elements are usually icons or icon buttons. The `InputGroup` will add padding
```
mannycarrera4, line 68:
```suggestion
 * The start and end elements are usually icons or icon buttons within the input. The `InputGroup` will add padding
```

modules/react/text-input/stories/examples/Icons.tsx
```diff
@@ -0,0 +1,143 @@
...
+      </InputGroup.End>
+    </InputGroup>
+  );
+};
+
+const CustomButton = styled(TertiaryButton)({
+  opacity: 0,
+  '[data-has-value] &': {
+    opacity: 1,
+  },
```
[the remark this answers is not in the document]
NicholasBoll: You can select a range of lines by holding the shift key to make sure the comment is considered outdated when fixed.

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -0,0 +1,167 @@
...
+    return undefined;
+  }
+  if (values.length === 1) {
+    return values[0];
+  }
+  return `calc(${values.map(toPx).join(' + ')})`;
+};
+
+/**
+ * An `InputGroup` is a container around a {@link TextInput} with optional start and end elements.
```
mannycarrera4, line 67:
```suggestion
 * An `InputGroup` is a container around a {@link TextInput} with optional inner start and inner end elements.
```

mannycarrera4: no idea if that suggestion makes sense, but trying to be consistent with the component name

modules/react/text-input/stories/TextInput.stories.mdx
```diff
@@ -81,6 +82,22 @@ Labels for required fields are suffixed by a red asterisk.
 
 <ExampleCodeBlock code={Required} />
 
+### Icons
+
+`InputGroup` is available to add icons to the `TextInput`. Internally, a container `div` element is
+used with relative position styling on the div and absolute position styling on the start and end
```
mannycarrera4, line 88:
```suggestion
used with relative position styling on the `div` and absolute position styling on the start and end
```

modules/react/text-input/stories/TextInput.stories.mdx
```diff
@@ -81,6 +82,22 @@ Labels for required fields are suffixed by a red asterisk.
...
+icons. `InputGroup.Start` and `InputGroup.End` are used to position elements at the start and end of
+the input. "start" and "end" are used instead of "left" and "right" to match
+[CSS Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties)
+and will be semantically correct in left-to-right and right-to-left languages.
+
+`InputGroup.Start` and `InputGroup.End` subcomponents can handle any child elements, but are built
```
mannycarrera4, line 94:
```suggestion
`InputGroup.InnerStart` and `InputGroup.InnerEnd` subcomponents can handle any child elements, but are built
```

modules/react/text-input/stories/TextInput.stories.mdx
```diff
@@ -81,6 +82,22 @@ Labels for required fields are suffixed by a red asterisk.
...
+for icons. The default width is `40px`, which is perfect for icons. If you need to use something
+else, be sure to set the `width` property of `InputGroup.Start` or `InputGroup.End` to match the
```
mannycarrera4, line 96:
```suggestion
else, be sure to set the `width` property of `InputGroup.InnerStart` or `InputGroup.InnerEnd` to match the
```

modules/react/text-input/lib/InputGroup.tsx
mannycarrera4: WWMA

modules/react/text-input/stories/TextInput.stories.mdx
NicholasBoll: Thanks for catching that.

PR #2190, opened 2023-05-02
fix: Return the correct props from composeHooks

NicholasBoll:
## Summary

`composeHooks` accidentally broke at some point during our compound component paradigm building. It no longer merges the output of all incoming hooks together properly. This fix changes the signature of `composeHooks` and `BehaviorHook` to allow for the correct extraction of output props.

## Release Category
Components

### BREAKING CHANGES
The type signature of `composeHooks` was changed to give more accurate return prop types. This may cause issues with Typescript if your code expected the incorrect return types.

---

## Checklist

## For the Reviewer

- [x] Code
- [x] Documentation
- [x] Testing

### DIFF
--- a/modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
+++ b/modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
@@ -29,6 +29,7 @@ any questions.
 - [Utility Updates](#utility-updates)
   - [useTheme and getTheme](#usetheme-and-gettheme)
   - [useThemedRing](#usethemedring)
+  - [composeHooks](#composehooks)
 - [Glossary](#glossary)
   - [Main](#main)
   - [Preview](#preview)
@@ -257,6 +258,38 @@ recommend using
 instead.
 
 ---
+
+### composeHooks
+
+The `composeHooks` types are now accurate. Before the types were incorrectly merged to equal `{}`.
+This also affects components created using `createContainer` or `createSubcomponent`. The
+`elemProps` type interface will now reflect all the incoming props from the hook properly. If you
+get an error when passing `elemProps` from a hook using `composeHooks`, you may get a Typescript
+error. Sometimes returning a generic object widens types and style or JSX attributes are more
+strict. This can cause problems with JSX attributes like `position` which expects values like
+`'relative' | 'absolute'` and doesn't accept a string.
+
+For example:
+
+```ts
+return {
+  position: 'relative',
+}; // { 'position': string }
+```
+
+Typescript doesn't know that this object interface cannot be mutated, so it will widen the
+`position` type to a `string` which is now allowed when you pass the prop list to a JSX element.
+You'll have to add an `as const` to either the property or the whole object to force Typescript to
+narrow the type.
+
+```ts
+return {
+  position: 'relative' as const, // forces the type to be `'relative'` instead of `string`
+} as const; // OR add `as const` here to narrow the whole object.
+```
+
+`as const` instructs Typescript the type is `readonly`. Typescript knows readonly values or objects
+cannot be changed and will therefore narrow the type for you.
 
 ## Token Updates
 
--- a/modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
+++ b/modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
@@ -277,7 +277,7 @@ return {
 }; // { 'position': string }
 ```
 
-Typescript doesn't know that this object interface cannot be mutated, so it will widen the
+TypeScript doesn't know that this object interface cannot be mutated, so it will widen the
 `position` type to a `string` which is now allowed when you pass the prop list to a JSX element.
 You'll have to add an `as const` to either the property or the whole object to force Typescript to
 narrow the type.
--- a/modules/react/common/lib/utils/components.ts
+++ b/modules/react/common/lib/utils/components.ts
@@ -705,7 +705,7 @@ export function createSubModelElemPropsHook<M extends () => Model<any, any>>(mod
   };
 }
 
-/** Simplify and speedup inference by capturing types in the signature itself */
+/** Simplify and speed up inference by capturing types in the signature itself */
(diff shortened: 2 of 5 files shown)


modules/react/common/lib/utils/components.ts
```diff
@@ -583,9 +583,7 @@ export const createComponent = <
  * })
  * ```
  */
-export const createElemPropsHook = <
-  TModelHook extends (config: any) => {state: Record<string, any>; events: Record<string, any>}
->(
+export const createElemPropsHook = <TModelHook extends (config: any) => Model<any, any>>(
```
NicholasBoll, line 586: It seemed like `Model` was cleaner

modules/react/common/lib/utils/components.ts
```diff
@@ -596,17 +594,20 @@ export const createElemPropsHook = <
...
+  return ((model, elemProps, ref) => {
+    const props = mergeProps(fn(model, ref, elemProps || ({} as any)), elemProps || ({} as any));
+    if (!props.hasOwnProperty('ref') && ref) {
+      // This is the weird "incoming ref isn't in props, but outgoing ref is in props" thing
+      props.ref = ref;
+    }
+    return props;
+  }) as BehaviorHook<
+    TModelHook extends (config: any) => infer TModel ? TModel : Model<any, any>,
+    PO
```
NicholasBoll, line 609: The only change here is the `as BehaviorHook...`. I had to wrap in a `return` which caused whitespace changes.

modules/react/common/lib/utils/components.ts
```diff
@@ -646,26 +647,26 @@ export const createElemPropsHook = <
...
       props.ref = ref;
     }
     return props;
-  };
+  }) as BehaviorHook<M, PO>;
 };
 
 /**
- * @deprecated use `subModel` instead
+ * @deprecated use `createSubModelElemPropsHook` instead
```
NicholasBoll, line 661: typo

modules/react/common/lib/utils/components.ts
```diff
@@ -718,13 +736,11 @@ export function createSubModelElemPropsHook<M extends () => Model<any, any>>(mod
...
-    model: M,
-    elemProps?: P,
-    ref?: React.Ref<R>
-  ): O & P & (R extends HTMLOrSVGElement ? {ref: React.Ref<R>} : {});
-}['bivarianceHack'];
+export interface BehaviorHook<M extends Model<any, any>, O extends {}> extends BaseHook<M, O> {
+  <P extends {}, R>(model: M, elemProps?: P, ref?: React.Ref<R>): O &
+    P &
+    (R extends HTMLOrSVGElement ? {ref: React.Ref<R>} : {});
+}
```
NicholasBoll, line 743: Removing the `bivarianceHack` part - it isn't allowed when extending

modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
```diff
@@ -258,6 +259,38 @@ instead.
...
+
+For example:
+
+```ts
+return {
+  position: 'relative',
+}; // { 'position': string }
+```
+
+Typescript doesn't know that this object interface cannot be mutated, so it will widen the
```
alanbsmith, line 280:
```suggestion
TypeScript doesn't know that this object interface cannot be mutated, so it will widen the
```

modules/react/common/lib/utils/components.ts
```diff
@@ -698,12 +699,29 @@ export function createSubModelElemPropsHook<M extends () => Model<any, any>>(mod
...
   ): BehaviorHook<ReturnType<M>, O> => {
-    return (model: ReturnType<M>, props: any, ref: React.Ref<any>) => {
+    return ((model: ReturnType<M>, props: any, ref: React.Ref<any>) => {
       return elemPropsHook(fn(model), props, ref);
-    };
+    }) as BehaviorHook<ReturnType<M>, O>;
   };
 }
 
+/** Simplify and speedup inference by capturing types in the signature itself */
```
alanbsmith, line 708:
```suggestion
/** Simplify and speed up inference by capturing types in the signature itself */
```

modules/react/common/lib/utils/components.ts
```diff
@@ -698,12 +699,29 @@ export function createSubModelElemPropsHook<M extends () => Model<any, any>>(mod
...
+interface BaseHook<M extends Model<any, any>, O extends {}> {
+  /**
+   * Capture the model type in Typescript only. Do not use in runtime!
```
alanbsmith, line 711:
```suggestion
   * Capture the model type in TypeScript only. Do not use in runtime!
```

modules/react/common/lib/utils/components.ts
```diff
@@ -698,12 +699,29 @@ export function createSubModelElemPropsHook<M extends () => Model<any, any>>(mod
...
+   *
+   * @private
+   */
+  __model: M;
+  /**
+   * Capture the hook's output type in Typescript only. Do not use in runtime! This is used to cache
```
alanbsmith, line 717:
```suggestion
   * Capture the hook's output type in TypeScript only. Do not use in runtime! This is used to cache
```

modules/react/common/lib/utils/components.ts
```diff
@@ -698,12 +699,29 @@ export function createSubModelElemPropsHook<M extends () => Model<any, any>>(mod
...
+   * and speedup the output types during inference
```
alanbsmith, line 718:
```suggestion
   * and speed up the output types during inference
```

modules/react/common/lib/utils/components.ts
```diff
@@ -698,12 +699,29 @@ export function createSubModelElemPropsHook<M extends () => Model<any, any>>(mod
...
+   *
+   * @private
+   */
+  __output: O;
+}
+
 // Typescript function parameters are contravariant while return types are covariant. This is a
```
alanbsmith, line 725:
```suggestion
// TypeScript function parameters are contravariant while return types are covariant. This is a
```

PR #2194, opened 2023-05-02
chore: Use index instead of identifier for cursor navigation

NicholasBoll:
## Summary

Virtual lists using dynamic data need to intercept navigation changes to trigger data loading. If data isn't loaded yet, an `id` isn't known yet. The navigation managers were updated to use indexes instead of ids to handle unloaded data. This should not effect most people as the navigation manager is provided by default. This is mostly internal.

## Release Category
Components

### BREAKING CHANGES
The signature of the `NavigationManaget` and `NavigationRequestor` was changed to use numeric indexes instead of string identifiers. This will only break for those who created a custom navigation manager.

---

## Checklist

## For the Reviewer

The tests are the best place to start

- [x] Code
- [x] Documentation
- [x] Testing

## Testing Manually

All code is covered by tests

### DIFF
--- a/modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
+++ b/modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
@@ -26,6 +26,7 @@ any questions.
 - [Component Updates](#component-updates)
   - [Button](#button)
   - [Toast](#toast)
+  - [Collection](#collection)
 - [Utility Updates](#utility-updates)
   - [useTheme and getTheme](#usetheme-and-gettheme)
   - [useThemedRing](#usethemedring)
@@ -291,6 +292,8 @@ return {
 `as const` instructs Typescript the type is `readonly`. Typescript knows readonly values or objects
 cannot be changed and will therefore narrow the type for you.
 
+---
+
 ## Token Updates
 
 ### Depth
@@ -377,6 +380,15 @@ previously used `actionText` or `onActionClick`. The codemod will also update im
 
 > **Note:** You will manually need to set `mode` to `alert` if your `Toast` conveys urgent and
 > important information such as an error.
+
+---
+
+### Collection
+
+Navigation was updated to use numberical indexes instead of string ids. The `model.state.cursorId`
+is left unchanged. The change is to support virtual lists where navigation knows where it needs to
+go, but the id may not be loaded yet. The mechanism for navigating is private and should not
+breaking anything. If you created a custom navigation manager, the signature has been changed.
 
 ## Utility Updates
 
--- a/modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
+++ b/modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
@@ -385,10 +385,11 @@ previously used `actionText` or `onActionClick`. The codemod will also update im
 
 ### Collection
 
-Navigation was updated to use numberical indexes instead of string ids. The `model.state.cursorId`
-is left unchanged. The change is to support virtual lists where navigation knows where it needs to
-go, but the id may not be loaded yet. The mechanism for navigating is private and should not
-breaking anything. If you created a custom navigation manager, the signature has been changed.
+Navigation was updated to use numerical indexes instead of string identifiers. The
+`model.state.cursorId` is left unchanged. The change is to support virtual lists where navigation
+knows where it needs to go, but the identifier may not be loaded yet. The mechanism for navigating
+is private and should not breaking anything. If you created a custom navigation manager, the
+signature has been changed.
 
 ## Utility Updates
 
--- a/modules/react/collection/lib/useBaseListModel.tsx
+++ b/modules/react/collection/lib/useBaseListModel.tsx
@@ -1,13 +1,13 @@
 import React from 'react';
 import {useVirtual} from './react-virtual';
 
-import {useUniqueId, assert, createModelHook, Generic} from '@workday/canvas-kit-react/common';
+import {useUniqueId, createModelHook, Generic} from '@workday/canvas-kit-react/common';
 
 export type Orientation = 'horizontal' | 'vertical';
 
 export const defaultGetId = (item: any): string => {
-  assert(item.id, 'A list item must have an `id` field or a `getId` function defined');
-  return item.id;
+  // assert(item.id, 'A list item must have an `id` field or a `getId` function defined');
+  return item.id || '';
 };
 
 export const defaultGetTextValue = (item: any): string => {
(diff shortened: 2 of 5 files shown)


modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
```diff
@@ -378,6 +381,15 @@ previously used `actionText` or `onActionClick`. The codemod will also update im
 > **Note:** You will manually need to set `mode` to `alert` if your `Toast` conveys urgent and
 > important information such as an error.
 
+---
+
+### Collection
+
+Navigation was updated to use numberical indexes instead of string ids. The `model.state.cursorId`
```
mannycarrera4, line 388:
```suggestion
Navigation was updated to use numerical indexes instead of string `id`s. The `model.state.cursorId`
```

modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
```diff
@@ -378,6 +381,15 @@ previously used `actionText` or `onActionClick`. The codemod will also update im
...
+is left unchanged. The change is to support virtual lists where navigation knows where it needs to
+go, but the id may not be loaded yet. The mechanism for navigating is private and should not
```
mannycarrera4, line 390:
```suggestion
go, but the `id` may not be loaded yet. The mechanism for navigating is private and should not
```

modules/react/collection/lib/useBaseListModel.tsx
```diff
@@ -1,13 +1,13 @@
...
 
-import {useUniqueId, assert, createModelHook, Generic} from '@workday/canvas-kit-react/common';
+import {useUniqueId, createModelHook, Generic} from '@workday/canvas-kit-react/common';
 
 export type Orientation = 'horizontal' | 'vertical';
 
 export const defaultGetId = (item: any): string => {
-  assert(item.id, 'A list item must have an `id` field or a `getId` function defined');
-  return item.id;
+  // assert(item.id, 'A list item must have an `id` field or a `getId` function defined');
```
mannycarrera4, line 9:
```suggestion
```

mannycarrera4, line 9:
```suggestion
  assert(item.id, 'A list item must have an `id` field or a `getId` function defined');
```

modules/react/collection/lib/useCursorListModel.tsx
```diff
@@ -333,6 +345,15 @@ export const useCursorListModel = createModelHook({
      * @private Use useGridModel instead to make a grid instead of a list
      */
     columnCount,
+    /**
+     * A React.Ref of the current page size. Either provided as config, or determined at runtime
+     * based on the size of the list container and the number of items fitting within the container.
+     */
+    pageSizeRef,
+    /**
+     * A
```
mannycarrera4, line 354: missing comment?

modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
NicholasBoll: Also, I guess my spell checker doesn't work on MDX files?

modules/react/collection/lib/useBaseListModel.tsx
NicholasBoll: It should be removed. We cannot make guarantees on unloaded data.

modules/docs/mdx/9.0-UPGRADE-GUIDE.mdx
mannycarrera4: can't believe you've made me correct your english

modules/react/collection/lib/useBaseListModel.tsx
mannycarrera4: ok that was my initial suggestion, just wasn't sure

modules/react/collection/lib/useBaseListModel.tsx
```diff
@@ -112,8 +115,17 @@ export const useBaseListModel = createModelHook({
 })(config => {
   const id = useUniqueId(config.id);
 
+  // Optimization to not redo items when `getId` and `getTextValue` references change. They will not
+  // likely change during the lifecycle and we don't want to recalculate items when a lamba is
+  // passed instead of a stable reference.
+  const getIdRef = React.useRef(defaultGetId);
+  const getTextValueRef = React.useRef(defaultGetTextValue);
```
mannycarrera4, line 122: nice! you're a master at refs, I feel like this would be a good topic of discussion/ learning. How to use refs more effectively

modules/react/collection/lib/useCursorListModel.tsx
```diff
@@ -69,90 +69,89 @@ export const createNavigationManager = (manager: NavigationManager) => manager;
...
-    return state.items[lastRowIndex + (currentIndex % state.columnCount)];
+    return lastRowIndex + (index % state.columnCount);
   }
-  return getLast(id, {state});
+  return getLast(index, {state});
 };
 
-const getItem: NavigationRequestor = (id, {state}) => {
-  const item = id ? state.items.find(item => getId(item) === id) : getFirst(id, {state}); // no id, return first item
+const getItem: (id: string, model: NavigationInput) => Item<Generic> = (id, {state}) => {
```
mannycarrera4, line 132: is ID still string her or number?

mannycarrera4 (approved):
minor comment, but lgtm

modules/react/collection/lib/useBaseListModel.tsx
NicholasBoll: I remember this being an issue for AngularJS as well. The rule was ["always have a 'dot' in your ngModel"](https://github.com/angular/angular.js/wiki/Understanding-Scopes).

It comes from how references and values work in JavasScript. `React.useRef` gives you access to a reference to a value - think of it as a gift that is wrapping a value. The reference to the gift itself never changes, but what's inside could be different every time. I think that's why they choose the name "ref" because it means a reference. Like the literal reference in memory that points to a mutable value.

In JavaScript, there's references and there's values. When you pass data around in JavaScript, the engine makes a decision on how the thing is passed around based on its type. Primitives like strings and numbers are passed by value (a copy) while everything else is by a reference. A reference refers to an identifier/pointer/location that then has to be looked up.

A React ref wraps everything in an object, forcing the user of that thing to do a lookup. This allows the ref to be passed via closures. Closures is how React hooks operate - a closure is a "stack" of scoped variables every time a function or scope is created. A closure access outer variables:

```ts
const foo = 'bar';

const onClick = () => {
  console.log(foo) // `foo` is "closed over" or `onClick` creates a closure that has `foo` in scope
}
```

But what happens here?
```ts
const foo = props.foo;

const onClick = React.useCallback(() => {
  console.log(foo) // what will be logged?
}, [])
```

`React.useCallback`, `React.useMemo`, `React.useEffect`, `React.useLayoutEffect` create memoized closures that are based on a dependency array of reference/value comparisons. In this example, if `props.foo` changes, the `console.log` will log out the old value (the outdated closed over value). This is why all these functions have dependency arrays so they can be re-evaluated. In the case of `use{Layout}Effect`, we want to do something when specific data changes. But this violates rules of hooks. If you create a reference using `React.useRef`, rules of hooks does not complain about a dependency missing from the dependency list because React knows `React.useRef` creates a reference wrapper:

```ts
const foo = React.useRef(props.foo);
foo.current = props.foo

const onClick = React.useCallback(() => {
  console.log(foo.current) // what will be logged?
}, [])
```

Now the reference of `onClick` is stable (will not change each render) AND the console.log will always reflect the latest changes to `props.foo`.

modules/react/collection/lib/useCursorListModel.tsx
NicholasBoll: `getItem` is given a string identifier. Everything else gets an index. In the current context, `getItem` doesn't make much sense because the `items` is always an array, so `find` can be used in its place.

PR #2202, opened 2023-05-05
feat(collection): Add support for string children

NicholasBoll:
## Summary

Add string children support to the collection system via `useListItemAllowChildStrings` elemProps hook.

## Release Category
Components

---

## Checklist

## For the Reviewer

- [x] Code
- [x] Documentation
- [x] Testing

## Testing Manually

Play with the `StringChildren` example. Also tests cover this functionality

### DIFF
--- a/modules/react/collection/index.ts
+++ b/modules/react/collection/index.ts
@@ -10,7 +10,8 @@ export * from './lib/useListItemRovingFocus';
 export * from './lib/useListItemSelect';
 export * from './lib/useListModel';
 export * from './lib/useGridModel';
-export {ListBox} from './lib/ListBox';
+export * from './lib/useListItemAllowChildStrings';
+export {ListBox, ListBoxProps} from './lib/ListBox';
 export {
   singleSelectionManager,
   multiSelectionManager,
--- a/modules/react/collection/lib/useListItemAllowChildStrings.ts
+++ b/modules/react/collection/lib/useListItemAllowChildStrings.ts
@@ -1,0 +1,37 @@
+import {createElemPropsHook} from '@workday/canvas-kit-react/common';
+import {useListModel} from '@workday/canvas-kit-react/collection';
+
+/**
+ * This elemProps hook allows for children values to be considered identifiers if the children
+ * are strings. This can be useful for autocomplete or select components that allow string values.
+ * This hook must be defined _before_ {@link useListItemRegister} because it sets the `data-id`
+ * attribute if one hasn't been defined by the application.
+ *
+ * An example might look like:
+ * ```tsx
+ * const useMyListItem = composeHooks(
+ *   // any other hooks here
+ *   useListItemSelect,
+ *   useListItemRegister,
+ *   useListItemAllowChildStrings
+ * )
+ *
+ *
+ * <MyList onSelect={({id}) => {
+ *   console.log(id) // will be "First" or "Second"
+ * }}>
+ *   <MyList.Item>First</MyList.Item>
+ *   <MyList.Item>Second</MyList.Item>
+ * </MyList>
+ * ```
+ */
+export const useListItemAllowChildStrings = createElemPropsHook(useListModel)(
+  (_, __, elemProps: {'data-id'?: string; children?: React.ReactNode} = {}) => {
+    const props: {'data-id'?: string} = {};
+    if (!elemProps['data-id'] && typeof elemProps.children === 'string') {
+      props['data-id'] = elemProps.children;
+    }
+
+    return props;
+  }
+);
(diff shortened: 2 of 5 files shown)


modules/react/collection/stories/Collection.stories.mdx
```diff
@@ -34,7 +35,7 @@ yarn add @workday/canvas-kit-react
 The `ListBox` on its own isn't very useful. It registers each item with the model. The
 `ListBox.Item` only uses the `useListItemRegister` hook which handles registration of static items
 to the model. The `ListBox` uses `useListRenderItems` which handles rendering static items as well
-as [Dynamic List](#dynamic-list) example).
+as [Dynamic List](#dynamic-list) example.
```
NicholasBoll, line 38: I'm not sure how this parenthesis was missed.

RayRedGoose (approved):
👍

modules/react/collection/index.ts
```diff
@@ -10,7 +10,8 @@ export * from './lib/useListItemRovingFocus';
 export * from './lib/useListItemSelect';
 export * from './lib/useListModel';
 export * from './lib/useGridModel';
-export {ListBox} from './lib/ListBox';
+export * from './lib/useListItemAllowChildStrings';
+export {ListBox, ListBoxProps} from './lib/ListBox';
```
willklein, line 14: I didn't see where we needed `ListBoxProps` to be added as an export. Also thought we could use `ExtractProps` on the consumer side if necessary.

modules/react/collection/spec/useListItemAllowChildStrings.spec.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+import {renderHook} from '@testing-library/react-hooks';
+import {useListModel, useListItemAllowChildStrings} from '@workday/canvas-kit-react/collection';
+
+describe('useListItemAllowChildSiblings', () => {
```
willklein, line 6:
```suggestion
describe('useListItemAllowChildStrings
```suggestion
describe('useListItemAllowChildSiblings', () => {
```', () => {
```

modules/react/collection/lib/useListItemAllowChildStrings.ts
```diff
@@ -0,0 +1,37 @@
+import {createElemPropsHook} from '@workday/canvas-kit-react/common';
+import {useListModel} from '@workday/canvas-kit-react/collection';
+
+/**
+ * This elemProps hook allows for children values to be considered identifiers if the children
+ * are strings. This can be useful for autocomplete or select components that allow string values.
+ * This hook must be defined _before_ {@link useListItemRegister} because it sets the `data-id`
```
willklein, line 7: Small thing: "because it sets" -> "it" is ambiguous, do we mean "this hook" or "that hook"? Might be best to state either "this hook" or `useListItemRegister` to be clear.

willklein: Also, we say "defined before" but it seems to be the opposite in the `composeHooks` usage. I understand that what comes later takes precedence (or do I?). I think maybe we mean something like "listed after" and maybe have added explanation that it will be applied in reverse order with `composeHooks`.
```

**Review by willklein — APPROVED** [review]
Reviewed everything. All non-blocking feedback but please review before merging in case you agree something could be better. I think there's at least one error and other things might warrant changes.

File: modules/react/collection/index.ts (hunk 2 above)
Comment (NicholasBoll) [reply modules/react/collection/index.ts:14]: I did not mean to add this change to this PR. I think it is used in the Combobox changes. I think the reason why I did this was because `ExtractProps<typeof ListBox, never>` returns the model as part of the prop interface. `ExtractProps` was made before model components existed and it isn't clear what it should return with model components.

I don't know what will be most common - wanting the `model` on the prop interface or not wanting it. In the case of a combobox component, I wanted to use the `children` and other props of the `ListBox` without defining it again, but I need a new model. That would look like this:

```ts
export ComboboxListProps extends Omit<ExtractProps<typeof ListBox, never>, 'model'> {
```

Most other components that have the same interface as `ListBox` manually type the `children`:
- https://github.com/Workday/canvas-kit/blob/99b29390ab73aa15117d6fc147c23d814ffc43d3/modules/react/tabs/lib/TabsList.tsx#L34
- https://github.com/Workday/canvas-kit/blob/99b29390ab73aa15117d6fc147c23d814ffc43d3/modules/react/breadcrumbs/lib/BreadcrumbsList.tsx#L22
- https://github.com/Workday/canvas-kit/blob/99b29390ab73aa15117d6fc147c23d814ffc43d3/modules/preview-react/segmented-control/lib/SegmentedControlList.tsx#L15

I was experimenting with a different paradigm where extending a `ListBox` component would involve extending the props instead. Thoughts?

PR #2204, opened 2023-05-07
fix: Focusable elements not always focused

TylerCharlesworth:
## Summary

Some React component libraries (notably https://www.npmjs.com/package/react-select) have refs that are focusable but are not HTMLElements. Due to the ordering of the logic currently, these components are not focused correctly. 

Example: 
We have a Canvas Kit popup modal that sets initialFocusRef to a react-select component. Since it is typed differently than Canvas Kit expects here, it does not correctly receive keyboard focus and the keyboard focus remains on the background elements (behind the modal!). This is bad for accessibility.

## Release Category
Components

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods


TylerCharlesworth:
Apologies but I don't seem to be able to add the `ready for review` label myself. Can someone help with this?

alanbsmith:
Putting this on hold until we ship V9.

modules/react/common/lib/utils/changeFocus.ts
```diff
@@ -18,7 +18,8 @@ export const changeFocus = (element: unknown) => {
 
       element.dispatchEvent(event);
     }
-
-    element.focus();
   }
+
+  // Even if the element is not an HTMLElement, we should still attempt to focus it.
+  element.focus?.();
```
NicholasBoll, line 24: What other element type are you getting? If not `HTMLElement`, is there additional instances we should test? Also, why not remove the `if` block entirely? You've fixed the focus change, but not focus styling.

TylerCharlesworth: That's a good point! Why does this function care what `element` is an instanceof at all, so long as the methods are implemented? Updated.

The [react-select](https://www.npmjs.com/package/react-select) library returns a custom ref with an extended API for ["advanced" usages](https://react-select.com/advanced#accessing-internals), including the basics that you'd expect on a standard ref (like focus):
image

NicholasBoll: The `if(element instanceof ...)` is for Typescript because the type is `unknown`. It doesn't really matter what the value is, but Typescript wants to know that what is handed to it has a callable property called 'focus'.

NicholasBoll: The `if(element instanceof ...)` is for Typescript because the type is `unknown`. It doesn't really matter what the value is, but Typescript wants to know that what is handed to it has a callable property called 'focus'.

NicholasBoll: You're change works in JavaScript, but it fails the TypeScript compiler because TypeScript still thinks `element` is `unknown` and that the methods are not callable.

We need to either cast the instances of `element` OR place a proper guard in the place of `instanceof` to grant both runtime and compiletime safety.

NicholasBoll: I've added a guard to please TypeScript and avoid any runtime errors.

mannycarrera4 (approved):

PR #2207, opened 2023-05-08
feat(text-input): Add clear icon to InputGroup

NicholasBoll:
## Summary

It is error prone to add a clear input button to an `InputGroup`. This change makes `ClearInputButton` a subcomponent of `InputGroup` where `InputGroup` internally tracks the inputRef and can match our accessibility specifications.

## Release Category
Components

---

## Checklist

## For the Reviewer

- [x] Code
- [x] Documentation
- [x] Testing

## Testing Manually

Try typing in the input of this story: https://5d854c26ba934e0020f5e98a-twbtpktbtg.chromatic.com/?path=/story/components-inputs-text-input--icons

### DIFF
--- a/jest/verifyComponent.tsx
+++ b/jest/verifyComponent.tsx
@@ -21,7 +21,7 @@ export function verifyComponent(
   Component: React.ComponentType<any>,
   {modelFn, props}: {modelFn?: Function; props: object}
 ) {
-  describe('verifyComponent', () => {
+  describe(`verifyComponent "${Component.displayName}"`, () => {
     const Test = React.forwardRef(({...elemProps}: {as?: React.ElementType}, ref) => {
       const model = modelFn?.() || null;
 
--- a/modules/react/common/lib/utils/components.ts
+++ b/modules/react/common/lib/utils/components.ts
@@ -294,6 +294,11 @@ export const createContainer = <
           ? (elemPropsHook as any)(localModel, elemProps, ref)
           : elemProps;
 
+        // make sure there's always a ref being passed, even if there are no elemProps hooks to run
+        if (ref && !finalElemProps.hasOwnProperty('ref')) {
+          finalElemProps.ref = ref;
+        }
+
         return React.createElement(
           Context.Provider,
           {value: localModel},
--- a/modules/react/text-input/lib/InputGroup.tsx
+++ b/modules/react/text-input/lib/InputGroup.tsx
@@ -17,7 +17,7 @@ import {xSmallIcon} from '@workday/canvas-system-icons-web';
 
 import {TextInput} from './TextInput';
 
-const useInputGroupModel = createModelHook({})(() => {
+export const useInputGroupModel = createModelHook({})(() => {
   const inputRef = React.useRef<HTMLInputElement>(null);
 
   return {
--- a/modules/react/text-input/lib/InputGroup.tsx
+++ b/modules/react/text-input/lib/InputGroup.tsx
@@ -2,10 +2,12 @@ import React from 'react';
 
 import {space} from '@workday/canvas-kit-react/tokens';
 import {
-  createComponent,
+  createContainer,
+  createElemPropsHook,
+  createModelHook,
+  createSubcomponent,
   dispatchInputEvent,
   ExtractProps,
-  mergeProps,
   useForkRef,
   useIsRTL,
 } from '@workday/canvas-kit-react/common';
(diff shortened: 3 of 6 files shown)


modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -1,11 +1,24 @@
...
 import {Flex} from '@workday/canvas-kit-react/layout';
+import {TertiaryButton} from '@workday/canvas-kit-react/button';
+import {xSmallIcon} from '@workday/canvas-system-icons-web';
 
 import {TextInput} from './TextInput';
 
+const InputGroupContext = React.createContext<React.RefObject<HTMLInputElement>>({
+  current: null,
+});
+
```
mannycarrera4, line 21: I'm sure this is not how you intended it to be used, but if you created a model, and have `inputGroup` use `createContainer` would that be another approach and that ways all it's sup components can uuse `createSubComponent` and you'd have access to the model?

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -42,7 +55,69 @@ export const InputGroupInnerEnd = createComponent('div')({
...
+  },
+});
+
+export interface ClearInputButtonProps {}
+
+/**
+ * A clear input button. This can be a component later.
+ */
+export const ClearInputButton = createComponent(TertiaryButton)({
+  Component(elemProps: ClearInputButtonProps, ref, Element) {
```
mannycarrera4, line 70: instead of defining probs can we have something like `ExtractProps<typeof TertiaryButton>` ?

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -42,7 +55,69 @@ export const InputGroupInnerEnd = createComponent('div')({
...
+        },
+      },
+      elemProps
+    );
+
+    return (
+      <Element
+        ref={elementRef}
+        // This element does not need to be accessible via screen reader. The user can already clear
+        // an input
```
mannycarrera4, line 103: so no need for an `aria-label` ?

modules/react/text-input/lib/InputGroup.tsx
NicholasBoll: Good point

modules/react/text-input/lib/InputGroup.tsx
NicholasBoll: It is hidden from screen readers and keyboard users because of the `tabIndex={-1}` and `role="presentation"`.

This isn't specified yet, but the keyboard already has the ability to clear the input. Clearing the input is mostly for mouse and touch users. This could change in the future, however. Accessibility likes less tab stops if they are not necessary, but there may be an argument for making the button accessible via keyboard or screen reader.

modules/react/text-input/lib/InputGroup.tsx
NicholasBoll: That is a good point. I could use a model. I was thinking about model pollution. It isn't a problem in this case, and we have ways around it.

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -42,7 +55,69 @@ export const InputGroupInnerEnd = createComponent('div')({
...
+        role="presentation"
+        icon={xSmallIcon}
+        // "small" is needed to render correctly within a `TextInput`
+        size="small"
+        // A clear input button doesn't need focus. There's already keyboard keys to clear an input
+        tabIndex={-1}
+        transition="opacity 300ms ease"
+        // Use style attribute to avoid the cost of Emotion's styling solution that causes the
+        // browser to throw away style cache. The difference can be significant for large amount of
+        // elements (could be a 80ms difference)
```
mannycarrera4, line 113: 👍

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -42,7 +55,69 @@ export const InputGroupInnerEnd = createComponent('div')({
...
+        style={{
+          opacity: inputHasValue ? 1 : 0,
+          pointerEvents: inputHasValue ? 'auto' : 'none',
```
mannycarrera4, line 116: do we have to account for if the input is `disabled` for any reason. Will auto handle this use case?

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -42,7 +55,69 @@ export const InputGroupInnerEnd = createComponent('div')({
...
+      if (input && button) {
+        input.addEventListener('input', () => {
+          setInputHasValue(!!input.value);
+        });
+      }
+    }, [inputRef, localRef]);
+
+    const props = mergeProps(
+      {
+        onMouseDown(event: React.MouseEvent) {
```
mannycarrera4, line 89:
```suggestion
        // prevent a focus change to the button. Focus should stay in the input.
        onMouseDown(event: React.MouseEvent) {
```

modules/react/text-input/lib/InputGroup.tsx
```diff
@@ -164,5 +239,10 @@ export const InputGroup = createComponent('div')({
      * within the input.
      */
     InnerEnd: InputGroupInnerEnd,
+    /**
+     * A component that can be added to an input group that will clear the input. It will only render
+     * when the input has a value and will fade when a value is entered.
+     */
+    ClearInputButton: ClearInputButton,
```
mannycarrera4, line 246: I'm fine with this or `ClearButton` since it already has `InputGroup` in front when you use it. `InputGroup.ClearButton` . This is super nit, non blocking

modules/react/text-input/lib/InputGroup.tsx
NicholasBoll: I can't think of a reason someone would disable this button. Why include it at all if it is disabled?

But to the question, `disabled` supersedes `pointer-events`.

mannycarrera4 (approved):

mannycarrera4:
Gotta update unit tests

modules/react/common/lib/utils/components.ts
```diff
@@ -294,6 +294,11 @@ export const createContainer = <
           ? (elemPropsHook as any)(localModel, elemProps, ref)
           : elemProps;
 
+        // make sure there's always a ref being passed, even if there are no elemProps hooks to run
+        if (ref && !finalElemProps.hasOwnProperty('ref')) {
+          finalElemProps.ref = ref;
+        }
```
NicholasBoll, line 300: This was failing if a container didn't use an `elemPropsHook` and a `ref` was passed.

jest/verifyComponent.tsx
```diff
@@ -21,7 +21,7 @@ export function verifyComponent(
   Component: React.ComponentType<any>,
   {modelFn, props}: {modelFn?: Function; props: object}
 ) {
-  describe('verifyComponent', () => {
+  describe(`verifyComponent "${Component.displayName}"`, () => {
```
NicholasBoll, line 24: This helps identify multiple `verifyComponents` when it comes to compound components

Issue #2152, opened 2023-04-20
Tabs are 2 tab stops on first interaction, then it functions normally afterward


andrew-poblete:
## 🐛 Bug Report

When navigating tabs, the first and second tabs now receive focus, which is not how tabs should function. After the 2nd tab receives focus, the tab set reverts to acting like standard tabs.

## To Reproduce

1. https://canvas.workday.com/v8/assets/system-icons (v8 site)
2. Inspect the Guidance/Gallery/Create Your Own tabs (notice that they all have tabindex=0 set)
3. Tab to Guidance.
4. Tab one more time. Focus moves to Gallery and the tabindex in the code becomes -1.

## Expected Behavior
The tablist should be one tab stop and the tabs should only be accessible using the arrow keys.

## Actual Results
Focus moves to the first tab, then pressing tab again moves focus to the second tab instead of out of the tablist.

## Browser (if applicable)

Chrome

## Link to repl or repo (highly encouraged)

N/A

### Error Output

N/A

Include a screenshot of a visual bug
![Capture](https://user-images.githubusercontent.com/105951517/233490970-a352968d-0e6a-4f6f-87bc-c7f27e23ef03.png)

Issue #2218, opened 2023-05-16
Arrow keys don't visually move focus in Tabs in SSR/SSG environments


jamesfan:
## 🐛 Bug Report

SSR: Server Side Rendering ([Gatsby definition](https://www.gatsbyjs.com/docs/conceptual/rendering-options/#server-side-rendering-ssr), each page is rendered on the server at runtime with each page request)
SSG: Static Site Generation ([Gatsby definition](https://www.gatsbyjs.com/docs/conceptual/rendering-options/#static-site-generation-ssg), the entire site is pre-rendered on the server during the build process)

When focus is assigned to Tabs on a server-side rendered page (via SSR or SSG), the right and left arrow keys do not visually move focus between the tabs.

## To Reproduce

Steps to reproduce the behavior on the Canvas Site (which uses Gatsby and SSG):

1. Go to the [tabs page on the site](https://canvas.workday.com/components/containers/tabs/) directly _or_ navigate to it using the side navigation and hit refresh when you get there (either of these needs to happen to trigger the initial load of the server-side rendered page).
2. Left-click on the `Web` tab at the top of the page.
3. Use the left and right arrows to cycle through the platform tabs. Note how the blue focus outline doesn't cycle properly.
4. Navigate out of Tabs and back to it using the side navigation (e.g., go to Tables and then back to Tabs). This renders the Tabs page without server-side rendering + hydration. Clicking on the `Web` tab and using the arrow keys will work properly in this case.

This issue is reproducible on both the [v8](https://canvas.workday.com/v8/components/containers/tabs/) and [v9](https://canvas.workday.com/v9/components/containers/tabs/) versions of the site.

## Expected Behavior

The arrow keys should visually move focus between the tabs.

PR #2223, opened 2023-05-17
fix(collection): Fix SSR focus and multiple tab stops

NicholasBoll:
## Summary

Use HTML to extract id when changing focus to avoid issues when the server and client do not agree on id generation. Also, force-choose a default cursor ID when no `initialCursorId` is passed in the model config.

Fixes: #2152, #2218

## Release Category
Components

---

## Checklist

## For the Reviewer

- [x] Code
- [x] Testing

## Testing Manually

Go through the steps outlined in #2152
The SSR cannot be tested manually without a static-site generator.

### DIFF
--- a/modules/react/collection/lib/useCursorListModel.tsx
+++ b/modules/react/collection/lib/useCursorListModel.tsx
@@ -315,12 +315,12 @@ export const useCursorListModel = createModelHook({
   },
   requiredConfig: useBaseListModel.requiredConfig,
 })(config => {
-  const [cursorId, setCursorId] = React.useState(config.initialCursorId);
-  const columnCount = config.columnCount || 0;
   const list = useBaseListModel(config);
-  const initialCurrentRef = React.useRef(
+  const [cursorId, setCursorId] = React.useState(
     config.initialCursorId || (config.items?.length ? getId(list.state.items![0]) : '')
   );
+  const columnCount = config.columnCount || 0;
+  const initialCurrentRef = React.useRef(cursorId);
   const navigation = config.navigation;
 
   const state = {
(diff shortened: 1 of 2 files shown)


modules/react/collection/lib/useCursorListModel.tsx
```diff
@@ -315,12 +315,12 @@ export const useCursorListModel = createModelHook({
...
 })(config => {
-  const [cursorId, setCursorId] = React.useState(config.initialCursorId);
-  const columnCount = config.columnCount || 0;
   const list = useBaseListModel(config);
-  const initialCurrentRef = React.useRef(
+  const [cursorId, setCursorId] = React.useState(
     config.initialCursorId || (config.items?.length ? getId(list.state.items![0]) : '')
   );
+  const columnCount = config.columnCount || 0;
+  const initialCurrentRef = React.useRef(cursorId);
```
NicholasBoll, line 323: I'm not sure why the initial `cursorId` could only be `config.initialCursorId`, but the `initialCurrentRef` could be `config.initialCursorId` OR the first item in the items array. This change makes it so `cursorId` can also start as the first item in the array.

This fixes any component using the dynamic API (passing `items`) from the "multiple tab stop" problem where no `cursorId` is defined yet until the user interacts with the collection. You can see that further down the `registerItem` method calls `setCursorId` to make sure there's always a cursor, but the dynamic API doesn't call `registerItem`.

jamesfan (approved):
I've verified this PR fixes #2152 and #2218 with a local production build of the v8 site (`npm pack` the `canvas-kit-react` module in the PR branch and unpacked the tarball directly into my local canvas-site `node_modules`.)

Issue #2222, opened 2023-05-17
Example of Body Content Overflow cannot be scrolled with the keyboard


bstanton678:
## 🐛 Bug Report

In the example of Body Content Overflow on the Canvas Kit design site, the overflow text cannot be scrolled with the keyboard only. This will prevent sighted users from accessing all of the overflow content. 

## To Reproduce

1. Navigate to the Canvas Kit Design Site => Components => Popups => Modal => Body Content Overflow
2. Active the "Open License" button
3. Observe: Keyboard focus is set to the 'X' close icon at the top of the modal, 
4. Use the Tab key to focus the elements in the modal
5. Issue: the overflow text cannot be scrolled with the arrow keys. 

## Expected Behavior

Users should be able to focus the overflow content and scroll it using the up and down arrow keys. This is for sighted users that want to only use a keyboard. 

## Actual Results

A clear and concise description of what actually happens

Overflow content in the modal dialog cannot be scrolled using the keyboard. Users are required to scroll with a mouse or trackpad to access content visually. 

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

All of them. 

## Link to repl or repo (highly encouraged)

Please provide a reference to a repo or branch to help us reproduce this bug.

https://canvas.workdaydesign.com/components/popups/modal/#tab=web&heading=body-content-overflow 

Include a screenshot of a visual bug

<img width="1627" alt="ck site body content overflow" src="https://github.com/Workday/canvas-kit/assets/40372497/a90458cc-1121-46f2-b29f-19553f896a70">

bstanton678:
Should be very straightforward to fix by adding tabIndex="0" on the `<Modal.Body>` component.

CostaAspro708:
Hi William, I have not contribuited yet I think this would be an easy one for me to start with.

PR #2225, opened 2023-05-18
fix: Example Body Content Overflow with support for keyboard scroll

bstanton678:
Added tabindex to the scrollable area of this example.

## Summary

Fixes: #2222  

Keyboard accessibility: the scrollable area in the Body Content Overflow example of a modal could not be scrolled up/down using the keyboard only. 

## Release Category
Component Examples

### Release Note
Adding tabIndex 0 to Modal.Body will allow for content to be accessible and scrollable if the content overflows

---

## Checklist

## For the Reviewer

1. I would open the modal example, and inspect the scrollable. overflow container for a tabindex attribute. 
2. When the modal is opened, verify keyboard focus is set to the top of the dialog, 
3. Use the Tab key to focus the scrollable area, 
4. Verify the Up / Down arrow keys can be used to scroll all content into view. 

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods

This should not have much affect on the UI. There will be a visible focus border displayed around the scrollable area.


modules/react/modal/stories/examples/BodyOverflow.tsx
```diff
@@ -20,7 +20,7 @@ export const BodyOverflow = () => {
         <Modal.Card>
           <Modal.CloseIcon aria-label="Close" />
           <Modal.Heading>MIT License</Modal.Heading>
-          <Modal.Body>
+          <Modal.Body tabIndex="0">
```
mannycarrera4, line 23:
```suggestion
          <Modal.Body tabIndex={0}>
```

mannycarrera4:
I'm wondering if this be the default and always add tabIndex 0 to the body instead of in the example?

bstanton678:
> I'm wondering if this be the default and always add tabIndex 0 to the body instead of in the example?

I thought a little bit about that too, but I only want the Modal.Body to be keyboard focusable when the content overflows like in this example. The Basic example is great as-is, and I don't want to focus body content there. 

Maybe the move is to support a new boolean prop for Modal.Body for consumers?

mannycarrera4:
> > I'm wondering if this be the default and always add tabIndex 0 to the body instead of in the example?
> 
> I thought a little bit about that too, but I only want the Modal.Body to be keyboard focusable when the content overflows like in this example. The Basic example is great as-is, and I don't want to focus body content there.
> 
> Maybe the move is to support a new boolean prop for Modal.Body for consumers?

I think this is where a great call out in the documentation would exists saying if you need focus and have scrollable content, use this tabIndex. I agree that the example as is is good for now

PR #2250, opened 2023-06-12
fix: Popup import fix

MarcoASedano:
## Summary

Fixes a circular dependency that was caused by important from the category root. The circular dependency was throwing an error when writing cypress component tests for certain canvas kit components (i.e. SidePanel). The fix is to change the imports to be relative.

## Release Category
Components

---

## For the Reviewer

I searched the `modules/react/popup/lib` directory for any usage of `from '@workday/canvas-kit-react/popup'` and the 2 files changed were the only instances I found.


MarcoASedano:
New PR with support branch as base: https://github.com/Workday/canvas-kit/pull/2251

Issue #2257, opened 2023-06-15
`useFocusRedirect` does not handle focus leave from radio group in Popup


pablobirukov:
## 🐛 Bug Report

`useFocusRedirect` does not handle focus leave from Popup when a radio group is the last element in the popup

## To Reproduce

Have a [Popup with radio group](https://github.com/pablobirukov/canvas-kit/pull/1/files)

https://codesandbox.io/s/focus-redirect-with-radio-group-3562q2

Open popup, move focus with Tab to radio group, move focus with Tab further

## Expected Behavior

Popup is to close and focus is to move to the element following Popup trigger

Radio group behaves as it's a single focusable element, [spec](https://www.w3.org/wiki/RadioButton)

## Actual Results

focus leaves the document as if radio group was the last focusable element in the document (which it is)

## Browser (if applicable)

FF, Chrome

## Link to repl or repo (highly encouraged)

https://codesandbox.io/s/focus-redirect-with-radio-group-3562q2

### Error Output

https://github.com/Workday/canvas-kit/assets/1861546/735fd15b-95f7-4745-abe8-69f3ce074e93

NicholasBoll:
I've updated the issue to includes a code sandbox link

PR #2258, opened 2023-06-15
fix: Fix for useFocusRedirect

pablobirukov:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/2257

My naive approach: when radio is happens to be the last focusable element withing Popup, check whether it belongs to a radio group, and if so, ensure we handle focus leave for all radio group elements. 

[RadioButton focus management spec](https://www.w3.org/wiki/RadioButton)

## Release Category
Components

### Release Note
`useFocusRedirect` handles focus leave from radio group within Popup

---

## Checklist

## For the Reviewer

## Testing Manually

Can be verified in Storybook, with changes from repro https://github.com/pablobirukov/canvas-kit/pull/1/files

https://github.com/Workday/canvas-kit/assets/1861546/bc42efd3-19ed-4cbb-8796-b9a2fc253880


modules/react/common/lib/utils/elements.ts
```diff
@@ -43,15 +43,33 @@ export const getFirstFocusableElement = (container: HTMLElement): HTMLElement |
...
   const elements = container.querySelectorAll('*');
 
   for (let i = elements.length - 1; i >= 0; i--) {
     const element = elements.item(i);
     if (element && isFocusable(element) && element.getAttribute('tabindex') !== '-1') {
+      if (isRadioInput(element)) {
+        const radioGroup = getRadioGroup(container, element);
+
+        return radioGroup.length > 0 ? Array.from(radioGroup) : element;
+      }
```
NicholasBoll, line 72: Why not instead update `isFocusable` to exclude `input[type=radio]` that isn't checked?

NicholasBoll: I guess that wouldn't work because if non of the radios in the group are `checked`, none of them would be considered focusable.

pablobirukov: > I guess that wouldn't work because if non of the radios in the group are checked, none of them would be considered focusable.
Yes, it would not in that case.

Also if onChange is being slow (say, it invokes something else before finally updating radio group's value which causes the input to become selected), a certain input stays focused and not yet selected for a few frames. It's a developer to blame in that case, but we can guard this corner case too

modules/react/common/lib/utils/elements.ts
```diff
@@ -43,15 +43,33 @@ export const getFirstFocusableElement = (container: HTMLElement): HTMLElement |
...
+export const getLastFocusableElement = (container: HTMLElement): Element[] | Element | null => {
   const elements = container.querySelectorAll('*');
 
   for (let i = elements.length - 1; i >= 0; i--) {
     const element = elements.item(i);
     if (element && isFocusable(element) && element.getAttribute('tabindex') !== '-1') {
+      if (isRadioInput(element)) {
+        const radioGroup = getRadioGroup(container, element);
+
+        return radioGroup.length > 0 ? Array.from(radioGroup) : element;
```
pablobirukov, line 71: I suppose it should be
```suggestion
        return radioGroup.length > 1 ? Array.from(radioGroup) : element;
```

mannycarrera4 (approved):

Issue #2262, opened 2023-06-21
Typing on SelectProps does not match the descriptive comment


JeffreyRuder:
## 🐛 Bug Report

The description of the `options` property of the `SelectProps` interface in the react preview module does not match the actual type definition of the property.

The comment says:
```
   * The options of the Select. `options` may be an array of objects or an array of strings.
   *
   * If `options` is an array of objects, each object must adhere to the `Option` interface:
```

This comment indicates that all elements in the array should have the same type, either `string` or `Option`.

However, the type definition is `options: (string | Option)[]` which means the elements in the array could be a mix of `string` and `Option`.

## To Reproduce

N/A

## Expected Behavior

Either the typing of the `options` property should change to `option: string[] | Option[]` which would match the comment, or the comment should change to reflect that the elements in the array can be a mix of strings and Option objects.

## Actual Results

N/A
## Browser (if applicable)

N/A

## Link to repl or repo (highly encouraged)

https://github.com/Workday/canvas-kit/blob/9d5acdca5b46550d3a8063682859f30fa0d7d6ba/modules/preview-react/select/lib/Select.tsx#L14

### Error Output

N/A

PR #2264, opened 2023-06-21
fix: Make type declaration for SelectProps options match actual use

JeffreyRuder:
## Summary

Fixes: #2262  

The type declaration for `SelectProps['options']` did not match what was actually expected or how the options were actually used. The type declaration indicated that the options array could be a mixture of `strings` and objects matching the `Option` interface, but other code (and the comments on the options prop) indicated that it should be all of one or the other.

## Release Category
Components

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods

## Testing Manually

This change mainly affects developers who might want to use or extend `Select`.

In a code editor, in `Select.tsx`, add some test code like:

```
const isOptionArray = (arr: SelectProps['options']): arr is Option[] => {
  return arr.length > 0 && typeof arr[0] !== 'string'
}

const testOptions = (options: SelectProps['options']) => {
  if (isOptionArray(options)) {
    options.forEach(o => console.log(o.value))
  } else {
    options.forEach(o => console.log(o))
  }
}
```

Typescript should recognize that, in the `if` block, the options are guaranteed to satisfy the `Option` interface and have a `value`. It should not complain that `value` is not present on type `string`.

Also, there should not be any regressions in tests or in the functionality of `Select`.

This should not affect the UI.

## Thank You Gif (optional)

Thank you!

### DIFF
--- a/modules/preview-react/select/lib/Select.tsx
+++ b/modules/preview-react/select/lib/Select.tsx
@@ -27,7 +27,7 @@ export interface SelectProps extends CoreSelectBaseProps {
    *
    * The `data` object is carried over to the `option` passed into the `renderOption` function where it may then be used to customize how each option is rendered.
    */
-  options: (Option | string)[];
+  options: Option[] | string[];
 }
 
 interface SelectContainerProps extends SelectProps, StyledType {


jamesfan:
Hi @JeffreyRuder, thanks for the find!

Although we don't expect it to be a common use case, the Preview Select should in fact support an options array with a mixture of strings and objects matching the `Option` interface. The component contains [logic](https://github.com/Workday/canvas-kit/blob/master/modules/preview-react/select/lib/Select.tsx#L89) which normalizes every option to match the same format.

> ...but other code (and the comments on the options prop) indicated that it should be all of one or the other.

You're correct that the JSDocs for the `options` prop indicate it should be all of one or the other, but I can't find "other code" which indicates the same. Can you point out what code you're referring to?

Otherwise, this issue seems like it could be addressed by updating the comments for the `options` prop like you suggested as a possible solution under `Expected Behavior` in #2262.

JeffreyRuder:
Ok, I will close this PR and make another one updating the comments.

I think by other code I meant the comment on `setNormalizedOptions` in `Select.tsx` which also references `an array of strings or an array of objects`

Issue #1854, opened 2022-10-18
Investigate a precompiled CSS-in-JS framework vs a static styling framework


alanbsmith:
## 🔍 Spike /  Exploration

We've talked a lot about moving past Emotion, but we haven't had much direction. It would be helpful to limit the scope of our options and state the specific criteria we're considering.

## Prior Art

- [Breaking up with CSS-in-JS](https://dev.to/srmagura/why-were-breaking-up-wiht-css-in-js-4g9b)
- [Alan's response](https://gist.github.com/alanbsmith/e5b9fad745161e2b61d72dbe0d0cb79c)

## What questions would be answered once this spike was completed?

There are four main topics we need to consider:

- Theme Support
  - We need a theme system we can use robustly throughout Canvas Kit
  - How do we handle theming across multiple CK versions?
- Style Overrides
  - Teams override our styles all the time
  - It is impossible for us to offer CSS variables for everything, so direct overrides of styles is necessary
- Local Scoping
  - Basically naming CSS classnames with hashes to avoid clashing
- Migration Path
  - How do we handle the migration path for teams using Emotion?
  - How can we make it opt-in vs opt-out to reduce friction?

## Acceptance Criteria

- [ ] Explore the options
- [ ] Cover the topics in a GH discussion
- [ ] Share with the team

alanbsmith:
Closing. Nicholas completed this with #2273

PR #2273, opened 2023-06-28
feat: Add CSS styling function

NicholasBoll:
## Summary

Implements: #2208 
Related: #1854

Discussion: #2265 

This change adds a styling function to use CSS-in-JS with a minimum developer runtime and a transform process that can turn CSS-in-JS into static styling.

## Release Category
Components

### Release Note
This change introduces an intermediate step towards static styling. The style function takes in statically analyzable style definitions with modifiers, variables, and static styles to create styling to be used in components. It can later be interpreted at build time to create CSS packages.

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods

## Thank You Gif

Emotional Damage Meme


modules/styling/README.md
```diff
@@ -0,0 +1,8 @@
+# Canvas Kit React
+
+The bundle package containing all modules of the Canvas Kit React.
+
+> Note: By default, no fonts are included with Canvas Kit modules. To use official Canvas Kit fonts,
+> install and import the `@workday/canvas-kit-react-fonts` module in addition to
+> `@workday/canvas-kit-react`. By installing `@workday/canvas-kit-react-fonts` you are opting in to
+> using the Workday CDN.
```
NicholasBoll, line 8: This needs to be updated to mention styling instead

modules/react/common/lib/utils/slugify.ts
```diff
@@ -4,8 +4,11 @@
 export function slugify(input: string): string {
   return input
     .trim()
+    .replace(/[\.\s_-]+/g, '-')
     .replace(/[^\w\s-]/g, '')
-    .replace(/[\s_-]+/g, '-')
+    .replace(/([A-Z])/g, m => {
+      return '-' + m.toLowerCase();
```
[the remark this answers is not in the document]
NicholasBoll: For example: `borderRadius` became `borderradius` instead of `border-radius`

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,350 @@
+import React from 'react';
+import {generateUniqueId} from './uniqueId';
+// eslint-disable-next-line @emotion/no-vanilla
+import {css} from '@emotion/css';
+import {serializeStyles, Keyframes, SerializedStyles, CSSObject} from '@emotion/serialize';
+
+import {slugify} from './slugify';
+
+// future placeholder for token replacement
+const tokens = {};
```
mannycarrera4, line 10: can these be updated now?

NicholasBoll: I'm not convinced this is something we should do. It works with Vanilla Extract (sprinkles) because all token replacements are build-time only. If we add this as a runtime, it will send more info to the client and I'm not convinced that's a good thing.

NicholasBoll: The `replaceWithToken` recursive function is used to convert a raw variable with a `var()` wrapper if detected. The `tokens` can be removed and the functions could be renamed to avoid any confusion

modules/styling/lib/uniqueId.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Generates a unique and HTML5 compliant identifier every time it is called. Internally it uses a 4
+ * character random seed starting with a letter. This seed is unique to each instance of this
+ * package meaning different versions of Canvas Kit on the page will have a different seed. Each
+ * call will use a Base 36 string (10 numbers + 26 letters) based on an incremented number. The
+ * incremented number always starts at 0 and can be reset for testing purposes using
+ * [resetUniqueIdCount](#resetuniqueidcount). [setUniqueSeed](#setuniqueseed) can also be used for
+ * testing or server side rendering to get the same results during hydration.
+ */
+export const generateUniqueId = () => seed + (c++).toString(36);
```
mannycarrera4, line 19: lol `c++`

modules/styling/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@workday/canvas-kit-styling",
+  "version": "9.1.20",
```
mannycarrera4, line 3: i think the latest is 9.1.24

NicholasBoll: The value keeps changing and there isn't a good way to make sure it is always up to date :(

modules/styling/stories/Welcome.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+If you're using Canvas Kit and not directly using this package, there is nothing extra to do on your
+end. The Canvas Kit packages are using the static compilation as part of the build process. If you
+want to use this package for your own styles, you don't need to do anything special to use in
+development. Included is a small runtime to get styling working. If you wish to statically compile
+your CSS from your TypeScript files for faster page loading, visit the
+[Getting Started](/docs/features-styling-getting-started--page)
+
+## What is this?
+
+Canvas Kit Styling is two things:
```
mannycarrera4, line 25: I'm wondering if we should link out to the discussion of the future of css and why we built this?

mannycarrera4: yeah it seems like this is built BECAUSE we want to better support theming/ and styling and move away from emotion

modules/styling/stories/GettingStarted.stories.mdx
```diff
@@ -0,0 +1,89 @@
...
+yarn add @workday/canvas-kit-styling
+```
+
+## Usage
+
+```tsx
+import React from 'react';
+import {createRoot} from 'react-dom/client';
+
+import {cs} from '@workday/canvas-kit-styling';
```
mannycarrera4, line 15: is this suppose to be `createStyles` ?

modules/styling/stories/GettingStarted.stories.mdx
```diff
@@ -0,0 +1,89 @@
...
+```
+
+## Development
+
+Canvas Kit Styling comes with a runtime that doesn't need anything special for developement. The
+runtime uses `@emotion/css` to include your styles on the page.
+
+## Production
+
+If you wish to use the static compilation, you must use the `@workday/canvas-kit-styling-parser`
```
mannycarrera4, line 36: is this still part of this PR? the parser?

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+      if ((obj as Object).hasOwnProperty(key)) {
+        (converted as any)[key] = convertAllProperties(obj[key]);
+      }
+    }
+    return converted as T;
+  }
+  return convertProperty(obj);
+}
+
+export type CS = string | Record<string, string>;
```
mannycarrera4, line 48: do we want to stay as CS or CreateStyles?

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+
+export type CsVarsMap<T extends string, ID extends string | never> = [ID] extends [never]
+  ? Record<T, string>
+  : {[K in T]: `--${ID}-${K}`};
```
mannycarrera4, line 52: mind adding some jsdocs of what this type is doing? i need to be better about learning types like this, NB comment

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+ * be unique at runtime to avoid collisions. The return value is a function and a
+ * Map. The function can be used to pass in values from JavaScript. The function will
+ * return a map of variable keys to CSS Variable names.
+ *
+ * ```ts
+ * // creates a `color` and `background` variable
+ * const myVars = createVars('color', 'background')
+ *
+ * // 'color' is a typed property. The type is `string`
+ * console.log(myValues.color) // `'var(--{hash}-color)'`
```
mannycarrera4, line 98:
```suggestion
 * console.log(myVars.color) // `'var(--{hash}-color)'`
```

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+}
+
+/**
+ * Create temporary CSS variables to use in components. The CSS variable names will
+ * be unique at runtime to avoid collisions. The return value is a function and a
+ * Map. The function can be used to pass in values from JavaScript. The function will
+ * return a map of variable keys to CSS Variable names.
+ *
+ * ```ts
+ * // creates a `color` and `background` variable
```
mannycarrera4, line 94:
```suggestion
 * // creates a `color` and `background` CSS variable
```

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+ *
+ * // 'color' is a typed property. The type is `string`
+ * console.log(myValues.color) // `'var(--{hash}-color)'`
+ *
+ * // 'color' is a typed property. The type is `string?`
+ * // The returned object can be assigned to the `style` property of an element
+ * console.log(myVars({ color: 'red' })) //  `{'--{hash}-color': 'red'}`
+ *
+ * const div = document.createElement('div')
+ * div.style = myVars({ color: 'red' }) // <div style="--{hash}-color: red;" />
```
mannycarrera4, line 105: so you've used `createVars` and then here you're using your variable and assigning it a value to use in a component right?

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+ * ```
+ */
+export function createVars<T extends string, ID extends string>(input: {
+  id: ID;
+  args: T[];
+}): CsVars<T, ID>;
+export function createVars<T extends string>(...args: T[]): CsVars<T, never>;
+export function createVars<T extends string, ID extends string>(...args: T[]): CsVars<T, ID> {
```
mannycarrera4, line 113: curious why the argument of variables wouldn't be an array? `createVars(['background', 'color'...restVariables])`. No preference just asking!

mannycarrera4: ah `...args`??

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+type ModifierValues<T extends ModifierConfig> = {
+  [P in keyof T]: keyof T[P];
+};
+
+type ModifierReturn<T extends ModifierConfig> = T &
+  ((modifiers: Partial<ModifierValues<T>>) => string);
+
+/**
+ * Creates a modifier function that takes in a modifier config and will return a CSS class name that
+ * matches the result. Modifiers can be thought as `if` or `switch` statements. This function can be
```
mannycarrera4, line 147:
```suggestion
 * matches the result. Modifiers can be thought as `if` or `switch` statements when conditionally changing the styles of a component based on props. This function can be
```

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+ *     medium: createStyles({ fontSize: 14 })
+ *   }
+ * })
+ *
+ * // with the modifier function
+ * myModifiers({ size: 'medium' }) // returns the medium class name
+ *
+ * // manually without the function
+ * size === 'small' ? myModifiers.size.small : size === 'medium' ? myModifiers.size.medium : ''
+ * ```
```
mannycarrera4, line 165: i like the HTML reference that you have above, can we add something like:
```
// When passing the modifier of size to your component
<Component size="small" ...></Component>
// internal component code
<div className="default-classname modifier-small className"></div>
```
Very similar to how [BEM](https://getbem.com/introduction/) does it.

I'm ok if this doesn't live in the jsdoc but somewhere else

mannycarrera4: Also probably worth noting that the order here is really important

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+
+  return modifierFn as ModifierReturn<T>;
+}
+
+function isCsPropsReturn(input: {}): input is CsToPropsReturn {
+  return input.hasOwnProperty('style') || input.hasOwnProperty('className');
+}
+
+/**
+ * All acceptable values of the `cs` prop. It can be a CSS class name, any CSS properties, an object
```
mannycarrera4, line 188: `cs` or `createStyles` ?

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+  return input.map(csToProps).reduce((result, val) => {
+    return {
+      className: [result.className, val.className].filter(v => v).join(' '),
+      style: {...result.style, ...val.style},
+    };
+  }, {} as CsToPropsReturn);
+}
+
+export interface CSProps {
+  /** The `cs` prop takes in a single value or an array of values. You can pass the CSS class name
```
mannycarrera4, line 246:
```suggestion
  /** The `cs` prop takes in a single value or an array of values. You can pass the CSS class name
```
Just through out the file, are we using `cs` or `createStyles`?

modules/styling/lib/cs.ts
mannycarrera4: sorry, i think I'm confusing createStyles with cs

modules/styling/lib/uniqueId.ts
```diff
@@ -0,0 +1,44 @@
...
+export const setUniqueSeed = (s: string) => {
+  seed = s;
+};
+
+/**
+ * This should only be called for tests in an `beforeEach`
+ */
+export const resetUniqueIdCount = () => {
+  c = 0;
+};
```
mannycarrera4, line 44: do we want this to be in common? or do we not want the styling package to rely on any react packages?

mannycarrera4: prob the latter

NicholasBoll: We can't have this in common because then `@workday/canvas-kit-react` would depend on `@workday/canvas-kit-styling` and the reverse would be true. We'd have a circular dependency.

modules/styling/stories/Welcome.stories.mdx
NicholasBoll: I'm adding a "Why" section with a link to the discussion.

modules/styling/stories/GettingStarted.stories.mdx
NicholasBoll: Not part of this PR, but another PR will add it. It is difficult to maintain 2 separate PRs and fix this type of thing. I'd rather leave this alone for now.

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+ *
+ * ```ts
+ * const myStyles = createStyles({
+ *   backgroundColor: 'red'
+ * })
+ * ```
+ *
+ *
+ *
+ * The cs function is curried into 2 parts. The first function could be done at build time. The
```
mannycarrera4, line 277:
```suggestion
 * The `cs` function is curried into 2 parts. The first function could be done at build time. The
```

modules/styling/lib/cs.ts
```diff
@@ -0,0 +1,355 @@
...
+ * returned function combines classnames and will remain as a small runtime.
```
mannycarrera4, line 278:
```suggestion
 * returned function combines `classnames` and will remain as a small runtime.
```

modules/styling/lib/cs.ts
NicholasBoll: stylistic choice

mannycarrera4 (approved):
EMOTIONAL DAMAGE

Issue #2280, opened 2023-07-05
indeterminate checkbox not visually showing the correct state


vibdev:
## 🐛 Bug Report

Checkboxes in indeterminate state have the correct aria-checked value but don't visually display the correct state unless they are also 'checked.' This is different from a plain HTML checkbox that only needs indeterminate attribute set to display the '-' icon.

## To Reproduce

`<Checkbox indeterminate={true} checked={false}  />` <--- shows no visual update

## Expected Behavior

It should show the '-' icon

## Actual Results

It shows no visual state change

## Link to repl or repo (highly encouraged)

[Sandbox](https://codesandbox.io/s/indeterminate-checkbox-fxdpcp)

PR #2281, opened 2023-07-05
fix(checkbox): Fix indeterminate state not getting correctly applied

vibdev:
## Summary

Fixes: #2280   

The underlying checkbox was not getting the indeterminate attribute applied and we were relying on it being both checked and indeterminate to update the visual display. This change sets the attribute and updates the CSS to also check the `:indeterminate` pseudo class. Indeterminate checkboxes are a bit odd as [the state can't be directly changed via HTML and needs to be set with a ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes)

## Release Category
Components

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [x] Testing
- [ ] Codemods

## Thank You Gif (optional)

a Simpsons character holding up a diploma "With minor indetermination"

### DIFF
--- a/modules/react/checkbox/lib/Checkbox.tsx
+++ b/modules/react/checkbox/lib/Checkbox.tsx
@@ -220,7 +220,7 @@ const CheckboxInput = styled('input')<CheckboxProps & StyledType>(
           variant === 'inverse' ? `1px solid ${colors.soap300}` : `1px solid ${errorColors.inner}`,
         boxShadow: `0 0 0 1px ${errorColors.inner}, 0 0 0 2px ${errorColors.outer}`,
       },
-      '&:not(:checked):not(:indeterminate):not(:disabled):not(:focus):hover, &:not(:checked):not(indeterminate):not(:disabled):active': {
+      '&:not(:checked):not(:indeterminate):not(:disabled):not(:focus):hover, &:not(:checked):not(:indeterminate):not(:disabled):active': {
         '~ div:first-of-type': {
           borderColor: variant === 'inverse' ? `1px solid ${colors.soap300}` : errorColors.inner,
         },
--- a/modules/react/checkbox/lib/Checkbox.tsx
+++ b/modules/react/checkbox/lib/Checkbox.tsx
@@ -11,7 +11,6 @@ import {
   useTheme,
   Themeable,
   useLocalRef,
-  useForkRef,
 } from '@workday/canvas-kit-react/common';
 import {borderRadius, colors, inputColors, spaceNumbers} from '@workday/canvas-kit-react/tokens';
 import {SystemIcon} from '@workday/canvas-kit-react/icon';
--- a/modules/react/checkbox/lib/Checkbox.tsx
+++ b/modules/react/checkbox/lib/Checkbox.tsx
@@ -324,7 +324,7 @@ export const Checkbox = createComponent('input')({
     Element
   ) => {
     const inputId = useUniqueId(id);
-    const {localRef} = useLocalRef(useForkRef(ref));
+    const {localRef, elementRef} = useLocalRef(ref);
     React.useEffect(() => {
       if (typeof indeterminate === 'boolean' && localRef.current) {
         localRef.current.indeterminate = indeterminate;
@@ -339,7 +339,7 @@ export const Checkbox = createComponent('input')({
             checked={checked}
             disabled={disabled}
             id={inputId}
-            ref={localRef}
+            ref={elementRef}
             type="checkbox"
             variant={variant}
             aria-checked={indeterminate ? 'mixed' : checked}
--- a/modules/react/checkbox/lib/Checkbox.tsx
+++ b/modules/react/checkbox/lib/Checkbox.tsx
@@ -10,6 +10,8 @@ import {
   styled,
   useTheme,
   Themeable,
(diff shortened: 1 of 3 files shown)


modules/react/checkbox/lib/Checkbox.tsx
```diff
@@ -330,18 +339,18 @@ export const Checkbox = createComponent('input')({
...
             {...elemProps}
           />
           <CheckboxRipple variant={variant} />
           <CheckboxBackground variant={variant} checked={checked} disabled={disabled}>
-            <CheckboxCheck checked={checked}>
+            <CheckboxCheck checked={checked} indeterminate={indeterminate}>
               {indeterminate ? (
                 <IndeterminateBox variant={variant} />
-              ) : (
+              ) : checked ? (
```
vibdev, line 353: This fixes a different sneaky bug where the checkbox flashes the checked icon when moving between indeterminate true to false

jamesfan: Nice catch!

modules/react/checkbox/lib/Checkbox.tsx
jamesfan, line 1: Can you look into the regressions in the Chromatic tests? You can disregard the `Tooltips Placements` changes, but we should address the changes in `Checkbox States` and `Checkbox Themed States`. The regressions look specific to the active states:

image

modules/react/checkbox/lib/Checkbox.tsx
```diff
@@ -219,12 +220,12 @@ const CheckboxInput = styled('input')<CheckboxProps & StyledType>(
           variant === 'inverse' ? `1px solid ${colors.soap300}` : `1px solid ${errorColors.inner}`,
         boxShadow: `0 0 0 1px ${errorColors.inner}, 0 0 0 2px ${errorColors.outer}`,
       },
-      '&:not(:checked):not(:disabled):not(:focus):hover, &:not(:checked):not(:disabled):active': {
+      '&:not(:checked):not(:indeterminate):not(:disabled):not(:focus):hover, &:not(:checked):not(indeterminate):not(:disabled):active': {
```
jamesfan, line 223: Think I found the issue triggering the visual regressions mentioned above.
```suggestion
      '&:not(:checked):not(:indeterminate):not(:disabled):not(:focus):hover, &:not(:checked):not(:indeterminate):not(:disabled):active': {
```

vibdev: that could do it, nice catch

modules/react/checkbox/stories/examples/Indeterminate.tsx
```diff
@@ -45,8 +45,8 @@ export const Indeterminate = () => {
 
     const anyToppingChecked = newToppings.filter(topping => topping.checked).length > 0;
     const anyToppingUnchecked = newToppings.filter(topping => !topping.checked).length > 0;
-    setPizzaChecked(anyToppingChecked);
     setPizzaIndeterminate(anyToppingChecked && anyToppingUnchecked);
+    setPizzaChecked(anyToppingChecked && !anyToppingUnchecked);
```
jamesfan, line 49: We want the top-level pizza checkbox to only be checked if _all_ toppings are checked, correct? 

```
!anyToppingUnchecked == "not any toppings are unchecked" == "all toppings are checked"
```

```suggestion
    setPizzaChecked(!anyToppingUnchecked);
```

Alternatively, we could just set an `allToppingsChecked` variable for a less confusing way of putting this 😂

vibdev: oof yeah the first part is redundant, I like the idea of making a new var to be more clear on the intentions here

modules/react/checkbox/lib/Checkbox.tsx
```diff
@@ -360,6 +368,7 @@ export const Checkbox = createComponent('input')({
             disabled={disabled}
             variant={variant}
             paddingInlineStart={checkboxLabelDistance}
+            cursor="pointer"
```
vibdev, line 371: @jamesfan just noticed the label doesn't have the correct cursor even though it is clickable

modules/react/checkbox/lib/Checkbox.tsx
```diff
@@ -322,6 +323,13 @@ export const Checkbox = createComponent('input')({
     Element
   ) => {
     const inputId = useUniqueId(id);
+    const {localRef, elementRef} = useLocalRef(ref);
+    React.useEffect(() => {
+      if (typeof indeterminate === 'boolean' && localRef.current) {
+        localRef.current.indeterminate = indeterminate;
```
jamesfan, line 329: @vibdev Last bit: can you add a Cypress test ensuring `indeterminate` is being set properly on the checkbox?

I'm thinking something which loads the Indeterminate story, checks a topping (any of them), and then verifies the `Supreme Pizza Toppings` checkbox has the `indeterminate` property set. Something like this should do the trick:

```
cy.findByLabelText('Supreme Pizza Toppings').should('have.prop', 'indeterminate', true);
```

jamesfan (approved):
@vibdev Looks good, thanks for catching this issue!

Issue #2274, opened 2023-06-29
Button Styles Refactor


josh-bagwell:
## 🚀 Feature Proposal

A button style re-factor is needed in conjunction with the new `cs` prop that is being created. We will also need to update the Primary, Secondary, Tertiary, and Delete buttons (along with their inverses) to use the new system level tokens via CSS variables from the CanvasProvider.

**Acceptance Criteria:**
- All Canvas button variants and their inverses are tied to CSS variables
- Update the v10 upgrade guide

## Motivation

This will utilize the new `cs` prop and we now have access to new css properties we did not have when button was re-factored prior.

PR #2285, opened 2023-07-13
chore: Refactored Button styles to use `createStyles` utility

josh-bagwell:
## Summary

Style refactor for buttons to work with new theming, branding, tokens, createStyles utility and `cs` prop. 

Fixes: #2274  

We want to more easily apply theming to Buttons and the styles need to be refactored to more easily apply those themed styles with new tokens. 

## Release Category
Components

### BREAKING CHANGES
- We refactored how we styled Buttons to use our `createStyles` utility function. We don't anticipate
this as a breaking change but, there may be slight changes to visual test.
- Icons will no longer be "filled" on toggle. This decision was made to not have the existing icon
look different in the toggled state from default state.
- `PrimaryButton`: On the `inverse` variant, the focus ring is now consistent with the default variant of `PrimaryButton`. This will visually change the `inverse` variant to have a larger appearance when focused. 
- `colors` will no longer support the `focusRing` option:
  ```tsx
  import {focusRing} from '@workday/canvas-kit-react/common';

  // before
  <PrimaryButton
    colors={{
      // other colors
      focus: {
        // other colors
      focusRing: focusRing(/* options */)
      }
    }}
  />

  // after
  <PrimaryButton
    colors={{
      // other colors
      focus: {
        // other colors
      }
    }}
    css={{
      ':focus-visible': focusRing(/* options */)
    }}
  />;
  ```
---

## For the Reviewer

`/modules/react/button`


modules/react/button/lib/BaseButton.tsx
```diff
@@ -292,13 +133,295 @@ export const getPaddingStyles = (
...
+          buttonVars.active(colors?.active || {}),
+          buttonVars.default(colors?.default || {}),
+          buttonVars.disabled(colors?.disabled || {}),
+          buttonVars.focus(colors?.focus || {}),
+          buttonVars.hover(colors?.hover || {}),
+        ]}
+        {...elemProps}
+      >
         {children}
-      </ButtonContainer>
```
NicholasBoll, line 301: The `cs` prop should always go last. Otherwise if someone passes a `className`, all styling will be overridden.

modules/react/button/lib/BaseButton.tsx
```diff
@@ -292,13 +133,295 @@ export const getPaddingStyles = (
...
+            size: size,
+            iconPosition: getIconPosition(size, iconPosition),
+          }),
+          buttonVars.active(colors?.active || {}),
+          buttonVars.default(colors?.default || {}),
+          buttonVars.disabled(colors?.disabled || {}),
+          buttonVars.focus(colors?.focus || {}),
+          buttonVars.hover(colors?.hover || {}),
+        ]}
+        {...elemProps}
```
NicholasBoll, line 421: The `cs` prop should always go last. Otherwise if someone passes a `className`, all styling will be overridden.

josh-bagwell: Done

mannycarrera4:
image
 looks like icons action bar needs to be updated

mannycarrera4:
In the PR the focus button seems like it has a white background
image

In master, it should have a grey background
image

mannycarrera4:
In the PR toolbar icon button isn't filled:
image
What's in master is filled:
image

yarn.lock
```diff
@@ -18131,9 +18136,9 @@ windows-release@^3.1.0:
     execa "^1.0.0"
 
 word-wrap@^1.0.3, word-wrap@~1.2.3:
-  version "1.2.4"
-  resolved "https://registry.yarnpkg.com/word-wrap/-/word-wrap-1.2.4.tgz#cb4b50ec9aca570abd1f52f33cd45b6c61739a9f"
-  integrity sha512-2V81OA4ugVo5pRo46hAoD2ivUJx8jXmWXfUkY4KFNw0hEptvN0QfH3K4nHiwzGeKl5rFKedV48QVoqYavy4YpA==
+  version "1.2.3"
```
mannycarrera4, line 18139: did we want to update this?

modules/react/popup/lib/PopupCloseIcon.tsx
```diff
@@ -21,7 +21,7 @@ export const PopupCloseIcon = createSubcomponent('button')({
       icon={xIcon}
       type="button"
       position="absolute"
-      right={space.xxxs}
+      insetInlineEnd={space.xxxs}
```
mannycarrera4, line 24: nice!

modules/docs/mdx/10.0-UPGRADE_GUIDE.mdx
```diff
@@ -123,6 +128,22 @@ these sizes should change along with it. Since `px` are a fixed unit and do not
...
+
+#### Button Focus Ring Update
+
+We found that focus ring was not consistent across all Buttons. We have made some changes to create
+a consistent focus ring across all Buttons and variants. These will need to be taken note of due to
+some small changes visually with Buttons.
+
+- `PrimaryButton`: `inverse` variant now has the same focus ring as default variant and as
+  `SecondaryButton`.
+
```
NicholasBoll, line 146: We should highlight the breaking change that `colors` no longer supports the `focusRing` option.

```jsx
import {focusRing} from '@workday/canvas-kit-react/common';

// before
<PrimaryButton
  colors={{
    // other colors
    focus: {
      // other colors
     focusRing: focusRing(/* options */)
    }
  }}
/>

// after
<PrimaryButton
  colors={{
    // other colors
    focus: {
      // other colors
    }
  }}
  css={{
    ':focus-visible': focusRing(/* options */
  }}
/>
```

modules/preview-react/pill/lib/PillIconButton.tsx
```diff
@@ -63,6 +62,11 @@ const StyledIconButton = styled(BaseButton)<StyledType & PillIconButtonProps>({
     pointerEvents: 'all',
     cursor: 'pointer',
   },
+  '&:focus, &:focus-visible': {
```
NicholasBoll, line 65:
```suggestion
  '&.focus, &:focus-visible': {
```

modules/react/button/lib/BaseButton.tsx
```diff
@@ -292,13 +126,265 @@ export const getPaddingStyles = (
...
+      height: cssVar(system.space.x6, space.m),
+      minWidth: 'auto',
+      paddingInline: cssVar(system.space.x3, space.xs),
+      gap: cssVar(system.space.x1, space.xxxs),
+    }),
+  },
+  iconPosition: {
+    largeOnly: createStyles({
+      padding: '0',
+      minWidth: cssVar(system.space.x12, '3rem'),
```
NicholasBoll, line 295: I'd like to get rid of all fallbacks to system variables. The optimizer will be able to parse the CSS files from our tokens package and automatically fill in fallbacks. This would be a win for us in Canvas Kit because we don't have to lookup fallbacks manually, but still get the benefit of working in environments where the variables are not provided (most likely Storybook)

modules/react/common/lib/CanvasProvider.tsx
```diff
@@ -1,34 +1,60 @@
...
-    },
-  };
+const mappedKeys = {
+  lightest: 'lightest',
+  light: 'light',
+  main: 'base',
+  dark: 'dark',
+  darkest: 'darkest',
+  contrast: 'accent',
+};
```
mannycarrera4, line 18: is this something that consumer will have to do if they're using canvas provider?

.storybook/preview.js
```diff
@@ -34,7 +37,9 @@ function storySort(a, b) {
     prefix('basic', 'aa'),
     prefix('default', 'ab'),
     prefix('testing', 'zzz'),
-    prefix('examples', 'zz')
+    prefix('examples', 'zz'),
+    // Make sure upgrade guides follow chronological order by replacing `v9.0` with `v09.0`
+    value => value.replace(/v([1-9]\-0)/, '0$1')
```
mannycarrera4, line 42: 👍

modules/preview-react/pill/lib/Pill.tsx
```diff
@@ -158,6 +156,13 @@ const StyledNonInteractivePill = styled(StyledBasePill)<StyledType>({
   cursor: 'default',
   overflow: 'revert', // override BaseButton overflow styles so the click target exists outside the pill for removable
   position: 'relative',
+  '&.focus-visibile, &.focus': {
```
mannycarrera4, line 159:
```suggestion
  '&:focus-visibile, &.focus': {
```

modules/react/button/lib/Button.tsx
```diff
@@ -0,0 +1,67 @@
...
+            size={size}
+            iconPosition={iconPosition}
+            icon={icon}
+            shouldMirrorIcon={shouldMirrorIcon}
+          />
+        )}
+      </BaseButton>
+    );
+  },
+});
```
mannycarrera4, line 67: Is this file still needed or is it replacing `BaseButon`?

josh-bagwell:
> image looks like icons action bar needs to be updated

Should we consider making this a test as well in the future? I wouldn't have noticed this without you mentioning it to me. 

I had to reset a default `iconPosition`.

modules/react/popup/lib/PopupCloseIcon.tsx
josh-bagwell: This is what we'll be moving to with other components as well so we no longer have to worry about RTL.

modules/docs/mdx/10.0-UPGRADE_GUIDE.mdx
josh-bagwell: Done.

mannycarrera4:
> > In the PR toolbar icon button isn't filled: image What's in master is filled: image
> 
> This was a change we talked with design on. We decided that the icon shouldn't changed when toggled. This came up with segmented control on the tablet icon. It looks completely different so we decided to change it and not have it fill.

We should probably add this as change to the upgrade guide

modules/preview-react/pill/lib/Pill.tsx
```diff
@@ -147,8 +138,15 @@ const StyledBasePill = styled(BaseButton.as('button'))<StyledType & PillProps>(
     }),
   },
   ({variant}) => ({
-    '&:focus': {
+    '&:focus, &:focus-visible': {
```
mannycarrera4, line 141: does this need the `.focus` class?

modules/react/common/lib/CanvasProvider.tsx
josh-bagwell: They shouldn't have too. We're just mapping the existing structure to the new structure.

modules/react/button/lib/Button.tsx
josh-bagwell: Yup. This is the mediator for each button variant. Just allows us to clean stuff up.

modules/preview-react/pill/lib/Pill.tsx
josh-bagwell: Yeah this just got updated. I made `focus` a class and `focus-visible` the psuedo

modules/react/button/lib/DeleteButton.tsx
```diff
@@ -1,75 +1,54 @@
...
-    label: themeError.contrast,
-  },
-  hover: {
-    background: themeError.dark,
+/**
+ * Extends all the style properties from Box to our buttons as well as props from ButtonContainerProps.
+ * We omit `ref` since all of our buttons use `createComponent` and already give access to `ref`.
+ * Use this type to extend and customize any one off buttons that you want full control over styling.
+ */
+export interface DeleteButtonProps extends Omit<ButtonContainerProps, 'ref'> {}
```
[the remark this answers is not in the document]
josh-bagwell: Ahh. Yeah I left that in since it's an exportable interface and if anyone was using that, I didn't want it to cause an issue if I removed it. That goes for all the other buttons as well.

modules/react/button/lib/DeleteButton.tsx
```diff
@@ -1,75 +1,54 @@
...
+    [buttonVars.disabled.label]: cssVar(brand.error.accent),
+    opacity: 1,
   },
 });
 
-/**
- * Use sparingly for destructive actions that will result in data loss, can’t be undone, or will
- * have significant consequences. They commonly appear in confirmation dialogs as the final
- * confirmation before deleting.
- */
```
mannycarrera4, line 48: I think we want this comment back? above `export const DeleteButton...`

modules/react/button/lib/Button.tsx
```diff
@@ -0,0 +1,67 @@
...
+        {icon && iconPosition === 'end' && (
+          <BaseButton.Icon
+            size={size}
+            iconPosition={iconPosition}
+            icon={icon}
+            shouldMirrorIcon={shouldMirrorIcon}
+          />
+        )}
+      </BaseButton>
+    );
```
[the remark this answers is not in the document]
josh-bagwell: Yeah definitely. 

It's not accessible via the package since I didn't add it to the index. But I can callout what it's job is.

modules/react/button/lib/PrimaryButton.tsx
```diff
@@ -1,171 +1,104 @@
...
-  }
-};
-
-export interface PrimaryButtonProps extends Themeable, GrowthBehavior, BaseButtonProps {
+/**
+ * Extends all the style properties from Box to our buttons as well as props from ButtonContainerProps.
+ * We omit `ref` since all of our buttons use `createComponent` and already give access to `ref`.
+ * Use this type to extend and customize any one off buttons that you want full control over styling.
+ */
+export interface PrimaryButtonProps extends Omit<ButtonContainerProps, 'ref'> {
```
mannycarrera4, line 14: Same thing here, we can extend `ButtonProps` from `Button.tsx`

modules/react/button/lib/Button.tsx
```diff
@@ -0,0 +1,67 @@
...
+export interface ButtonProps extends Omit<ButtonContainerProps, 'ref'> {}
+
+export const Button = createComponent('button')({
+  displayName: 'Button',
+  Component: (
+    {
+      children,
+      icon,
+      colors,
+      iconPosition = children ? undefined : 'only',
```
mannycarrera4, line 20: what happens with the `start | end` positions?

josh-bagwell: This was just fixed. I fixed it when you called out the actionBar issue

modules/react/button/lib/DeleteButton.tsx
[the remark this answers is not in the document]
josh-bagwell: Ahh I see what you mean

modules/react/button/lib/DeleteButton.tsx
josh-bagwell: Yeah definitely. Re-added.

modules/react/button/lib/Button.tsx
```diff
@@ -0,0 +1,67 @@
...
+export interface ButtonProps extends Omit<ButtonContainerProps, 'ref'> {}
+
+export const Button = createComponent('button')({
+  displayName: 'Button',
+  Component: (
+    {
+      children,
+      icon,
+      colors,
+      iconPosition = icon ? (children ? 'start' : 'only') : undefined,
```
mannycarrera4, line 20: Does `only` do anything? In the primary button storybook example, iconPosition isn't reference, there's just no children.

mannycarrera4: image

mannycarrera4:
Focus seems to stay visible even after clicking the button.
Repo:
1. Go to primary button storybook example
2. Hit tab to focus the first button
3. Click down on the first button
4. Notice that the focus ring stays visible even after clicking with the mouse

[Link](https://recordit.co/98AmXD4E7R)

modules/react/button/lib/Button.tsx
josh-bagwell: Yup. 

`only` refers to when a Button has only an icon with no children. So it maps to largeOnly, mediumOnly, smallOnly and extraSmallOnly

mannycarrera4: Where does that value of the prop be used? In the icon only variant, iconPosition isn't defined

mannycarrera4:
I think we should update the `CustomStyles` storybook example to use `Button` and maybe an example or two using the `cs` prop and `createStyles` to show people how to override things

modules/react/button/lib/Button.tsx
josh-bagwell: Which shouldn't cause an issue for users

modules/react/button/lib/Button.tsx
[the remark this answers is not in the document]
mannycarrera4: so no one can consume Button.tsx?

modules/react/button/lib/Button.tsx
mannycarrera4: It seems like I can still pass `only` to a button that has children and an icon and it breaks the padding. I'm just curious why we need `only`

modules/react/button/lib/DeleteButton.tsx
```diff
@@ -1,43 +1,44 @@
...
-    background: themeError.main,
-    label: themeError.contrast,
-  },
-  hover: {
-    background: themeError.dark,
+/**
+ * Extends all the style properties from Box to our buttons as well as props from ButtonContainerProps.
+ * We omit `ref` since all of our buttons use `createComponent` and already give access to `ref`.
+ * Use this type to extend and customize any one off buttons that you want full control over styling.
+ */
```
mannycarrera4, line 13: I don't think you need this comment here, since the original prop interface has it. Same for the other buttons

modules/react/button/lib/Button.tsx
```diff
@@ -0,0 +1,67 @@
...
+    Element
+  ) => {
+    return (
+      <BaseButton
+        as={Element}
+        ref={ref}
+        size={size}
+        icon={icon}
+        iconPosition={iconPosition}
+        type="button"
```
mannycarrera4, line 35:
```suggestion
```

mannycarrera4: this is already added on base button https://github.com/Workday/canvas-kit/pull/2285/files#diff-417ee64d3f922776d0d7fc16914309849f3820955d4110431c675c0f4043d885R35

modules/react/button/lib/DeleteButton.tsx
```diff
@@ -48,28 +49,11 @@ const getDeleteButtonColors = ({
...
-      colors={getDeleteButtonColors(theme)}
-      size={size}
-      {...elemProps}
-    >
-      <BaseButton.Label>{children}</BaseButton.Label>
-    </BaseButton>
-  ),
+  Component: ({children, colors, size, ...elemProps}: DeleteButtonProps, ref, Element) => {
+    return (
+      <Button as={Element} ref={ref} type="button" size={size} cs={[deleteStyles]} {...elemProps}>
```
mannycarrera4, line 54:
```suggestion
      <Button as={Element} ref={ref} size={size} cs={[deleteStyles]} {...elemProps}>
```

modules/react/button/lib/DeleteButton.tsx
josh-bagwell: If this is removed, then DeleteButtonProps won't have JSDocs for it. You'd have to hover over ButtonProps directly. Is that what we want to do ?

modules/react/button/lib/Button.tsx
mannycarrera4: Ok i think i know what's going on! this line here in `Button.tsx`:
```
iconPosition = icon ? (children ? 'start' : 'only') : undefined,
```
Is doing some funky stuff. You're changing the value here, but down below you have this logic:
```
 {icon && iconPosition === 'only' && (
          <BaseButton.Icon
            size={size}
            iconPosition={iconPosition}
            icon={icon}
            shouldMirrorIcon={shouldMirrorIcon}
          />
        )}
```
But it currently works because when you deconstruct the props, you set it to undefined if there's no children which trumps the logic below in the render function. So adding `iconPosition="only"` does nothing on an icon only button

NicholasBoll:
> Focus seems to stay visible even after clicking the button. Repo:
> 
> 1. Go to primary button storybook example
> 2. Hit tab to focus the first button
> 3. Click down on the first button
> 4. Notice that the focus ring stays visible even after clicking with the mouse
> 
> [Link](https://recordit.co/98AmXD4E7R)

@mannycarrera4 We're no longer using the `InputProvider` to determine if the focus ring should be visible. That was our own heuristics. We're now using [:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible):

> The :focus-visible pseudo-class applies while an element matches the [:focus](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus) pseudo-class and the UA ([User Agent](https://developer.mozilla.org/en-US/docs/Glossary/User_agent)) determines via heuristics that the focus should be made evident on the element. (Many browsers show a "focus ring" by default in this case.)

.storybook/preview.js
```diff
@@ -5,6 +5,9 @@ import routes from './routes';
 import {CanvasProviderDecorator} from '../utils/storybook';
 import theme from './theme';
 import {defaultCanvasTheme} from '@workday/canvas-kit-react/common';
+import "@workday/canvas-tokens-web/dist/css/base/_variables.css";
+import "@workday/canvas-tokens-web/dist/css/brand/_variables.css";
+import "@workday/canvas-tokens-web/dist/css/system/_variables.css";
```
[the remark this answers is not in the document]
NicholasBoll: For now, the exports of Canvas Tokens includes the `dist` folder. I think we wanted to remove it in the future, but we don't control that here.

package.json
```diff
@@ -176,6 +176,7 @@
   "browserslist": "> 5%, last 2 firefox versions, last 2 chrome versions, last 2 safari versions, last 2 edge versions, Firefox ESR, not ie 11",
   "dependencies": {
     "@emotion/react": "^11.7.1",
-    "@emotion/styled": "^11.6.0"
+    "@emotion/styled": "^11.6.0",
```
mannycarrera4, line 179: this shouldn't be down graded

mannycarrera4 (approved):

mannycarrera4:
EMOTIONAL DAMAGE

modules/react/button/lib/BaseButton.tsx
```diff
@@ -45,260 +56,267 @@ export interface ButtonContainerProps extends Partial<SystemIconProps>, GrowthBe
...
-          ...baseStyles,
-          ...hoverStyles,
-          ...activeStyles,
-          outline: 'none',
-          boxShadow: 'none',
-          animation: 'none',
-        },
-      }),
-    };
+  '&:disabled, &:active:disabled, &:focus:disabled, &:hover:disabled': {
```
NicholasBoll, line 168:
```suggestion
  '&:disabled, &:active:disabled, &:focus-visible:disabled, &:hover:disabled': {
```

NicholasBoll (approved):

Issue #1932, opened 2022-11-29
Convert Select to a Compound Component


willklein:
## 🚀 Feature Proposal

The Select component in Preview is atomic (single element API) and styled with custom HTML (`<button />` + `<input />`), while the Select in Main is compound-like and uses the native `<select>` element with `<option />` elements.

We would likely have an API akin to the one in Main but styled like the one in Preview, in the form of a compound component.

```[tasklist]
### Tasks
- [ ] #780 
```

PR #2309, opened 2023-08-03
feat: Convert Select from Main into a compound component

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/1932

## Release Category
Components

### BREAKING CHANGES
- We've converted `Select` in Main into a compound component. This component matches our pattern of providing access to lower-level elements and allows for more flexibility. 
- The spacing between menu and its target element will increase from 0px to 4px
- We've also [deprecated](https://canvas.workday.com/getting-started/canvas-glossary#deprecation) the `Select` in Preview. You may still consume this component but suggest migrating over to the one in Main.

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods


modules/react/combobox/lib/ComboboxInput.tsx
```diff
@@ -77,7 +79,10 @@ export const useComboboxInput = composeHooks(
           model.events.setWidth(event.currentTarget.clientWidth);
         }
       },
-      value: model.state.value,
+      value:
+        model.state.selectedIds.length > 0 && model.state.cursorId !== ''
+          ? model.state.items[currentValue].textValue
+          : model.state.value,
```
mannycarrera4, line 85: I had to do this because it was rendering the `id` as the value instead of the `textValue` in the input.

modules/react/combobox/lib/ComboboxMenuItem.tsx
```diff
@@ -37,6 +37,8 @@ export const useComboboxMenuItem = composeHooks(
       }
       if (event.currentTarget.getAttribute('aria-disabled') !== 'true') {
         model.events.select({id});
+        model.events.goTo({id});
```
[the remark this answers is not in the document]
mannycarrera4: `goTo` sets the cursorId

modules/react/combobox/lib/useComboboxTypeAhead.tsx
```diff
@@ -0,0 +1,105 @@
...
+        if (event.key.length === 1 && event.key.match(/\S/)) {
+          handleKeyboardTypeAhead(event.key, model.state.items.length);
+        }
+      },
+      style: {caretColor: 'transparent', cursor: 'default'},
+    };
+  }),
+
+  useComboboxInput
+);
```
mannycarrera4, line 105: Ignore this file for now, still WIP

modules/react/combobox/stories/examples/Autocomplete.tsx
```diff
@@ -88,7 +88,7 @@ export const Autocomplete = () => {
     },
     useComboboxModel
   );
-
+  console.log(model);
```
mannycarrera4, line 91:
```suggestion
```

modules/react/combobox/lib/ComboboxMenuPopper.tsx
```diff
@@ -0,0 +1,35 @@
...
+})<MenuPopperProps>(({children, ...elemProps}) => {
+  return (
+    <Popper
+      placement="bottom-start"
+      popperOptions={{
+        modifiers: [
+          {
+            name: 'offset',
+            options: {
+              offset: () => [0, 4],
```
mannycarrera4, line 25: TBD

modules/react/action-bar/lib/useActionBarModel.tsx
```diff
@@ -44,8 +44,6 @@ export const useActionBarModel = createModelHook({
   let nonInteractiveIds = model.state.nonInteractiveIds;
   const totalSize = model.state.items.length;
 
-  console.log(config.maximumVisible);
```
mannycarrera4, line 47: i think this was meant to be removed?

modules/react/combobox/lib/ComboboxMenu.tsx
```diff
@@ -25,7 +26,7 @@ export const ComboboxMenu = createSubcomponent()({
      * of React. This means no extra props given to this component will be forwarded to the `div`
      * element, but the `ref` will be forwarded.
      */
-    Popper: Menu.Popper,
+    Popper: ComboboxMenuPopper,
```
mannycarrera4, line 29: In case we want to change the popup behavior or style (changing the space between the menu and input for focus rings) I created a component to handle this change.

modules/react/combobox/lib/ComboboxMenuItem.tsx
```diff
@@ -63,6 +66,9 @@ export const useComboboxMenuItem = composeHooks(
 export const ComboboxMenuItem = createSubcomponent('li')({
   modelHook: useComboboxModel,
   elemPropsHook: useComboboxMenuItem,
+  subComponents: {
+    Icon: styled(SystemIcon)({alignSelf: 'start'}),
```
mannycarrera4, line 70: This allows Icons to be used in menu items.

modules/react/form-field/lib/FormField.tsx
```diff
@@ -231,7 +232,13 @@ export class FormField extends React.Component<React.PropsWithChildren<FormField
       );
     } else {
       return (
-        <FormFieldContainer useFieldset={useFieldset} labelPosition={labelPosition} {...elemProps}>
+        <FormFieldContainer
+          id="foo"
+          useFieldset={useFieldset}
+          labelPosition={labelPosition}
+          grow={grow}
+          {...elemProps}
```
mannycarrera4, line 240: idk why this prop wasn't forwarded here. If usefieldset is not used, consumers should still be able to set width to 100%. Look at `grow` story for select for reference.

modules/react/select/lib/Select.tsx
```diff
@@ -1,129 +1,220 @@
...
+ * and interaction as defined by [WAI](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/).
+ * All valid attributes are spread to the underlying `<input>` element.
+ */
+export const SelectInput = createSubcomponent(TextInput)({
+  modelHook: useSelectModel,
+  elemPropsHook: useSelectInput,
+})<SelectInputProps>(
+  ({placeholder = 'Choose an Option', width, inputStartIcon, ...props}, Element, model) => {
+    return (
+      <InputGroup width={width}>
```
mannycarrera4, line 39: Because the parent container wraps the input, I want to be sure to pass the width to it in case they want to expand the width of it. Everything else gets spread to the lower level input element.

modules/react/select/lib/Select.tsx
```diff
@@ -1,129 +1,220 @@
...
-  position: 'absolute',
-  top: space.xxxs,
-  right: space.xxxs,
-  padding: space.xxxs,
-  pointerEvents: 'none',
-  '& path': {
-    transition: '100ms fill',
+export const SelectItem = createSubcomponent('li')({
+  modelHook: useSelectModel,
+  elemPropsHook: useSelectItem,
```
mannycarrera4, line 62: this just adds `role="option"` to the menu item.

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,209 @@
+import React from 'react';
+import {composeHooks, createElemPropsHook} from '@workday/canvas-kit-react/common';
+import {useComboboxInput} from '@workday/canvas-kit-react/combobox';
+import {useSelectModel} from './useSelectModel';
+
+/**
+ * `useSelectInput` extends `useComboboxInput` while adding type ahead functionality and adding select specific [keyboard support](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/)
```
mannycarrera4, line 7: update description

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,209 @@
...
+ */
+export const useSelectInput = composeHooks(
+  useComboboxInput,
+  createElemPropsHook(useSelectModel)((model, ref) => {
+    const keySofar = React.useRef('');
+    const timer = React.useRef<ReturnType<typeof setTimeout>>();
+
+    React.useLayoutEffect(() => {
+      model.events.setWidth(model.state.targetRef.current?.clientWidth || 0);
```
mannycarrera4, line 16: Before the popup is open, set the width to the the width of the input. In a case where the consumer sets `initialVisibility` to `visible` for some odd reason, the popup will have the width and the items will be visible.

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,209 @@
...
+          handleKeyboardTypeAhead(event.key, model.state.items.length);
+        }
+
+        // If the dropdown is NOT visible and ArrowUp, ArrowDown, Enter and Spacebar is typed, when the dropdown opens
+        // it should go to the current selected item in the dropdown.
+        if (keysToOpenSelect.includes(event.key) || event.key === ' ') {
+          if (model.state.visibility === 'hidden') {
+            model.events.goTo({id: model.state.items[foundIndex].id});
+
+            if (model.state.isVirtualized && model.state.selectedIds.length > 0) {
```
mannycarrera4, line 143: what if it's not virtualized?

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,209 @@
...
+        // where the cursor is located in the list and close the dropdown
+        if (event.key === 'Enter' && !event.metaKey && model.state.visibility === 'visible') {
+          const element = document.querySelector(`[data-id="${model.state.cursorId}"]`);
+          if (element && element?.getAttribute('aria-disabled') !== 'true') {
+            model.events.select({id: model.state.cursorId});
+            if (model.state.mode === 'single') {
+              model.events.hide(event);
+            }
+          }
+        }
```
mannycarrera4, line 159: this might be handled by combobox already, need to verify.

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,209 @@
...
+          if (keySofar.current !== '') {
+            handleKeyboardTypeAhead(' ', model.state.items.length);
+          }
+        }
+      },
+
+      onClick(event: React.MouseEvent) {
+        if (model.state.visibility === 'hidden') {
+          model.events.setWidth(event.currentTarget.clientWidth);
+        }
```
mannycarrera4, line 195: again, may not need this since I have something similar in useEffect

modules/react/select/lib/hooks/useSelectModel.tsx
```diff
@@ -0,0 +1,27 @@
...
+    ...useComboboxModel.defaultConfig,
+  },
+  requiredConfig: {
+    ...useComboboxModel.requiredConfig,
+  },
+  contextOverride: useComboboxModel.Context,
+})(config => {
+  const model = useComboboxModel(config);
+  return model;
+});
```
mannycarrera4, line 27: ah composability, what a beauty

modules/react/form-field/lib/FormField.tsx
```diff
@@ -231,7 +232,13 @@ export class FormField extends React.Component<React.PropsWithChildren<FormField
       );
     } else {
       return (
-        <FormFieldContainer useFieldset={useFieldset} labelPosition={labelPosition} {...elemProps}>
+        <FormFieldContainer
+          id="foo"
```
mannycarrera4, line 236: remove

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,204 @@
...
+          const selectedItem = model.state.items.findIndex(
+            (item: {id: string}) => item.id === model.state.selectedIds[0]
+          );
+          model.events.goTo({id: model.state.items[selectedItem].id});
+          model.events.select({id: model.state.items[selectedItem].id});
+        }
+      }
+
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
```
mannycarrera4, line 32: This whole effect can prob go itno combobox. It's just selecting the item if there is one, if not,  moving the cursor to the first item and setting the widht of it to the traget.

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,204 @@
...
+          } else {
+            return item.id === model.state.cursorId;
+          }
+        });
+
+        // Prevent the keys from being enter in the input
+        event.preventDefault();
+
+        // Select should open if Enter, ArrowUp, ArrowDown and Spacebar is typed
+        const keysToOpenSelect = ['Enter', 'Spacebar', 'ArrowUp', 'ArrowDown'];
```
mannycarrera4, line 121: i believe combobox handles arrow up and arrow down

modules/react/select/stories/Select.stories.mdx
jamesfan, line 1: I noticed some strange scrolling behavior when clicking on examples in the Storybook docs.

1. Load the Select docs in SB ([latest build](https://5d854c26ba934e0020f5e98a-kyqhsbtnco.chromatic.com/?path=/docs/components-inputs-select--basic) as of the writing of this comment)
2. Scroll to an example further down on the page, such as `With Icons`.
3. Click on the Select Input and select any item in the menu **other than the first one**. The menu will collapse and the selected item will be represented in the Select Input.
4. Click on the Select Input again. The page will scroll to the top.

Strangely, if your **first and only** selection is the first item in the menu, the scrolling will not occur. For example, if you select `Activity Stream` in the `With Icons` example, no scrolling will occur when you click on the Select Input again.

mannycarrera4: interesting....

mannycarrera4: I wonder if that has anything to do with virtualization, which I set to false in the model

mannycarrera4: Looks like the dom element for the popup is still in the dom even after it is closed

modules/docs/mdx/10.0-UPGRADE-GUIDE.mdx
```diff
@@ -70,6 +73,10 @@ and browsers can compute the name of the Banner from the text given inside.
 We've deprecated `InputIconContainer` from [Main](#main) because it doesn't handle
 bidirectionality or icons at the start of an input. Please consider using [`InputGroup`](https://workday.github.io/canvas-kit/?path=/story/components-inputs-text-input--icons) instead.
 
+### Select Preview
+
+We've deprecated `Select` from [Preview](#preview). Please consider using [`Select`](https://workday.github.io/canvas-kit/?path=/story/components-inputs-select--basic) instead.
```
mannycarrera4, line 78: Select in Main?

modules/preview-react/select/lib/SelectBase.tsx
```diff
@@ -32,36 +32,58 @@ interface OptionData {
   // function without encountering TypeScript errors
   [key: string]: any;
 }
-
+/**
+ * @deprecated
+ */
```
mannycarrera4, line 37: Do we attach the same messaging here?

modules/react/select/stories/Select.stories.mdx
```diff
@@ -34,79 +34,200 @@ yarn add @workday/canvas-kit-react
...
+[dynamic API](/getting-started/for-developers/resources/collection-api/#dynamic-items) where you
+pass an array of `items` in the `model` config and provide a render function to display the items.
+It's important to set up the object in your items array in a way for the
+[collection system](/getting-started/for-developers/resources/collection-api/) to understand.
+
+#### Array of strings
+
+If you have an array of `strings`, the collection system will handle asigning an `id`, `index` and
+`textValue` to each item in the array. This ensures that both type ahead and keyboard navigation
+work correclty.
```
jamesfan, line 136:
```suggestion
work correctly.
```

modules/react/select/stories/Select.stories.mdx
```diff
@@ -34,79 +34,200 @@ yarn add @workday/canvas-kit-react
...
 
-### Ref Forwarding
+Set `nonInteractiveIds` to an array of `id`s that you want to disable interaction as well as
+searchability when typing into the `Select`.
 
-Select supports [ref forwarding](https://reactjs.org/docs/forwarding-refs.html). It will forward
-`ref` to its underlying `<select>` element.
+You will also need to set the aria attribute `aria-disabled` on the `Select.Item` if the item is
+disabled, otherwise set the attribute to `undefined`. `disabled` in this case is a property on the
+item's object. This ensures that the item is styled as disabled.
```
jamesfan, line 79: GH isn't allowing me to include a multi-line suggestion for the entire `Disabled Items` section, so I'm pasting it here:

```
### Disabled Items

In order to disable items and prevent users from interacting with them:

1. Set the `nonInteractiveIds` prop of `Select` to an array of disabled item `id`s. This will
   disable interaction for those items and exclude them from type-ahead.

2. Set the `aria-disabled` attribute of all disabled `Select.Item`s to `true`. This ensures the
   items are styled as disabled.

The following example assigns a `disabled` property to the disabled items in the `items` array.
`disabled` is used to populate `nonInteractiveIds` and set `aria-disabled` for the disabled items.
```

I did a few things here:

1. Updated language for consistency.
2. Replaced "searchability" with "type-ahead." I lifted the "type-ahead" nomenclature from the [WAI Listbox](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/) page.
3. Moved the bit about the `disabled` property in `items` to the very end.

mannycarrera4: Nice, i like the clear numbered steps, and the separation of the info of objects.

modules/react/combobox/lib/ComboboxInput.tsx
```diff
@@ -69,7 +88,11 @@ export const useComboboxInput = composeHooks(
         }
       },
       onBlur(event: React.FocusEvent) {
-        model.events.hide(event);
+        if (model.state.nonInteractiveIds.length > 0) {
+          return;
+        } else {
+          model.events.hide(event);
+        }
```
RayRedGoose, line 95:
```suggestion
        if (model.state.nonInteractiveIds.length <= 0) {
          model.events.hide(event);
        }
        return;
```
Will this work instead of full if/else?

modules/react/combobox/lib/ComboboxMenuItem.tsx
```diff
@@ -26,8 +29,22 @@ export interface ComboboxMenuItemProps {
...
+    // focus on the item with the cursor
+    React.useLayoutEffect(() => {
+      if (model.state.mode === 'single') {
+        if (model.state.cursorId === id) {
+          // delay focus changes to allow PopperJS to position
+          requestAnimationFrame(() => {
+            localRef.current?.focus();
+          });
+        }
+      }
```
RayRedGoose, line 45:
```suggestion
      if (model.state.mode === 'single' && model.state.cursorId === id) {
          // delay focus changes to allow PopperJS to position
          requestAnimationFrame(() => {
            localRef.current?.focus();
          });
      }
```

No need to use nested if statements, as you have only one block

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,162 @@
...
+        if (label.indexOf(startString.toLowerCase()) === 0) {
+          if (
+            !ignoreDisabled ||
+            (ignoreDisabled && !model.state.nonInteractiveIds.includes(model.state.items[i].id))
+          ) {
+            return i;
+          }
+        }
+      }
+      return -1;
```
RayRedGoose, line 46: maybe to use ternary and set condition as variable:
```
const isNotDisabled = label.indexOf(startString.toLowerCase()) === 0 && (!ignoreDisabled || !model.state.nonInteractiveIds.includes(model.state.items[i].id))

return isNotDisabled ? i : -1;

```

you have nested `if`s, so you can combine them. on line 40 you don't need to check `ignoreDisabled` because you're checking this on line 39

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,162 @@
...
+    // just use the index you dummy squid
+    const currentItem = model.navigation.getItem(model.state.cursorId, model);
+    const cursorFocusedIndex = model.state.items.findIndex(item => item.id === currentItem.id);
+
+    const handleKeyboardTypeAhead = (key: string, numOptions: number) => {
+      // If the starting point is beyond the list of options, reset it
+      // to the beginning of the list
+      let start = keySofar.current.length === 0 ? cursorFocusedIndex + 1 : cursorFocusedIndex;
+
+      start = start === numOptions ? 0 : start;
```
RayRedGoose, line 57:
```suggestion
      const number = keySofar.current.length === 0 ? cursorFocusedIndex + 1 : cursorFocusedIndex;
      const start = number === numOptions ? 0 : number;
```
Maybe you can do not have changeable var. Do you reassign this var in different place?

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    return {
+      onKeyDown(event: React.KeyboardEvent) {
+        const foundIndex = model.state.items.findIndex((item: {id: string}) => {
+          if (model.state.selectedIds.length > 0) {
+            return item.id === model.state.selectedIds[0];
+          } else {
+            return item.id === model.state.cursorId;
+          }
+        });
```
RayRedGoose, line 97:
```suggestion
        const foundIndex = model.state.items.findIndex(
          (item: {id: string}) => 
            (model.state.selectedIds.length > 0 && item.id === model.state.selectedIds[0]) || item.id === model.state.cursorId
        );
```
another way to set condition

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,162 @@
...
+          (event.key === 'Spacebar' || event.key === ' ') &&
+          model.state.visibility === 'hidden' &&
+          keySofar.current !== ''
+        ) {
+          // If the user is in the middle of typing a string, treat
+          // space key as type-ahead rather than option selection
+          if (keySofar.current !== '') {
+            handleKeyboardTypeAhead(' ', model.state.items.length);
+          }
+        }
```
RayRedGoose, line 157: nested ifs, you can combine them

mannycarrera4: master code cleaner 🙏

modules/react/collection/index.ts
```diff
@@ -15,9 +15,14 @@ export * from './lib/useGridModel';
...
 export {
   singleSelectionManager,
   multiSelectionManager,
   isSelected,
 } from './lib/useSelectionListModel';
-export {wrappingNavigationManager, navigationManager} from './lib/useCursorListModel';
+export {
+  wrappingNavigationManager,
+  navigationManager,
+  createNavigationManager,
```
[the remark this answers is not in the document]
mannycarrera4: i don't think i need to export these anymore

modules/react/combobox/lib/ComboboxInput.tsx
mannycarrera4: i think so?

modules/react/select/stories/Select.stories.mdx
```diff
@@ -34,79 +34,209 @@ yarn add @workday/canvas-kit-react
...
 #### Error
 
+Use the error state when the selection is no longer valid.
+
+> **Note**
+>
+> The blade-old package has been moved to it's own repo.
+> [github.com/razorpay/blade-old](https://github.com/razorpay/blade-old) (Only accessible to Razorpay employees)
+>
+> It is under **maintenance** and it won't have any major releases.
```
jamesfan, line 236: This is a strange note. An inadvertent copy-and-paste?

```suggestion
Use the error state when the selection is no longer valid.
```

modules/react/select/stories/examples/WithIcons.tsx
```diff
@@ -0,0 +1,53 @@
...
+    items: customOptions,
+  });
+  const currentItem = model.navigation.getItem(model.state.selectedIds[0], model);
+  return (
+    <Flex>
+      <Select model={model}>
+        <FormField label="Contact" inputId="with-icons-select">
+          <Select.Input
+            width="300px"
+            inputStartIcon={currentItem.value.icon}
```
jamesfan, line 30: What do you think about replacing `currentItem` with `selectedItem` for clarity?
```suggestion
  const selectedItem = model.navigation.getItem(model.state.selectedIds[0], model);
  return (
    <Flex>
      <Select model={model}>
        <FormField label="Contact" inputId="with-icons-select">
          <Select.Input
            width="300px"
            inputStartIcon={selectedItem.value.icon}
```

modules/react/select/stories/Select.stories.mdx
```diff
@@ -34,79 +34,209 @@ yarn add @workday/canvas-kit-react
...
+
+- Register your `items` with the `useSelectModel`
+- Get current item by
+  `const currentItem = model.navigation.getItem(model.state.selectedIds[0], model)`
+- Pass the value to the input in the following way:
+  `<Select.Input inputStartIcon={currentItem.value.icon} id="form-field-input-id" />`
+
+**Notes: The `Select.Input` will only render an icon when there is an item selected.**
+
+<ExampleCodeBlock code={WithIcons} />
```
jamesfan, line 103: A few updates for the `With Icons` example:

1. Cleaned up and updated language for consistency.
2. Linked to the `/assets/system-icons/` route. This content isn't great in SB, but it's better on the Canvas Site.
3. Moved the note about `Select.Input` to the end after the example and removed the bold since it didn't seem necessary to emphasize this.

```suggestion
### With Icons

Use `Select.Item.Icon` to render an icon for a `Select.Item`. The `icon` prop for `Select.Item.Icon` accepts
[system icons](/assets/system-icons/) from `@workday/canvas-system-icons-web`.

In order to render the icon for the selected item in the `Select.Input`:

1. Obtain a reference to the `model` by registering your `items` with `useSelectModel`.
4. Get the selected item: `const selectedItem = model.navigation.getItem(model.state.selectedIds[0], model)`
5. Pass the icon for the selected item to the input: `<Select.Input inputStartIcon={selectedItem.value.icon}>`

<ExampleCodeBlock code={WithIcons} />

Note that `Select.Input` will only render an icon if an item is selected.
```

modules/react/select/stories/Select.stories.mdx
mannycarrera4: lol yes, I was testing out creating a note, sorry!

modules/react/select/lib/hooks/useSelectInput.ts
mannycarrera4: This one turned a little tricky. It didn't quite work as I expected

modules/react/combobox/lib/ComboboxPopper.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const useMenuPopper = usePopupPopper;
+
+export const ComboboxMenuPopper = createSubcomponent('div')({
+  displayName: 'Combobox.Menu.Popper',
+  modelHook: useMenuModel,
+  elemPropsHook: useMenuPopper,
+})<MenuPopperProps>(({children, ...elemProps}) => {
+  return (
+    <Menu.Popper fallbackPlacements={['top', 'bottom']} {...elemProps}>
```
mannycarrera4, line 18: I specifically created this file to update the fallback placements. The popper should only open at the top or bottom, otherwise the fallbacks include left and right and this is not intended behavior

modules/react/avatar/lib/Avatar.tsx
```diff
@@ -38,7 +38,7 @@ export interface AvatarProps extends React.ButtonHTMLAttributes<HTMLButtonElemen
   as?: 'div';
   /**
    * The object-fit CSS property sets how the content of a replaced element,
-   * such as an <img> or <video>, should be resized to fit its container.
+   * such as an `<img>` or `<video>`, should be resized to fit its container.
```
mannycarrera4, line 41: The canvas site interpreted these as elements, so I put them in ticks to avoid that

modules/preview-react/select/lib/SelectMenu.tsx
```diff
@@ -228,6 +228,11 @@ const generatePopperOptions = (
   };
 };
 
+/**
+ * ### ⚠️ Select in Preview has been deprecated and will be removed in a future major version. ⚠️
```
mannycarrera4, line 232:
```suggestion
 * ### ⚠️ `SelectMenu` in Preview has been deprecated and will be removed in a future major version. ⚠️
```

modules/preview-react/select/lib/SelectMenu.tsx
```diff
@@ -228,6 +228,11 @@ const generatePopperOptions = (
...
+ * - Please consider using [`Select`](https://workday.github.io/canvas-kit/?path=/docs/components-inputs-select--basic) in Main
```
mannycarrera4, line 233:
```suggestion
 * - Please use [`Select` in Main](https://workday.github.io/canvas-kit/?path=/docs/components-inputs-select--basic) instead.
```

modules/react/collection/index.ts
```diff
@@ -15,6 +15,7 @@ export * from './lib/useGridModel';
 export * from './lib/useListActiveDescendant';
 export * from './lib/useListItemAllowChildStrings';
 export {ListBox, ListBoxProps} from './lib/ListBox';
+export {keyboardEventToCursorEvents} from './lib/keyUtils';
```
mannycarrera4, line 18: export to use in comboboxInput

modules/react/collection/stories/examples/RegisteringItems.tsx
```diff
@@ -0,0 +1,46 @@
...
+];
+
+export const RegisteringItems = () => {
+  return (
+    <Flex gap="m" display="inline-flex">
+      <Flex flexDirection="column">
+        <Text marginBottom="m" fontWeight="bold">
+          An Array of Objects
+        </Text>
+        <ListBox items={items} width="200px" maxHeight={200}>
```
mannycarrera4, line 28:
```suggestion
        <ListBox items={items} width="200px">
```

modules/react/collection/stories/examples/RegisteringItems.tsx
```diff
@@ -0,0 +1,46 @@
...
+export const RegisteringItems = () => {
+  return (
+    <Flex gap="m" display="inline-flex">
+      <Flex flexDirection="column">
+        <Text marginBottom="m" fontWeight="bold">
+          An Array of Objects
+        </Text>
+        <ListBox items={items} width="200px">
+          {item => (
+            <ListBox.Item paddingBottom="s" data-id={item.id}>
```
mannycarrera4, line 30: technically I don't need `data-id` here

modules/react/combobox/lib/ComboboxInput.tsx
```diff
@@ -4,20 +4,36 @@ import {
...
+      if (model.state.visibility === 'visible') {
+        const handled = keyboardEventToCursorEvents(event, model, isRTL);
+        if (handled) {
+          event.preventDefault();
+        }
+      }
+    },
+  };
+});
+
```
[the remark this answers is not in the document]
mannycarrera4: Also should these hooks move to `./hook` ?

modules/react/combobox/lib/ComboboxInput.tsx
```diff
@@ -40,21 +60,20 @@ export const useComboboxInput = composeHooks(
               `[id="${listboxId}"] [data-id="${model.state.cursorId}"]`
             );
             if (menuItem) {
-              menuItem.scrollIntoView({block: 'nearest'});
+              requestAnimationFrame(() => {
+                menuItem.scrollIntoView({block: 'nearest'});
```
mannycarrera4, line 64: This prevents the popup from scrolling to the top of the page, re force another re render so that the browser knows where the popup is localted

modules/react/combobox/lib/ComboboxInput.tsx
```diff
@@ -40,21 +60,20 @@ export const useComboboxInput = composeHooks(
...
+              });
             }
           }
         }
       }
 
       // we only want to run this effect if the cursor changes and not any other time
```
mannycarrera4, line 71: update comment

modules/react/combobox/lib/ComboboxMenuItem.tsx
```diff
@@ -26,8 +29,20 @@ export interface ComboboxMenuItemProps {
...
     const id = elemProps['data-id'] || '';
+    const {localRef, elementRef} = useLocalRef(ref as React.Ref<HTMLElement>);
+
+    // focus on the item with the cursor
+    React.useLayoutEffect(() => {
+      if (model.state.mode === 'single' && model.state.cursorId === id) {
+          // delay focus changes to allow PopperJS to position
+          requestAnimationFrame(() => {
+            localRef.current?.focus();
+          });
```
mannycarrera4, line 42: investigate if I need these here

modules/react/combobox/lib/hooks/useSetPopupWidth.ts
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import {createElemPropsHook} from '@workday/canvas-kit-react/common';
+import {useComboboxModel} from './useComboboxModel';
+
+export const useSetPopupWidth = createElemPropsHook(useComboboxModel)(model => {
+  React.useLayoutEffect(() => {
+    // If there is no selected item and items exists we want to set the cursor to the first item in the array
```
mannycarrera4, line 7: update comment

modules/react/menu/lib/MenuPopper.tsx
```diff
@@ -15,7 +15,20 @@ export const MenuPopper = createSubcomponent('div')({
...
   return (
-    <Popper placement="bottom-start" {...elemProps}>
+    <Popper
+      placement="bottom-start"
+      popperOptions={{
+        modifiers: [
+          {
+            name: 'offset',
+            options: {
+              offset: () => [0, 4],
```
mannycarrera4, line 25: this is the `4px` offset

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -0,0 +1,65 @@
...
+  createElemPropsHook,
+  useForkRef,
+  useLocalRef,
+  useResizeObserver,
+} from '@workday/canvas-kit-react/common';
+import {useComboboxInput, useKeyboardTypeAhead} from '@workday/canvas-kit-react/combobox';
+import {useSelectModel} from './useSelectModel';
+
+/**
+ * `useSelectInput` extends {@link useComboboxInput useComboboxInput}  and adds type ahead functionality and select specific [keyboard support](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/).
```
mannycarrera4, line 13: and type ahead

modules/react/select/stories/Select.stories.mdx
```diff
@@ -34,79 +35,148 @@ yarn add @workday/canvas-kit-react
...
-also need to set the `hintId` and `hintText` props on the Form Field to meet accessibility
-standards.
+### Registering Items
 
-The `error` prop may be applied directly to the Select with a value of `Select.ErrorType.Alert` or
-`Select.ErrorType.Error` if Form Field is not being used.
+You can register items in two way.
+
+1. An Array of Strings
+2. An Array objects
```
mannycarrera4, line 139: fix

modules/react/avatar/lib/Avatar.tsx
jamesfan: This fix should go into a separate PR against the `support` branch since the PR which introduced the issue (#2337) was merged into `support`.

modules/react/collection/stories/Collection.stories.mdx
```diff
@@ -53,13 +55,42 @@ item for selection, maintaining a cursor, or anything else.
...
-statically, provide a render function instead. The function is called with an `items` value that is
-the same was what's provided to the model. By default, providing items will enable virtualization.
-This example adds a `maxHeight` to ensure overflow. Virtualization uses absolute positioning of each
-item, which could cause problems for popup menus. If your item count is low, pass
-`shouldVirtualize={false}` to disable virtualization.
+statically, provide a render function instead. The function is called with an `item` value that is
+the same as what's provided to the model.
 
-<ExampleCodeBlock code={DynamicItems} />
+#### Registering Items
```
jamesfan, line 61: @mannycarrera4 [Originally](https://5d854c26ba934e0020f5e98a-bdjnqbgofl.chromatic.com/?path=/docs/features-collections--basic), you had `Registering Items` as an h3 alongside `Dynamic Items`. I moved it to an h4 _under_ `Dynamic Items` since it applies specifically to the dynamic API. I also condensed some of the text and replaced the code snippets with your strings/objects examples. Snippets seemed redundant given the examples.

modules/react/select/lib/Select.tsx
```diff
@@ -101,19 +95,23 @@ export const Select = createContainer()({
...
      * **Note: An `id` must be provided that matches the `inputId` attribute on the `FormField` component from Main in order for the `label` element to be associated with the `input` element.**
      *
      * ```tsx
      * <Select items={options}>
-     *   <Select.Input id="matching-formfield-inputId" onChange={(event) => handleChange(event)}>
-     *   ...
+     *   <FormField label="Contact" inputId="matching-formfield-inputId">
+     *     <Select.Input id="matching-formfield-inputId" onChange={(event) => handleChange(event)}>
+     *     ...
+     *   </FormField>
```
mannycarrera4, line 107: so you opted in to adding this! makes sense!

modules/react/collection/stories/Collection.stories.mdx
```diff
@@ -53,13 +55,42 @@ item for selection, maintaining a cursor, or anything else.
...
+be included in the object and will be accessible in the render function.
+
+<ExampleCodeBlock code={DynamicItemsObjects} />
+
+> **Note: The `id` should match your text value for each item. This ensures proper selection of
+> items in components like `Select` and `Autocomplete`. If your `id` is different from your text
+> value, you'll have to add a
+> [translation layer](https://codesandbox.io/s/combobox-object-items-forked-mc7tm4). **
+
+#### Virtualization
```
jamesfan, line 86: I made `Virtualization` into its own subsection underneath `Dynamic Items`.

modules/react/select/stories/Select.stories.mdx
```diff
@@ -34,79 +34,145 @@ yarn add @workday/canvas-kit-react
...
 ### Basic Example
 
-Select should be used in tandem with [Form Field](/components/inputs/form-field/) to meet
-accessibility standards.
+`Select` supports a
+[dynamic API](/getting-started/for-developers/resources/collection-api/#dynamic-items) where you
+pass an array of items via the `items` prop and provide a render function to display the items. The
+items may be provided as an
+[array of strings](/getting-started/for-developers/resources/collection-api/#array-of-strings) or an
+[array of objects](/getting-started/for-developers/resources/collection-api/#array-of-objects).
```
jamesfan, line 42: [Your original Select docs](https://5d854c26ba934e0020f5e98a-bdjnqbgofl.chromatic.com/?path=/docs/components-inputs-select--basic), for reference.

Originally, you had a separate `Registering Items` section in the Select MDX with examples. I decided to move that content into the `Basic Example` and link out to the `Array of Strings` and `Array of Objects` sections in the Collection MDX. It seemed like that would be sufficient instead of repeating that content here.

modules/react/select/stories/Select.stories.mdx
```diff
@@ -34,79 +34,145 @@ yarn add @workday/canvas-kit-react
...
+Our example uses [React state](<(https://react.dev/learn/state-a-components-memory)>) to track the
+value of the `Select`.
+
+### Hoisted Model
+
+By default, `Select` will create and use its own model internally. Alternatively, you may configure
+your own model with `useSelectModel` and pass it to `Select` via the `model` prop. This pattern is
+referred to as
+[hoisting the model](/getting-started/for-developers/resources/compound-components/#configuring-a-model)
+and provides direct access to its `state` and `events` outside of the `Select` component.
```
jamesfan, line 60: [Your original Select docs](https://5d854c26ba934e0020f5e98a-bdjnqbgofl.chromatic.com/?path=/docs/components-inputs-select--basic), for reference.

I updated the `Hoisted Model` section and modeled it after the [`Hoisted Model` section in Tabs](https://github.com/Workday/canvas-kit/blob/master/modules/react/tabs/stories/Tabs.stories.mdx#L59). I also updated the example itself to provide a use case for why you might want to hoist the model (example is contrived, but it should get the job done).

modules/react/select/stories/Select.stories.mdx
```diff
@@ -34,79 +34,145 @@ yarn add @workday/canvas-kit-react
...
 
-### Error States
+`Select.Card` has a default maximum height of `300px` to restrict the height of the dropdown menu.
+Set its `maxHeight` prop to override this value.
 
-Set the `error` prop of the wrapping Form Field to `FormField.ErrorType.Alert` or
-`FormField.ErrorType.Error` to set the Select to the Alert or Error state, respectively. You will
-also need to set the `hintId` and `hintText` props on the Form Field to meet accessibility
-standards.
+<ExampleCodeBlock code={MenuHeight} />
```
jamesfan, line 134: I renamed the `Scrollable` section and example to `Menu Height` to better reflect its content related to `maxHeight`.

modules/react/select/stories/Select.stories.mdx
```diff
@@ -34,79 +34,145 @@ yarn add @workday/canvas-kit-react
...
+
 <ExampleCodeBlock code={Error} />
 
 ## Component API
 
 <SymbolDoc name="Select" fileName="/react/" />
-
-## Specifications
-
-<Specifications file="Select.spec.ts" name="Select" />
```
jamesfan, line 112: We should add the Specifications back in, no?

modules/react/select/stories/examples/Basic.tsx
```diff
@@ -1,21 +1,49 @@
...
+  {id: 'E-mail', data: {textValue: 'foo'}},
+  {id: 'Phone'},
+  {id: 'Fax (disabled)', disabled: true},
+  {id: 'Mail'},
+  {id: 'Mobile Phone'},
+  {
+    id: 'The Ontologically Anthropocentric Sensory Immersive Simulation',
+    disabled: false,
+  },
+];
```
jamesfan, line 16: Where possible, can we use an array of strings instead of objects for our examples? Some examples will need to use objects (e.g., `Disabled Items` needs to use the `disabled` key), but many of the examples should work with simple strings. This also means we should be able to delete the disabled-related code (`disabledItems` and `nonInteractiveIds`) from most of our examples.

For example, I don't think the `Basic Example` needs to include any disabled items given the amount of code the disabled feature introduces. Best to limit the disabled code to the `Disabled Items` example.

modules/react/select/stories/examples/Basic.tsx
```diff
@@ -1,21 +1,49 @@
 import React from 'react';
 import {FormField} from '@workday/canvas-kit-react/form-field';
-import {Select, SelectOption} from '@workday/canvas-kit-react/select';
+import {Select} from '@workday/canvas-kit-react/select';
+import {Flex} from '@workday/canvas-kit-react/layout';
+
+const options = [
+  {id: 'E-mail', data: {textValue: 'foo'}},
```
jamesfan, line 7: This may be a moot point after [simplifying the examples](https://github.com/Workday/canvas-kit/pull/2309#discussion_r1350931690), but we should remove `data: {textValue: 'foo'}` if there are any lingering around after simplifying the examples.

modules/react/collection/stories/Collection.stories.mdx
```diff
@@ -53,13 +55,42 @@ item for selection, maintaining a cursor, or anything else.
...
+Use objects when your data contains additional information. Each object must have an `id` to ensure
+it's registered and searchable. Additional information such as icons or if the item is disabled can
+be included in the object and will be accessible in the render function.
+
+<ExampleCodeBlock code={DynamicItemsObjects} />
+
+> **Note: The `id` should match your text value for each item. This ensures proper selection of
+> items in components like `Select` and `Autocomplete`. If your `id` is different from your text
+> value, you'll have to add a
+> [translation layer](https://codesandbox.io/s/combobox-object-items-forked-mc7tm4). **
```
jamesfan, line 84: I'd suggest removing `Autocomplete` since that implies we have an `Autocomplete` component in canvas-kit. The linked Codesandbox also has a few issues:

1. "Submist" should be "Submit".
2. The item `label`s should be fruits instead of Foo, Bar, etc.
3. I'm hoping this is a bug with the example and not the component, but I noticed the up/down arrow keys don't work for navigating the menu.

modules/react/select/stories/examples/Basic.tsx
mannycarrera4: yes!

modules/react/select/stories/examples/Basic.tsx
mannycarrera4: fair point

modules/react/select/stories/examples/Basic.tsx
mannycarrera4: fixed

modules/react/collection/stories/Collection.stories.mdx
mannycarrera4: 3. ... not sure how to fix that. The keyboard handler is using `id` and not the label

mannycarrera4: @NicholasBoll I'm not sure this will ever work since keyboard handling depends on the id

NicholasBoll: The example was missing the `getId` function. It is fixed now.

modules/react/collection/stories/Collection.stories.mdx
NicholasBoll: >  By default, providing items will enable virtualization. This example adds a `maxHeight` to ensure overflow. Virtualization uses absolute positioning of each item, which could cause problems for popup menus

Menus default virtualization to `false`.

Also, what issues? It might be useful to describe what issues may be encountered. For example, an HTML container element can render to the width of children, but if the children are not always in the DOM, the min width may not be calculated according to the widest item. A `width` or `minWidth` should be set on virtualized lists.

modules/react/collection/stories/Collection.stories.mdx
```diff
@@ -53,13 +55,42 @@ item for selection, maintaining a cursor, or anything else.
...
+
+#### Virtualization
+
+By default, providing items dynamically will enable virtualization. This example adds a `maxHeight`
+to ensure overflow. Virtualization uses absolute positioning of each item, which could cause
+problems for popup menus. If your item count is low, pass `shouldVirtualize={false}` to disable
+virtualization.
```
NicholasBoll, line 91: >  By default, providing items will enable virtualization. This example adds a `maxHeight` to ensure overflow. Virtualization uses absolute positioning of each item, which could cause problems for popup menus

Menus default virtualization to `false`.

Also, what issues? It might be useful to describe what issues may be encountered. For example, an HTML container element can render to the width of children, but if the children are not always in the DOM, the min width may not be calculated according to the widest item. A `width` or `minWidth` should be set on virtualized lists.

modules/react/collection/stories/Collection.stories.mdx
NicholasBoll: This also raises the issue that `getId` and `getTextValue` are never mentioned in the documentation.

mannycarrera4: i can add that

modules/react/collection/lib/useListItemRegister.tsx
```diff
@@ -94,7 +94,7 @@ export const useListItemRegister = createElemPropsHook(useListModel)(
       disabled: elemProps.disabled || state.nonInteractiveIds.includes(localId) ? true : undefined,
       item: null, // remove `item` from prop list
       virtual: null, // remove `virtual` from prop list
-      'aria-setsize': elemProps.virtual?.size,
+      'aria-setsize': elemProps.virtual ? elemProps.virtual.size : state.items.length,
```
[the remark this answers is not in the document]
NicholasBoll: `aria-setsize` should not be present if it is not virtualized. What was here before was correct.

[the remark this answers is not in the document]
NicholasBoll: Ooops. This is a bug. #2362

modules/docs/mdx/10.0-UPGRADE-GUIDE.mdx
```diff
@@ -102,13 +113,18 @@ Example using `space` with calc instead of deprecated `spaceNumbers`.
...
-For more information on how to handle this migration, you can reference our [discussion](https://github.com/Workday/canvas-kit/discussions/2343).
+For more information on how to handle this migration, you can reference our
+[discussion](https://github.com/Workday/canvas-kit/discussions/2343).
 
 ### Table
 
 **PR:** [#2344](https://github.com/Workday/canvas-kit/pull/2344)
 
-We've deprecated `Table` and `TableRow` as well as all of their exported members. Please use [`Table`](https://workday.github.io/canvas-kit/?path=/docs/preview-table--basic) in [Preview](#preview) instead.
+as all of their exported members. Please use
```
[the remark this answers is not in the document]
mannycarrera4: yeah bad manny bad merge

modules/react/collection/stories/examples/DynamicItemsObjects.tsx
```diff
@@ -0,0 +1,25 @@
...
+  {id: 'Berlin (Europe)'},
+  {id: 'Boston (United States)'},
+  {id: 'Boulder (United States)'},
+  {id: 'Chicago (United States)'},
+];
+
+export const DynamicItemsObjects = () => (
+  <ListBox items={items}>
+    {item => (
+      <ListBox.Item aria-disabled={item.disabled ? item.disabled : undefined}>
```
RayRedGoose, line 20: why do you do like this: `item.disabled ? item.disabled : undefined`? isn't `disabled already undefined if it's not listed?

modules/react/combobox/lib/hooks/useComboboxInput.ts
```diff
@@ -0,0 +1,94 @@
...
+          // We don't want to submit forms while the combobox is open
+          event.preventDefault();
+        }
+      },
+      onBlur(event: React.FocusEvent) {
+        if (model.state.nonInteractiveIds.length > 0) {
+          return;
+        } else {
+          model.events.hide(event);
+        }
```
RayRedGoose, line 70:
```suggestion
        if (model.state.nonInteractiveIds.length < 1) {
          model.events.hide(event);
        }
```

modules/react/combobox/lib/hooks/useKeyboardTypeAhead.ts
```diff
@@ -0,0 +1,132 @@
...
+      const label = model.state.items[i].textValue.toLowerCase();
+      if (label.indexOf(startString.toLowerCase()) === 0) {
+        if (
+          !ignoreDisabled ||
+          (ignoreDisabled && !model.state.nonInteractiveIds.includes(model.state.items[i].id))
+        ) {
+          return i;
+        }
+      }
+    }
```
RayRedGoose, line 57:
```suggestion
      if (
        label.startsWith(startString.toLowerCase()) &&
        (!ignoreDisabled || !model.state.nonInteractiveIds.includes(model.state.items[i].id))
      ) {
        return i;
      }
```
`startsWith` will check if label starts from startString (i guess it is what you check)
`ignoreDisabled` is boolean, so there is only two variations. You already check one of this, so you don't need to check again here:
`ignoreDisabled && !model.state.nonInteractiveIds.includes(model.state.items[i].id))`

modules/react/combobox/lib/hooks/useKeyboardTypeAhead.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    // Keeps track of the current key types and adds to it
+    // if you type `de` vs `d` for denver
+    keySofar.current += key;
+    setKeyTypedSofar(keySofar.current);
+    startClearKeysSoFarTimer();
+
+    // First, look for a match from start to end
+    let matchIndex;
+    matchIndex = getIndexByStartString(start, keySofar.current);
```
RayRedGoose, line 79: why do not assign at the same time with initialization?

```js
let matchIndex = getIndexByStartString(start, keySofar.current);
```

modules/react/combobox/lib/hooks/useKeyboardTypeAhead.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    // If a match isn't found between start and end, wrap the search
+    // around and search again from the beginning (0) to start
+    if (matchIndex === -1) {
+      matchIndex = getIndexByStartString(0, keySofar.current, start);
+    }
```
RayRedGoose, line 85: So after this step `matchIndex` will be always more than `-1` or there is any possibility `matchIndex` will be `0`?

modules/react/combobox/lib/hooks/useKeyboardTypeAhead.ts
```diff
@@ -0,0 +1,132 @@
...
+        // go to that item and select based on its id
+        model.events.goTo({id: model.state.items[matchIndex].id});
+        model.events.select({id: model.state.items[matchIndex].id});
+      } else {
+        // Otherwise the menu is visible
+        // move the cursor to that matched item
+        model.events.goTo({id: model.state.items[matchIndex].id});
+      }
+    }
+  };
```
RayRedGoose, line 100:
```suggestion
    if (matchIndex > -1) {
      model.events.goTo({id: model.state.items[matchIndex].id});
      if (model.state.visibility === 'hidden') {
        // If the menu is closed, fire the change event
        // go to that item and select based on its id
        model.events.select({id: model.state.items[matchIndex].id});
      }
    }
  };
```

will this work?

modules/react/combobox/lib/hooks/useMoveCursorToIndex.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * If there is nothing selected, move cursor to the first item, otherwise, move the cursor to the selected item
+ */
+export const useMoveCursorToIndex = createElemPropsHook(useComboboxModel)(model => {
+  const visible = model.state.visibility !== 'hidden';
+  React.useLayoutEffect(() => {
+    const cursorItemId = model.state.items.find(item => item.id === model.state.cursorId)?.id;
+    // If there is no selected item and items exists we want to set the cursor to the first item in the array
+    if (model.state.selectedIds.length === 0 && model.state.items.length > 0 && !cursorItemId) {
+      model.events.goTo({id: model.state.items[0].id});
+    } else if (model.state.selectedIds.length > 0 && !cursorItemId) {
```
RayRedGoose, line 15:
```suggestion
    } else if (model.state.selectedIds.length && !cursorItemId) {
```

modules/react/combobox/lib/hooks/useMoveCursorToIndex.tsx
```diff
@@ -0,0 +1,36 @@
...
+    model.state.selectedIds,
+    model.events,
+    model.state.items,
+    model.state.cursorId,
+    model.state.visibility,
+  ]);
+
+  React.useEffect(() => {
+    // when closed, the cursor should reset to the selected item if something is selected
+    if (!visible && model.state.selectedIds.length !== 0) {
```
RayRedGoose, line 29:
```suggestion
    if (!visible && model.state.selectedIds.length) {
```

modules/react/combobox/lib/hooks/useMoveCursorToIndex.tsx
```diff
@@ -0,0 +1,36 @@
...
+      console.log();
```
RayRedGoose, line 30:
```suggestion
```

modules/react/select/lib/Select.tsx
```diff
@@ -1,129 +1,192 @@
...
-      width: '100%',
-    }
+export const SelectInput = createSubcomponent(TextInput)({
+  modelHook: useSelectModel,
+  elemPropsHook: useSelectInput,
+})<SelectInputProps>(
+  ({placeholder = 'Choose an option', inputStartIcon, ...elemProps}, Element, model) => {
+    return (
+      <InputGroup>
+        {inputStartIcon && model.state.selectedIds.length > 0 && (
```
RayRedGoose, line 32:
```suggestion
        {inputStartIcon && model.state.selectedIds.length && (
```

modules/react/select/stories/stories_VisualTesting.tsx
```diff
@@ -0,0 +1,150 @@
...
+              {label: '', value: false},
+              {label: 'Disabled', value: true},
+            ],
+          },
+          props => {
+            if (props.disabled && !['', 'hover'].includes(props.className)) {
+              return false;
+            }
+            return true;
+          }
```
RayRedGoose, line 65:
```suggestion
          props => {
            return !props.disabled || !props.className || props.className === 'hover';
          }
```
This conditional will always return true or false, false will be for all items that disabled with class name hover

modules/react/select/stories/stories_VisualTesting.tsx
```diff
@@ -0,0 +1,150 @@
...
+        ]}
+        columnProps={[{label: 'Default', props: {}}]}
+      >
+        {props => (
+          <FormField label="Contact" style={{marginBottom: '250px'}}>
+            <Select model={model}>
+              <Select.Input {...props} id="contact-select" />
+              <Select.Popper>
+                <Select.Card maxHeight="200px">
+                  {model.state.items.length > 0 && (
```
RayRedGoose, line 117:
```suggestion
                  {model.state.items.length && (
```

modules/react/collection/stories/examples/DynamicItemsObjects.tsx
mannycarrera4: Updated, ty

modules/react/combobox/lib/hooks/useKeyboardTypeAhead.ts
mannycarrera4: it's possible it's 0

modules/react/combobox/lib/hooks/useMoveCursorToIndex.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * If there is nothing selected, move cursor to the first item, otherwise, move the cursor to the selected item
+ */
+export const useMoveCursorToIndex = createElemPropsHook(useComboboxModel)(model => {
+  const visible = model.state.visibility !== 'hidden';
+  React.useLayoutEffect(() => {
+    const cursorItemId = model.state.items.find(item => item.id === model.state.cursorId)?.id;
+    // If there is no selected item and items exists we want to set the cursor to the first item in the array
+    if (model.state.selectedIds.length === 0 && model.state.items.length > 0 && !cursorItemId) {
```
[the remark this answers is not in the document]
mannycarrera4: updated, ty!

modules/react/combobox/lib/hooks/useComboboxInput.ts
NicholasBoll: You'd need to do `<=` to maintain the same functionality

modules/react/combobox/lib/hooks/useKeyboardTypeAhead.ts
mannycarrera4: i updated this with starts with, but the rest of the logic didn't work

mannycarrera4:
```if (
        label.startsWith(startString.toLowerCase()) &&
        (!ignoreDisabled ||
          (ignoreDisabled && !model.state.nonInteractiveIds.includes(model.state.items[i].id)))
      ) {
        return i;
      }
      ```

Comment (NicholasBoll) [reply modules/react/combobox/lib/hooks/useKeyboardTypeAhead.ts:57]: @RayRedGoose I think you missed the second conditional. In English "if label starts with `startString` AND either `ignoreDisabled` is `false` OR `ignoreDisabled` is `true` AND the current item is interactable". Your logic is missing the "`ignoreDisabled` is `true`" part.

Here's logic that's the same:
```suggestion
      if (
        label.startsWith(startString.toLowerCase()) &&
        (!ignoreDisabled ||
          (ignoreDisabled && !model.state.nonInteractiveIds.includes(model.state.items[i].id)))
      ) {
        return i;
      }
```

modules/react/combobox/lib/hooks/useKeyboardTypeAhead.ts
mannycarrera4: yep good catch!

modules/react/combobox/lib/hooks/useKeyboardTypeAhead.ts
mannycarrera4: if it's the first item

NicholasBoll: I think @RayRedGoose means to say "is there any possibility `matchIndex` will be `-1`"? A `0` means match found and it is the first item. This logic forces a wrap around from the current cursor, but it is possible a match is still not found.

NicholasBoll (approved):

Issue #2253, opened 2023-06-13
Update our Component MDX Template


jamesfan:
## 💡 Suggestion

Update our component MDX template documented in https://github.com/Workday/canvas-kit/discussions/1131 to better reflect the current state of our component documentation and provide a more consistent experience for both the document author and reader.

## Motivation

To maintain consistency in our component documentation, we document our component MDX structure in #1131.

This template has grown out of date since it was last updated in 2021 (particularly with the introduction of our custom docgen parsing in #1962) and should be updated to reflect the current state of our component documentation.

Also, the current template includes three separate use cases for `Simple Component`, `Component Family with Multiple Components`, and `Compound Component`. We may be able to consolidate/simplify this.

This work does **not** include updating all component MDXes to meet the updated template, although it may include updating one or two component MDXes to validate the updated template (Button is a strong candidate).

PR #2310, opened 2023-08-03
docs: Add documentation guidelines

jamesfan:
## Summary

Fixes: #2253 

## Release Category
Documentation

---

## Checklist

- [x] Update `PULL_REQUEST_TEMPLATE.md` to reference the new `DOCUMENTATION_GUIDELINES.mdx` file instead of the [old discussion](https://github.com/Workday/canvas-kit/discussions/1131)

## For the Reviewer


modules/docs/mdx/DOCUMENTATION_GUIDELINES.mdx
```diff
@@ -0,0 +1,238 @@
...
+<!--
+Import components which will be used to render documentation in the MDX file.
+The Specifications import is optional; only include it if the Specifications
+section exists (see below).
+-->
+import {Specifications, SymbolDoc} from '@workday/canvas-kit-docs';
+
+<!-- TODO: Figure out if we need to import the component -->
+<!-- Import component -->
+import {[Component]} from '@workday/canvas-kit-[preview-|labs-]react/[component]';
```
jamesfan, line 51: As far as I can tell, the imported component is only used in the `<Meta>` Storybook tag a few lines down and even there it doesn't seem to be required (SB renders fine without it). Button.stories.mdx doesn't even include the `component` prop in its [Meta](https://github.com/Workday/canvas-kit/blob/master/modules/react/button/stories/button/Button.stories.mdx#L19) tag -- I'm guessing we left it out because there are multiple Button components, but maybe it's not required in the first place?

modules/react/button/stories/button/Button.stories.mdx
```diff
@@ -47,50 +46,38 @@ when you need to place a Primary Button on a dark or colorful background such as
 
 <ExampleCodeBlock code={PrimaryInverse} />
 
-#### Component API
-
-<SymbolDoc name="PrimaryButton" fileName="/react/" hideHeading hideDescription headingStart={4} />
```
jamesfan, line 52: Note the difference in the Button docs. Instead of displaying the Component API section in the Usage section underneath each component, we now display a single Component API section for all the buttons.

My reasoning for this:

1. Button is an exception (see the [Component Families](https://github.com/Workday/canvas-kit/pull/2310/files#diff-5eed4904a92938d88cb229dd58f9bc3e48fea2a1aab1b7e0944261b680e5959eR134) section in the Documentation Guidelines). Most of our component docs only document a single top-level component whereas Button documents multiple top-level components. This updated structure brings Button more in line with the structure of our other component docs.
2. Although the props tables for each button are no longer colocated with the corresponding example for that button, I think the overall Button doc reads more clearly now that props tables no longer interrupt the flow between examples for each button.

mannycarrera4: I still think since buttons is the only family mdx file we have, we should just break it out into each mdx file. We do this for text component. If we had a unified `<Button/>` component then that would make sense. But since this is the only outlier I say we break them out into their own mdx files. That would be however, that maybe on the side we break it out to `primaryButton` `secondary` pages? I see the con of this is as well. Just posting my thought here but not a blocker for this PR.

modules/docs/mdx/DOCUMENTATION_GUIDELINES.mdx
```diff
@@ -0,0 +1,238 @@
...
+  }}
+/>
+
+# Documentation Guidelines
+
+## Introduction
+
+### Writing Style
+
+Google provides [helpful resources](https://developers.google.com/tech-writing/overview) for how to
```
mannycarrera4, line 17: wohoooo

modules/docs/mdx/DOCUMENTATION_GUIDELINES.mdx
```diff
@@ -0,0 +1,238 @@
...
+
+<!-- TODO: Figure out if we need to import the component -->
+<!-- Import component -->
+import {[Component]} from '@workday/canvas-kit-[preview-|labs-]react/[component]';
+
+<!-- Import component examples -->
+import {Basic} from './examples/Basic'; import {AnotherExample} from './examples/AnotherExample';
+import ...
+
+<!-- TODO: Figure out if we need the component prop for the Meta tag -->
```
mannycarrera4, line 57: remove before merging

modules/docs/mdx/DOCUMENTATION_GUIDELINES.mdx
```diff
@@ -0,0 +1,238 @@
...
+The Specifications section is optional. Only include it if the component has
+Cypress integration tests (i.e., a spec.ts file in the cypress/integration
+folder).
+-->
+## Specifications
+
+<Specifications file="[Component].spec.ts" name="[Component]" />
+````
+
+> **Note:** Refer to [Tabs](/components/containers/tabs/) and [Modal](/components/popups/modal/) for
```
mannycarrera4, line 131: yesss i was about to suggest this, glad we have a gold standard for reference!

mannycarrera4 (approved):
non blocker comments!

josh-bagwell (approved):

modules/docs/mdx/DOCUMENTATION_GUIDELINES.mdx
jamesfan: Never mind, the `component` prop is necessary in order to get the component to render in the Storybook navigation. For example, `<Meta title="Components/Containers/Tabs" />` (without the `component={Tabs}` prop) won't render the `Tabs` section under `Components/Containers`.

modules/react/button/stories/button/Button.stories.mdx
jamesfan: After discussion with the rest of the team, it sounds like we may break the Button components into their own MDXes in a future PR.

modules/react/button/stories/button/Button.stories.mdx
```diff
@@ -1,12 +1,5 @@
 import {Specifications, SymbolDoc, SymbolDescription} from '@workday/canvas-kit-docs';
 
-import {
-  PrimaryButton,
-  SecondaryButton,
-  TertiaryButton,
-  DeleteButton,
-} from '@workday/canvas-kit-react/button';
```
jamesfan, line 8: These component imports are not used anywhere in the Buttons MDX (since they're not being used by the `Meta` SB component).

PR #2364, opened 2023-10-20
fix: Force PrimaryButton text color in all states

anicholls:
## Summary

The `PrimaryButton` styling works great when there are no conflicting styles. However, the state pseudo selectors take an unopinionated approach on the text color of the button, which makes it easily overridden by globally scoped css (particularly when using `as="a"`). 

In platforms like [Docusaurus](https://docusaurus.io/), it's common to have styling for `a:hover`. Since `PrimaryButton` does not have any `color` set for `:hover` (or other states), this gets overridden by a simple global specifier, resulting in something that looks like this:

image

While I could add another style for `a[type="button"]:hover { color: #fff }`, this would cause other problems because we have no way of determining via DOM a `PrimaryButton` from a `SecondaryButton`. If this rule were applied to a SecondaryButton, the text would be invisible on a white background.

The simplest approach here IMO was to take a slightly more assertive approach in specifying the color for all states. I don't believe this should have any negative effects, and I don't think it should be a breaking change.

I believe buttons have been largely re-written in v10 👑 , so this may need to be ported to the prerelease branch as well.

Thanks for the consideration!

## Release Category
Components

---

## Checklist

## For the Reviewer

### DIFF
--- a/modules/react/button/lib/PrimaryButton.tsx
+++ b/modules/react/button/lib/PrimaryButton.tsx
@@ -31,15 +31,19 @@ const getPrimaryButtonColors = (variant: 'inverse' | undefined, theme: EmotionCa
         },
         hover: {
           background: themePrimary.dark,
+          label: themePrimary.contrast,
         },
         active: {
           background: themePrimary.darkest,
+          label: themePrimary.contrast,
         },
         focus: {
           background: themePrimary.main,
+          label: themePrimary.contrast,
         },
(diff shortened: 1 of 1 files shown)


NicholasBoll (approved):

NicholasBoll:
v10 is explicit about label color in `PrimaryButton`, so this fix is only required for v9.

PR #2379, opened 2023-10-30
fix: Fix Style prop merging with cs prop

NicholasBoll:
## Summary

Fixes and issue where using the `cs` prop does not override styles when using style props. Many of our components still use style props, so this fix allows people to use the `cs` prop themselves to override our components.

## Release Category
Components

### Release Note
This fix adds Emotion's `CacheProvider` to the `CanvasProvider`. Any application that uses the `CacheProvider` will not see this fix within the render tree of the custom `CacheProvider`.  For this fix to be applied everywhere, consider removing any use of Emotion's `CacheProvider`.

Some instances of use of `CacheProvider` were to set the `compat` mode of the cache. You can now do the following instead and not use `CacheProvider` at all:

```ts
cache.compat = true
```

Also note this fix will break automatic server side rendering because style merging is not creating server-side only `style` tags. Since `createStyles` isn't compatible with automatic server side rendering, the merge style fix isn't compatible either. Use Emotion's solution for server-side `@emotion/css`: https://emotion.sh/docs/ssr#when-using-emotioncss

The only modification is to import `cache` from `@emotion/css` instead of creating a new cache. Their documentation doesn't work:

```patch
- import createCache from '@emotion/cache'
- 
- const key = 'custom'
- const cache = createCache({ key })
+ import {cache} from '@emotion/css';
+ const {key} = cache
```
---

## Checklist

## For the Reviewer

Start with the new example and the Styling API documentation.

- [x] Code
- [x] Documentation
- [x] Testing


modules/react/layout/lib/StylingBox.tsx
```diff
@@ -0,0 +1,180 @@
...
+  'wordBreak',
+];
+
+const shouldForwardProp = (prop: string) => {
+  return isPropValid(prop) && !omittedProps.includes(prop);
+};
+
+const stylingBoxStyles = createStyles({
+  boxSizing: 'border-box',
+});
```
mannycarrera4, line 64: This is the tidbit you mentioned, using createStyles so that things can be overridden
 yeah?

mannycarrera4 (approved):

modules/react/button/lib/BaseButton.tsx
```diff
@@ -314,40 +314,36 @@ export const BaseButton = createComponent('button')({
   Component: (
     {
       children,
-      cs,
       size,
       fillIcon,
       iconPosition,
       icon,
+      colors,
```
mannycarrera4, line 321: I think we removed this for this reason: https://github.com/Workday/canvas-kit/pull/2381

mannycarrera4: Unless mergeStyles handles this

modules/react/button/lib/TertiaryButton.tsx
```diff
@@ -173,16 +173,14 @@ export const TertiaryButton = createComponent('button')({
         icon={icon}
         iconPosition={iconPosition}
         size={size}
-        cs={[
+        {...mergeStyles(elemProps, [
```
mannycarrera4, line 176: curious in secondary we remove the props iconPositon and size and just pass it as elemProps, do we wanna do the same here?

NicholasBoll: No, because `iconPosition` and `size` are used inside this render function. `size` has a default being set in the deconstructed props and `iconPosition` is used in `baseIconPosition` on line 161.

modules/react/button/lib/BaseButton.tsx
NicholasBoll: I have no idea why `colors` was removed in #2381. `colors` should be removed from the prop list in `BaseButton` because it isn't a prop that `Box` knew how to handle. It happened to be removed by `Box` because when `Element` was a string because `Box` uses `@emotion/is-prop-valid` and `colors` isn't a valid prop. But if `Element` is not a string, but a component, the `colors` prop would be forwarded to that component which is bad.

My suggestion for the fix was only to delete the line from `Button.tsx` that removed `colors` from `elemProps` passed to `BaseButton`. `colors` should never have been removed from the deconstruction list in `BaseButton`. I'm simply fixing that bug here because I'm no longer using `Box` and no longer using `@emotion/is-prop-valid`

Issue #2352, opened 2023-10-10
Modal covered by bottom address bar on iPhone


mike-kerley:
## 🐛 Bug Report

The `Modal` component is partially covered by the bottom address bar on mobile browsers, making it impossible to click/tap on action buttons at the bottom.

## To Reproduce

Create a `<Modal>` like the one in https://workday.github.io/canvas-kit/?path=/docs/components-popups-modal--basic. View it from a mobile simulator/emulator.

## Expected Behavior

Modal is fully visible and buttons can be clicked/tapped

## Actual Results

Modal is partially covered as shown below

## Browser (if applicable)

* iPhone Safari (iOS 17)
* Default Samsung mobile browser, using latest android on Samsung S23

## Link to repl or repo (highly encouraged)

https://github.com/mike-kerley/canvas-kit-modal-demo

### Error Output

<image src="https://github.com/Workday/canvas-kit/assets/58498830/469f6f07-afd0-41b9-9923-ef3683f2c40b" width="325">

mike-kerley:
I believe this is caused by the use of `100vh` and `alignItems: 'end'` in `ModalOverlay`:
https://github.com/Workday/canvas-kit/blob/2c0010f61e0f2dbeeb1436ac2b79eb806d8bb91c/modules/react/modal/lib/ModalOverlay.tsx#L58-L61

Safari (and apparently Samsung's browser) include the toolbar area in their viewport height calculation.

mannycarrera4:
When testing, make sure to open storybook in a new iframe and then set the viewport, otherwise storybook knows to shrink the viewport and the bug isn't reproducible.

PR #2385, opened 2023-11-08
fix: Prevent modal from overlapping on mobile devices

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/2352

## Release Category
Components

---

### DIFF
--- a/modules/react/modal/lib/ModalOverlay.tsx
+++ b/modules/react/modal/lib/ModalOverlay.tsx
@@ -26,20 +26,30 @@ const fadeIn = keyframes`
   }
 `;
 
-const Container = styled(Box)<StyledType>({
-  position: 'fixed',
-  top: 0,
-  left: 0,
-  width: '100vw',
-  height: '100vh',
-  background: 'rgba(0,0,0,0.65)',
-  animationName: `${fadeIn}`,
-  animationDuration: '0.3s',
-  // Allow overriding of animation in special cases
-  '.wd-no-animation &': {
-    animation: 'none',
+const Container = styled(Box)<StyledType>(
+  {
+    position: 'fixed',
+    top: 0,
+    left: 0,
+    width: '100vw',
+    height: '100vh',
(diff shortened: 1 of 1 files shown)


mannycarrera4:
## Before Fix
image


## After Fix
image

modules/react/modal/lib/ModalOverlay.tsx
```diff
@@ -26,20 +26,30 @@ const fadeIn = keyframes`
...
+    '.wd-no-animation &': {
+      animation: 'none',
+    },
   },
-});
+  ({theme}) => {
+    const {canvas: canvasTheme} = getTheme(theme);
+    return {
+      [canvasTheme.breakpoints.down('s')]: {
+        height: '100%',
```
mannycarrera4, line 48: This solution comes from this [article](https://dev-tips.com/css/overlapping-bottom-navigation-bar-despite-100vh-in-ios-safari)

modules/react/modal/lib/ModalOverlay.tsx
```diff
@@ -48,14 +58,14 @@ const Container = styled(Box)<StyledType>({
...
-    maxHeight: '100vh',
+    maxHeight: '100%',
     display: 'flex',
     position: 'absolute',
     left: 0,
     top: 0,
     justifyContent: 'center',
     alignItems: 'center',
-    height: '100vh',
+    height: '100%',
```
mannycarrera4, line 68: Not sure if this might break other applications

josh-bagwell: The time it would be a problem is if the parent container of the overlay isn't 100% of the height of the viewable area since `100%` sets the height of that element to `100%` of the height of it's parent element.

josh-bagwell (approved):

josh-bagwell:
Looked good to me on Simulator when tested. 

Just need to approve the visual tests.

PR #2408, opened 2023-11-16
fix: Update cs prop to create a class instead of style

NicholasBoll:
## Summary

Adding static styles to the `cs` prop creates a `style` attribute. This fix updates `cs` prop handling to create a CSS class instead.

The following is now supported officially:

```tsx
cs={{ padding: 10 }}
```

Instead of creating `style={{padding:10}}`, it will now invoke `@emotion/css` and will return a className.

## Release Category
Components

---

## Checklist

## For the Reviewer

`cs.spec.ts`

- [ ] Code
- [ ] Testing

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -3,7 +3,6 @@ module.exports = {
     test: {
       presets: ['@babel/preset-env', '@babel/preset-typescript', ['@babel/preset-react']],
       plugins: [
-        '@emotion',
         '@babel/plugin-transform-runtime',
         ['@babel/proposal-class-properties', {loose: true}],
       ], // https://github.com/storybookjs/storybook/issues/14805 and https://github.com/storybookjs/storybook/issues/14805#issuecomment-889504884 ],
--- a/modules/styling/lib/cs.ts
+++ b/modules/styling/lib/cs.ts
@@ -268,7 +268,7 @@ export function csToProps(input: CSToPropsInput): CsToPropsReturn {
     const staticStyles: CSSObject = {};
     const cssVars: Record<string, string> = {};
     for (const key in input) {
-      if (key.charAt(0) === '-') {
+      if (key.startsWith('--')) {
         cssVars[key] = (input as any)[key];
       } else {
         (staticStyles as any)[key] = (input as any)[key];
--- a/modules/styling/lib/cs.ts
+++ b/modules/styling/lib/cs.ts
@@ -260,6 +260,26 @@ export function csToProps(input: CSToPropsInput): CsToPropsReturn {
   // At this point, `input` is either an array or a `Record<T, string>`. If it isn't an array, it
   // must be a style object for setting CSS variables. So set the `style` attribute
   if (!Array.isArray(input)) {
+    // The input is an object. We'll separate variables from static styles. If static styles are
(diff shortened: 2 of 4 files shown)


babel.config.js
```diff
@@ -3,7 +3,6 @@ module.exports = {
     test: {
       presets: ['@babel/preset-env', '@babel/preset-typescript', ['@babel/preset-react']],
       plugins: [
-        '@emotion',
```
mannycarrera4, line 6: no need?

mannycarrera4:
Only feedback I have is to add to https://workday.github.io/canvas-kit/?path=/story/features-styling-api--styling-overrides an example of doing this but NB

package.json
```diff
@@ -14,7 +14,6 @@
     "@babel/preset-typescript": "^7.1.0",
     "@commitlint/cli": "^8.3.5",
     "@commitlint/config-conventional": "^8.1.0",
-    "@emotion/babel-plugin": "^11.9.2",
```
mannycarrera4, line 17: shouldn't the yarn lock file be updated?

babel.config.js
NicholasBoll: It breaks the spec file if this is included. We haven't needed it for a while, but it didn't really matter until I try to compare style hashes created in different files.

package.json
NicholasBoll: The babel plugin is also a transitive dependency, so the lockfile isn't updated in this case

modules/styling/lib/cs.ts
```diff
@@ -260,6 +260,26 @@ export function csToProps(input: CSToPropsInput): CsToPropsReturn {
...
   // must be a style object for setting CSS variables. So set the `style` attribute
   if (!Array.isArray(input)) {
+    // The input is an object. We'll separate variables from static styles. If static styles are
+    // detected here, a dynamic mode should be detected and done for development. The transform
+    // should take care of extracting static styles out of a render function and this function
+    // should not be run.
+    let hasStaticStyles = false;
+    const staticStyles: CSSObject = {};
+    const cssVars: Record<string, string> = {};
+    for (const key in input) {
```
RayRedGoose, line 270: Isn't safer to check if key starts with '--' for css vars instead of just '-'? will something like `-webkit-transform` be recognized as css var? I don't know if there is any possibility of passing this though

NicholasBoll: Yes. Good catch

NicholasBoll: I think it is unlikely that someone is using a vendor prefix anymore, but a vendor prefix is still a valid style key. I've updated the check to look for `--`

NicholasBoll:
> Only feedback I have is to add to https://workday.github.io/canvas-kit/?path=/story/features-styling-api--styling-overrides an example of doing this but NB

StyleOverrides covers the merging order of different ways of styling which includes the `cs` prop. This way of styling still uses the `cs` prop. Merging the `cs` prop internally follows external rules. When the babel plugin is made to statically hoist out static styles, the plugin will need to ensure it is being hoisted to the right place so that runtime style merging is the same as static style merging. So the order it is declared matters. Without the plugin, it will merge in the order it appears on the `cs` prop array.

mannycarrera4 (approved):

PR #2427, opened 2023-11-21
fix: Check for undefined error in radio group

mannycarrera4:
## Summary

When `error` is undefined, the styled attribute was still being applied. This only adds the style attribute with the box shadow when there's an error.

## Release Category
Components

---

### DIFF
--- a/cypress/integration/Autocomplete.spec.ts
+++ b/cypress/integration/Autocomplete.spec.ts
@@ -184,7 +184,7 @@ describe('Autocomplete', () => {
                 cy.findByRole('combobox').should('have.value', 'Red Apple 12');
               });
 
-              it('should change the filtered results', () => {
+              it.skip('should change the filtered results', () => {
                 cy.findByRole('option', {name: 'Red Apple 121'}).should('be.visible');
               });
             });
@@ -332,7 +332,7 @@ describe('Autocomplete', () => {
           });
         });
       });
-      
+
       context('when the user types in a value not found', () => {
         beforeEach(() => {
           cy.findByRole('combobox').type('Peach');
--- a/modules/preview-react/radio/lib/RadioGroup.tsx
+++ b/modules/preview-react/radio/lib/RadioGroup.tsx
@@ -62,7 +62,6 @@ export const RadioGroup = createContainer(Flex)({
   },
 })<RadioGroupProps>(({children, error, theme, ...elemProps}, Element, model) => {
   const errorColors = getErrorColors(error, theme);
-  console.log(errorColors);
   return (
     <Flex
       as={Element}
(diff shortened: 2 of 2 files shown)


modules/preview-react/radio/lib/RadioGroup.tsx
```diff
@@ -62,6 +62,7 @@ export const RadioGroup = createContainer(Flex)({
   },
 })<RadioGroupProps>(({children, error, theme, ...elemProps}, Element, model) => {
   const errorColors = getErrorColors(error, theme);
+  console.log(errorColors);
```
vibdev, line 65: stray log

mannycarrera4: timber

vibdev (approved):

cypress/integration/Autocomplete.spec.ts
```diff
@@ -184,7 +184,7 @@ describe('Autocomplete', () => {
                 cy.findByRole('combobox').should('have.value', 'Red Apple 12');
               });
 
-              it('should change the filtered results', () => {
+              it.skip('should change the filtered results', () => {
```
josh-bagwell, line 187: is this another skip or since it's part of `support` you had to put this in?

mannycarrera4: yeah it was support, still failing, so I added, same one we did on master!

josh-bagwell (approved):

modules/preview-react/radio/lib/RadioGroup.tsx
```diff
@@ -76,12 +76,16 @@ export const RadioGroup = createContainer(Flex)({
...
       marginX={`-${space.xs}`}
       width="fit-content"
-      style={{
-        boxShadow:
-          errorColors.outer !== errorColors.inner
-            ? `inset 0 0 0 1px ${errorColors.outer}, inset 0 0 0 3px ${errorColors.inner}`
-            : `inset 0 0 0 2px ${errorColors.inner}`,
-      }}
+      style={
+        error !== undefined
```
RayRedGoose, line 80:
```suggestion
        !!error
```

What if somebody put `false` as a value for error?

RayRedGoose: or just `error` as you have ternary

mannycarrera4: I think error can only be `ErrorType` or undefined

RayRedGoose: so `ErrorType` is always true, so no need to check specific type

vibdev: ErrorType is an enum so it can equal `0` (when Error) or `1` (when Alert). 
TS enums are so misleading, ```!!ErrorType.Error === false```

PR #2449, opened 2023-12-07
feat: The one with Stencils

NicholasBoll:
## Summary

Resolves: #2464

Adds stencils to the `styling` package to help organize component styles.

Discussion: #2412

## Release Category
Components

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing

### DIFF
--- a/modules/styling/lib/cs.ts
+++ b/modules/styling/lib/cs.ts
@@ -147,14 +147,26 @@ export function createVars<T extends string, ID extends string>(input: {
   args: T[];
 }): CsVars<T, ID>;
 export function createVars<T extends string>(...args: T[]): CsVars<T, never>;
-export function createVars<T extends string, ID extends string>(...args: T[]): CsVars<T, ID> {
+export function createVars<
+  T extends Record<string, string> | Record<string, Record<string, string>>,
+  ID extends string = never
+>(input: T, id?: ID): DefaultedVars<T, ID>;
+export function createVars<
+  T extends string | Record<string, string> | Record<string, Record<string, string>>,
+  ID extends string
+>(...args: T[]): any {
+  const arg0 = args[0];
+  if (typeof arg0 === 'object' && !(arg0 as any).id) {
+    // this is a defaulted var
+    return createDefaultedVars(arg0 as any, args[1] as string);
+  }
   const id = (args[0] as any).id || generateUniqueId(); //?
   // eslint-disable-next-line no-param-reassign
   args = (args[0] as any).args || args;
-  const result = (input: Record<T, string>) => {
+  const result = (input: Record<ToString<T>, string>) => {
     const vars = {};
     args.forEach(key => {
-      if (input[key]) {
+      if (input[key as ToString<T>]) {
         // @ts-ignore TS complains about `key` not in object `{}`
         vars[result[key]] = input[key];
       }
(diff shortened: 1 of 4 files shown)


modules/styling/lib/cs.ts
```diff
@@ -168,7 +180,169 @@ export function createVars<T extends string, ID extends string>(...args: T[]): C
...
+//         // [myVars.default.color]: 'red',
+//         // ...myVars.$$defaults,
+//         color: myVars.default.color,
+//       }),
+//     });
+// }
+// <style>.css-123abc {
+//   --myVars-default-color: red;
+//   color: var(--myVars-default-color);
+//}</style>
```
alanbsmith, line 213: Do you still need this?

modules/styling/stories/Stencils.stories.mdx
```diff
@@ -0,0 +1,270 @@
+import {SymbolDoc} from '@workday/canvas-kit-docs';
+
+import {CreateStencil} from './examples/CreateStencil';
+
+<Meta title="Features/Styling/Stencils" />
+
+## Stencil
+
+Stencils are a reusable function that returns styles based on modifiers passed in. A Stencil should
```
mannycarrera4, line 9: does it return styles or a class name with your styles ?

modules/styling/lib/cs.ts
```diff
@@ -539,3 +719,198 @@ export function handleCsProp<
...
+};
+
+type VariableValuesStencil<
+  V extends Record<string, string> | Record<string, Record<string, string>>
+> = V extends Record<string, string>
+  ? {[K in keyof V]?: V[K]}
+  : {[K1 in keyof V]?: {[K2 in keyof V[K1]]: string}};
+
+/**
+ * Creates a reuseable Stencil for styling elements. It takes base styles, modifiers, and compound
```
mannycarrera4, line 852:
```suggestion
 * Creates a reuseable Stencil for styling elements. It takes vars, base styles, modifiers, and compound
```

modules/styling/stories/Stencils.stories.mdx
NicholasBoll: It returns `{ style: CSSObject, className: string}`

alanbsmith (approved):

modules/styling/stories/Stencils.stories.mdx
```diff
@@ -0,0 +1,270 @@
...
+
+```html
+<style>
+  .css-abc123 {
+    --abc123-defaultColor: red;
+    color: var(--abc123-defaultColor);
+  }
+</style>
+<div class="css-123abc" style="--abc123-defaultColor: blue;"></div>
+```
```
mannycarrera4, line 59: are you showing how `red` get override here?. The `this will produce the following html` is for the above code block, but it looks like below you're showing that applying a different value to the variable in the style attribute overrides it?

modules/styling/stories/Stencils.stories.mdx
```diff
@@ -0,0 +1,270 @@
...
+}
+
+.css-abc123 {
+  --abc123-defaultColor: red;
+  color: var(--abc123-defaultColor); // blue
+}
+```
+
+Variables are automatically added to the config of a Stencil. They share the same namespace as
+modifiers, so do not have a modifier with the same name as a variable.
```
mannycarrera4, line 77:
```suggestion
modifiers, so **do not have a modifier with the same name as a variable**.
```

modules/styling/stories/Stencils.stories.mdx
```diff
@@ -0,0 +1,270 @@
...
+
+**Note:** Variables should be used sparingly. Style properties can be easily overridden without
+variables. Variables are useful if you want to expose changing properties regardless of selectors.
+For example, Buttons use a variables for colors of all states (hover, active, focus, disabled, and
```
mannycarrera4, line 81:
```suggestion
For example, Buttons use variables for colors of all states (hover, active, focus, disabled, and
```

modules/styling/stories/Stencils.stories.mdx
```diff
@@ -0,0 +1,270 @@
...
+Modifiers are modifications to base styles. It should be used to change the appearance of a base
+style. For example, a button may have a modifier for "primary" or "secondary" which may change the
+visual emphasis of the button. Each modifier has its own CSS class name and the stencil will return
+the correct CSS classes to apply to an element based on what modifiers are active.
+
+```tsx
+const buttonStencil = createStencil({
+  base: {
+    padding: 5
+    // base styles
```
mannycarrera4, line 96:
```suggestion
     background: 'white',
    // base styles
```
In the above codeblock you mention that modifiers are used to change some base style, i think adding background here, shows how when a modifier is active, it will change the base style, in this case the background color

modules/styling/stories/Stencils.stories.mdx
```diff
@@ -0,0 +1,270 @@
...
+
+<div {...elemProps} />
+```
+
+The HTML may look something like this:
+
+```html
+<style>
+  .css-a0 {
+    padding: 5px;
```
mannycarrera4, line 123:
```suggestion
    padding: 5px;
    background-color: white;
```

modules/styling/stories/Stencils.stories.mdx
```diff
@@ -0,0 +1,270 @@
...
+[CSS specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity). If two matching
+selectors have the same specificity, the last defined property wins. Stencils take advantage of this
+by making all styles have the same specificity of `0-1-0` and inserting `base` styles, then
+`modifiers` (in order), then `compound` modifiers (in order). This means if a property was defined
+in `base`, a `modifier`, and a `compound` modifier, the `compound` modifier would win because it is
+the last defined. This should be the expected order.
+
+#### `handleCsProp`
+
+But what about when using components that use `@emotion/react` or `@emotion/styled`. Those libraries
```
mannycarrera4, line 240:
```suggestion
But what about when using components that use `@emotion/react` or `@emotion/styled`? Those libraries
```

modules/styling/stories/Stencils.stories.mdx
```diff
@@ -0,0 +1,270 @@
...
+</style>
+<div class="css-a0 css-a1"></div>
+```
+
+The optional `defaultModifiers` config property will default modifiers to a value. If a modifier is
+not passed to the stencil, the default will be used.
+
+```tsx
+myStencil(); // className will be `'css-a0 css-a2'`
+```
```
mannycarrera4, line 140: NB: We should prob add the function call on base & modifers when you need access to your variables

mannycarrera4:
Looks like `stencils` lives outside of the `API` folder, is that intended?

modules/styling/stories/Stencils.stories.mdx
NicholasBoll: If the modifier is optional, a base property value might make sense. If a modifier is not optional, there's no point in having a property that will always be overridden. It just takes more bytes.

mannycarrera4 (approved):

Issue #2452, opened 2023-12-07
`handleCsProp` was returning a class name when the style attribute was added to an element instead of the style object


mannycarrera4:
## 🐛 Bug Report

When a component was using the `style` attribute, `handleCSProp` was returning a `className` instead of merging in those styles. Also, `handleCSprop` was not merging the `style` object passed to a component and instead, returning an empty object.

PR #2453, opened 2023-12-07
fix: Fix forwarding of the style attribute

NicholasBoll:
## Summary

Fixes: #2452

The `style` attribute was being converted into a class name. The styles were being added and passing visual tests, but doing so in an unoptimized way. We use the `style` attribute for styles that change frequently to avoid expensive Style recalculations and we accidentally broke that.

## Release Category
Components

---

## Checklist

## For the Reviewer

- [x] Code
- [x] Testing

## Testing Manually

You could try adding a `style` attribute to any example. Without this change, the style attribute gets removed and the styles get merged into the element's main CSS class.

### DIFF
--- a/modules/styling/lib/cs.ts
+++ b/modules/styling/lib/cs.ts
@@ -1,8 +1,8 @@
-import React from 'react';
 // eslint-disable-next-line @emotion/no-vanilla
 import {cache, css} from '@emotion/css';
 import {getRegisteredStyles} from '@emotion/utils';
 import {serializeStyles, Keyframes, SerializedStyles, CSSObject} from '@emotion/serialize';
+import {Properties} from 'csstype';
 
 import {generateUniqueId} from './uniqueId';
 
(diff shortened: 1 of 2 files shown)


modules/styling/lib/cs.ts
```diff
@@ -1,8 +1,8 @@
-import React from 'react';
```
NicholasBoll, line 1: Remove the React dependency. It was unintentional.

modules/styling/lib/cs.ts
```diff
@@ -489,7 +489,7 @@ export function handleCsProp<
   // _does_ care because it uses the `classList` order to determine style merging. Therefore we
   // should always be careful to order class names knowing we need to support runtime merging if
   // runtime styles are detected.
-  const returnProps = csToProps([localCs, className, cs, style]);
+  const returnProps = csToProps([localCs, className, cs]);
```
NicholasBoll, line 492: Adding `style` to the `csToProps` was always weird. But before we added support for static inline styles, it worked. Afterwards, `csToProps` couldn't differentiate between a `style` prop given from `elemProps` or one given by a `createVars` function or a Stencil. So I removed `style` from the array and merge it below instead.

Issue #2463, opened 2023-12-12
Add style transformation support for stencils


NicholasBoll:
## 🚀 Feature Proposal

The style transform handles `createStyles`, but not `createStencils`. The style transform should be updated to support stencils as well.

## Motivation

If we merge the stencils PR and start using it, the style transform will not work on stencils. Stencils include variables which are used by other stencils or `createStyles`. This isn't urgent, but will be needed for complete style transform support.

PR #2489, opened 2023-12-21
feat: Add style transform support for stencils

NicholasBoll:
## Summary

Fixes: #2463 

* Refactors the style transform to make it easier to understand the parts.
* Add automated fallback support
* Removed unnecessary fallbacks
* Add style transform support for `createStencils`
* Add auto-wrapping for CSS variables in `createStyles`, `createStencils`, and the transformer.

## Release Category
Components

### Release Note

CSS variables can now be unwrapped anywhere in a property definition when using `createStyles` or `createStencil`. Previously, unwrapped variables could only be used if they started the property. For example, the following is now valid:

```ts
const myVar = '--my-var';

const myStyles= createStyles({
  boxShadow: `0 0 0 2px ${myVar}`
})
```

You no longer need to use `cssVar` around `myVar` to be valid. This should help with the transition from hard-coded variables (`const myVar = 'red' => const myVar = '--my-var'`) without needing to wrap in a `cssVar`. It allows us to think about `myVar` as a JS variable and the system will convert to a CSS variable for us.

---

## Checklist

## For the Reviewer

- [x] Code
- [x] Documentation
- [x] Testing

## Testing Manually

Tests should suffice, but you could always do a `(cd modules/react; yarn clean && yarn build:es6)` and verify the `dist` directory. Remember to do a `yarn clean` when you're done otherwise your environment is hosed.


modules/preview-react/form-field/lib/FormFieldLabel.tsx
```diff
@@ -25,7 +25,7 @@ const asteriskStyles = createStyles({
   fontSize: type.properties.fontSizes[20],
   fontWeight: type.properties.fontWeights.regular,
   textDecoration: 'unset',
-  color: cssVar(brand.error.base, '#de2e21'),
+  color: brand.error.base,
```
NicholasBoll, line 28: This type of change is in many places. Auto-fallback works and manually including a fallback is a maitenance issue.

mannycarrera4 (approved):
You write code, sometimes I agree with the code you write

modules/styling-transform/lib/utils/handleFocusRing.ts
```diff
@@ -0,0 +1,130 @@
...
+            defaults[property.name.text as keyof typeof defaults] = property.initializer;
+          }
+        });
+      }
+
+      const inset = parseNodeToStaticValue(defaults.inset, checker, prefix, vars);
+
+      let boxShadow: ts.TemplateExpression;
+      switch (inset) {
+        case 'outer':
```
josh-bagwell, line 53: Would `outer` be considered to be both focus rings on the "outside" of the container? I noticed an `inset` here or this something we need to talk about when re-factoring `focusRing`?

NicholasBoll: This is why I think `focusRing` will need an update. `outer` is actually both inner:

https://github.com/Workday/canvas-kit/blob/8817a0a5d258a0f1302ed563beeaa70b40883e5e/modules/react/common/lib/styles/focusRing.ts#L10-L13

I'm supporting `focusRing` in the static optimizer, not changing it.

Issue #2508, opened 2024-01-08
Font weight value as variable doesn't recognized in stencil by TS


RayRedGoose:
## 🐛 Bug Report

The below code produces TS error, because `font-weight` value as variable doesn't recognized by TS.

```
const labelTextStencil = createStencil({
  base: {
    ...system.type.subtext.large,
    color: base.blackPepper300,
  },
})
```

PR #2509, opened 2024-01-08
fix: Fix stencil types to accept a value as CSS variables

RayRedGoose:
## Summary

Fixes: #2508 

## Release Category
Infrastructure

### DIFF
--- a/modules/styling/lib/cs.ts
+++ b/modules/styling/lib/cs.ts
@@ -733,16 +733,16 @@ export function handleCsProp<
 
 type StylesReturn<V extends Record<string, string> | Record<string, Record<string, string>> = {}> =
   | SerializedStyles
-  | CSSObject
-  | ((vars: OptionalVars<V>) => SerializedStyles | CSSObject);
+  | CSSObjectWithVars
+  | ((vars: OptionalVars<V>) => SerializedStyles | CSSObjectWithVars);
 
 export type StencilModifierConfig<
   V extends Record<string, string> | Record<string, Record<string, string>> = {}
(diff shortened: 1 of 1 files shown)


modules/styling/lib/cs.ts
```diff
@@ -733,16 +733,16 @@ export function handleCsProp<
...
 
 export type StencilModifierConfig<
   V extends Record<string, string> | Record<string, Record<string, string>> = {}
 > = Record<string, Record<string, StylesReturn<V>>>;
 
 export type StencilCompoundConfig<M> = {
-  modifiers: {[K in keyof M]?: MaybeBoolean<keyof M[K]>};
-  styles: SerializedStyles | CSSObject;
+  modifiers: {[K in keyof M]?: keyof M[K]};
+  styles: SerializedStyles | CSSObjectWithVars;
```
NicholasBoll, line 745: Did you mean to remove the `MaybeBoolean` here?

mannycarrera4 (dismissed):

modules/styling/lib/cs.ts
RayRedGoose: Seems like my file hasn't updated 🤔 I pushed the fix

RayRedGoose: It should be only `CSSObject` replacements

josh-bagwell (approved):

modules/styling/lib/cs.ts
josh-bagwell: This looks updated and now has the `MaybeBeelean`

mannycarrera4 (approved):

Issue #2518, opened 2024-01-11
Keyframe causes build issue with style parser in v11


RayRedGoose:
## 🐛 Bug Report

Update to the new style transformer caused a build error for v11, because parser can't correctly recognize keyframe code

```
const grow = keyframes`
  from {
    transform: scale(0.85);
  }
  to {
    transform: scale(1.0);
  }
`;
```

```
Error: Unknown type at: "grow". Received "string"
File: /home/runner/work/canvas-kit/canvas-kit/modules/react/badge/lib/CountBadge.tsx:43:18.
    alignItems: 'center',
    animation: `${grow} 0.2s ease`,
    borderRadius: system.shape.round,
```

RayRedGoose:
Fixed by #2526

PR #2520, opened 2024-01-11
fix: Fix keyframe parsed with style transformer

RayRedGoose:
## Summary

Fixes: #2518

Replaced template syntax for keyframe by object syntax to better parser of style transformer on the build.

## Release Category
Components


RayRedGoose:
Actually it's not fixing an issue

PR #2549, opened 2024-02-01
fix: Only run focusRing transform inside static styles

NicholasBoll:
## Summary

#2489 updated style transform to support stencils, but also allowed other style utilities to be transformed outside a `createStyles` or `createStencil` call. Since `createStyles` and `createStencil` have strict requirements for statically analyzable values and other places do not, this can result in odd behavior. For example, calling the following is an incorrect transformation, but present in `Checkbox.tsx`:

```tsx
      ...focusRing({
        width: variant === 'inverse' ? 2 : 0,
        separation: 0,
        animate: false,
        innerColor: variant === 'inverse' ? colors.blackPepper400 : undefined,
        outerColor: variant === 'inverse' ? colors.frenchVanilla100 : undefined,
      }),
```

This will result in the following:
```tsx
   ... { boxShadow: variant === `0 0 0 ${2} ${variant === 'inverse' ? colors.blackPepper400 : undefined}, ...`
```

Since `focusRing` allows `undefined` as an argument with a default fallback, this works at runtime, but the transform doesn't understand the context of runtime. There's no strict check outside static style functions, so the fix is to only run this transform from inside a static style function.

## Release Category
Components

---

## Note

Fixing this change to use node transforms for nested children was causing issues with other TypeScript AST transformations. Transformations want to go from parent to child instead of the other way around. Trying to inverse this process caused random nodes to not have a `parent` or a source file which would cause archaic error messages deep inside TypeScript code. I've had to refactor transformations (again) to use the outside-in approach to AST transformations. This means style transformation is explicit. There are 2 types of custom style transformations:
  - Style object
  - Style value
I had to merge some of the work I was doing on `prerelease/major`. It wasn't worthwhile to untangle the work, so I'm moving it to `master`.

## Checklist

## For the Reviewer

Always start from the spec file


modules/styling-transform/spec/utils/handleFocusRing.spec.ts
```diff
@@ -7,68 +7,78 @@ describe('handleFocusRing', () => {
...
   });
 
+  it('should not rewrite focusRing if not called inside a correct static style function', () => {
+    const program = createProgramFromSource(`
+      const styles = styled({ ...focusRing() })
+    `);
+
+    const result = transform(program, 'test.ts', {transformers: [handleFocusRing]});
+
+    expect(result).toContain('styles = styled({ ...focusRing() })');
```
mannycarrera4, line 17: is not happy about this

mannycarrera4: looks like this test needs to needs to be updated? https://github.com/Workday/canvas-kit/blob/8651a7204522376431c7d25a7ffe7aecc5c8dac2/modules/styling-transform/spec/utils/handleCreateStyles.spec.ts#L218

NicholasBoll: There's something in the transform that's causing the parent to be `undefined` and failing the lookup test

mannycarrera4 (approved):
🤯

PR #2573, opened 2024-02-12
ci: Fix package.json fix syntax error [skip release]

NicholasBoll:
## Summary

Fix a syntax error in `forward-merge.js` that updates version strings in `package.json` files

## Release Category
Infrastructure
---

### DIFF
--- a/utils/forward-merge.js
+++ b/utils/forward-merge.js
@@ -92,7 +92,9 @@ async function main() {
   let hasConflicts = false;
   let hasUnresolvedConflicts = false;
   const {GITHUB_REF: currentBranch = defaultBranch} = process.env;
-  const [branch, nextBranch] = getBranches(currentBranch.replace('refs/heads/', ''));
+  const [branch, nextBranch] = getBranches(
+    currentBranch.replace('refs/heads/', '').replace('\n', '')
+  );
 
   // create a merge branch
   if (!alreadyMerging) {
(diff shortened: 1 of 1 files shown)


utils/forward-merge.js
```diff
@@ -285,7 +287,7 @@ function parseContents(lines) {
  * @param file {string}
  */
 async function resolveJsonFile(file) {
-  const contents = await fs.readFile(file).toString();
+  const contents = (await fs.readFile(file)).toString();
```
NicholasBoll, line 290: Without this fix, it will `toString` the promise instead of the result of the promise. The `package.json` files were getting replaced with `[object Promise]` and that won't do.

utils/forward-merge.js
```diff
@@ -92,7 +92,9 @@ async function main() {
   let hasConflicts = false;
   let hasUnresolvedConflicts = false;
   const {GITHUB_REF: currentBranch = defaultBranch} = process.env;
-  const [branch, nextBranch] = getBranches(currentBranch.replace('refs/heads/', ''));
+  const [branch, nextBranch] = getBranches(
+    currentBranch.replace('refs/heads/', '').replace('\n', '')
+  );
```
NicholasBoll, line 97: This change makes it possible to test locally. The CI doesn't seem to add a newline at the end of the branch while locally it does.

mannycarrera4 (approved):

Issue #2628, opened 2024-03-05
Allow a nonce to be set during application bootstrapping


NicholasBoll:
## 🐛 Bug Report

If a Content Security Policy has a [nonce requirement](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce), styles will fail to load from static styling. There is no way to add a `nonce` to styling either. Static styling takes over the cache of the page, preventing any styling from getting a `nonce` set. There should be a way to create an Emotion instance with a `nonce`.

Canvas Kit cannot bootstrap an application, but it can expose a way to create an instance during bootstrapping so that a nonce can be set.

For more information, see [the nonce discussion](https://github.com/Workday/canvas-kit/discussions/2625).

PR #2629, opened 2024-03-05
chore: Add nonce support to styling

NicholasBoll:
## Summary

Add a way to create an Emotion instance with initialization options, including a `nonce` or custom `stylis` plugins. Before, the Emotion instance created by `@emotion/css` was used. Now `@emotion/css` is used as a fallback if an instance hasn't been created by the time a style utility is called.

Fixes: #2628 

This change adds the following opt-in features:
- Change the default Emotion cache instance used by Canvas Kit and the application
- Allow use of `injectGlobal` from `@workday/canvas-kit-styling` that uses the shared instance.

## Release Category
Infrastructure

### Release Note
This change does not introduce any breaking changes, but creating a custom Emotion instance can introduce a breaking change. A custom instance should only be used if all instances of Canvas Kit on the page are above the version this change is released in and no application code is imported directly from `@emotion/css`. This change updates all internal Canvas Kit styling to use the Emotion instance created in `@workday/canvas-kit-styling`. If no custom instance is created, the one created by `@emotion/css` will be used. If the default instance is used, there should be no breaking changes, but everyone should update their application code to use styling functions from `@workday/canvas-kit-styling` and not `@emotion/css`. SSR using `@emotion/css` is unaffected since server to client hydration only cares about the cache key ("css") and the style's hash, which should be the same even with a custom cache instance.

---

## Checklist

## For the Reviewer

https://5d854c26ba934e0020f5e98a-yttfiqwybi.chromatic.com/?path=/docs/styling-basics--create-modifiers, scroll to bottom of the page.
Also the `cs.ts` file.

- [ ] Code
- [ ] Documentation

## Testing Manually

There is no new feature demonstrated. We don't actually create a custom Emotion instance in our Storybook and we
cannot create a `nonce` on Chromatic's servers.

### DIFF
--- a/modules/react/layout/lib/utils/mergeStyles.ts
+++ b/modules/react/layout/lib/utils/mergeStyles.ts
@@ -1,4 +1,3 @@
-import {css} from '@emotion/css';
 import {CSToPropsInput, handleCsProp} from '@workday/canvas-kit-styling';
 import {boxStyleFn} from '../Box';
 import {backgroundStyleFnConfigs} from './background';
@@ -71,20 +70,13 @@ export function mergeStyles<T extends {}>(
     return result;
   }, {});
 
-  // We need to determine if style props have been used. If they have, we need to merge all the CSS
-  // classes into a single class name in the order that the class names are listed. This variable
-  // will collect the CSS class name created by Emotion if we detect style props.
-  let stylePropsClassName = '';
+  let styles = {};
 
   // We have style props. We need to create style and merge with our `csToProps` to get the correct
   // merging order for styles
   if (shouldRuntimeMergeStyles) {
-    const styles = boxStyleFn(styleProps);
-    stylePropsClassName = css(styles);
+    styles = boxStyleFn(styleProps);
   }
 
-  return handleCsProp(elemProps, [localCs, stylePropsClassName]) as Omit<
-    T,
-    'cs' | keyof CommonStyleProps
-  >;
+  return handleCsProp(elemProps, [localCs, styles]) as Omit<T, 'cs' | keyof CommonStyleProps>;
 }
--- a/modules/react/layout/spec/mergeStyles.spec.tsx
+++ b/modules/react/layout/spec/mergeStyles.spec.tsx
@@ -1,18 +1,17 @@
 import React from 'react';
 
-import {cache} from '@emotion/css';
-import {jsx, CacheProvider} from '@emotion/react';
+import {jsx} from '@emotion/react';
 import styled from '@emotion/styled';
 import {render as rtlRender, screen} from '@testing-library/react';
 
-// We need to force Emotion's cache wrapper to use the cache from `@emotion/css` for tests to pass
-const CacheWrapper = props => <CacheProvider value={cache} {...props} />;
-// @ts-ignore We want the types to be the same, but I don't care to fix the error
-const render: typeof rtlRender = (ui, options) =>
-  rtlRender(ui, {wrapper: CacheWrapper, ...options});
-
 import {mergeStyles} from '@workday/canvas-kit-react/layout';
 import {createStyles} from '@workday/canvas-kit-styling';
+import {CanvasProvider} from '@workday/canvas-kit-react/common';
+
+// We need to force Emotion's cache wrapper to use the cache from `@workday/canvas-kit-styling`
+// @ts-ignore We want the types to be the same, but I don't care to fix the error
+const render: typeof rtlRender = (ui, options) =>
+  rtlRender(ui, {wrapper: CanvasProvider, ...options});
 
 describe('mergeStyles', () => {
   const padding = {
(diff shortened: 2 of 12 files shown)


modules/react/layout/lib/utils/mergeStyles.ts
```diff
@@ -71,20 +70,13 @@ export function mergeStyles<T extends {}>(
...
-    const styles = boxStyleFn(styleProps);
-    stylePropsClassName = css(styles);
+    styles = boxStyleFn(styleProps);
   }
 
-  return handleCsProp(elemProps, [localCs, stylePropsClassName]) as Omit<
-    T,
-    'cs' | keyof CommonStyleProps
-  >;
+  return handleCsProp(elemProps, [localCs, styles]) as Omit<T, 'cs' | keyof CommonStyleProps>;
```
NicholasBoll, line 81: I don't know why we used the `css` function here. Probably because this function existed before `handleCsProp` handled dynamic style overrides. But now that `handleCsProp` does this, we can just pass the style object.

modules/react/layout/spec/mergeStyles.spec.tsx
```diff
@@ -1,18 +1,17 @@
...
-  rtlRender(ui, {wrapper: CacheWrapper, ...options});
-
 import {mergeStyles} from '@workday/canvas-kit-react/layout';
 import {createStyles} from '@workday/canvas-kit-styling';
+import {CanvasProvider} from '@workday/canvas-kit-react/common';
+
+// We need to force Emotion's cache wrapper to use the cache from `@workday/canvas-kit-styling`
+// @ts-ignore We want the types to be the same, but I don't care to fix the error
+const render: typeof rtlRender = (ui, options) =>
+  rtlRender(ui, {wrapper: CanvasProvider, ...options});
```
NicholasBoll, line 14: The `CanvasProvider` handles setting the correct cache object, so we should use `CanvasProvider` instead a custom `CacheProvider`

modules/styling/lib/cs.ts
```diff
@@ -722,25 +730,7 @@ export function handleCsProp<
...
-    returnProps.className = getRegisteredStyles(
-      cache.registered,
-      registeredStyles,
-      returnProps.className || ''
-    );
-
-    // The `className` is mutated again, but with a brand new CSS class name of all the merged
-    // registered styles. This is how Emotion merges styles from the `css` prop and `styled`
-    // components.
-    returnProps.className += css(registeredStyles);
```
NicholasBoll, line 743: I didn't realize before that the `cx` function from `@emotion/css` does this exact thing. All I need to do is detect if there are runtime dynamic styles and call the `cx` function from the Emotion instance if there is and Emotion will handle style merging for me. This reduces the code we maintain and have to explain and still passes all our tests.

modules/styling/spec/cs.spec.tsx
```diff
@@ -22,8 +22,10 @@ import {
   handleCsProp,
   keyframes,
   CSProps,
+  getCache,
 } from '../lib/cs';
 
+const cache = getCache();
```
NicholasBoll, line 28: I'm not 100% convinced there should be a `getCache` function if there's already a `getInstance` function. Most of the time, people should not use `getInstance`, but may need to use `getCache`, so I don't know...

Dealing with the cache and the Emotion instance are advance use-cases.

modules/styling/stories/MergingStyles.stories.mdx
```diff
@@ -1,6 +1,7 @@
 import {SymbolDoc} from '@workday/canvas-kit-docs';
 
 import {CreateStencil} from './examples/CreateStencil';
+import {StylingOverrides} from './examples/StylingOverrides';
```
NicholasBoll, line 4: This was accidentally removed in #2498 from another file and not added here. The `ExampleCode` block below did not error, but instead could not find the story so the example rendered as "No code available" and a blank example:

empty example

Importing the example fixes the documentation.

modules/styling/stories/Basics.stories.mdx
```diff
@@ -312,3 +312,55 @@ The HTML will look something like this:
 Notice the stencil adds all the class names that match the base, modifiers, and compound modifiers.
 
 <ExampleCodeBlock code={CreateStencil} />
+
+### Keyframes
+
+The `keyframes` function re-exports the [Emotion CSS keyframes](https://emotion.sh/docs/keyframes)
+function, but is compatible with a custom Emotion instance and is understood by the Static style
+transformer.
+
```
vibdev, line 321: Should you also call out the injectGlobal function

NicholasBoll: I don't have support in the style transform yet, but I think it makes sense to add it.

NicholasBoll: Added it.

RayRedGoose (approved):

Issue #2630, opened 2024-03-06
Run the style transform in Webpack for development and visual regression


NicholasBoll:
## 🚀 Feature Proposal

Add the style transformer to webpack so we can verify everything is working in development and in visual regression.

## Motivation

There's not really a way to know the CSS class names or if the style transform is working correctly in Storybook. The scariest is lack of visual regression. We should verify the transform is working in visual regression rather than relying on our users to report visual regression issues.

PR #2643, opened 2024-03-13
ci: Run style transform in Storybook

NicholasBoll:
## Summary

Fixes: #2630 

Runs the Style transform as part of Webpack so we can test the style transform both visually in Chromatic and in development. We'll also get any errors caused by incorrect styling.

- The style transform in Webpack shares the same TypeScript program as the doc parser. The Webpack processing time is about the same as just running the Doc Parser (doc parser is slow with all those type checker calls)
- Style transform is on by default in Storybook in development. We can see errors the style transform will give during a CI build in Webpack
- Chromatic will run visual regression against our style transform! So we can see any style transform issues BEFORE our users see them.
- `SKIP_DOCGEN=true` controls both the processing by the docgen as well as the style transform. Disabling these preprocessors is a much faster experience for both getting Storybook started as well as hot reloading. But docgen/style transform should be run for accuracy at least before pushing changes to a PR

## Release Category
Infrastructure

---

## Checklist

## For the Reviewer

Test files

- [ ] Code
- [ ] Testing

## Testing Manually

Run storybook locally using `yarn start`. You cannot skip docs.


modules/docs/docgen/createDocProgram.js
```diff
@@ -34,8 +34,10 @@ function getTSConfig(basePath = '.') {
     const contents = ts.readConfigFile(tsconfigPath, ts.sys.readFile);
     if (contents.config) {
       // https://github.com/microsoft/TypeScript/issues/5276#issuecomment-149369652
-      config = ts.convertCompilerOptionsFromJson(contents.config.compilerOptions, tsconfigPath)
-        .options;
+      config = ts.convertCompilerOptionsFromJson(
+        contents.config.compilerOptions,
+        tsconfigPath
+      ).options;
```
NicholasBoll, line 40: This is a prettier change since we upgraded to Prettier v2. First change to this file since then.

.storybook/main.js
```diff
@@ -64,8 +67,18 @@ module.exports = {
         include: [modulesPath],
         exclude: /examples|stories|spec|codemod|docs/,
         loaders: [
+          // loaders are run in reverse order. symbol-doc-loader needs to be done first
+          {
+            loader: path.resolve(__dirname, 'style-transform-loader'),
+            options: {
+              Doc,
+            },
+          },
```
mannycarrera4, line 76: is this what adds the transform as part of webpack?

.storybook/symbol-doc-loader.js
```diff
@@ -1,6 +1,6 @@
 // @ts-check
-const {createDocProgram} = require('../modules/docs/docgen/createDocProgram');
```
mannycarrera4, line 2: why was this removed?

josh-bagwell: Just making a note here, I think it's because Nicholas is creating the program now in `main.js`. Could be wrong though

NicholasBoll: Yes. The program is created in `main.js`. I think eventually it makes sense to have a Webpack plugin that manages the program instead though.

.storybook/main.js
NicholasBoll: Yes

PR #2677, opened 2024-04-03
ci: Run style transform in Storybook

NicholasBoll:
## Summary

Fixes: #2630 

Runs the Style transform as part of Webpack so we can test the style transform both visually in Chromatic and in development. We'll also get any errors caused by incorrect styling.

- The style transform in Webpack shares the same TypeScript program as the doc parser. The Webpack processing time is about the same as just running the Doc Parser (doc parser is slow with all those type checker calls)
- Style transform is on by default in Storybook in development. We can see errors the style transform will give during a CI build in Webpack
- Chromatic will run visual regression against our style transform! So we can see any style transform issues BEFORE our users see them.
- `SKIP_DOCGEN=true` controls both the processing by the docgen as well as the style transform. Disabling these preprocessors is a much faster experience for both getting Storybook started as well as hot reloading. But docgen/style transform should be run for accuracy at least before pushing changes to a PR

## Release Category
Infrastructure

---

## Checklist

## For the Reviewer

Test files

- [ ] Code
- [ ] Testing

## Testing Manually

Run storybook locally using `yarn start`. You cannot skip docs.


modules/styling-transform/lib/utils/handleCreateStencil.ts
```diff
@@ -62,7 +45,36 @@ export const handleCreateStencil: NodeTransformer = (node, context) => {
...
+              return (
+                result ||
+                fileStyles.reduce((_, rule) => {
+                  if (rule.includes(`.${className}`)) {
+                    return true;
+                  }
+                  return result;
+                }, false)
+              );
+            }, false)
```
RayRedGoose, line 62:
```suggestion
            !Object.values(context.styles).some(fileStyles => {
              return fileStyles.slice(-1).includes(`.${className}`);
            });
```
As it returns true only for last item that includes class name, I don't think it's good to use 2 reduce functions.

modules/styling-transform/spec/createProgramFromSource.ts
```diff
@@ -96,7 +103,10 @@ export function createProgramFromSource(...args: any[]) {
     getDirectories: path => ts.sys.getDirectories(path),
     // This should be kept up to date with getSourceFile()
     fileExists: fileName => {
-      return !!sourceFiles.find(s => s.fileName === fileName) || ts.sys.fileExists(fileName);
+      return (
+        !!sourceFiles.find(s => s.fileName === getLocalFileName(sources, fileName)) ||
```
RayRedGoose, line 107:
```suggestion
        sourceFiles.some(s => s.fileName === getLocalFileName(sources, fileName)) ||
```

Can we use `.some` here as we are looking for a boolean value?

modules/react/combobox/lib/ComboboxMenuList.tsx
```diff
@@ -38,7 +37,7 @@ export const ComboboxMenuList = createSubcomponent('ul')({
       as={Element}
       model={model}
       marginY={cssVar(system.space.x2)}
-      {...mergeStyles(elemProps, comoboxMenuListStencil({orientation: model.state.orientation}))}
+      {...handleCsProp(elemProps, comboboxMenuListStencil({orientation: model.state.orientation}))}
```
RayRedGoose, line 40: Will this broke anything if anyone use style props with this component?

modules/styling-transform/lib/utils/handleCreateStencil.ts
NicholasBoll: You're right that the outer loop is a `.some`, which is more clear than a `.reduce`. But the `.slice(-1)` doesn't make any sense to me.

Both our code are wrong in different ways. The following is correct:

```ts
!Object.values(context.styles).some(fileStyles => {
  return fileStyles.some(rule => rule.includes(`.${className}`));
})
```

modules/react/combobox/lib/ComboboxMenuList.tsx
NicholasBoll: When this was changed to `mergeStyles`, the `Combobox` broke. This fixes that issue by allowing `marginY` to be handled correctly by `ListBox`, otherwise `marginY` was treated as a style prop.

RayRedGoose (approved):
🚀

PR #2690, opened 2024-04-09
ci: Upgrade node version for all Actions

NicholasBoll:
## Summary

Updates the node version of all our GitHub action workflows to use node v18

## Release Category
Infrastructure


NicholasBoll:
This PR won't necessarily work. We'd have to update our dependencies to work with node 18

PR #2706, opened 2024-04-18
chore: Update @workday/canvas-tokens-web version

RayRedGoose:
## Summary

Update @workday/canvas-tokens-web version to 1.3.1

## Release Category
Infrastructure

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,2 +1,2 @@
 workspaces-experimental true
---install.frozen-lockfile true
+
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,2 +1,2 @@
 workspaces-experimental true
---install.frozen-lockfile false
+--install.frozen-lockfile true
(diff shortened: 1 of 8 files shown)


.yarnrc
```diff
@@ -1,2 +1,2 @@
 workspaces-experimental true
---install.frozen-lockfile true
+
```
alanbsmith, line 2:
```suggestion
--install.frozen-lockfile false
```

RayRedGoose: thanks!

.yarnrc
```diff
@@ -1,2 +1,2 @@
 workspaces-experimental true
---install.frozen-lockfile true
+--install.frozen-lockfile false
```
alanbsmith, line 2:
```suggestion
--install.frozen-lockfile true
```

alanbsmith (approved):

Issue #2716, opened 2024-05-01
`onFocus` isn't passed to the visual input on Select. 


mannycarrera4:
## 🐛 Bug Report

There's no clear way to trigger a focus event on Select. I think we can forward `onFocus` to the visual input so that consumers can trigger evenets/ side effects based on when the select is focused.

Here's the onFocus on the hidden input https://github.com/Workday/canvas-kit/blob/master/modules/react/select/lib/Select.tsx#L87

Pass the same to the visual input.

PR #2717, opened 2024-05-01
fix: Pass onFocus to visible Select

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/2716

## Release Category
Components

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods

### DIFF
--- a/modules/react/select/lib/Select.tsx
+++ b/modules/react/select/lib/Select.tsx
@@ -8,7 +8,7 @@ import {
 import {Combobox} from '@workday/canvas-kit-react/combobox';
 
 import {InputGroup, TextInput} from '@workday/canvas-kit-react/text-input';
-import {mergeStyles} from '@workday/canvas-kit-react/layout';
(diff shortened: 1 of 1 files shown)


modules/react/select/lib/Select.tsx
```diff
@@ -94,6 +94,7 @@ export const SelectInput = createSubcomponent(TextInput)({
           disabled={disabled}
           placeholder={placeholder}
           error={error}
+          onFocus={onFocus}
```
drschulz, line 97: Can we also add ref support for this input as well? We also need to be able to get a ref to this element (not the hidden one)

alanbsmith: Unless the hidden input is inert, I _think_ this will double-fire the focus event.
How did the previous Select work around this problem?

alanbsmith: Just tested this. The hidden input is inert, so the event only fires once.

Issue #2720, opened 2024-05-02
Add `box-sizing: border-box` to createStyles util


RayRedGoose:
## 🚀 Feature Proposal

A clear and concise description of what the feature is.

## Motivation

Please outline the motivation for the proposal.

## Example

Please provide an example for how this feature would be used.

PR #2719, opened 2024-05-01
fix: Style merging in compat mode

NicholasBoll:
## Summary

Fixes: #2718

Fixed style merging in compat mode that can result in incorrect styles

## Release Category
Components

## Release Note

This fixes an issue where styles would merge incorrectly in compatibility mode. Compatibility mode is triggered when our component is augmented by a style prop, wrapped with `styled`, passed a `css` prop, or passed a `cs` prop with object styles. This fix ensures that `modifiers` via `createModifiers` or `createStencil` merge styles the same way in both static and compat modes. See the linked issue for more details.

---

## Checklist

- [x] Label `ready for review` has been added to PR

## For the Reviewer

- [x] PR title is short and descriptive
- [x] PR summary describes the change (Fixes/Resolves linked correctly)

## Where Should the Reviewer Start?

Test file

## Testing Manually

You can verify the test fails if the cs file is reverted

PR #2721, opened 2024-05-02
feat: Add `box-sizing:border-box` automatically to all styles created by `createStyles`

RayRedGoose:
## Summary

Resolves: #2720 

This PR adds `box-sizing:border-box` to all style created by `createStyles` as we do the same for stencils. Also, `createModifiers` has been updated to support styles passed as an object to remove usage of `createStyles`.

## Release Category
Infrastructure

### BREAKING CHANGES
Adds `box-sizing: border-box` to all styles created by `createStyles`. If your did not add this style already, it may change the way width works for the component. Our intent is to make all elements use border box layouts to make width calculations more predictable. This change may change the way your component works if you use the width style property.

## Thank You Gif (optional)

Car Fixing GIF By The Back Roads


modules/docs/lib/InformationHighlight/index.tsx
```diff
@@ -55,7 +54,7 @@ export const InformationHighlight = createContainer('section')({
   return (
     <Base
       as={Element}
-      cs={[containerStyles, containerModifiers({variant: model.state.variant})]}
+      {...informationHighlightStencil({variant: model.state.variant})}
```
NicholasBoll, line 57: While a stencil does return a spreadable object, this pattern doesn't handle merging. There's a very simple way to break `InformationHighlight`:

```tsx
<InformationHighlight className="whoops" />
```

The `className` prop returned by the stencil will be overwritten by the incoming prop. At this point, `Base` isn't doing anything useful for us.

This is also why I'm not a fan of us using a `Base` component because even though we can use the `cs` prop, our component cannot handle the `cs` prop because it will get overridden instead  of merged. I think the safest way these components work is by using `handleCsProp` and dropping `Base`:


```tsx
return (
  <Element {...handleCsProp(props, informationHighlightStencil({variant: model.state.variant}))} />
```

modules/styling/lib/cs.ts
```diff
@@ -421,13 +423,21 @@ export function createModifiers<M extends ModifierConfig>(input: M): ModifierRet
   const modifierFn = (modifiers: Partial<ModifierValues<M>>) => {
     return Object.keys(modifiers)
       .filter(key => input[key] && (input as any)[key][modifiers[key]])
-      .map(key => (input as any)[key][modifiers[key]])
+      .map(key => (modifierFn as any)[key][modifiers[key]])
```
NicholasBoll, line 426: This will conflict with #2719

NicholasBoll:
There's something that's missing here: A `createModifiers` style transform. Originally there wasn't one, because `createModifiers` didn't call `createStyles` implicitly. But now you'll find the style transform no longer processes `createModifiers`.

### Current

```tsx
// source
const myModifiers = createModifiers({
  size: {
    large: createStyles({padding: 20}),
    small: createStyles({padding: 10})
  }
})

// transform
const myModifiers = createModifiers({
  size: {
    large: createStyles({name: 'abcde1', styles: 'padding:20px;'}),
    small: createStyles({name: 'abcde2', styles: 'padding:10px;'}),
  }
})
```

### With this change
```tsx
// source
const myModifiers = createModifiers({
  size: {
    large: {padding: 20},
    small: {padding: 10}
  }
})

// transform
const myModifiers = createModifiers({
  size: {
    large: {padding: 20}, // should be {name: 'abcde1', styles: 'padding:20px;'} like stencils
    small: {padding: 10}
  }
})
```

This change means `createModifiers` lost transformation.

PR #2741, opened 2024-05-15
feat: Allow stencil modifiers-in-selectors

NicholasBoll:
## Summary

Add support for using parent class names in Stencil selectors to support sharing of modifiers to subcomponents. Allows support for https://github.com/Workday/canvas-kit/discussions/2711

## Release Category
Styling

---

Overall, this change does the following things:
- Remove the need for `makeEmotionSafe`
  - The hash is now the last part of the variables, so there's no more need for it to be Emotion safe
  - `*-emotion-safe` was being added to the CSS Kit output. CSS Kit doesn't use Emotion and doesn't need to be Emotion safe
- Separate the `names` resolution cache from the `extractedNames` resolution cache
  - Variables resolved with the same value in React Kit and CSS Kit. This is not necessary and there's less chance of collisions if we use hashes in React Kit.
  - Before React Kit: `--cnvs-button-label-emotion-safe`
  - After React Kit: `--label-button-{hash}`
  - Before CSS Kit: `--cnvs-button-label-emotion-safe`
  - After CSS Kit: `--cnvs-button-label`
- Add class names created by `createStyles` and `createStencils` to the `names` and `extractedNames` cache. This allows a reference to class names to be used in selectors.

For example, this is not possible before this change:

```ts
const formFieldStencil = createStencil({
  base: { /* ... */ },
  modifiers: {
    orientation: {
      horizontal: { /* ... */ },
      vertical: { /* ... */ }
    }
  }
})

const formFieldLabelStencil = createStencil({
  base: {
    // base styles
    [`.${formFieldStencil.modifiers.orientation.horizontal} :where(&)`] {
      // styles if the formFieldStencil has the horizontal orientation
    }
  }
})
```

This was possible at runtime, but the transform did not know how to handle it because the modifier class name wasn't resolved yet. This addition removes the need to share `orientation` with the `FormField`'s model and have a modifier in `formFieldLabel` which would require CSS Kit authors to add the orientation modifier to both `FormField` AND `FormField.Label`.

## Checklist

## For the Reviewer


modules/css/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@workday/canvas-kit-css",
-  "version": "10.3.24",
+  "version": "10.3.36",
```
NicholasBoll, line 3: The build process automates the syncing of these version strings.

modules/styling-transform/lib/styleTransform.ts
```diff
@@ -20,7 +20,8 @@ export interface StyleTransformerOptions extends TransformerContext {
   transformers?: NodeTransformer[];
 }
 
-let vars: TransformerContext['variables'] = {};
+let vars: TransformerContext['names'] = {};
+let extractedNames: TransformerContext['extractedNames'] = {};
```
NicholasBoll, line 24: `names` and `extractedNames` are used by the static resolution system. The transformation process populates these caches with resolved names.

For example:
```ts
const foo = createStencil({
  modifiers: {
    size: {
      small: {},
      large: {}
    }
  }
})

const bar = createStencil({
  base: {
    // The type of `foo.modifiers.size.small` is `string` which the static
    // transform doesn't like. It will now be added to `names` so that
    // static resolution uses the actual class name
    [`.${foo.modifiers.size.small} :where(&)`]: {
    }
  }
})
```

NicholasBoll: `names` are a cache of the resolved names for React Kit. It will include hashes. The `extractedNames` are the names used in CSS Kit.

modules/styling-transform/lib/styleTransform.ts
```diff
@@ -83,8 +85,11 @@ export default function styleTransformer(
     const fallbackVars = getVariablesFromFiles(files);
     console.log(`Found ${Object.keys(fallbackVars).length} variables.`);
 
-    // eslint-disable-next-line no-param-reassign
-    vars = {...variables, ...fallbackVars};
+    // eslint-disable-next-line guard-for-in
+    for (const key in fallbackVars) {
+      names[key] = fallbackVars[key];
+    }
+    // vars = {...names, ...fallbackVars};
```
NicholasBoll, line 92: This can be removed

modules/styling-transform/lib/styleTransform.ts
```diff
@@ -83,8 +85,11 @@ export default function styleTransformer(
     const fallbackVars = getVariablesFromFiles(files);
     console.log(`Found ${Object.keys(fallbackVars).length} variables.`);
 
-    // eslint-disable-next-line no-param-reassign
-    vars = {...variables, ...fallbackVars};
+    // eslint-disable-next-line guard-for-in
+    for (const key in fallbackVars) {
+      names[key] = fallbackVars[key];
+    }
```
NicholasBoll, line 91: There was a bug where passing in `names` to the `transform` did not return the same reference. That's because a new reference was used internally. That means I couldn't properly test changes. This update instead mutates the name cache that was passed in so the cache can be properly tested against.

RayRedGoose (approved):

PR #2752, opened 2024-05-21
fix: Support shorthand properties in Stencils

NicholasBoll:
## Summary

Support shorthand properties in Stencils for variables.

## Release Category
Styling

---

For example:

```ts
const myStencil = createStencil({
  vars: {
    color: 'red',
  },
  base({ color }) => ({
    color // shorthand property wasn't supported
  })
})
```

### DIFF
--- a/modules/styling-transform/lib/utils/parseObjectToStaticValue.ts
+++ b/modules/styling-transform/lib/utils/parseObjectToStaticValue.ts
@@ -105,6 +105,15 @@ function parsePropertyToStaticValue(node: ts.Node, context: TransformerContext):
     }
   }
 
+  // { name }
+  if (ts.isShorthandPropertyAssignment(node)) {
+    const key = node.name.text;
+
+    styleObj[key] = parseNodeToStaticValue(node.name, context).toString() || '';
+
+    return styleObj;
+  }
+
   // { ...value }
   if (ts.isSpreadAssignment(node)) {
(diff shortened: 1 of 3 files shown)


modules/styling-transform/lib/utils/parseObjectToStaticValue.ts
```diff
@@ -105,6 +105,15 @@ function parsePropertyToStaticValue(node: ts.Node, context: TransformerContext):
     }
   }
 
+  // { name }
+  if (ts.isShorthandPropertyAssignment(node)) {
```
alanbsmith, line 109: The name of this function is clear enough, but I can't find any docs on this function. Bizarre.

alanbsmith (approved):

modules/styling-transform/lib/utils/parseObjectToStaticValue.ts
NicholasBoll: https://ts-ast-viewer.com/#code/MYewdgzgLgBAZiEMC8MDkCRoNwCheiSwBGAhgE4owDeuM98iMA9MzAMoAWI5UnpYACYAFciAAOAU14BPAIIQIASwDmYALaSwUXAF8gA

NicholasBoll: There is almost no documentation on any TypeScript AST or Type Checker functions. There's no JSDoc. The source code is all I get: https://github.com/microsoft/TypeScript

Issue #2756, opened 2024-05-23
Select component does not accept undefined for initialSelectedIds


drobota:
## 🐛 Bug Report

useSelectInput throws error if initialSelectedIds is set to undefined.

(property) initialSelectedIds?: SelectedIds | undefined

## Link to repl or repo (highly encouraged)
https://github.com/Workday/canvas-kit/blob/60ab75c33942c15bb1fd116ca19bdebeb005a99d/modules/react/select/lib/hooks/useSelectInput.ts

##Error Output
TypeError: Cannot read properties of undefined (reading 'length')

```bash
    <Select initialSelectedIds={id ? [id]: undefined} items={items} getTextValue={getTextValue}>
      <Select.Input {...props} />
      <Select.Popper>
        <Select.Card>
          <Select.List>
             ....
          </Select.List>
        </Select.Card>
      </Select.Popper>
    </Select>
```

PR #2766, opened 2024-05-30
fix: Prevent undefined from overriding model defaultConfig

NicholasBoll:
## Summary

Fixes: #2756

Models use the object spread operator to allow `config` to override `defaultConfig`. If `undefined` is passed to a key, it would cause `undefined` to be assigned to config even if the model didn't allow for that. Instead of using an object spread, `config` keys are looped over to check for `undefined` values.

## Release Category
Components

---

Extended change information:

```tsx
const useMyModel = createModelHook({
  defaultConfig: {
    foo: 'bar'
  },
  requiredConfig: {}
})(config => {
  return {
    state: {
      foo: config.foo // type is `{foo: string}`
    },
    events: {}
  }
})


const model = useMyModel({
  foo: undefined
})
```

**Before**
```tsx
model.state.foo // undefined
```

**After**
```tsx
model.state.foo // 'bar'
```

### DIFF
--- a/modules/react/common/spec/models.spec.tsx
+++ b/modules/react/common/spec/models.spec.tsx
@@ -271,4 +271,40 @@ describe('createModelHook', () => {
       shouldSelect?: (value: string, state: {baz: string}) => boolean;
     }>();
   });
+
+  it('should override defaultConfig values with `undefined` if `undefined` is allowed', () => {
+    const useTestModel = createModelHook({
+      defaultConfig: {
+        foo: undefined,
+      },
+    })(config => {
+      return {state: {foo: config.foo}, events: {}};
+    });
+
+    const {result} = renderHook(() =>
+      useTestModel({
+        foo: undefined,
+      })
+    );
+
+    expect(result.current.state).toHaveProperty('foo', undefined);
+  });
+
+  it('should not override defaultConfig values with `undefined` if `undefined` is not allowed', () => {
+    const useTestModel = createModelHook({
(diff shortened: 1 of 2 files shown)


mannycarrera4 (approved):

modules/react/common/spec/models.spec.tsx
```diff
@@ -271,4 +271,40 @@ describe('createModelHook', () => {
...
+      })
+    );
+
+    expect(result.current.state).toHaveProperty('foo', undefined);
+  });
+
+  it('should not override defaultConfig values with `undefined` if `undefined` is not allowed', () => {
+    const useTestModel = createModelHook({
+      defaultConfig: {
+        foo: 'foo',
```
NicholasBoll, line 296:
```suggestion
        foo: 'bar',
```

modules/react/common/spec/models.spec.tsx
```diff
@@ -271,4 +271,40 @@ describe('createModelHook', () => {
...
+      return {state: {foo: config.foo}, events: {}};
+    });
+
+    const {result} = renderHook(() =>
+      useTestModel({
+        foo: undefined,
+      })
+    );
+
+    expect(result.current.state).toHaveProperty('foo', 'foo');
```
NicholasBoll, line 308:
```suggestion
    expect(result.current.state).toHaveProperty('foo', 'bar');
```

josh-bagwell (approved):

Issue #2797, opened 2024-06-27
Tooltip type describe should add `aria-description` strings to element for better screen reader support


mannycarrera4:
## 🐛 Bug Report

We want to be able to provide the describe tool tip on the description of the child component regardless of whether the tool tip is visible.

Since aria-describedby references another element in the DOM, this’ll only work when the tool tip is visible. And that is, only visible when the child component is focused.

But… screen readers can access things on the page without actually focusing them. So, the describe tool tip doesn’t always work for screen reader users. Only sometimes, when the child component is focused.

## Expected Behavior

Voiceover can describe the element regardless if the tooltip is open or not.

## To Reproduce
With VoiceOver / MacOS:
Launch VoiceOver, open Safari
Canvas Design Site: Tooltip popups > [Describing an element](https://canvas.workdaydesign.com/components/popups/tooltip#tab=web&heading=describing-an-element)
Reading commands: Ctrl + Opt + Right
Issue: observe tool tip text on Delete button is not described
Press Tab to focus “Show Code”
Read backward: Ctrl + Opt + Left Arrow
Issue: observe tool tip text on Delete button is not described
Set Rotor to “Form Fields”: Ctrl + Opt + Cmd + Right Arrow
Jump buttons on the page: Ctrl + Opt + Cmd + UP / DOWN arrow
Issue: observe tool tip text on Delete button is not described
With NVDA on Windows 11:
Launch NVDA, open Chrome
Canvas site => Tooltips => [Describing an element](https://canvas.workdaydesign.com/components/popups/tooltip#tab=web&heading=describing-an-element)
In virtual mode, use quick key ‘b’ to jump to the Delete Button
Issue: observe the tool tip text is not described
## Expected Behavior
Screen readers should be able to automatically describe the Tool tip description while reading elements without keyboard focus.

PR #2814, opened 2024-06-28
feat: Add new `description` variant to Tooltip and deprecate `describe` variant for better accessibility.

williamjstanton:
## Summary

Fixes: #2797  

## Release Category
Components

### Release Note
- We've added a new `description` variant and **deprecated** the `describe` variant.

The `describe` variant of the tool tip relies on `aria-describedby` attribute to assign the tool tip text to the accessible description of the target element. This only works when the tool tip is visible and present in the DOM, which only occurs when the element has focus. Screen readers can access web page elements without necessarily using keyboard focus, therefore, this variant didn't appear to always work reliably for screen reader users. 

The preferred type to use is `description` to provide better accessible support.

## Checklist

## For the Reviewer

For the `describe` variant of `Tooltip`, I replaced the `aria-describedby` attribute with `aria-description` and set the `title` prop value string. Also, remove the check for if visible, so we can always assign this string to the description. 

- in the `useTooltip` custom hook
- check out the updates to the Describe examples in storybook
- check out the new "Tooltip" section under "Examples" in Storybook

- [X] Code
- [X] Documentation
- [X] Testing
- [ ] Codemods

## Testing Manually

In the following scenarios, validate the issues cannot be reproduced and the tool tip text is always described without focusing the button elements. 

With VoiceOver / MacOS:
1. Launch VoiceOver, open Safari
2. Canvas Design Site: Tooltip popups > [Describing an element](https://canvas.workdaydesign.com/components/popups/tooltip#tab=web&heading=describing-an-element)
3. Reading commands: Ctrl + Opt + Right
4. Issue: observe tool tip text on Delete button is not described
5. Press Tab to focus “Show Code”
6. Read backward: Ctrl + Opt + Left Arrow
7. Issue: observe tool tip text on Delete button is not described
8. Set Rotor to “Form Fields”: Ctrl + Opt + Cmd + Right Arrow
9. Jump buttons on the page: Ctrl + Opt + Cmd + UP / DOWN arrow
10. Issue: observe tool tip text on Delete button is not described

With NVDA on Windows 11:
1. Launch NVDA, open Chrome
2. Canvas site => Tooltips => [Describing an element](https://canvas.workdaydesign.com/components/popups/tooltip#tab=web&heading=describing-an-element)
3. In virtual mode, use quick key ‘b’ to jump to the Delete Button
4. Issue: observe the tool tip text is not described

No visual changes. Just `aria` attribute change.


modules/react/tooltip/lib/useTooltip.tsx
```diff
@@ -145,10 +145,11 @@ export function useTooltip<T extends Element = Element>({
...
   };
 
+  // deprecate: I don't want to remove the aria-description string
   // remove `aria-describedby` if undefined to not override what's provided
-  if (targetProps['aria-describedby'] === undefined) {
-    delete targetProps['aria-describedby'];
-  }
+  // if (targetProps['aria-describedby'] === undefined) {
+  //   delete targetProps['aria-describedby'];
+  // }
```
mannycarrera4, line 152: If we always want `aria-description` you can remove these lines of code.

williamjstanton: Sounds good, I got rid of it.

modules/react/tooltip/lib/useTooltip.tsx
```diff
@@ -135,7 +135,7 @@ export function useTooltip<T extends Element = Element>({
 
   const targetProps = {
     // extra description of the target element for assistive technology
-    'aria-describedby': type === 'describe' && visible ? id : undefined,
+    'aria-description': type === 'describe' ? titleText : undefined,
```
RayRedGoose, line 138: I think this is potentially breaking change, because now tooltip will require to have `titleText` passed, but it's optional now.

josh-bagwell: Agree with Raisa. This would be breaking. 

Also, shouldn't both be present since `aria-describedby` points to the target of the element with the id and `aria-description` is just describing what the element is with nothing pointing to it?

williamjstanton: > I think this is potentially breaking change, because now tooltip will require to have `titleText` passed, but it's optional now.

Does this mean `title` prop was optional? Why?

williamjstanton: > Also, shouldn't both be present since `aria-describedby` points to the target of the element with the id and `aria-description` is just describing what the element is with nothing pointing to it?

No, for similar reasons why we discourage using both `aria-labelledby` references and `aria-label` strings on the same element. Browsers follow an algorithm to define a 1) `accessible name` and 2) `accessible description` for an element. 

If `aria-describedby` is used on an element, then an `aria-description` string will be ignored and discarded by the browser.

NicholasBoll: `titleText` is optional in the `useTooltip` hook, but not the `title` in the `Tooltip` component. `titleText` was optional in `useTooltip` because `type="describe"` did not use the config. This change makes it so `type="describe"` also uses the config. There is still the `type="muted"` that doesn't use the `titleText` config.

Now, `title` is required by `Tooltip` because it is the contents of the tooltip. `OverflowTooltip` does not use `title` because the contents of the target are the contents of the tooltip. An `OverflowTooltip` uses the `type="muted"` because a screen reader doesn't understand the concept of the ellipsis - to the screen reader, the full contents are exposed already. A screen dictation user will be able to target based on what content is already visible and not cut off - from there it is part of the design to make sure the most significant part of the target text is visible when overflowing.

modules/react/tooltip/spec/Tooltip.spec.tsx
```diff
@@ -28,10 +28,9 @@ describe('Tooltip', () => {
       fireEvent.mouseEnter(screen.getByText('Test Text')); // triggers the tooltip
 
       jest.advanceTimersByTime(300); // advance the timer by the amount of delay time
-      expect(screen.getByText('Test Text')).toHaveAttribute('aria-describedby');
+      expect(screen.getByText('Test Text')).toHaveAttribute('aria-description');
```
NicholasBoll, line 31: I think it makes sense to assert the contents of the attribute and not just the presence of it. The second parameter is the text that is expected.

modules/react/tooltip/spec/Tooltip.spec.tsx
```diff
@@ -28,10 +28,9 @@ describe('Tooltip', () => {
...
 
-      const id = screen.getByText('Test Text').getAttribute('aria-describedby');
-      expect(screen.getByRole('tooltip')).toHaveAttribute('id', id);
+      expect(screen.getByRole('tooltip')).toHaveAttribute('id');
```
NicholasBoll, line 33: I don't think it is important for the tooltip to have an `id` at this point. This line should be removed. The `id` was only used for the `aria-describedby`.

mannycarrera4:
closing to revisit after v12

cypress/component/Tooltip.spec.tsx
```diff
@@ -138,7 +139,7 @@ describe('Tooltip', () => {
         cy.checkA11y();
       });
 
-      it('the "Delete" button should have an aria-describedby linking to the role="tooltip" element', () => {
+      it('the "Delete" button should have an accessible description equal to the tooltip text', () => {
```
mannycarrera4, line 142:
```suggestion
      it('the "Delete" button should have an aria-describedby linking to the role="tooltip" element', () => {
```

modules/react/_examples/stories/examples/Tooltips/ListOfUploadedFiles.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const ListOfUploadedFiles = () => {
+  return (
+    <>
+      <Heading size="medium">Uploaded Files:</Heading>
+      <Flex as="ul" gap="1rem" flexDirection="column">
+        {files.map(i => (
+          <Flex as="li" style={listStyles}>
+            <Text>{i}</Text>
+            <Tooltip type="describe" title={i}>
```
mannycarrera4, line 28:
```suggestion
            <Tooltip type="description" title={i}>
```

modules/react/tooltip/lib/Tooltip.tsx
```diff
@@ -45,7 +45,7 @@ export interface TooltipProps extends Omit<React.HTMLAttributes<HTMLDivElement>,
    * - `label`: Sets the accessible name for the wrapped element. Use for icons or if tooltip
    *   `title` prop is the same as the text content of the wrapped element. E.g. TertiaryButtons that render an icon or
    *   Ellipsis tooltips.
-   * - `describe`: Sets `aria-describedby` of the wrapped element. Use if the tooltip has additional
+   * - **Deprecated: `describe` is deprecated, please use `description`**.`describe`: Sets `aria-describedby` of the wrapped element. Use if the tooltip has additional
```
mannycarrera4, line 48: not sure of a better way to deprecate a string union

NicholasBoll (approved):

Issue #2905, opened 2024-09-04
Support updated focusRing through `colors` prop


josh-bagwell:
## 🐛 Bug Report

There's an issue with using `boxShadowInner` and `boxShadowOuter` within `focus` in our `colors` prop. This threw a TS error. 

## Expected Behavior

The expected result of this: 
```tsx
    <PrimaryButton colors={{focus: {boxShadowInner: 'red'}}}>Primary</PrimaryButton>
```
Should be that the inner box-shadow is updated to "red". This threw a TS error. 

## Actual Results

It throws a TS error.

PR #2906, opened 2024-09-04
fix: Updated ButtonColors interface to deprecated focusRing from focus

josh-bagwell:
## Summary

Fixes: #2905  

This fixes a TS error in buttons that wouldn't allow the update of the focusRing styles through the colors prop. Also, deprecated the use of focusRing within focus in the colors prop on buttons and added support for boxShadowInner and boxShadowOuter within focus in colors prop. 

## Release Category
Components

### Release Note
Deprecated the use of focusRing within focus in the colors prop on buttons as this does not work with our current styling methods. Added support for boxShadowInner and boxShadowOuter within focus in colors prop. 

---

`/modules/react/button/lib/types.ts`

### DIFF
--- a/modules/react/button/lib/types.ts
+++ b/modules/react/button/lib/types.ts
@@ -17,7 +17,7 @@ export interface ButtonColors {
   active?: ButtonStateColors;
   focus?: ButtonStateColors & {
     /**
-     * @deprecated `focusRing` within focus, is deprecated and will not work with our current method of styling. This will be removed in a later version. If you want to customize the focus ring, use `boxShadowInner` and `boxShadowOuter` to update the inner and outer colors of the focus ring. If you decide to use these, use with caution.
+     * @deprecated `focusRing` within focus, is deprecated and will not work with our current method of styling. This will be removed in a later version. If you want to customize the focus ring, use `boxShadowInner` and `boxShadowOuter` to update the inner and outer colors of the focus ring. Use with caution.
      */
     focusRing?: CSSObject;
     /**
--- a/modules/react/button/lib/types.ts
+++ b/modules/react/button/lib/types.ts
@@ -18,9 +18,6 @@ export interface ButtonColors {
   focus?: ButtonStateColors & {
     /**
      * @deprecated This option is no longer supported at run time and will be removed from the type interface in a v12. If you want to customize the focus ring, use `boxShadowInner` and `boxShadowOuter` to update the inner and outer colors of the focus ring. Use with caution.
-     * ```tsx
-     * colors={{focus: {boxShadowInner: COLOR_HERE}}}
-     * ```
      */
     focusRing?: CSSObject;
     /**
--- a/modules/react/button/lib/types.ts
+++ b/modules/react/button/lib/types.ts
@@ -17,12 +17,16 @@ export interface ButtonColors {
   active?: ButtonStateColors;
   focus?: ButtonStateColors & {
     /**
-     * FocusRing within focus, is deprecated in v12 and will not work with our current method of styling.
-     * This will be removed in a later version.
-     * @deprecated
+     * @deprecated `focusRing` within focus, is deprecated and will not work with our current method of styling. This will be removed in a later version. If you want to customize the focus ring, use `boxShadowInner` and `boxShadowOuter` to update the inner and outer colors of the focus ring. If you decide to use these, use with caution.
      */
     focusRing?: CSSObject;
+    /**
+     * Updates the color of the inner `box-shadow` within a focus ring.
+     */
     boxShadowInner?: string;
+    /**
+     * Updates the color of the outer `box-shadow` within a focus ring.
+     */
     boxShadowOuter?: string;
   };
   disabled?: ButtonStateColors;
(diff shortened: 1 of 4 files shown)


KevinLee00:
Hi, I'm the developer that ran into the bug.

 In addition to this code change, I'd love to see the official Canvas Kit documentation updated to no longer reference `focusRing`, and to add `boxShadowInner` and `boxShadowOuter` in order to prevent developers in the future from running into this.

Screenshot 2024-09-04 at 4 49 26 PM

josh-bagwell:
> Hi, I'm the developer that ran into the bug.
> 
> In addition to this code change, I'd love to see the official Canvas Kit documentation updated to no longer reference `focusRing`, and to add `boxShadowInner` and `boxShadowOuter` in order to prevent developers in the future from running into this.
> 
> Screenshot 2024-09-04 at 4 49 26 PM

Hey Kevin,

`focusRing` will be deprecated before it's removed. This is now deprecated and when hovered, it shows the deprecation message. boxShadowInner and boxShadowOuter are now in the documentation as well.

modules/react/button/lib/types.ts
```diff
@@ -16,7 +16,14 @@ export interface ButtonColors {
   hover?: ButtonStateColors;
   active?: ButtonStateColors;
   focus?: ButtonStateColors & {
+    /**
+     * FocusRing within focus, is deprecated in v12 and will not work with our current method of styling.
+     * This will be removed in a later version.
+     * @deprecated
```
mannycarrera4, line 22:
```suggestion
     *  @deprecated `focusRing` within focus is deprecated and will not work with our current method of styling. This will be removed in a future major version. If you want to customize the focus ring, use `boxShadowInner` and `boxShadowOuter` to change the colors, but use this with caution.
```

modules/react/button/lib/types.ts
```diff
@@ -16,7 +16,14 @@ export interface ButtonColors {
...
+     */
     focusRing?: CSSObject;
+    boxShadowInner?: string;
+    boxShadowOuter?: string;
```
mannycarrera4, line 26: i would add comments above these saying what they style

modules/react/button/stories/button/examples/CustomStyles.tsx
```diff
@@ -20,7 +20,9 @@ const getDropdownColors = () => {
       label: colors.frenchVanilla100,
     },
     active: {},
-    focus: {},
+    focus: {
+      boxShadowOuter: colors.berrySmoothie400,
```
mannycarrera4, line 24: i would add boxShadadow inner as well just to showcase

modules/react/button/lib/types.ts
josh-bagwell: I added some additional items with that as well

modules/react/button/lib/types.ts
josh-bagwell: Agreed.

modules/react/button/stories/button/examples/CustomStyles.tsx
josh-bagwell: Agreed

modules/react/button/lib/types.ts
```diff
@@ -16,7 +16,18 @@ export interface ButtonColors {
   hover?: ButtonStateColors;
   active?: ButtonStateColors;
   focus?: ButtonStateColors & {
+    /**
+     * @deprecated `focusRing` within focus, is deprecated and will not work with our current method of styling. This will be removed in a later version. If you want to customize the focus ring, use `boxShadowInner` and `boxShadowOuter` to update the inner and outer colors of the focus ring. If you decide to use these, use with caution.
```
mannycarrera4, line 20: nit: `Use these with caution`. Just less words

josh-bagwell: That makes sense

mannycarrera4:
image
in the API docs, focus ring doesn't have a tooltip describing that it's deprecated, wondering how we can achieve that

NicholasBoll:
> image in the API docs, focus ring doesn't have a tooltip describing that it's deprecated, wondering how we can achieve that

That processing happens in the `SymbolDoc` parser. The `boxShadowOuter` has an underline on it with a tooltip containing the JSDoc description. We can add support for `@depreacted` tags via a strikethrough styling.

NicholasBoll:
The other option is making this be an `interface` that's exported that has the `focusRing`, `boxShadowInner`, and `boxShadowOuter` keys. Interfaces are rendered via tables instead of inlined.

mannycarrera4:
I think we can update `value.tsx` in the docs package to render the deprecated stuff:

```{p.description || p.tags.deprecated ? (
              <DescriptionTooltip
                type="describe"
                style={{maxWidth: '50em'}}
                title={<MdxJSToJSX>{p.description || p.tags.deprecated}</MdxJSToJSX>}
              >
                <span
                  className="token property"
                  style={{
                    cursor: 'pointer',
                    textDecoration: p.tags.deprecated ? 'line-through' : 'underline',
                    textDecorationStyle: 'dotted',
                  }}
                >
                  {p.name}
                </span>
              </DescriptionTooltip>
            )
            ```

mannycarrera4:
image

mannycarrera4 (approved):

PR #3006, opened 2024-10-28
fix: Update Select to open with spacebar

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/3007

## Release Category
Components

---

### DIFF
--- a/cypress/integration/Select.spec.ts
+++ b/cypress/integration/Select.spec.ts
@@ -361,7 +361,7 @@ describe('Select', () => {
       });
       it('should select phone and the selected value should be visible', () => {
         // should open the menu
-        cy.findByRole('listbox');
+        cy.findByRole('listbox').should('exist');
       });
     });
   });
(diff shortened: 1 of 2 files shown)


cypress/integration/Select.spec.ts
```diff
@@ -353,6 +353,17 @@ describe('Select', () => {
...
     });
+
+    context('when spacebar is types and no value exists', () => {
+      beforeEach(() => {
+        cy.findByRole('combobox').focus();
+        cy.focused().realType(' ');
+      });
+      it('should select phone and the selected value should be visible', () => {
+        // should open the menu
+        cy.findByRole('listbox');
```
NicholasBoll, line 364: I don't like implicit assertions. Add the assertion to be explicit about intent.

```suggestion
        cy.findByRole('listbox').should('exist');
```

modules/react/select/lib/hooks/useSelectInput.ts
```diff
@@ -150,6 +149,7 @@ export const useSelectInput = composeHooks(
           if (event.key !== 'Tab') {
             event.preventDefault();
           }
+          console.log(elemProps.keySofar);
```
NicholasBoll, line 152:
```suggestion
```

NicholasBoll (approved):

PR #3029, opened 2024-11-04
fix: Add gap to overflow calculations

NicholasBoll:
## Summary

Adds the setting of the container gap to the `useOverflowListMeasure` hook

## Release Category
Components

---


modules/react/collection/lib/useOverflowListMeasure.ts
```diff
@@ -9,12 +14,22 @@ import {useOverflowListModel} from './useOverflowListModel';
  * overflow detection.
  */
 export const useOverflowListMeasure = createElemPropsHook(useOverflowListModel)((model, ref) => {
-  const localRef = React.useRef(null);
-  const {ref: resizeRef} = useResizeObserver({
+  // const localRef = React.useRef(null);
```
NicholasBoll, line 17:
```suggestion
```

Issue #3025, opened 2024-11-04
Add removal support to the Collection System


NicholasBoll:
## 🚀 Feature Proposal

Add a elemProps hook for deleting an item from a list that is compatible with the Collection system. Focus management should be handled by the hook. I couldn't find a specification for the focus management that I could point to. The [Tabs pattern example](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#keyboardinteraction) covers some specifications:

> Delete (Optional): If deletion is allowed, deletes (closes) the current tab element and its associated tab panel, sets focus on the tab following the tab that was closed, and optionally activates the newly focused tab. If there is not a tab that followed the tab that was deleted, e.g., the deleted tab was the right-most tab in a left-to-right horizontal tab list, sets focus on and optionally activates the tab that preceded the deleted tab. If the application allows all tabs to be deleted, and the user deletes the last remaining tab in the tab list, the application moves focus to another element that provides a logical work flow. As an alternative to Delete, or in addition to supporting Delete, the delete function is available in a context menu.

Other examples have the same focus management: https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/

## Motivation

We have a Tabs example that allows removal of tabs. We've received requests and questions around removing items from a list. We need to do this on `MultiSelect`. The focus management can be difficult to do correctly in user-land. It makes sense to cover this functionality in Canvas Kit via a hook.

PR #3036, opened 2024-11-05
feat(collection): Add removable support

NicholasBoll:
## Summary

Fixes: #3025

Adds `remove` event to the `ListModel`. An `onRemove` config should be added to dynamic lists to remove the item from the collection.

The `MultiSelect.Input` uses this new remove event to handle removing items from the Selected pill list when the user uses the “Delete” key. Focus is managed by the collection system when an item is removed.

## Release Category
Components

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing

### DIFF
--- a/cypress/component/Tabs.spec.tsx
+++ b/cypress/component/Tabs.spec.tsx
@@ -75,7 +75,7 @@ describe('Tabs', () => {
 
         context('when the tab key is pressed', () => {
           beforeEach(() => {
-            cy.tab();
+            cy.realPress('Tab');
           });
 
           it('should move focus to the tabpanel', () => {
@@ -148,7 +148,7 @@ describe('Tabs', () => {
 
           context('when the tab key is pressed', () => {
             beforeEach(() => {
-              cy.tab();
+              cy.realPress('Tab');
             });
 
             it('should focus on the tab panel of the first tab', () => {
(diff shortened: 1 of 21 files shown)


modules/preview-react/multi-select/lib/MultiSelectInput.tsx
```diff
@@ -228,34 +186,25 @@ export const MultiSelectSearchInput = createSubcomponent(TextInput)({
...
+            {...formInputProps}
+          />
           <InputGroup.Input
             data-part="user-input"
             as={Element}
             aria-labelledby={ariaLabelledBy}
             {...elemProps}
           />
-          <InputGroup.InnerEnd>
+          <InputGroup.InnerEnd width={system.space.x4}>
```
mannycarrera4, line 200: do we care if this is a style prop or should we use `cs` here?

josh-bagwell (approved):

josh-bagwell:
Seeing two failed Tabs specs

modules/preview-react/multi-select/lib/MultiSelectInput.tsx
NicholasBoll: `width` is a CSS variable in inputgroup, so `cs={{width...}}` would now work correctly.

mannycarrera4 (approved):

cypress/component/Tabs.spec.tsx
```diff
@@ -75,7 +75,7 @@ describe('Tabs', () => {
 
         context('when the tab key is pressed', () => {
           beforeEach(() => {
-            cy.tab();
+            cy.realPress('Tab');
```
josh-bagwell, line 78: realPress or realType?

NicholasBoll: There's no `realType` for the tab key. There's only a `realPress`. It probably makes more sense to use `realPress` for keyboard commands and `readType` or just `type` for entering in text.

Issue #3048, opened 2024-11-08
feat: Add BaseTable Component


josh-bagwell:
## 🚀 Feature Proposal

The current `Table` component uses CSS Grid for it's cell and header sizing since it controls it via the `Table.Row`. But, the standard HTML `Table` element, will set the size of each cell to the largest cell within the Table. So if you were to look at this from a "column"(Table's do not build in terms of columns but in rows, this is why I used quotes to denote that) perspective, each "column" of the table would automatically be set to the width of the largest cell in that "column" using the older `Table` component prior to v11. The new HTMLTable would be a new component. This new component can give a user an experience similar to the older Table that they may have been using before but, it will be styled the same as the Table that exists now. This will work alongside the current Table in main that uses CSS Grid.
## Motivation

This had come up with someone in support as they wanted a Table to auto set the widths of each "column" to it's largest cell. The current Table does not do this automatically as each Table is made up of rows and those rows dictate that sizing needed automatically through CSS Grid's `grid-template-columns` property. This gives the user a lot of flexibility when it comes to sizing the Table however, it does not set the width of the cells automatically to the largest cell within that "column" of that Table.

PR #3126, opened 2025-02-06
feat: Add New Component `BaseTable`

josh-bagwell:
## Summary

Fixes: #3048  

Add a new base component for the Table component: `BaseTable`. This will be a standard Html `Table` without any  `CSS Grid` features. This will be the "base" level component that will be used to build other Tables.  

## Release Category
Components

### Release Note
Addition of a new base level component: `BaseTable`. This will be a standard HTML `Table` with no `CSS Grid` features. 

---

`/modules/react/table`


modules/react/table/lib/Table.tsx
```diff
@@ -8,18 +8,67 @@ import {TableHeader} from './TableHeader';
...
-import {createStencil, px2rem} from '@workday/canvas-kit-styling';
+import {calc, createStencil, px2rem} from '@workday/canvas-kit-styling';
 import {system} from '@workday/canvas-tokens-web';
 
-const tableStencil = createStencil({
+export interface TableProps extends GridProps {
+  /**
+   * Variant has an option for `simple` which will have the Table act as a HTML Table without Grid styles.
+   */
+  variant?: 'html';
```
mannycarrera4, line 18: Maybe this is `table | grid` with the default being gird? Not sure html conveys that it behaves like a default html table element.

josh-bagwell: Yup. Going to be addressing this.

modules/react/table/lib/BaseTableBody.tsx
```diff
@@ -29,11 +28,11 @@ const tableBodyStencil = createStencil({
   },
 });
 
-export const TableBody = createComponent('tbody')({
+export const BaseTableBody = createComponent('tbody')({
```
mannycarrera4, line 31: this is a breaking change

modules/react/table/lib/Table.tsx
```diff
@@ -1,29 +1,137 @@
...
+export interface TableProps extends GridProps {}
+
 const tableStencil = createStencil({
+  extends: baseTableStencil,
+  base: {
+    display: 'grid',
+  },
+});
+
+const tableBodyStencil = createStencil({
```
mannycarrera4, line 23: we should export these

modules/react/table/lib/Table.tsx
```diff
@@ -1,29 +1,137 @@
...
+      </Element>
+    );
+  },
+});
+
+export const TableCaption = createComponent('caption')({
+  displayName: 'Table.Caption',
+  Component: ({children, ...elemProps}: FlexProps, ref, Element) => {
+    return (
+      <Element ref={ref} {...mergeStyles(elemProps, tableCaptionStencil())}>
```
mannycarrera4, line 81: Question for @NicholasBoll . If `TableCaption` extends `baseTableCaption` should this also render `BaseTableCaption`?

modules/react/table/lib/BaseTableBody.tsx
mannycarrera4: oh i see.. nvm, github made this view confusing

modules/react/table/lib/Table.tsx
josh-bagwell: You mean in terms of rendering `<Element/>` or `<BaseTableCaption/>`

modules/react/table/lib/Table.tsx
josh-bagwell: Agreed.

modules/react/table/index.ts
```diff
@@ -1 +1,3 @@
 export * from './lib/Table';
+export * from './lib/BaseTable';
+export * from './lib/parts';
```
mannycarrera4, line 3: i think you also want to add `TableRow` since that folder doesn't have an `index.tsx`

modules/react/table/lib/Table.tsx
```diff
@@ -66,11 +174,11 @@ export default function App() {
  */
 export const Table = createComponent('table')({
   displayName: 'Table',
-  Component: ({children, ...elemProps}: GridProps, ref, Element) => {
+  Component: ({children, ...elemProps}: TableProps, ref, Element) => {
     return (
-      <Element ref={ref} {...mergeStyles(elemProps, tableStencil())}>
+      <BaseTable as={Element} ref={ref} {...mergeStyles(elemProps, tableStencil())}>
```
mannycarrera4, line 179: I think Element is fine here after talking to Nicholas no?

josh-bagwell: Yeah probably. I can update it.

mannycarrera4 (approved):

modules/react/table/index.ts
josh-bagwell: Yup added an index file so I can export from the folder.

PR #3128, opened 2025-02-08
chore: Expandable component style refactor

williamjstanton:
## Summary

Fixes: #2954   

Expandable style re-factor is needed with the new cs prop. We will need to update this component to use the new system level tokens via CSS variables.

## Release Category
Tokens

### Release Note
- The `ExpandableTarget` component is now rendering the `BaseButton` component, and may have some small impact on the visual styling of the keyboard focus ring. 
- The `ExpandableIcon` stencil is extending the `SystemIcon` stencil, and may have had an impact on the chevron icon color. 
- Fixed an accessibility bug in the `Avatar` component where an `aria-label` string was rendered on a generic HTML element that did not have an explicit or implicit `role`

## Checklist

## For the Reviewer

modules/labs-react/expandable/lib/Expandable.tsx 
(all subcomponents) 
Line 274: modules/react/avatar/lib/Avatar.tsx 

- [X] Code
- [X] Documentation
- [ ] Testing
- [ ] Codemods

expandable-focus

expandable-hover


modules/labs-react/expandable/lib/ExpandableAvatar.tsx
```diff
@@ -3,29 +3,32 @@ import React from 'react';
...
+import {createStencil} from '@workday/canvas-kit-styling';
+import {mergeStyles} from '@workday/canvas-kit-react/layout';
 
 export interface ExpandableAvatarProps extends AvatarProps {}
 
-const expandableAvatarStyles = createStyles({
-  marginRight: system.space.x2,
-  flexShrink: 0,
+export const expandableAvatarStencil = createStencil({
+  base: {
```
mannycarrera4, line 12: since this renders an avatar, maybe we extend the `avatarStencil` as well

modules/labs-react/expandable/lib/ExpandableAvatar.tsx
```diff
@@ -3,29 +3,32 @@ import React from 'react';
...
+    marginRight: system.space.x2,
+    flexShrink: 0,
+  },
 });
 
 // When the component is created, it needs to be a button element to match AvatarProps.
 // Once Avatar becomes a `createComponent` we can default the element type to a `div`
 // and the types should be properly extracted
+// Setting altText prop to a default empty string for decorative purposes
```
mannycarrera4, line 21: should we set `aria-hidden` as well?

williamjstanton:
No, I don't believe aria-hidden is necessary. Setting the alt attribute on
 elements to an empty string is the best practice, and very effective
for hiding images. In the case of an <svg>, we already use ARIA role
"presentation" which strips the implied 'img' role in an <svg> element.
This gives us the same outcome (hiding the graphic), if we can assume there
isn't any <text> rendered in the <svg>.

modules/labs-react/expandable/lib/ExpandableAvatar.tsx
williamjstanton: No, I don't believe aria-hidden is necessary. Setting the alt attribute on
 elements to an empty string is the best practice, and very effective
for hiding images. In the case of an <svg>, we already use ARIA role
"presentation" which strips the implied 'img' role in an <svg> element.
This gives us the same outcome (hiding the graphic), if we can assume there
isn't any <text> rendered in the <svg>

modules/labs-react/expandable/lib/ExpandableAvatar.tsx
williamjstanton: I did try this, and then noticed that the `` didn't work in the Avatar examples. I didn't spend a lot of time digging in and trying to understand why.

modules/labs-react/expandable/lib/ExpandableAvatar.tsx
mannycarrera4: Gotcha, i just asked because we had an instance where a user can add a url, and cypress complains that you can't have an image with empty alt text, so we had to add aria-hidden

modules/labs-react/expandable/lib/ExpandableAvatar.tsx
mannycarrera4: I found a bug, we have to switch the order of the stencil if we move `false` above `true` it works as intended: https://github.com/Workday/canvas-kit/blob/master/modules/react/avatar/lib/Avatar.tsx#L221

My guess is true gets called first and never gets to false

modules/labs-react/expandable/lib/ExpandableAvatar.tsx
williamjstanton: "... can add a url" - what do you mean? Does this turn the image into a linked image... ?

mannycarrera4: Yeah if you pass a `url` to say an image that's hosted, it renders an `img` element which requires an `alt` attribute.

mannycarrera4:
Looks like the gap / padding seems off

## Your PR

image


## Master

image

modules/labs-react/expandable/lib/ExpandableIcon.tsx
```diff
@@ -30,51 +25,56 @@ export interface ExpandableIconProps extends Omit<ExtractProps<typeof SystemIcon
...
+    },
+    {
+      modifiers: {position: 'start', isExpanded: 'true'},
+      styles: {
+        marginRight: system.space.x2,
+        transform: 'rotate(180deg)',
+      },
+    },
+  ],
+});
```
josh-bagwell, line 64:
```suggestion
  compound: [
    {
      modifiers: {position: 'end', isExpanded: 'false'},
      styles: {
        marginInlineStart: 'auto',
        transform: 'rotate(180deg)',
        paddingInlineEnd: system.space.x3, // do I need these padding changes?
      },
    },
    {
      modifiers: {position: 'end', isExpanded: 'true'},
      styles: {
        marginInlineStart: 'auto',
        paddingInlineStart: system.space.x3, // do I need these padding changes?
      },
    },
    {
      modifiers: {position: 'start', isExpanded: 'false'},
      styles: {
        marginInlineEnd: system.space.x2,
        transform: 'rotate(90deg)',
      },
    },
    {
      modifiers: {position: 'start', isExpanded: 'true'},
      styles: {
        marginInlineEnd: system.space.x2,
        transform: 'rotate(180deg)',
      },
    },
  ],
});
```

modules/labs-react/expandable/lib/ExpandableAvatar.tsx
williamjstanton: Hey thanks! It looks like I got it!

modules/react/avatar/lib/Avatar.tsx
```diff
@@ -271,8 +271,7 @@ export const Avatar = createComponent('button')({
     return (
       <Element
         ref={ref}
-        aria-label={altText}
-        role={Element === 'button' ? 'button' : 'img'}
+        aria-label={Element === 'button' ? altText : undefined}
```
mannycarrera4, line 274: nice

modules/react/icon/lib/SystemIcon.tsx
```diff
@@ -191,6 +191,7 @@ export const systemIconStencil = createStencil({
       ),
     },
   }),
+  modifiers: {},
```
mannycarrera4, line 194: defaulting this to an empty object. If you extend a stencil and it's missing a property, typescript doesn't know to default to an empty object and crashes.

mannycarrera4 (approved):

PR #3141, opened 2025-02-20
fix(popup): Fix useReturnFocus to respect tabindex=-1 buttons

### LINKED ISSUE #3140: Return Focus doesn't respect buttons with a tabindex of -1
## 🐛 Bug Report

The `useReturnFocus` hook does not respect buttons with a `tabindex` of `-1`. These buttons are focusable via mouse/touch, but not via keyboard. If one of these buttons is clicked, focus should go to that button that was clicked and not the return focus target.

If you click on a button with a `tabindex` of `-1`, the focus ring will go to the button for a frame, then the focus is redirected to the popup's target.

### Error Output

Image

NicholasBoll:
## Summary

Separate the concept of focusability between keyboard and mouse. Buttons with a `tabindex` of `-1` are focusable via a pointer device, but not via a keyboard. A user clicking on a button that isn't focusable by keyboard is still a valid target for avoiding focus stealing.

Fixes #3140

Fixes: #1234 

## Release Category
Components

---

## For the Reviewer

Test

- [ ] Code
- [ ] Testing

## Testing Manually

Go to the `ReturnFocusTest` and open the popup and click on the tabindex button. The Cypress Test also does this.

### DIFF
--- a/modules/react/button/lib/BaseButton.tsx
+++ b/modules/react/button/lib/BaseButton.tsx
@@ -169,7 +169,7 @@ export const buttonStencil = createStencil({
       opacity: cssVar(buttonColorPropVars.default.opacity, cssVar(opacity, system.opacity.full)),
     },
     // Focus Styles
-    '&:focus-visible, &.focus': {
+    '&:focus, &.focus': {
       backgroundColor: cssVar(
         buttonColorPropVars.focus.background,
         cssVar(background, 'transparent')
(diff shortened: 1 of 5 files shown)


modules/react/button/lib/BaseButton.tsx
```diff
@@ -169,7 +169,7 @@ export const buttonStencil = createStencil({
       opacity: cssVar(buttonColorPropVars.default.opacity, cssVar(opacity, system.opacity.full)),
     },
     // Focus Styles
```
mannycarrera4, line 171: wow, really?

modules/react/button/lib/BaseButton.tsx
```diff
@@ -169,7 +169,7 @@ export const buttonStencil = createStencil({
...
-    '&:focus-visible, &.focus': {
+    '&:focus, &.focus': {
```
mannycarrera4, line 172: is this a breaking change? I thought we moved away from `:focus` to `focus-visibile`.

mannycarrera4 (approved):

Issue #3148, opened 2025-02-24
Add support for urls to our `Graphic` component to support SAS assets.


mannycarrera4:
## 🚀 Feature Proposal

Our `Graphic` component should support rendering `urls` so that we can provide support for assets loaded from SAS. In order to do this, we should update the `CanvasGraphics` type and the rendering of `Graphic`s to render an image where the `url` is passed.

PR #3149, opened 2025-02-24
feat: Add support for SAS URL to Graphic component

mannycarrera4:
## Summary

Fixes: https://github.com/Workday/canvas-kit/issues/3148

## Release Category
Components

### Release Note
We've updated the `Graphic` component to better support images and `url`s including changing the underlying dom structure  to remove the use of our `SVG` component and leverage the `img` api.

The following changes have been made:
- Deprecated `graphicStencil` which extended `svgStencil` since the component no longer renders an `svg`.
- Added the following props to better support images: `alt`, `srcset`, `sizes`.
- Changed from rendering `SVG` which was using `dangerouslySetInnerHTML` to use `img`.

The component should still be backwards compatible with old graphics while providing a better API for urls and images of different type.

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods


modules/react/icon/lib/Graphic.tsx
```diff
@@ -79,23 +84,56 @@ export const graphicStencil = createStencil({
...
+  },
+  base: ({width, height}) => ({
+    width,
+    height,
+    '& [data-part="graphic-img"]': {
+      maxWidth: '100%',
+      maxHeight: '100%',
+    },
+  }),
+});
```
mannycarrera4, line 100: I'm happy to come up with a different solution. I just didn't want the `img` to extend `svgStencil` because `width` and `height` didn't do anything. I created a new stencil to break apart those styles. Does that make sense?

modules/styling/lib/number2rem.ts
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ *
+ * Function that converts a number to a string. If the value is a number, it will convert it to a `rem` string.
+ */
+export function number2rem(value: number | string | undefined): string {
+  if (typeof value === 'number') {
+    return px2rem(value);
+  }
+  return value || '';
```
mannycarrera4, line 11: we do this check in a lot of places, so I just wrote a helper function to help with these scenarios:
```tsx
width: typeof width === 'number' ? px2rem(width) : width,
```

modules/react/icon/lib/Graphic.tsx
```diff
@@ -26,7 +26,17 @@ export interface GraphicStyles {
...
-export interface GraphicProps extends GraphicStyles, Pick<SvgProps, 'shouldMirror' | 'cs'> {
+export interface CanvasGraphic {
+  name: string;
+  type: CanvasIconTypes;
+  svg?: string;
+  filename: string;
+  category?: string;
+  tags?: string[];
+  url?: string;
+}
```
mannycarrera4, line 37: Rewrote the type interface to make URL and svg optional. cc @RayRedGoose

modules/react/icon/lib/Graphic.tsx
```diff
@@ -79,23 +89,62 @@ export const graphicStencil = createStencil({
...
+          >
+            <img data-part="graphic-img" src={src.url} alt="" />
+          </Element>
+        ) : typeof src.svg === 'string' && src.svg ? (
+          <Svg
+            type={CanvasIconTypes.Graphic}
+            as={Element}
+            ref={ref}
+            shouldMirror={shouldMirror}
+            src={src as OldCanvasGraphic}
```
mannycarrera4, line 136: Couldn't find away around not casting this to the old SVG since it requires `svg` as a property on `src`.

modules/react/icon/lib/Graphic.tsx
```diff
@@ -79,23 +89,62 @@ export const graphicStencil = createStencil({
...
+
 export const Graphic = createComponent('span')({
   displayName: 'Graphic',
-  Component: ({grow, width, height, ...elemProps}: GraphicProps, ref, Element) => {
+  Component: (
+    {grow, width, height, src, shouldMirror, ...elemProps}: GraphicProps,
+    ref,
+    Element
+  ) => {
+    console.log(src.svg);
```
alanbsmith, line 114:
```suggestion
```

modules/react/icon/lib/Graphic.tsx
```diff
@@ -79,23 +89,62 @@ export const graphicStencil = createStencil({
...
+              elemProps,
+              graphicStencil({
+                grow,
+                width: number2rem(width),
+                height: number2rem(height),
+              })
+            )}
+          />
+        ) : null}
+      </>
```
RayRedGoose, line 147: You don't really need `<></>` wrapper here, each step has only one element. Also, you can remove null, by making this:
```
typeof src.svg === 'string' && src.svg
->
typeof src.svg === 'string' && !!src.svg
```

modules/react/icon/stories/Icons.stories.tsx
```diff
@@ -117,7 +126,7 @@ export const GraphicStory = {
   component: Graphic,
   render: () => (
     <div className="story">
-      <Graphic src={graphicExample} />
+      <Graphic src={graphicExample} cs={{width: 40}} />
```
RayRedGoose, line 129: should it be `px2rem(40)`

modules/react/icon/lib/Graphic.tsx
```diff
@@ -79,23 +89,62 @@ export const graphicStencil = createStencil({
...
+            as={Element}
+            ref={ref}
+            shouldMirror={shouldMirror}
+            src={src as OldCanvasGraphic}
+            {...handleCsProp(
+              elemProps,
+              graphicStencil({
+                grow,
+                width: number2rem(width),
+                height: number2rem(height),
```
RayRedGoose, line 142: should we just change props and do not allow numbers?

modules/styling/lib/number2rem.ts
alanbsmith: Nit: `numberToRem` isn't really what's happening though, or that's not really the purpose of this abstraction. Why not make `px2rem` able to handle `string` values?

modules/react/icon/lib/Graphic.tsx
alanbsmith: I'm still seeing type issues with it. 🤔

alanbsmith (changes requested):
I made a first pass, but I need to look at this more tomorrow. I created a [StackBlitz](https://stackblitz.com/edit/graphics-component-prototype?file=src%2FGraphic.tsx,src%2FApp.tsx,src%2Fsearch-graphic.ts,src%2Findex.css,src%2Fmain.tsx) to test the component, and I'm getting errors with the url prop.

modules/react/icon/stories/Icons.stories.tsx
mannycarrera4: i was just testing that the CS prop works.  Will update!

modules/react/icon/lib/Graphic.tsx
mannycarrera4: `Graphic` only expects a `src` prop which is an object. I've updated the type to have an additional property of `url`.  My thought was that the graphic package would also export a object not just a URL. If that's not the case, I can add an additional `url` prop to Graphic to support the URL.

modules/styling/lib/px2rem.ts
```diff
@@ -1,11 +1,18 @@
...
  * // returns '0.125rem'
  * ```
  */
-export function px2rem(px: number, base = 16): string {
-  return `${px / base}rem`;
+export function px2rem(px: number | string | undefined, base = 16): string {
+  if (typeof px === 'number') {
+    return `${px / base}rem`;
+  } else if (typeof px === 'string') {
+    const value = parseFloat(px);
```
mannycarrera4, line 13: if the consumer passes something like `30ch` this is not the same, so I'm not sure if this is worth it, they have to pass a pixel value.

modules/styling/spec/px2rem.spec.ts
```diff
@@ -1,6 +1,6 @@
 import {px2rem} from '../lib/px2rem';
 
-describe('px2rem', () => {
+describe.only('px2rem', () => {
```
alanbsmith, line 3:
```suggestion
describe('px2rem', () => {
```

modules/styling/lib/px2rem.ts
NicholasBoll: I don't understand this change. `px2rem` is supposed to be run to convert a pixel number to a `rem`. If you accept a string, you'll have to forward that string.

Something like:
```ts
export function px2rem(px: number | string | undefined, base = 16): string {
  return typeof px === 'number' ? `${px / base}rem` : px ? px : ''
}
```

I don't know why `undefined` is supported. Returning an empty string vs `undefined` is different to a style block. In Emotion, `undefined` means "omit the property". I don't think that decisions should be made in this function.

Any changes to `px2rem` will require changes to https://github.com/Workday/canvas-kit/blob/master/modules/styling-transform/lib/utils/handlePx2Rem.ts.

You're overloading this function. This change is not needed for the only place I see it used in this PR.

modules/react/icon/lib/Graphic.tsx
```diff
@@ -79,23 +109,56 @@ export const graphicStencil = createStencil({
...
+      },
+    },
+  },
+});
+
 export const Graphic = createComponent('span')({
   displayName: 'Graphic',
-  Component: ({grow, width, height, ...elemProps}: GraphicProps, ref, Element) => {
+  Component: (
+    {grow = false, width, height, src, srcset, alt, sizes, ...elemProps}: GraphicProps,
```
NicholasBoll, line 139: `srcset` is not used anywhere. Is it supposed to be forwarded to the `img` element?

mannycarrera4: updated.

modules/styling/lib/px2rem.ts
mannycarrera4: fair

modules/react/icon/lib/Graphic.tsx
```diff
@@ -26,11 +25,39 @@ export interface GraphicStyles {
...
+  /**
+   * The width of the graphic
+   */
+  width?: number | string;
+  /**
+   * If true, expand the Graphic to fit its container. `grow` takes precedence over both `width` and `height`.
+   */
+  grow?: boolean;
+  /**
+   * The source of the graphic. If `svg` is provided, it will create a Base64-encoded ASCII string from a binary string (i.e., a string in which each character in the string is treated as a byte of binary data) and pass it to an image `src`. If `url` is provided, it will be rendered as an image via the `src` property..
```
josh-bagwell, line 42:
```suggestion
   * The source of the graphic. If `svg` is provided, it will create a Base64-encoded ASCII string from a binary string (i.e., a string in which each character in the string is treated as a byte of binary data) and pass it to an image `src`. If `url` is provided, it will be rendered as an image via the `src` property.
```

PR #3153, opened 2025-02-26
docs: Fix Firefox hard-fail on markdown-to-jsx bug

NicholasBoll:
## Summary

The `markdown-to-jsx` seems to have a bug in Firefox that causes improperly formatted HTML. This was causing an edge case that isn't supposed to be possible in our documentation system. Instead of Firefox erroring on proper parsing, we instead guard an access. Firefox will now show improperly formatted content rather than erroring out with no content. We will address the Firefox bug with our Vite upgrade where we drop `markdown-to-jsx` which should fix the issue.

## Release Category
Documentation

---

### DIFF
--- a/modules/docs/lib/MDXElements.tsx
+++ b/modules/docs/lib/MDXElements.tsx
@@ -25,7 +25,7 @@ export const MDX = createComponent('div')({
  */
 const Button = (props: any) => {
   const components = useMDXComponents();
-  if (props['data-symbol'] !== undefined) {
+  if (props['data-symbol'] !== undefined && props.children) {
     return (
       <code>
         <SymbolDialog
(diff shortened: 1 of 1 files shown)


modules/docs/lib/MDXElements.tsx
```diff
@@ -25,7 +25,7 @@ export const MDX = createComponent('div')({
  */
 const Button = (props: any) => {
   const components = useMDXComponents();
-  if (props['data-symbol'] !== undefined) {
+  if (props['data-symbol'] !== undefined && props.children) {
```
NicholasBoll, line 28: `markdown-to-jsx` will rewrite the following:
```jsx
<button data-symbol="" className="token symbol">createStyles</button>
```

Firefox:
```jsx
<button data-symbol="" className="token symbol"></button>createStyles&gt;/button>
```

Everything else:
```jsx
<button data-symbol="" className="token symbol">createStyles</button>
```

In Firefox, that means `props.children` is `undefined` which will fail a few lines down because we access a property on `props.children`. Instead of Firefox failing to render with an error message, we simply skip it. This results in improperly formatted content in Firefox, but that is better than failing to render anything on the page.

alanbsmith (approved):

PR #3160, opened 2025-02-28
fix: Support negative values

NicholasBoll:
## Summary

Add support for negative numbers in `createStyles` and `createStencil`

## Release Category
Styling

---


modules/styling-transform/lib/utils/parseNodeToStaticValue.ts
```diff
@@ -37,6 +37,15 @@ export function parseNodeToStaticValue(
     return parseFloat(node.text);
   }
 
+  // -12
```
josh-bagwell, line 40:
```suggestion
```

josh-bagwell (approved):

PR #3165, opened 2025-03-07
fix: Export collection types used in unexported models

NicholasBoll:
## Summary

Exporting types allows TypeScript definition files to use correct paths. We use `declarationMap` that includes source files which TypeScript can use to find non-exported types, but that also means a user's TypeScript compiler is reading our source files...

## Release Category
Infrastructure


package.json
```diff
@@ -4,11 +4,11 @@
...
-    "@babel/plugin-transform-modules-commonjs": "^7.2.0",
-    "@babel/plugin-transform-runtime": "^7.4.4",
-    "@babel/preset-env": "^7.8.0",
-    "@babel/preset-react": "^7.8.0",
-    "@babel/preset-typescript": "^7.1.0",
+    "@babel/plugin-transform-modules-commonjs": "^7.26.3",
+    "@babel/plugin-transform-runtime": "^7.26.3",
+    "@babel/preset-env": "^7.26.3",
+    "@babel/preset-react": "^7.26.3",
+    "@babel/preset-typescript": "^7.26.0",
```
josh-bagwell, line 11: Does this need to be: `7.26.3` or is it good at `7.26.0`?

josh-bagwell (approved):

PR #3166, opened 2025-03-07
docs: Add Snapshot serializers to ignore hashes

NicholasBoll:
## Summary

Adds Snapshot serializers to the testing documentation with ways to ignore hashes injected by our Static Styling package.

## Release Category
Documentation

---

### DIFF
--- a/modules/docs/mdx/TESTING.mdx
+++ b/modules/docs/mdx/TESTING.mdx
@@ -162,6 +162,31 @@ This will ensure snapshot tests have stable ids for each snapshot. It is still p
 changing if you add an additional component that uses id generation - subsequent ids will be
 different, but this will prevent snapshot tests that don't have any changes from showing diffs.
 
+The Canvas Kit Styling package uses CSS Variables and multiple class names with unique hashes.
+The following snapshot serializers handle styling. Setting unique seeds will not effect static style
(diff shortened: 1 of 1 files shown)


josh-bagwell (approved):

modules/docs/mdx/TESTING.mdx
```diff
@@ -162,6 +162,31 @@ This will ensure snapshot tests have stable ids for each snapshot. It is still p
...
+
+```ts
+// Handle `css-{hash}` class names
+const emotionClassnameRegex = /(css-[a-zA-Z0-9]{1,7})/g;
+expect.addSnapshotSerializer({
+  test: (val) => typeof val === "string" && emotionClassnameRegex.test(val),
+  print: (val) => {
+    return `"${val.replaceAll(emotionClassnameRegex, "css-xxxxx")}"`;
+  },
+});
```
mannycarrera4, line 178: How would you use these functions?

mannycarrera4 (approved):
Just for my own sanity, I would like to have an example of how to use these functions

modules/docs/mdx/TESTING.mdx
NicholasBoll: The `expect.addSnapshotSerializer` is the function call. It is used for test runners that have snapshots like Jest and Vitest.

Issue #2259, opened 2023-06-15
Support grouping for Menu in Main


jaclynjessup:
## 🚀 Feature Proposal

Support grouping within the Menu in Main

Tasks:
- [ ] Update the collection system to support group for menu items
- [ ] Add storybook example for menu grouping
- [ ] Update upgrade guide

## Motivation

To have feature parody with the Menu in Preview we should support grouping within the Menu in Main.

## Example

[Headers](https://workday.github.io/canvas-kit/?path=%2Fstory%2Fpreview-menu--headers) from the Menu in Preview

PR #3212, opened 2025-03-26
feat(menu): Add static grouping

NicholasBoll:
## Summary

Add a `Menu.Group` for menu grouping within the `Menu` component.

Implements: #2259 

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing

Menu with grouping


modules/react/collection/lib/ListBox.tsx
```diff
@@ -122,6 +122,7 @@ export const ListBox = createContainer('ul')({
         {...handleCsProp(
           {
             style: {
+              height,
```
mannycarrera4, line 125: was this for virtualization ?

NicholasBoll: No. If you set `height`, it doesn't work since it is removed from the prop list, but not added anywhere. It was a bug I noticed

modules/react/menu/lib/MenuGroup.tsx
```diff
@@ -0,0 +1,121 @@
...
+  displayName: 'Menu.Group',
+  modelHook: useMenuGroupModel,
+  elemPropsHook: useMenuGroup,
+  subComponents: {
+    Heading: MenuGroupHeading,
+  },
+})<MenuGroupProps>(({children, title, ...elemProps}, Element) => {
+  return (
+    <Element {...mergeStyles(elemProps)}>
+      {title ? <MenuGroupHeading>{title}</MenuGroupHeading> : null}
```
mannycarrera4, line 117: It looks like `MenuGroupHeading` is a sub component of `MenuGroup` so is the `title` prop still valid? 

If the title prop is the only way to add a heading I would either add a data-part attribute so people can style it, or add some css variables for `color` so people can style it. It looks like there's two different APIs here, one where you pass `title` to the `MenuGroup` and one where you can define a `Menu.Group.Heading`.

mannycarrera4: nvm

NicholasBoll: The JSDoc says you can do the following:

```tsx
// default
<Menu.Group title="Group One">
  <Menu.Item>Item One</Menu.Item>
</Menu.Group>

// direct access to the heading
<Menu.Group>
  <Menu.Group.Heading cs={...}>Group One</Menu.Group.Heading>
  <Menu.Item>Item One</Menu.Item>
</Menu.Group>
```

It's kind of a nice API. If you only care about text, you can give a string. If you want access to the underlying component, you have declare the component.

NicholasBoll: I don't know if `title` is the best prop name. It could be `heading` or `label`

mannycarrera4: title is fine, yeah I just saw the jsdoc. I guess we're rolled out "config" APIs for ease of use which I like, it's just not common so yeah I like it! Also `heading` makes sense since it's call `Menu.Group.Heading` but that's nit.

modules/react/menu/lib/MenuGroup.tsx
```diff
@@ -0,0 +1,121 @@
...
+  };
+});
+
+const MenuGroupHeading = createSubcomponent('div')({
+  modelHook: useMenuGroupModel,
+  elemPropsHook: useMenuGroupHeading,
+})(({children, ...elemProps}, Element) => {
+  return (
+    <OverflowTooltip>
+      <Element {...mergeStyles(elemProps, menuGroupHeadingStencil())}>{children}</Element>
```
mannycarrera4, line 102: Since this is a net new component, thoughts on just extending `CS` prop and using `handleCsProp` here?

NicholasBoll: I'm not sure. I like the idea of a common API for all components. It is annoying to use a component that doesn't allow the pattern you use in the app. I like the idea of a single codemod that changes the style:

```tsx
// before
<PrimaryButton padding={system.space.x4} margin={system.space.x2}>
// after
<PrimaryButton cs={{padding: system.space.x4, margin: system.space.x2}}>
```

We've generally used `handleCsProp` on components that render another component and `mergeStyles` on components that render elements. That's the closest to how style props work (style props are run on the base component).

modules/react/menu/lib/MenuGroup.tsx
NicholasBoll: I thought about `heading`, but it is the heading text. It doesn't replace the heading component. I think I got the `title` name from React Spectrum's [Section](https://react-spectrum.adobe.com/react-spectrum/Menu.html#sections) which are like groups. `label` could work. I don't want `headingText`. Something short and descriptive. We use `title` for Tooltips which is meant to have text. `title` implies text.

mannycarrera4 (approved):

josh-bagwell (approved):

PR #3258, opened 2025-04-16
feat: Use css transitions for sidepanel instead of animation

mannycarrera4:
## Summary

Use css `transition` instead of animation in SidePanel (Preview). 

## Release Category
Components

### Release Note
When we supported IE 11 we needed to use animation to support animation events. Since we dropped support for IE11, we can now use css transitions to animate the width of the SidePanel when it's expanding or collapsing.  

As part of this change, we also removed the `onAnimationStart` prop. Since moving to transitions, React doesn't support `onTransitionStart`, therefore the prop is no longer needed.

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods


modules/preview-react/side-panel/lib/SidePanel.tsx
```diff
@@ -192,70 +140,44 @@ export const SidePanel = createComponent('section')({
...
-      }
-
-      if (typeof onAnimationStart !== 'undefined') {
-        onAnimationStart(event);
+    const handleAnimationStart = () => {
+      if (state === 'collapsed' || state === 'collapsing') {
+        setState('expanding');
+      } else if (state === 'expanded' || state === 'expanding') {
+        setState('collapsing');
       }
```
RayRedGoose, line 158: Same here:
`setState(state.replace(/ed$/, 'ing'));`

modules/preview-react/side-panel/lib/SidePanel.tsx
```diff
@@ -192,70 +140,44 @@ export const SidePanel = createComponent('section')({
...
       <Element
         ref={ref}
-        onAnimationEnd={handleAnimationEnd}
-        onAnimationStart={handleAnimationStart}
+        onTransitionEnd={handleAnimationEnd}
         {...handleCsProp(elemProps, [
           panelStencil({
-            expanded: expanded ? 'true' : 'false',
-            touched: touched ? 'true' : 'false',
+            expanded: expanded,
```
RayRedGoose, line 167:
```suggestion
            expanded,
```

modules/preview-react/side-panel/lib/hooks.ts
```diff
@@ -3,6 +3,8 @@ import * as React from 'react';
 export const SidePanelContext = React.createContext({
   state: 'expanded',
   origin: 'left' as 'left' | 'right',
+  // eslint-disable-next-line no-empty-function
+  handleAnimationStart: () => {},
```
RayRedGoose, line 7: Will this shut down eslint:
`() => undefined`

modules/preview-react/side-panel/lib/SidePanel.tsx
mannycarrera4: Al though this is less code, I find it less clear to read. I prefer the explicit logic so that whomever jumps into this code again can read through the logic.

modules/preview-react/side-panel/lib/hooks.ts
mannycarrera4, line 7:
```suggestion
  handleAnimationStart: () => undefined,
```

modules/preview-react/side-panel/lib/SidePanel.tsx
NicholasBoll: That's a little too clever. I'd be okay with a little less syntax, but still clear in what's going on:

```tsx
      if (['collapsed', 'collapsing'].includes(state)) {
        setState('expanding');
      } else if (['expanded', 'expanding'].includes(state)) {
        setState('collapsing');
      }
```

modules/preview-react/side-panel/lib/SidePanel.tsx
```diff
@@ -192,70 +140,44 @@ export const SidePanel = createComponent('section')({
...
-          setState('collapsed');
-        }
-
-        if (event.animationName === sidepanelExpandedlKeyFrames) {
+    const handleAnimationEnd = (event: React.TransitionEvent<HTMLDivElement>) => {
+      if (event.propertyName === 'width') {
+        if (state === 'expanding') {
           setState('expanded');
+        } else if (state === 'collapsing') {
+          setState('collapsed');
```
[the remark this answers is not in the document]
NicholasBoll: I'm okay with the more verbose syntax. The structure is more clear on intent. This is a state machine and we're transitioning between valid states. The type of `state` is also a disjoint union of string literals. The `String.replace` function returns a `string` which doesn't fit into `'expanded' | 'expanding' | 'collapsed' | 'collapsing'` and we'd have to cast the result to keep TypeScript happy. That would be less code even with the type casting, but takes a bit more to understand what's happening.

NicholasBoll (approved):

PR #3269, opened 2025-04-28
fix(select): Remove excess props from `SelectInputProps`

NicholasBoll:
## Summary

Removing unintentional props from `SelectInputProps`. Any component that uses `create*` component utility functions and uses `ExtractProps` for the prop interface should use `never` as the second parameter. The `never` means "don't include the default HTML attributes in the prop interface". The `as` prop is used to determine the final HTML attributes. Forgetting to use the `never` means the prop interfaces can be incompatible unintentionally, causing TypeScript issues for developers doing things we say is supported.

## Release Category
Components

---

## Checklist

## For the Reviewer

### DIFF
--- a/modules/react/common/lib/utils/components.ts
+++ b/modules/react/common/lib/utils/components.ts
@@ -77,8 +77,6 @@ type ExtractHTMLAttributes<T extends React.DetailedHTMLProps<any, any>> =
  * - `ExtractProps<typeof Card, 'aside'>`: `CardProps & React.HTMLAttributes<HTMLElement>`
  * - `ExtractProps<typeof Card, never>`: `CardProps`
  *
- * **Note:** The `never` means "don't include the default HTML attributes in the prop interface". The `as` prop is used to determine the final HTML attributes. Forgetting to use the `never` means the prop interfaces can be incompatible unintentionally, causing TypeScript issues for developers doing things we say is supported.
- *
  * @template TComponent The component you wish to extract props from. Needs 'typeof` in front:
  * `typeof Card`
  * @template TElement An optional override of the element that will be used. Define this if you use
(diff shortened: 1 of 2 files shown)


modules/react/common/lib/utils/components.ts
```diff
@@ -77,6 +77,8 @@ type ExtractHTMLAttributes<T extends React.DetailedHTMLProps<any, any>> =
  * - `ExtractProps<typeof Card, 'aside'>`: `CardProps & React.HTMLAttributes<HTMLElement>`
  * - `ExtractProps<typeof Card, never>`: `CardProps`
  *
+ * **Note:** The `never` means "don't include the default HTML attributes in the prop interface". The `as` prop is used to determine the final HTML attributes. Forgetting to use the `never` means the prop interfaces can be incompatible unintentionally, causing TypeScript issues for developers doing things we say is supported.
+ *
```
mannycarrera4, line 81:
```suggestion
```

modules/react/select/lib/SelectInput.tsx
```diff
@@ -55,37 +55,26 @@ const hiddenSelectInputStencil = createStencil({
...
+        placeholder="Choose an option"
+        {...selectInputStencil.parts.visualInput}
+        {...elemProps}
+      />
+      <InputGroup.InnerEnd {...selectInputStencil.parts.caretContainer}>
+        <SystemIcon {...selectInputStencil.parts.caret} icon={caretDownSmallIcon} />
+      </InputGroup.InnerEnd>
+    </InputGroup>
+  );
+});
```
mannycarrera4, line 80: Looks like this was pulled from master which has these stencil changes, but in support we still have the old stencils.

mannycarrera4 (approved):

modules/react/select/lib/SelectInput.tsx
```diff
@@ -52,40 +52,34 @@ const hiddenSelectInputStencil = createStencil({
   },
 });
 
-export const SelectInput = createSubcomponent(TextInput)({
+export const SelectInput = createSubcomponent(TextInput.as('input'))({
```
mannycarrera4, line 55:
```suggestion
export const SelectInput = createSubcomponent(TextInput)({
```

PR #3271, opened 2025-04-29
ci: Pad canary build ids with zeros

NicholasBoll:
## Summary

We have issues with pre-ids in our canary builds where package managers sort alphabetically. This is an issue when we go from `999` to `1000`. When sorting alphabetically, "1000" is before "999" instead of after. If we prefix with a zero, we can handle up to 9999 before we have this issue. `0999` is properly sorted before `1000`

## Release Category
Infrastructure

---

### DIFF
--- a/utils/publish-canary.mjs
+++ b/utils/publish-canary.mjs
@@ -10,11 +10,13 @@ const exec = promisify(childProcess.exec);
 const {
   SLACK_WEBHOOK,
   GITHUB_REF,
+  GITHUB_RUN_NUMBER = '0',
   BUILD_URL = 'https://github.com/Workday/canvas-kit/actions',
 } = process.env;
 
 console.log('GITHUB_REF', GITHUB_REF);
 const branch = GITHUB_REF.replace('refs/heads/', '');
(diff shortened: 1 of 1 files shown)


alanbsmith (approved):

mannycarrera4 (approved):

utils/publish-canary.mjs
```diff
@@ -10,11 +10,13 @@ const exec = promisify(childProcess.exec);
 const {
   SLACK_WEBHOOK,
   GITHUB_REF,
+  GITHUB_RUN_NUMBER = '0',
   BUILD_URL = 'https://github.com/Workday/canvas-kit/actions',
 } = process.env;
 
 console.log('GITHUB_REF', GITHUB_REF);
 const branch = GITHUB_REF.replace('refs/heads/', '');
+const prefixedBuildNumber = GITHUB_RUN_NUMBER.padStart(4, '0');
```
mannycarrera4, line 19:
```suggestion
const prefixedBuildNumber = GITHUB_RUN_NUMBER.padStart(5, '0');
```

mannycarrera4: just to be sure

PR #3345, opened 2025-06-17
chore: Update chromatic action settings

RayRedGoose:
## Summary

Fixes: #3344

Updated inputs for chromatic.

## Release Category
Infrastructure

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -82,7 +82,7 @@ jobs:
       - uses: chromaui/action@main
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          projectToken: dlpro96xybh
           storybookBuildDir: docs
           exitOnceUploaded: false
           exitZeroOnChanges: true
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yml
```diff
@@ -82,7 +82,7 @@ jobs:
       - uses: chromaui/action@main
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          appCode: dlpro96xybh
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
```
NicholasBoll: I don't understand how this works. We already used to use a GH secret. The problem is pull requests from forks are not trusted, so GH Actions sends an empty string for any secret. This change will simply send an empty string to Chromatic.

The documentation isn't clear:
> ### Continuous Integration
> To set up CI, store your project token as the CHROMATIC_PROJECT_TOKEN environment variable or secret. The CLI automatically recognizes this variable, so you won’t have to provide it explicitly.

Here's the GHA documentation: https://www.chromatic.com/docs/github-actions/#forked-repositories

> ### Forked repositories
> GitHub secrets work at a repository level. Forked repositories will not have access to them. If you want to run Chromatic on cross-repository (forked) PRs, you’ll have to expose the projectToken by including it as plaintext in your chromatic.yml workflow file. Be aware that anyone with access to this file will be able to run Chromatic builds on your project, consuming your snapshot quota. You can reset the project token on the Manage > Configure screen at any time if you think it may have been compromised.

mannycarrera4:
```suggestion
          projectToken: dlpro96xybh
```

NicholasBoll (approved):

PR #3356, opened 2025-07-03
fix: Update styling transform to use deterministic hashes

NicholasBoll:
## Summary

The `styling-transform` module did not package properly and couldn't be used outside of CK source code.

- Updated `package.json` `files` to ensure all files are in the node package
- Updated build process to create commonjs and es6 builds so users don't need to do any special TypeScript processing
- Updated styling config to use a version-based seed so each version of Canvas Kit has a unique set of styles so they don't have style merge collisions

## Release Category
Infrastructure

---

### DIFF
--- a/modules/styling-transform/lib/styleTransform.ts
+++ b/modules/styling-transform/lib/styleTransform.ts
@@ -1,18 +1,18 @@
 /// <reference types="node" />
-import ts from 'typescript';
 import path from 'node:path';
+import ts from 'typescript';
 
 import {getVariablesFromFiles} from './utils/getCssVariables';
-import {handleCreateVars} from './utils/handleCreateVars';
-import {handleCreateStyles} from './utils/handleCreateStyles';
-import {handleCreateStencil} from './utils/handleCreateStencil';
 import {handleCalc} from './utils/handleCalc';
-import {handlePx2Rem} from './utils/handlePx2Rem';
+import {handleCreateStencil} from './utils/handleCreateStencil';
+import {handleCreateStyles} from './utils/handleCreateStyles';
+import {handleCreateVars} from './utils/handleCreateVars';
 import {handleCssVar} from './utils/handleCssVar';
-import {Config, NodeTransformer, ObjectTransform, TransformerContext} from './utils/types';
-import {handleKeyframes} from './utils/handleKeyframes';
 import {handleInjectGlobal} from './utils/handleInjectGlobal';
+import {handleKeyframes} from './utils/handleKeyframes';
 import {handleParentModifier} from './utils/handleParentModifier';
+import {handlePx2Rem} from './utils/handlePx2Rem';
+import {NodeTransformer, ObjectTransform, TransformerContext} from './utils/types';
 
 export type NestedStyleObject = {[key: string]: string | NestedStyleObject};
 
@@ -26,8 +26,6 @@ let extractedNames: TransformerContext['extractedNames'] = {};
 let styles: TransformerContext['styles'] = {};
 let cache: TransformerContext['cache'] = {};
 let loadedFallbacks = false;
-let configLoaded = false;
-let config: Config = {};
 
 /**
  * The reset is used in tests and should not be called normally.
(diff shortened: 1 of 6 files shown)


modules/styling-transform/lib/styleTransform.ts
```diff
@@ -161,22 +146,26 @@ export function transform(
...
-    }, undefined as ts.Node | void);
+    return transformers.reduce(
+      (result, transformer) => {
+        if (result) {
+          return result;
+        }
+        return transformer(node, context);
+      },
+      undefined as ts.Node | void
+    );
```
NicholasBoll: Prettier reformatting

modules/styling-transform/lib/styleTransform.ts
```diff
@@ -1,18 +1,18 @@
...
+import {handleCreateStyles} from './utils/handleCreateStyles';
+import {handleCreateVars} from './utils/handleCreateVars';
 import {handleCssVar} from './utils/handleCssVar';
-import {Config, NodeTransformer, ObjectTransform, TransformerContext} from './utils/types';
-import {handleKeyframes} from './utils/handleKeyframes';
 import {handleInjectGlobal} from './utils/handleInjectGlobal';
+import {handleKeyframes} from './utils/handleKeyframes';
 import {handleParentModifier} from './utils/handleParentModifier';
+import {handlePx2Rem} from './utils/handlePx2Rem';
+import {NodeTransformer, ObjectTransform, TransformerContext} from './utils/types';
```
NicholasBoll, line 15: Prettier reformatting

modules/styling-transform/lib/utils/createStyleObjectNode.ts
```diff
@@ -64,10 +65,9 @@ export function serializeStyles(
 ) {
   const {getFileName, styles, cache, names, extractedNames} = context;
   const fileName = getFileName(node.getSourceFile().fileName);
-  const hash = generateUniqueId(); //getHash(node, context);
-  const serialized = {...serializedStylesEmotion([input]), name: hash} as ReturnType<
-    typeof serializedStylesEmotion
-  >;
+  const serialized = serializedStylesEmotion([input]);
+  const hash = getHash(node, context);
+  serialized.name = hash;
```
NicholasBoll, line 70: Instead of using a non-deterministic auto-incremented ID, we use a deterministic hash based on the AST node (start and end character index)

modules/styling-transform/lib/utils/createStyleObjectNode.ts
```diff
@@ -92,3 +92,14 @@ export function serializeStyles(
...
+
+/**
+ * Generate a deterministic hash that is unique to the seed +
+ */
+export function getHash(node: ts.Node, context: TransformerContext) {
+  const content = context.seed + node.getStart() + node.getEnd();
+  // Convert content to a sha256 hash and then convert to a base36 number for better compression
+  return Number(
+    `0x${crypto.createHash('sha256').update(content).digest('hex').slice(0, 7)}`
+  ).toString(36);
```
NicholasBoll, line 104: We assuming the start/end character index is enough to be both deterministic, yet unique.

There is a non-zero chance that a style from two separate files could have a style block that contain the same start and end character indices. This would result in a bug that is almost impossible to track down - if you add a console log or anything to the file that produced a duplicate hash above the code in question, the hash would change. Adding the filename to the hash would prevent this from happening, but also comes with the problem of determining an environment-agnostic repeatable filename extractor. If we could know the root folder, we could extract the relative file path. Just adding this comment to my future self or anyone else for the small chance that this is actually a problem and I can say "yeah, I thought that was a possibility, but highly improbable"...

jamesfan: What do you think about adding the content of this PR comment to the source code itself? That way, someone digging through the code at least has a chance to see the warning about adding `console.log` or any other debugging code to the file which produced the duplicate hash.

NicholasBoll: I'm not sure where to add that comment. Would someone track it down to here? That might be something to put in the docs, but it only will happen when running the transform.

modules/styling-transform/lib/styleTransform.ts
```diff
@@ -161,22 +146,26 @@ export function transform(
...
-        return result;
-      }
-      return transformer(node, context);
-    }, undefined as ts.Node | void);
+    return transformers.reduce(
+      (result, transformer) => {
+        if (result) {
+          return result;
+        }
+        return transformer(node, context);
```
RayRedGoose:
```suggestion
        return result || transformer(node, context);
```

modules/styling-transform/lib/webpack-loader.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const printer = ts.createPrinter(program.getCompilerOptions());
+
+  const transformed = printer.printFile(
+    ts
+      .transform(sourceFile, transformers, program.getCompilerOptions())
+      .transformed.find(s => s.fileName === id) || sourceFile
+  );
+
+  return postTransform ? postTransform(transformed, id) || transformed : transformed;
```
RayRedGoose: Can it be like below?

```
return (postTransform && postTransform(transformed, id)) || transformed;
```

modules/styling-transform/lib/utils/createStyleObjectNode.ts
NicholasBoll: I added this to the styling documentation

modules/styling-transform/lib/webpack-loader.ts
NicholasBoll: Could be
```ts
return postTransform?.(transformed, id) || transformed;
```

Issue #3451, opened 2025-08-15
chore: Add component tracking


alanbsmith:
## 💡 Suggestion

Add data attributes to our components for client-side analytics.

## Motivation

We'd like to track how our components are used in our products.

## Example

```
data-uxi-canvas-component-type="PrimaryButton"
data-uxi-canvas-kit-version="13.1.0"
```

PR #3452, opened 2025-08-15
chore: Add component tracking tags

alanbsmith:
## Summary

Fixes: #3451 

This PR adds component tagging for analytics.

## Release Category

Infrastructure

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods

### DIFF
--- a/modules/react/button/lib/parts/ButtonLabelIcon.tsx
+++ b/modules/react/button/lib/parts/ButtonLabelIcon.tsx
@@ -26,7 +26,6 @@ const iconSizes: Record<ButtonSizes, number> = {
 };
 
 export const ButtonLabelIcon = createComponent('span')({
-  displayName: 'ButtonLabelIconNew',
   Component: (
     {icon, size = 'medium', shouldMirrorIcon = false, ...elemProps}: ButtonLabelIconProps,
     ref,
--- a/modules/react/common/lib/utils/insights.ts
+++ b/modules/react/common/lib/utils/insights.ts
@@ -1,0 +1,18 @@
+import {version} from '@workday/canvas-kit-react/version';
+import {slugify} from '@workday/canvas-kit-styling';
+
+/**
+ * Functionality in this file is only for our own internal analytics and is not intended for external consumers.
+ */
+const versionTag = process.env.NODE_ENV === 'production' ? version : version.split('.')[0];
+
+/**
+ * This function returns data attributes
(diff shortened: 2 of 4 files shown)


modules/react/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.json",
   "exclude": ["node_modules", "ts-tmp", "dist", "**/spec", "**/stories"],
   "compilerOptions": {
+    "resolveJsonModule": true,
```
alanbsmith: Needed to pull in the `package.json` to access the version string.

modules/react/common/lib/utils/insights.ts
```diff
@@ -0,0 +1,27 @@
+import {version} from '../../../package.json';
```
NicholasBoll: We probably need to add the version string somewhere so people can easily figure out what version they are using, but this won't work. This will work for Storybook which uses Webpack which is configured to understand `*.json` imports, but we build out packages and publish on `npm` using the TypeScript compiler which translates. The CommonJS version will look something like:

```js
const {version} = require('../../../package.json');
```

This will move the requirement to understand how to import a JSON file to the consumer. Usually `node_modules` folders are excluded for special processing (`loaders` in Webpack. It may work by default in Vite... Not sure).

To do a proper version string, we need a line somewhere like:
```ts
export version = "%VERSION%";
```

From there, we need both a loader to replace the `%VERSION%` string and a build step that changes the text either before or after the TS compiler.

modules/react/button/lib/parts/ButtonLabelIcon.tsx
```diff
@@ -26,7 +26,6 @@ const iconSizes: Record<ButtonSizes, number> = {
 };
 
 export const ButtonLabelIcon = createComponent('span')({
-  displayName: 'ButtonLabelIconNew',
```
alanbsmith, line 29: Removing this displayName as we don't really need it, and it's messing up some component tracking.

modules/react/common/lib/utils/insights.ts
```diff
@@ -0,0 +1,29 @@
+/// <reference types="@types/node" />
```
mannycarrera4, line 1: @alanbsmith this should fix it. I'm not sure why we need it in some files and we don't need it in others, that is a mystery to me. But this should tell the compiler that we're using node.

PR #3477, opened 2025-08-28
feat: Add isElementRTL and deprecate useIsRTL

NicholasBoll:
## Summary

Deprecates the `useIsRTL` function. `useIsRTL` gets the RTL from the Emotion theme which will be deprecated. Use [CSS logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) for styles, `isElementRTL` for events, and `shouldMirrorInRTL` for icons.

## Release Category
Infrastructure

### BREAKING CHANGES
`keyboardEventToCursorEvents` no longer takes an `isRTL` input. It will instead get the content direction from the `event` that is passed in. You will need to remove this third parameter.

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods

### DIFF
--- a/.storybook/routes.js
+++ b/.storybook/routes.js
@@ -42,6 +42,7 @@ const routes = {
   '/components/text/text/': 'components-text-text--docs',
   '/components/text/title/': 'components-text-title--docs',
   '/examples/layout/': 'examples-layouts--docs',
+  '/get-started/for-developers/theming/overview/': 'features-theming-overview--docs',
   '/getting-started/for-developers/contributing/': 'guides-contributing--docs',
   '/getting-started/for-developers/resources/api-pattern-guidelines/':
     'guides-api-pattern-guidelines--docs',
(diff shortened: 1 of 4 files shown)


modules/react/common/lib/utils/isElementRTL.ts
```diff
@@ -0,0 +1,27 @@
...
+ *       //"advance" in RTL. Previous is to the right of the current in RTL.
+ *       model.events.goToPreviousRow();
+ *     } else {
+ *       model.events.goToNextRow();
+ *     }
+ *   }
+ * ```
+ */
+export const isElementRTL = (element: Element): boolean => {
+  const direction = getComputedStyle(element).direction;
```
mannycarrera4, line 25: should we have a fail safe checking for `window` here. I know we've had issues with server side  rendering where things like `getComputedStyle` fail

NicholasBoll: This is intended to be used in events which don't run on servers during server-side rendering. It should be safe to assume `window` exists in event handlers.

mannycarrera4 (approved):

modules/react/button/lib/BaseButton.tsx
```diff
@@ -131,6 +126,7 @@ export const buttonStencil = createStencil({
     lineHeight: 'normal',
     letterSpacing: '0.015rem',
     fontWeight: system.fontWeight.bold,
+    test: system.color.text.default,
```
NicholasBoll:
```suggestion
```

Issue #3538, opened 2025-10-14
Create fallback for input border radius CSS token


mannycarrera4:
## 🐛 Bug Report

When going from v2 - v3, we used a new token for our input border radius. This breaks consumers. We should add a fallback to ensure a smoother transition during the brand refresh.

PR #3539, opened 2025-10-14
fix: Add fallback to our inputs for border radius

mannycarrera4:
## Summary
Fixes: https://github.com/Workday/canvas-kit/issues/3538

## Release Category
Components

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods


modules/react/text-input/lib/TextInput.tsx
```diff
@@ -23,7 +23,8 @@ export const textInputStencil = createStencil({
     display: 'block',
     border: `${px2rem(1)} solid ${cssVar(system.color.border.input.default)}`,
     backgroundColor: system.color.bg.default,
-    borderRadius: system.shape.x1Half,
+    //TODO: Replace string with new token in v15/v4
+    borderRadius: cssVar('--cnvs-sys-shape-m', system.shape.x1Half),
```
mannycarrera4:
```suggestion
    borderRadius: cssVar('--cnvs-sys-shape-sm', system.shape.x1Half),
```

modules/react/common/lib/forwardFitTokens.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * This file is NOT INTENDED to be used by consumers, only internally by Canvas Kit.
+ */
+export const forwardFitTokens = {
+  system: {
+    shape: {
+      //TODO: in v4 Tokens and v15, we should replace this forward fit token with the actual token value
+      sm: '--cnvs-sys-shape-sm', // this should be set to 8px
```
alanbsmith: This is a non-blocking question, and more of a design decision I'm trying to understand.

Why did you choose to do it like the above instead of like this below?

```ts
sm: cssVar(forwardFitTokens.system.shape.sm, system.shape.x1Half),
```

I don't have a preference, but I was wondering if you do.

mannycarrera4: ahh I can do that, i think in my head i can see the style property and then know that it is forward fitted, no preference

alanbsmith: That would make sense too. I don't think there's a right or wrong here. The way I wrote it above prevents someone from accidentally putting a different default value for the token.

Issue #3544, opened 2025-10-17
fix: Update spec tables to not extend past width of viewport


josh-bagwell:
## 🐛 Bug Report

Spec tables extend past the width of the viewport. This is because of some code blocks and the container not being within the `<Table.Cell/>` element.

## Expected Behavior

The code block should be within the `<Table.Cell/>` container.

mannycarrera4:
fixed in #3545

PR #3545, opened 2025-10-17
docs: Updated the spec tables and examples to wrap

josh-bagwell:
## Summary

Fixes: #3544  

This PR fixes an issue with the spec tables overlapping on smaller viewports.

## Release Category
Documentation

---

`/modules/docs/lib/MDXElements.tsx`
`/modules/docs/lib/Value.tsx`

### DIFF
--- a/modules/docs/lib/ExampleCodeBlock.tsx
+++ b/modules/docs/lib/ExampleCodeBlock.tsx
@@ -25,7 +25,7 @@ const cardStencil = createStencil({
       boxShadow: system.depth[1],
       borderRadius: system.shape.x1,
       position: 'relative',
-      overflow: 'auto',
+      overflow: 'auto', // This allows for the entire ExampleCodeBlock to scroll on smaller viewports
     },
     '[data-part="code-block"]': {
       display: 'none',


mannycarrera4:
Instead of updating each button story to have `flexWrap` maybe we should try making the codeblock scrollable. https://github.com/Workday/canvas-kit/blob/master/modules/docs/lib/ExampleCodeBlock.tsx#L28 try adding overflow scroll here.

There's gonna be more stories where the content overflow the container, not just buttons. Having the codeblock scrollable is fine.

But test it out!

modules/docs/lib/ExampleCodeBlock.tsx
```diff
@@ -122,7 +123,8 @@ export const ExampleCodeBlock = ({code}: any) => {
   return (
     <div {...cardStencil({opened: isCodeDisplayed})}>
       <Card data-part="example-block" className="sb-unstyled">
-        <Card.Body>
+        {/* This allows for the div to scroll on smaller viewports while not allowing the components to overflow over the container */}
+        <Card.Body cs={{overflow: 'auto'}}>
```
mannycarrera4, line 127: I've been trying to move styles to a class, you can add it to the `cardStencil` and use `data-part`. I know cs is convenient but until we have a processor that cleans this up, let's keep the styles in one place.

mannycarrera4 (approved):

modules/docs/lib/ExampleCodeBlock.tsx
josh-bagwell: Updated.

PR #3655, opened 2025-12-22
fix: Add missing heading component and update to remove aria labels on toggle button

williamjstanton:
## Summary

Reviewing the examples for SidePanel, The `<Heading>` component depended on properly using the model's `id` and also needed logic for hiding it while SidePanel is collapsed. I thought this would be a good candidate for a new sub-component where we could build some of that into the compound component. 

## Release Category
Components

---

## Checklist

## For the Reviewer

- [X] Code
- [X] Documentation
- [ ] Testing
- [ ] Codemods

## Testing Manually

1. I want to validate the `SidePanel.Heading` element is still rendered in the DOM, even when in collapsed state
2. I want to make sure the `aria-labelledby` reference on the `SidePanel` `section` element is still properly naming the `section`
3. I want to make sure screen readers can still detect the `SidePanel` landmark region while it is collapsed, and can still announce the `SidePanel`'s heading even though it is hidden. 

This is a net-new subcomponent for `SidePanel` and **should** maintain same styles.


modules/labs-react/side-panel/lib/SidePanelToggleButton.tsx
```diff
@@ -139,8 +137,8 @@ export const SidePanelToggleButton = createSubcomponent('button')({
     {
       variant = undefined,
       icon = transformationImportIcon,
-      tooltipTextExpand = 'Expand',
-      tooltipTextCollapse = 'Collapse',
+      tooltipTextExpand = 'Expand View',
```
williamjstanton, line 142: How are translations going to be supported for these Tooltip strings?

mannycarrera4: We should remove these default values really. But teams can pass the prop value, we can set them as empty strings and update the examples so showcase what teams should pass

modules/labs-react/side-panel/stories/SidePanel.mdx
```diff
@@ -132,13 +137,12 @@ logic or styling for bidirectional support.
...
-for screen readers.
+`SidePanel`'s `<section>` element container should always have an accessible name to help screen
+reader users understand the purpose of the panel. For this reason, we recommend using the
+`SidePanel.Heading` component and setting the `hidden` prop to `true`. This will visually hide the
+heading while keeping it accessible to screen readers.
 
-<ExampleCodeBlock code={HiddenName} />
+<ExampleCodeBlock code={Heading} />
 
 ### Alternate Variant
```
williamjstanton, line 148: I don't think the Overlay use case is covered in the Examples section of Storybook anymore. Should we update this example to demo the overlay (and depth 6) use case?

mannycarrera4 (approved):

Issue #3812, opened 2026-03-04
Restore the Scorecard Actions workflow by upgrading actions/upload-artifact


desrosj:
## 🐛 Bug Report

The Scorecard GitHub Actions workflow is currently broken due to the use of an outdated version of the actions/upload-artifact action.

Version 4 is a significant rewrite with up to 90% performance increase at worst. Because of the required changes to support this degree of improvement, GitHub has [blocked usage of `v3` entirely as of January 30, 2025](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). 

## To Reproduce

Not steps required to reproduce. Just observe the [list of runs for this workflow](https://github.com/Workday/canvas-kit/actions/workflows/scorecard.yml).

## Expected Behavior

The workflow should succeed and an artifact should be uploaded.

## Actual Results

The workflow fails before it can even perform any analysis.

### Error Output

Paste the error output here

```bash
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: 3cea5372237819ed00197afe530f5a7ea3e805c8`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
```

PR #3813, opened 2026-03-04
ci: Update `actions/upload-artifact` to `7.0.0`.

desrosj:
## Summary

Fixes: #3812.

The Scorecard workflow is currently broken because `v3` of the `actions/upload-artifact` action is deprecated and disabled.

## Release Category
Infrastructure

---

## Checklist

## For the Reviewer

This updates the third-party action to the latest version. There are [some breaking changes](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes) in `v4`, but this workflow should not be impacted by any of them. The [breaking changes in `v6`](https://github.com/actions/upload-artifact?tab=readme-ov-file#v6---whats-new) are also not relevant here because they only impact self-hosted runners.

Since this workflow only runs on `push` and `schedule` events, it's difficult to test before merging. 

This workflow only runs on `push` and `schedule` events, so the [action's documentation](https://github.com/actions/upload-artifact) is likely best.

## Thank You Gif (optional)

[A black number 7 on a white background](https://media1.giphy.com/media/v1.Y2lkPTc5MGI3NjExdDRnMXVxdTJ4MGIxa2E0eWNsd2lkMW0zMmU1eTVsZ2I0a3ZmcjZwMiZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/3o7TKpGT1RLRHlP7s4/giphy.gif)

### DIFF
--- a/.github/workflows/lint-pull-request.yml
+++ b/.github/workflows/lint-pull-request.yml
@@ -15,7 +15,7 @@ jobs:
 
       - name: Get Package Scopes
         id: scope-vars
-        run: echo "::set-output name=scopes::$(node utils/get-package-names.js)"
+        run: echo "scopes=$(node utils/get-package-names.js)" >> "$GITHUB_OUTPUT"
 
       # Please look up the latest version from
       # https://github.com/amannn/action-semantic-pull-request/releases


.github/workflows/scorecard.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # v3.1.0
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
```
RayRedGoose, line 62:
```suggestion
        uses: actions/upload-artifact@v7
```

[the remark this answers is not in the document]
desrosj: I just double checked this and it is incorrect. That hash does not seem to exist on the action's repository.

desrosj: I do not recommend using version qualifiers when using third-party actions as it goes against [GitHub's Secure Use recommendations](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions).

Full-length commit SHA values is the only way to make the action immutable. If a bad actor gains control to an action's repository, they can force overwrite a given tag, branch, etc. and the malicious code will automatically be used in any dependent workflows when they run next ([two](https://www.stepsecurity.io/blog/reviewdog-github-actions-are-compromised) [examples](https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/) of this from 2025). Pinning to a hash is the only way to protect against that.

**Note:** GitHub was working on the concept of Immutable Releases and [did announce a public preview](https://github.blog/changelog/2025-08-26-releases-now-support-immutability-in-public-preview/), but they [removed this from their public roadmap](https://github.com/github/roadmap/issues/592#issuecomment-3643300839) late last year and it's not really clear exactly how they work in practice as they feature currently stands.

That said, specifying a tag is easier and more convenient. And the guidelines do state that specifying a tag should be fine so long as you trust the creator. If this is the case here, happy to merge this! I personally take the cautious route and assume that no external code can be trusted, just in case.

When I looked at the various workflows in the repository, I noticed that the use of pinned hashes vs. versions is pretty mixed. So I chose to go with the pinned hash instead. The inline comment at the end helps give a human-readable version reference that each hash represents. It's also easy enough to add a `dependabot.yml` file configured to monitor for updates to the actions used within a repository. This helps quite a bit with the burden of needing to update on each release (or not if the given update has nothing of tangible value to a given repository), and it's even smart enough to update the inline comments denoting the hash's version.

alanbsmith: Yeah, that's a good point. I agree. We should keep the sha1.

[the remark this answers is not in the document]
alanbsmith: Yeah, I don't know where Copilot is pulling that from. I see the sha1 here:
https://github.com/actions/upload-artifact/releases/tag/v7.0.0

`bbbca2ddaa5d8feaa63e36b76fdaad77386f024f` is the full commit for that tag.

PR #3938, opened 2026-05-14
chore: Add exports field in package json

### LINKED ISSUE #3937: chore: Add exports field to package json
## 🚀 Feature Proposal

This is to add a `exports` field to our `package.json` file for each package (i.e. `react`, `preview`, `labs`, etc.)

josh-bagwell:
## Summary

Fixes: #3937  

## Release Category
Components

---

`/modules/react/package.json`


josh-bagwell:
Files have been updated to reflect the changes of adding an `exports` field to the `package.json` file for each package.

Following updates:
yarn build — handleFocusRing.ts now imports focusRing via a relative path instead of @workday/canvas-kit-react/common, so the Node.js CJS loader no longer needs the built dist to exist during compilation.

yarn test (canvas-kit-react) — vitest.config.ts aliases for canvas-kit-react now resolve to absolute file paths (modules/react/), bypassing the exports field check entirely.

yarn test (canvas-kit-preview-react) — Same alias fix applied for canvas-kit-preview-react, and its exports field was corrected to only list the components that actually live in that package.

josh-bagwell:
I'm going to throw this on hold for now and close it.

This could come back up and would like this PR to be here just in case.

Issue #3980, opened 2026-06-12
feat: Sana Canvas Tabs


jaclynjessup:
🚀 Feature Proposal

Update the Tabs component to align with v16 token and style changes, including the addition of a new item variant.

**Components in this ticket:**
- Tabs

---

## Prerequisites

- [ ] Design review: Confirm Figma component is marked ready for dev
- [ ] A11y review completed for focus ring / selected state overlap (see below)
- [ ] Design decision confirmed on outlined vs. filled (ghost) variant guidance

---

## Token Updates Required

| Category | Updated? |
|----------|----------|
| Space    | ✅ Yes |
| Size     | ✅ Yes |
| Shape    | ✅ Yes |
| Color    | ✅ Yes |
| Depth    | No |
| Theme    | No |

---

## Visual Changes

- Token changes across space, size, shape, and color
- Two `Tabs.Item` button-style variants currently unsupported — need to be added

---

## API Changes

- **Addition of `Tabs.Item` variant** to support the two button-style tab variants in Sana Figma.

---

## ⚠️ Accessibility Concern

**Overlapping focus ring & selected state border:** The selected state uses a bottom border that visually conflicts with the focus ring.

- Use `:focus-visible` (not `:focus`) to reduce false positives
- **Must be resolved before shipping.**

---

## Effort & Timeline

- **Priority:** High
- **Effort:** Large (token changes, addition of `Tabs.Item` variant)

---

## Dependencies

- **No blockers**
- **A11y review required** before shipping

---

## Acceptance Criteria

- [ ] Token updates applied (space, size, shape, color)
- [ ] New `Tabs.Item` variant added
- [ ] Focus ring / selected state border overlap resolved with `:focus-visible`
- [ ] A11y review signed off
- [ ] Usage guidance documented in Storybook
- [ ] Visual regression tests pass

RayRedGoose:
blocked by a11y

RayRedGoose:
Going to preview/labs

mannycarrera4:
Consider adding this accessibility feature here: https://github.com/Workday/canvas-kit/pull/3790

Issue #3558, opened 2025-10-23
Overflow Tabs: The "More" button is not included with the tabs in the tab list container


williamjstanton:
## 🐛 Bug Report

- DOM inspection: button elements are not allowed within tab list containers according to WAI-ARIA specification
- Users cannot access the "More" button using the Left/Right arrow keys like the other tabs in the list; users are required to press Tab key; this is inconsistent and not predictable. 

## To Reproduce

Test case: https://workday.github.io/canvas-kit/?path=/story/components-containers-tabs--overflow-tabs 
Click "360px" width button

Keyboard: 
1. Use the Tab key to focus the list of tabs, 
2. Use the Right arrow key to navigate tabs, 
3. Issue: Only the 2 tab items can be accessed with the arrows

Screen reader experience: 
1. Turn on VoiceOver (macOS) Cmd + Power btn 3x
2. Use Safari or Chrome
3. Use Tab key to focus the Tab list
4. Issue: Observe VoiceOver output is describing a tab list of 2 items, while the visual representation shows 3 tabs in the list. 

## Expected Behavior

1. Keyboard: Users should be able to more easily access the overflow "More" tab with the Right / left arrows, 
2. Screen reader: Should include the overflow "More" tab in the total number of tabs on the list. 

## Actual Results

A clear and concise description of what actually happens
1. Users need to learn to use the Tab key to focus the "More" overflow button
2. Screen readers don't count the "More" overflow button in the total count of tabs; users may not expect there to be more hidden page tabs. 

## Browser (if applicable)

If this was a visual bug, what browser did you spot it on?

All browsers

## Link to repl or repo (highly encouraged)

Please provide a reference to a repo or branch to help us reproduce this bug.

Storybook > Components > Containers > Tabs > Overflow 
https://workday.github.io/canvas-kit/?path=/story/components-containers-tabs--overflow-tabs 

Include a screenshot of a visual bug

<img width="1697" height="849" alt="Image" src="https://github.com/user-attachments/assets/1c5cccab-0882-4e34-9494-8d660d01dbba" />

## Goals & Recommendations

- Append the "More" overflow button to the collection list when needed so it may be accessed with Right/Left arrows. 
- Make sure overflow button has an ARIA `role="tab"` and `aria-haspopup="true"` and `aria-expanded={true | false}` properties
- Enter and Space keys can be used to open the attached menu on "More"

sheelah:
Repopening as a post-v16 todo, for our preview `Tabs` component.

PR #4062, opened 2026-07-15
feat: Add new Tabs component

RayRedGoose:
## Summary

Fixes: #3980

New Tabs component has been added to preview.

## Release Category
Components

### Release Note
New Tabs component has been added to preview.

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods

## Summary by CodeRabbit

* **New Features**
  * Introduced/expanded a React Tabs compound component with accessible roles, selectable panels, RTL support, disabled and icon-based tabs, dynamic add/delete, and single-panel usage.
  * Added overflow handling with an overflow button and menu popper for hidden tabs.
* **Documentation**
  * Added new Storybook/MDX docs and multiple usage examples, including API documentation updates and licensing/README content.
* **Tests**
  * Added automated specs and an SSR smoke test validating tab roles, panel rendering, and selection behavior.


modules/preview-react/tabs/lib/TabsItem.tsx
```diff
@@ -0,0 +1,264 @@
...
+    '&:disabled, &.disabled, &[aria-disabled]': {
+      opacity: system.opacity.disabled,
+      '&:hover': {
+        cursor: 'auto',
+        backgroundColor: system.legacy.color.surface.transparent,
+        [systemIconStencil.vars.color]: system.color.fg.default,
+      },
+    },
+
+    '&[aria-selected=true]': {
```
williamjstanton, line 141: Can we add a non-color selected indicator for the filled variant that survives Windows High Contrast / forced-colors themes? Right now the selected state appears to rely on background color only.

(Regression from our Tabs in main because they use a bottom-border on selected state.)

This is a red flag for me. The color difference between selected and unselected states does not meet contrast guidelines and much too difficult to see. CC: @purvas12

modules/preview-react/tabs/lib/TabsList.tsx
```diff
@@ -0,0 +1,181 @@
...
+      <Element
+        {...mergeStyles(
+          elemProps,
+          tabsListStencil({
+            modality: modality,
+          })
+        )}
+      >
+        {useListRenderItems(model, children)}
+        {overflowButton}
```
williamjstanton, line 177: Since this renders a non-tab menu button directly inside role="tablist", can we confirm the ARIA pattern here or document the exception? Existing tests suppress aria-required-children for this structure.”

The changes that I'd really like to see are: 
1. Include this overflow button into the collection system so the roving tabindex right / left arrow key interaction can work on this "More" tab, 
2. Set required `role="tab"` 
3. Set `aria-haspopup="menu"`

I worked on these changes here in my branch: 
`williamjstanton/canvas-kit/tree/william-3558-overflow-tabs-fixes`

Issue: https://github.com/Workday/canvas-kit/issues/3558

modules/preview-react/tabs/stories/Tabs.mdx
```diff
@@ -0,0 +1,134 @@
...
+and provides direct access to its `state` and `events` outside of the `Tabs` component.
+
+In this example, we set up external observation of the model state and create an external button to
+trigger an event to change the active tab.
+
+<ExampleCodeBlock code={HoistedModel} />
+
+### Named Tabs
+
+`Tabs.Item` and `Tabs.Panel` both take an optional `data-id` attribute that is used for the
```
williamjstanton, line 83: Should this say onSelect instead of onActivate to match the model API?

modules/preview-react/tabs/lib/TabsItem.tsx
```diff
@@ -0,0 +1,264 @@
...
+   */
+  id?: string;
+  /**
+   * Part of the ARIA specification for tabs. This attributes links a `role=tab` to a
+   * `role=tabpanel`. This value must be the same as the associated `id` attribute of the tab panel.
+   * This is automatically set by the component and should only be used in advanced cases.
+   */
+  'aria-controls'?: string;
+  /**
+   * Part of the ARIA specification for tabs. Lets screen readers know which tab is active. This
```
williamjstanton, line 73: This says aria-selected should never be false, but useTabsItem returns false for inactive tabs. I think the implementation is correct according to W3C spec. We just need to update this comment and any related docs. 

Spec: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

williamjstanton (changes requested):

modules/preview-react/tabs/lib/TabsItem.tsx
sheelah: Thanks William! That sounds right to me -- Removed that part of the comment.

modules/preview-react/tabs/stories/Tabs.mdx
sheelah: Good catch! Fixed 👍🏻

modules/preview-react/tabs/lib/TabsItem.tsx
sheelah: Thanks for the callout @williamjstanton - left a question for design on this.  I assume that the border that shows on selected tabs is enough of a differentiation for the `outlined` variant?  If so, then would adding a border that's visible on selected tabs for the default (`filled`) variant when we detect high contrast mode (via `'@media (forced-colors: active)'` which we use elsewhere) be enough of a visual indication?

modules/preview-react/tabs/lib/TabsList.tsx
sheelah: @williamjstanton Given our tight timeline (Jul 31st launch for v16), we'd like to put this on hold and instead add it in a fast-follow release after that deadline.  This way we can make sure we allow adequate time to implement this in a performant way & have thorough testing in place.  Does that work for you?

modules/preview-react/tabs/stories/Tabs.mdx
```diff
@@ -0,0 +1,134 @@
...
+
+<ExampleCodeBlock code={DynamicTabs} />
+
+## Component API
+
+<SymbolDoc name="Tabs" fileName="/react/" />
+
+## Specifications
+
+<Specifications file="./cypress/component/Tabs.spec.tsx" name="Tabs" />
```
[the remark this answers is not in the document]
sheelah: This is reusing current Tabs tests right now -- no functionality differences.

modules/preview-react/tabs/lib/TabsList.tsx
williamjstanton: OK, I'm comfortable with that. I realize this is a big change. How are we tracking this follow up? Should we create an issue?

modules/preview-react/tabs/lib/TabsItem.tsx
williamjstanton: > Thanks for the callout @williamjstanton - left a question for design on this. I assume that the border that shows on selected tabs is enough of a differentiation for the `outlined` variant? If so, then would adding a border that's visible on selected tabs for the default (`filled`) variant when we detect high contrast mode (via `'@media (forced-colors: active)'` which we use elsewhere) be enough of a visual indication?

Yes, the outline variant decoration for selected state negates the color contrast requirement. Adding it into a media query for the `filled` variant will certainly address the concern while using high contrast mode. 

But without high contrast mode, this filled variant design falls short of our standards and guidelines because there isn't enough contrast (minimum 3:1) between both states. (Selected tab, not selected tabs)

SC 1.4.11 Non-text contrast: https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html

modules/preview-react/tabs/lib/TabsList.tsx
sheelah: I've reopened https://github.com/Workday/canvas-kit/issues/3558 for this and it's part of our Sana roadmap, as a post-v16 launch item.  Thanks for your understanding!

modules/preview-react/tabs/lib/TabsItem.tsx
sheelah: Ah ok - didn't realize the `filled` variant's selected state itself had an issue as well. I will sync with design on this. Agree that it's super light.

sheelah: Outline on selected state for the filled variant has been added for high contrast mode.

sheelah: Filled variant now has a visible border for tab selected state ✅

modules/preview-react/tabs/lib/TabsList.tsx
williamjstanton: Amazing, that works for me.

modules/preview-react/tabs/lib/TabsItem.tsx
williamjstanton: OK, that works for me. 

Does this make the outline variant redundant then...?

williamjstanton (approved):
LG!

modules/preview-react/tabs/lib/TabsItem.tsx
sheelah: No, since the default styling is still different for both.  We aren't clear on the use case for the `outline` variant but it's needed to align with Sana.

modules/preview-react/tabs/lib/Tabs.tsx
```diff
@@ -0,0 +1,140 @@
...
+     *     <Tabs.OverflowButton>More</Tabs.OverflowButton>
+     *   }>
+     *     {dynamic tab items}
+     *   </Tabs.List>
+     * </Tabs>
+     * ```
+     */
+    OverflowButton: TabsOverflowButton,
+    /**
+     * @deprecated ⚠️ `MenuPopper` has been deprecated and will be removed in a future major version. Please use `Tabs.Menu.Popper` instead.
```
mannycarrera4: is this stll true?

mannycarrera4: since it's a new component we can just remove it?

sheelah: You mean remove the deprecation message, or remove MenuPopper as a subcomponent & call that out in the upgrade guide?

mannycarrera4: yeah if `MenuPopper` was deprecated in main tabs, why add it, we can just remove it since this is a net new component

Issue #3986, opened 2026-06-12
feat: Sana Canvas Navigation


jaclynjessup:
🚀 Feature Proposal

Update Navigation components to align with v16 token and style changes.

**Components in this ticket:**
- Breadcrumbs
- Hyperlink / External Hyperlink
- Pagination

---

## Prerequisites

- [ ] Design review: Confirm all Figma components are marked ready for dev
- [ ] A11y reviews completed for Breadcrumbs (chevron size) and Hyperlink (default state styling)

---

## Token Updates Required

All components: Space ✅, Size ✅, Shape ✅, Color ✅, Depth No, Theme No

---

## Visual Changes

| Component | Changes |
|-----------|---------|
| Breadcrumbs | Token changes; underline added on hover |
| Hyperlink | Token changes; `inverse` variant removed |
| Pagination | Token changes; icon sizes; corner radius |

---

### Breadcrumbs
- **No API changes.**

### Hyperlink / External Hyperlink
- **`inverse` variant removed.** Codemod required.

### Pagination
- **No API changes.**

---

### Breadcrumbs
The chevron icon button size has decreased and may fall below the recommended click target size (44×44px WCAG 2.5.5).
- **A11y review required before shipping.**

### Hyperlink
The updated default state has no styling to indicate link text (relying only on hover/active/focus), which creates a differentiation issue when links appear within a paragraph of black body text.
- **A11y review required before shipping.**

---

## Effort & Timeline

- **Priority:** Low 
- **Breadcrumbs Effort:** Medium (token changes, hover underline)
- **Hyperlink Effort:** Medium (token changes, inverse variant removal)
- **Pagination Effort:** Small (token changes, icon sizes, corner radius)

---

## Dependencies

- **No blockers**
- **A11y reviews required** for Breadcrumbs and Hyperlink before shipping

---

## Acceptance Criteria

- [ ] Breadcrumbs token updates applied; underline-on-hover added; chevron a11y reviewed
- [ ] Hyperlink `inverse` variant removed with codemod; default state a11y reviewed
- [ ] Pagination token updates applied (icon sizes, corner radius)
- [ ] Codemods written and documented for Hyperlink
- [ ] Visual regression tests pass
- [ ] Storybook updated

PR #4068, opened 2026-07-15
feat: Add Sana Canvas Navigation component updates

sheelah:
## Summary

Fixes: #3986. 

Note: I filed #4070 for the Breadcrumbs overflow button rendering issues we're seeing in Storybook.

## Release Category
Components

### Release Note
This introduces a new `type` variant for `Hyperlink` and `ExternalHyperlink`, defaulting to "inline". Set it to "standalone" to remove a link's underline.

### BREAKING CHANGES
The `Hyperlink` and `ExternalHyperlink` `standalone` and `standaloneInverse` variants have been removed, in favor of using a new `type` prop which is either `inline` (default) or `standalone`.  A codemod is available to migrate to the new component API.

---

## Checklist

## For the Reviewer

- [ ] Code
- [ ] Documentation
- [ ] Testing
- [ ] Codemods

## Summary by CodeRabbit

* **New Features**
  * Hyperlinks now separate color variants and underline styling via `variant` (color) and `type` (underline).
  * Updated ExternalHyperlink to apply the same `type`-driven underline behavior, plus new Secondary hyperlink examples and visual-state coverage.
* **Bug Fixes**
  * Refined Breadcrumbs and Pagination visuals (colors, icon sizing, focus/hover/active states, and GoTo layout/alignment).
* **Documentation**
  * Updated the v16 upgrade guides with the new Hyperlink semantics and migration instructions.
* **Chores**
  * Expanded the v16 codemod to automatically migrate legacy Hyperlink prop configurations.


modules/react/button/lib/Hyperlink.tsx
```diff
@@ -23,24 +32,24 @@ export const hyperlinkStencil = createStencil({
...
     },
     '&:focus, &.focus, &:focus-visible': {
-      boxShadow: `0 0 0 ${px2rem(2)} ${system.legacy.color.brand.focus.primary}`,
+      boxShadow: `0 0 0 ${px2rem(2)} ${system.legacy.color.brand.border.primary}`,
       outline: 'none',
     },
     '&:active, &.active': {
-      color: system.legacy.color.fg.info.strong,
-      background: system.legacy.color.surface.alt.default,
+      color: system.sana.color.brand.fg.link.hover,
```
mannycarrera4: should this be a brandable token? same for above?

modules/docs/llm/upgrade-guides/16.0-UPGRADE-GUIDE.md
```diff
@@ -73,17 +77,10 @@ our team. We'd be very happy to walk you through the process to set you up for s
 
 ### Instructions
 
-The easiest way to run our codemod is to use `npx` in your terminal.
-
-<<<<<<< HEAD
-
-```sh
-npx @workday/canvas-kit-codemod v16 [path]
```
mannycarrera4, line 88: sorry...

mannycarrera4 (approved):
LGTM

modules/react/button/lib/Hyperlink.tsx
sheelah: @mannycarrera4 That's a good question. Still wrapping my head around our current token hierarchy (legacy vs non-legacy etc). At the end of the day, who decides what gets a brandable token?

modules/react/button/lib/Hyperlink.tsx
```diff
@@ -23,24 +32,24 @@ export const hyperlinkStencil = createStencil({
   base: {
     fontFamily: system.fontFamily.default,
     textDecoration: 'underline',
-    color: system.legacy.color.fg.info.default,
+    color: system.sana.color.brand.fg.link.default,
```
sheelah: Using `system.sana.*` tokens for `color` in this component as there was no `system.legacy.color.*.fg.link.*` or `system.color.brand.fg.link.*`  equivalent

modules/react/button/lib/Hyperlink.tsx
sheelah: Left the `color` as-is for the Hyperlink states as there is no system.legacy.* or system.* equivalent, but I updated the focus ring color to be brandable.

`system.legacy.color.brand.border.primary` is not brandable but `system.legacy.color.brand.focus.primary` is according to `CanvasProvider` 's  focusOutline theme, so I swapped that in.

mannycarrera4: anything that has the `brand` word in it is brandable

mannycarrera4: I just don't think that hyperlinks should be brandable if someone choose purple, i think a hyperlink should still be blue

modules/react/button/lib/Hyperlink.tsx
mannycarrera4: i think we want `info` here since info is non brandable and its a link

sheelah: All set 👍🏻

modules/react/button/stories/button/Hyperlink.mdx
```diff
@@ -38,16 +39,21 @@ the current page. By default, `Hyperlink`s have an underline for accessibility.
...
 
-`Hyperlink`s also have a `standalone` variant. Use this variant when you need to place a `Hyperlink`
-outside of a paragraph or body text. This will remove the underline.
+<ExampleCodeBlock code={SecondaryLink} />
+
+#### Standalone
 
-This variant can also be used with the `ExternalHyperlink` component.
+Use `type="standalone"` when you need to place a `Hyperlink` outside of paragraph or body text. This
+removes the underline. By default, `type` is `"inline"` and links are underlined.
```
williamjstanton: Docs request: the new secondary variant and type="standalone" guidance should include a stronger WCAG 1.4.1 note. Removing underline is only safe when the link has another persistent non-color affordance or is clearly identified by context. A hover-only underline is not sufficient for inline/body text links per WCAG F73. 

Could we add guidance warning against type="standalone" for links embedded in body copy? I already have this noted in line 67 below. 

Does this example for 'standalone' duplicate the "in body text' example below? Can we consolidate the content? 
1. Introduce the new standalone type here, 
2. Use the example from 'in body text' below with the new type, 
3. Include / polish the 'accessibility note' in 'in body text' below? 

Guideline: https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html 
Failure: https://www.w3.org/WAI/WCAG22/Techniques/failures/F73

modules/react/button/stories/button/Hyperlink.mdx
```diff
@@ -38,16 +39,21 @@ the current page. By default, `Hyperlink`s have an underline for accessibility.
...
 
-#### Standalone
+#### Secondary
+
+`Hyperlink`s also have a `secondary` variant. This is a dark gray alternate to the default blue
+hyperlink. Use this variant when you need to place a `Hyperlink` on a colored background.
 
-`Hyperlink`s also have a `standalone` variant. Use this variant when you need to place a `Hyperlink`
-outside of a paragraph or body text. This will remove the underline.
+<ExampleCodeBlock code={SecondaryLink} />
```
williamjstanton, line 51: How come we are using the new standalone type here in secondary, before we have introduced it? Should we just focus on showcasing the new Secondary?

modules/react/button/lib/Hyperlink.tsx
```diff
@@ -6,14 +6,23 @@ import {system} from '@workday/canvas-tokens-web';
...
+  variant?: 'inverse' | 'secondary';
   /**
-   * attribute for the hyperlink URL
+   * Sets underline behavior for Hyperlink
+   * - `inline` (default): keeps the underline. Use when the hyperlink appears in body text.
+   * - `standalone`: removes the underline. Use when a hyperlink is outside paragraph or body text.
+   *
+   * @default 'inline'
+   */
+  type?: 'inline' | 'standalone';
```
williamjstanton: API/a11y edge case: the new Canvas type prop controls underline behavior, but it also shadows the native HTML type attribute if Hyperlink is rendered polymorphically as a button. Docs discourage links-as-buttons, which is good, but should we explicitly document that Hyperlink as="button" cannot use native type="button" / submit" through this prop?

modules/react/button/lib/Hyperlink.tsx
```diff
@@ -26,21 +35,21 @@ export const hyperlinkStencil = createStencil({
...
+    padding: `0 ${px2rem(2)}`,
     margin: '0 -2px',
     transition: 'color 0.15s,background-color 0.15s',
     wordBreak: 'break-word',
     '&:hover, &.hover': {
       color: system.legacy.color.fg.info.strong,
-      background: system.legacy.color.surface.alt.default,
+      background: system.legacy.color.surface.overlay.hover.default,
     },
     '&:focus, &.focus, &:focus-visible': {
```
williamjstanton, line 46: I validated this on a Windows machine using High Contrast themes, and the Hyperlink focus indicator is not visible. Since boxShadow is not reliable in forced-colors mode and this rule also sets outline: none, can we add a forced-colors-safe focus style here? A transparent outline or an explicit @media (forced-colors: active) outline/border fallback would make keyboard focus perceivable in WHCM.

modules/react/breadcrumbs/lib/BreadcrumbsLink.tsx
```diff
@@ -49,7 +59,8 @@ export const BreadcrumbsLink = createComponent('a')({
           ref={ref}
           as={Element}
           role="link"
-          variant="standalone"
+          variant="secondary"
+          type="standalone"
```
williamjstanton: I validated Breadcrumbs in Windows High Contrast and the focus indicator is not visible for breadcrumb links. Because Breadcrumbs.Link now renders Hyperlink with type="standalone" and inherits Hyperlink’s box-shadow-only focus treatment, can we ensure breadcrumb links get a forced-colors-safe focus indicator as well? This could be fixed in Hyperlink globally, but Breadcrumbs should be included in the verification since this PR changes its default link styling.

modules/react/pagination/lib/Pagination/PageButton.tsx
williamjstanton, line 44: For Windows High Contrast desktop themes, the current page is technically distinguishable by bold text, but the background color change won't survive. Can we strengthen this selected-state affordance by adding a more visually apparent indication such as a 2px border, for the current page state? (I think we did something similar in Segmented Control.)

williamjstanton (changes requested):
Finished reviewing accessibility. We need to polish up some of the focus states for high contrast modes.

modules/react/button/lib/Hyperlink.tsx
sheelah: Great point -- thanks! Talked this over with the team and we've decided to rename `type` to `linkType`. I've made that update.

modules/react/button/lib/Hyperlink.tsx
sheelah: Now we have `fg.link` tokens available in 4.4 beta so I've updated this to use those 👍🏻

modules/react/button/stories/button/Hyperlink.mdx
sheelah: Updated!

modules/react/button/stories/button/Hyperlink.mdx
sheelah: Yes, good point. I've updated this as well.

modules/react/button/lib/Hyperlink.tsx
sheelah: I've added an outline that shows when forced colors are active, which applies to all variants (also applies to Breadcrumbs which use this component) 👍🏻

modules/react/breadcrumbs/lib/BreadcrumbsLink.tsx
sheelah: Yes- this is covered by the change above which adds an outline to Hyperlink upon focus when forced colors are active. Thanks for the thorough testing!

modules/react/pagination/lib/Pagination/PageButton.tsx
sheelah: 👍🏻 I added the same approach here as SegementedControl -- a 2px outline visible when `'@media (prefers-contrast: more)'`.

sheelah:
Thanks @williamjstanton! This is ready for another look.

mannycarrera4 (approved):

modules/react/button/lib/Hyperlink.tsx
williamjstanton: Awesome!

modules/react/button/lib/Hyperlink.tsx
williamjstanton: Thanks, this works!

modules/react/pagination/lib/Pagination/PageButton.tsx
williamjstanton: Nice, I like it!

williamjstanton (approved):
LG!

